PR #8, opened 2022-02-18
Use custom type list

bmcpt:


gnunicorn (approved):
need to hold off until the ffigen-code is merged and switch this to `main`, but other than this looks fine.

q: ~~do we know if this still builds with the current SDK/app?~~ CI passed, so .... we assume this is fine...

PR #18, opened 2022-02-27
do not patch cargo toml

bmcpt:


effektio_flutter_sdk/lib/effektio_flutter_sdk_ffi.dart
```diff
@@ -655,52 +657,6 @@ class Api {
     return tmp9;
   }
 
-  String echo(
```
gnunicorn, line 658: this seems unrelated, but not a problem.

gnunicorn (approved):

effektio_flutter_sdk/lib/effektio_flutter_sdk_ffi.dart
bmcpt: It must be outdated. Should I add generated files to gitignore?

PR #33, opened 2022-03-06
Enforce and verify dart formatting

gtalha07:


.github/workflows/dart.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # Note: This workflow uses the latest stable version of the Dart SDK.
+      # You can specify other versions if desired, see documentation here:
+      # https://github.com/dart-lang/setup-dart/blob/main/README.md
+      # - uses: dart-lang/setup-dart@v1
+      - uses: dart-lang/setup-dart@9a04e6d73cca37bd455e0608d7e5092f881fd603
```
gnunicorn, line 25: any reason we use that specific thing rather than latest stable? Anything we need that isn't in stable? If so, please document - otherwise, let's stick with stable.

.github/workflows/dart.yml
```diff
@@ -0,0 +1,55 @@
...
+        working-directory: app
+        
+      - name: Apply Fix
+        run: dart fix --apply
+        working-directory: app
+        
+      # Uncomment this step to verify the use of 'dart format' on each commit.
+      - name: Verify formatting
+        run: dart format --output=show . 
+        working-directory: app/lib
```
gnunicorn, line 38: we aren't really enforcing the formatting here, do we? We only format and show, but don't stop CI runs if they differ... But the latter is what we really want, no?

gnunicorn (approved):
Please fix the comments and feel free to merge after.

.github/workflows/dart.yml
gtalha07: Yes, CI pass will fail on formatting because of unformatted files in main branch. So we have to overwrite files and commit on it.

PR #38, opened 2022-03-09
use latest matrix-sdk and ffi-gen

bmcpt:


effektio_flutter_sdk/lib/effektio_flutter_sdk_ffi.dart
```diff
@@ -7,7 +7,7 @@ library effektio;
 import "dart:async";
```
gnunicorn, line 7: isn't this file autogenerated, too? Why are we keeping this and not the others?

gnunicorn (approved):
Just one question

bmcpt:
For some reason build steps in ci aren't generating this files resulting in ci failure.

PR #56, opened 2022-04-04
FAQ area

gnunicorn:
- [x] Mock Data coming from native
- [x] Refactor Mocks in News
- [x] FAQ section
- [x] FAQ single item page
- [ ] multimedia FAQ content
- [ ] WYSIWYG editor for content


app/lib/common/widget/FaqListItem.dart
```diff
@@ -0,0 +1,319 @@
...
+                                        child: Text(
+                                          'Support',
+                                          style: GoogleFonts.montserrat(
+                                            color: Colors.white,
+                                            fontSize: 16,
+                                          ),
+                                        ),
+                                      )
+                                    ],
+                                  ),
```
gnunicorn, line 117: This PR introduced the idea of mocking the data right from the native side, as it makes it easier to replace it with the actual code later. Please let's not re-introduce mocked data right in the frontend (as this often requires significant changes when using actual async-function calls to the SDK). Let's mock this native side instead.

app/lib/common/widget/FaqListItem.dart
```diff
@@ -0,0 +1,319 @@
...
+                                    ),
+                                  },
+                                  child: Image.asset(
+                                    'assets/images/bookmark.png',
+                                  ),
+                                )
+                              ],
+                            ),
+                          ),
+                          Padding(
```
gnunicorn, line 81: This entire element is a shortcut to display a "team"/group, you belong to. It should be its own component. Please refactor this and make this into a resuable component.

app/lib/common/widget/FaqListItem.dart
```diff
@@ -0,0 +1,319 @@
...
+        child: Column(
+          children: <Widget>[
+            // ListTile(
+            // title: Text(
+            //   widget.faq.title(),
+            //   style: optionStyle,
+            //   maxLines: 2,
+            // ),
+            // trailing: Icon(Icons.turned_in_outlined, color: Colors.grey),
+            // subtitle:
```
gnunicorn, line 43: why is there outcommented code in here? Can't this just be removed instead?

app/lib/common/widget/FaqListItem.dart
```diff
@@ -0,0 +1,319 @@
...
+                              mainAxisAlignment: MainAxisAlignment.center,
+                              children: [
+                                Container(
+                                  height: 40.0,
+                                  padding: EdgeInsets.all(8.0),
+                                  margin: EdgeInsets.only(left: 12.0),
+                                  decoration: BoxDecoration(
+                                    borderRadius: BorderRadius.circular(10),
+                                    border: Border.all(
+                                      color: Color(0XFFFFFFFFF),
```
gnunicorn, line 94: Why are there static color definitions in this? We are using the Material Theme system, please use that instead of static colors. This goes for the entire view.

app/lib/common/widget/FaqListItem.dart
```diff
@@ -0,0 +1,319 @@
...
+                                          ),
+                                        ),
+                                      ],
+                                    ),
+                                  ),
+                                )
+                              ],
+                            ),
+                          ),
+                          Padding(
```
gnunicorn, line 181: this, too, is a list of comments and as such should a) be mocked from the native and not in the ui and b) be a separate component that can be reused in other places (and c) use the material theme system not static colors).

app/lib/screens/faq/Item.dart
```diff
@@ -0,0 +1,30 @@
...
+  const FaqItemScreen({Key? key, required this.client, required this.faq})
+      : super(key: key);
+  final Client client;
+  final Faq faq;
+
+  @override
+  _FaItemqScreenState createState() => _FaItemqScreenState();
+}
+
+class _FaItemqScreenState extends State<FaqItemScreen> {
```
gnunicorn, line 21: typo.

```suggestion
  _FaqItemScreenState createState() => _FaqItemScreenState();
}

class _FaqItemScreenState extends State<FaqItemScreen> {
```

app/lib/screens/faq/Overview.dart
```diff
@@ -0,0 +1,56 @@
...
+
+class FaqOverviewScreen extends StatefulWidget {
+  const FaqOverviewScreen({Key? key, required this.client}) : super(key: key);
+  final Client client;
+
+  @override
+  _FaOverviewqScreenState createState() => _FaOverviewqScreenState();
+}
+
+class _FaOverviewqScreenState extends State<FaqOverviewScreen> {
```
gnunicorn, line 19: Same typo.

```suggestion
  _FaqOverviewScreenState createState() => _FaqOverviewScreenState();
}

class _FaqOverviewScreenState extends State<FaqOverviewScreen> {
```

app/lib/screens/faq/Item.dart
```diff
@@ -0,0 +1,30 @@
+// ignore_for_file: prefer_const_constructors
```
gnunicorn, line 1: I prefer to call them "Details" as "ListItem" and "Item" are not easy to distinguish. Let's please refer to the single object Screens as "XYZDetailsScreen". Please rename the file and the component accordingly.

app/lib/screens/faq/Item.dart
```diff
@@ -0,0 +1,30 @@
...
+
+class _FaItemqScreenState extends State<FaqItemScreen> {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+        appBar: AppBar(
+          title: Text(widget.faq.title()),
+        ),
+        body: Center(child: Text(widget.faq.body())));
+  }
```
gnunicorn, line 29: no back, no sidebar? What does this screen actual look like? Please provide a screenshot.

app/pubspec.yaml
```diff
@@ -48,6 +48,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   flutter_launcher_icons: ^0.9.2
+  fluttertoast: ^8.0.9
```
gnunicorn, line 51: why introduce another external component, when material design [already has a snackbar component](https://material.io/components/snackbars)? Unless there is an important feature that provides (and we are using), let's stick with the material already included components. Please replace the flutter toast with the material snackbar.

app/lib/screens/faq/Item.dart
inderjeet-developer7: image

gnunicorn: that image does not show the screen of the code we are talking about.

PR #57, opened 2022-04-09
flutter app icon and name

inderjeet-developer7:
App icon changed


app/android/app/build.gradle
```diff
@@ -44,8 +44,8 @@ android {
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
         applicationId "com.example.effektio"
```
gnunicorn, line 46:
```suggestion
        applicationId "org.effektio.app"
```

app/android/app/build.gradle
```diff
@@ -44,8 +44,8 @@ android {
...
-        minSdkVersion flutter.minSdkVersion
-        targetSdkVersion flutter.targetSdkVersion
+        minSdkVersion 21
+        targetSdkVersion 29
```
gnunicorn, line 48: what is this about?

gnunicorn (changes requested):
The file app/ios/Runner/Assets.xcassets/AppIcon.appiconset/Icon-App-1024x1024@1x.png should be replaced with a higher resolution one

app/android/app/build.gradle
inderjeet-developer7: For generating the icons for android we need min SDK version as 21 or higher but the version here is lesser so did the changes, Flutter commands give error to these..

gnunicorn (approved):

PR #60, opened 2022-04-20
Chat functionality

gnunicorn:


gtalha07:
We would need extend TimelineStream class functionality so we can push messages in a conversation, ideally some function:

```
Future<RoomMessage> sendMessage (String messageText){
     ///code
} 
```
**https://docs.effektio.org/api/main/dart-sdk/effektio/TimelineStream-class.html**

gtalha07:
Updated hamburger icon with SVG format. Closing #66

gtalha07:
According to #3, the tasks have been implemented and can be reviewed in latest commit of PR. In general the basic chat is integrated. @gnunicorn you can have a look at CI check, then we can merge the PR in main. Some things which haven't been implemented due to restrictions: 
1. Not all chats are being shown, I believe its suppressed from the backend.
2. Seen tick indicator not changing its color dynamically, we would need some async/await function like ```next()``` to update it.
3. Room Info, as we need more functions to integrate it properly.

app/lib/screens/HomeScreens/News.dart
```diff
@@ -28,8 +29,10 @@ class _NewsScreenState extends State<NewsScreen> {
           builder: (BuildContext context) {
             return IconButton(
               icon: Container(
+                height: MediaQuery.of(context).size.height,
+                width: MediaQuery.of(context).size.width,
                 margin: const EdgeInsets.only(bottom: 10, left: 10),
-                child: Image.asset('assets/images/hamburger.png'),
+                child: SvgPicture.asset('assets/images/hamburger.svg'),
```
gnunicorn, line 35: this is not the right image now. We wanted the menu-lined-`e`, this is now a classic hamburger menu. Please revert.

app/lib/common/widget/emptyMessagesPlaceholder.dart
```diff
@@ -0,0 +1,34 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.start,
+      children: <Widget>[
+        SizedBox(height: MediaQuery.of(context).size.height / 6),
+        SvgPicture.asset('assets/images/emptyPlaceholder.svg'),
+        const SizedBox(height: 10),
+        const Text(
+          'No Messages here yet ...',
```
gnunicorn, line 16: Localisation!?!

app/lib/common/widget/emptyMessagesPlaceholder.dart
```diff
@@ -0,0 +1,34 @@
...
+          style: TextStyle(
+            color: Colors.white,
+            fontSize: 16,
+            fontWeight: FontWeight.w600,
+          ),
+        ),
+        const SizedBox(height: 20),
+        const Text(
+          'Start the conversation by sending the message',
```
gnunicorn, line 25: Localisation!?!

app/lib/main.dart
```diff
@@ -39,6 +40,13 @@ class Effektio extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      //To avoid redundancy, the whole textTheme of app is setup as Montserrat,
+      //so we can avoid using it explicitly in every Text Widget
```
gnunicorn, line 44: This would be a useful comment _on the PR description_ as it is an unrelated but useful change and would have helped me understand all the changes in the other files, that didn't seem to belong here.

app/lib/common/widget/NewsSideBar.dart
```diff
@@ -1,5 +1,6 @@
 // ignore_for_file: prefer_const_constructors
 
+import 'package:cached_network_image/cached_network_image.dart';
```
gnunicorn, line 3: I don't understand what these changes have to do with the chat-PR. Please revert this. If they are relevant please raise them as a separate PR.

app/lib/main.dart
```diff
@@ -147,8 +79,8 @@ class _EffektioHomeState extends State<EffektioHome> {
   late Future<Client> _client;
   int tabIndex = 0;
 
-  static const TextStyle optionStyle =
-      TextStyle(fontSize: 30, fontWeight: FontWeight.bold, color: Colors.white);
+  // static const TextStyle optionStyle =
+  //     TextStyle(fontSize: 30, fontWeight: FontWeight.bold, color: Colors.white);
```
gnunicorn, line 83:
```suggestion
```

app/lib/main.dart
```diff
@@ -182,75 +114,77 @@ class _EffektioHomeState extends State<EffektioHome> {
...
-                  ),
-                ),
-              ),
-            );
-          }
-        },
-      ),
+
+      ChatList(client: _client),
+      // FutureBuilder<Client>(
```
gnunicorn, line 119: please remove outcommented code.

app/lib/repository/client.dart
```diff
@@ -11,3 +14,45 @@ Future<Client> login(String username, String password) async {
   Client client = await sdk.login(username, password);
   return client;
 }
+
+//Fetch user id from sdk
+Future<String> getUser(Future<Client> client) async {
+  Client _client = await client;
+  final String userId = await _client.userId();
+  return userId;
+}
```
gnunicorn, line 23: I've seen these a few times now and I don't understand them. You can just as well write `await (await client).user_id()` wherever you are currently calling `getUser`. What's the point of putting that into another function?

app/lib/repository/client.dart
```diff
@@ -11,3 +14,45 @@ Future<Client> login(String username, String password) async {
...
+    isSeen = !isSeen;
+  }
+  return _messages;
+}
+
+//Push message
+Future<String> sendMessage(Conversation convo, String message) async {
+  var res = await convo.sendPlainMessage(message);
+  return res;
+}
```
gnunicorn, line 58: Same here? What is the purpose of a function doing exactly one function call and returning its result? Why?

app/lib/screens/HomeScreens/ChatList.dart
```diff
@@ -0,0 +1,192 @@
...
+        leadingWidth: MediaQuery.of(context).size.width / 5,
+        toolbarHeight: MediaQuery.of(context).size.height / 15,
+        backgroundColor: Color.fromRGBO(36, 38, 50, 1),
+        elevation: 0,
+        leading: Container(
+          margin: const EdgeInsets.only(left: 10),
+          child: TextButton(
+            onPressed: () {},
+            child: Text(
+              'Select',
```
gnunicorn, line 52: Localisation!?!

app/lib/screens/HomeScreens/ChatList.dart
```diff
@@ -0,0 +1,192 @@
...
+        child: SingleChildScrollView(
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: <Widget>[
+              Container(
+                margin: const EdgeInsets.only(
+                  left: 18,
+                ),
+                child: Text(
+                  'Chat',
```
gnunicorn, line 84: Localisation!?!

app/lib/screens/HomeScreens/ChatList.dart
```diff
@@ -0,0 +1,192 @@
...
+                  style: TextStyle(
+                    color: Colors.white,
```
gnunicorn, line 86: Someone should explain to me, how theming is supposed to work with `MaterialTheme` in flutter because again and again I see people _hard define colors and sizes_ in the widgets rather than on the Theme. How are we suppose to replace the theme later (for branded versions of the App) if they are hard coded? Can we please have a guide on how to use the Theming and enforce that?!?

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+    _handleEndReached();
+    _updateState();
+    super.initState();
+  }
+
+  //will detect if any new event is arrived and will re-render the screen
+  void _updateState() async {
+    var timeline = await widget.room.timeline();
+    await timeline.next();
+    var newEvent = await widget.room.latestMessage();
```
gnunicorn, line 55: `timeline.next()` already returns the new event.
```suggestion
    var newEvent = await timeline.next();
```

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+      final textMessage = types.TextMessage(
+        id: newEvent.eventId(),
+        author: user,
+        text: newEvent.body(),
+      );
+      setState(() {
+        _messages.insert(0, textMessage);
+      });
+    }
+    _updateState();
```
gnunicorn, line 69: this is recursive, right? Won't that run out of the stack eventually? Shouldn't rather put this into a loop?

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+      setState(() {
+        _messages[index] = updatedMessage;
+      });
+    });
+  }
+
+  //push messages in conversation
+  void _handleSendPressed(types.PartialText message) async {
+    await widget.room.typingNotice(false);
+    var eventId = await sendMessage(widget.room, message.text);
```
gnunicorn, line 95: This pattern isn't super clean. We will receive the event through the `timeline.next()` and thus might add it to the view, twice. This is fine for now, but in a later version we want to have the SDK take the event and handle its `sending`-state, too and then replace the data once it has been confirmed by the server. But the SDK isn't there yet.

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+    );
+  }
+
+  void _handleImageSelection() async {
+    final result = await ImagePicker().pickImage(
+      imageQuality: 70,
+      maxWidth: 1440,
+      source: ImageSource.gallery,
+    );
+
```
gnunicorn, line 182: This is confusing. We should have a FIXME to a ticket and, more importantly, alert the user that this isn't actually send to the server yet. Let's show a snackbar alerting the user that this isn't actually implemented yet.

```suggestion
FIXME: actually send the image to the server #NEW_TICKET
```

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+      _addMessage(message);
+    }
+  }
+
+  void _handleFileSelection() async {
+    final result = await FilePicker.platform.pickFiles(
+      type: FileType.any,
+    );
+
```
gnunicorn, line 206: Same as for imageMessage

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+    }
+  }
+
+  //Load messages at start
+  void _getMessages() async {
+    setState(() {
+      isLoading = true;
+    });
+    var stream = await widget.room.timeline();
+    List<types.Message> messages = await getMessages(stream, 10, widget.room);
```
gnunicorn, line 233: Your "message appear twice when scrolling back" stems from here:

The `timeline` holds the `index` of message looked up. When calling you create a new `stream` everytime and that always starts from `0`. Instead you need to hold the stream and keep calling `getMessages(stream, 10, widget.room)` on the same instance over and over again, when scrolling back in time.

Either fix it here or convert this comment into a ticket and add a `FIXME` linking to it.

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+    setState(() {
+      _messages = messages;
+      isLoading = false;
+    });
+  }
+
+  //Pagination Control
+  Future<void> _handleEndReached() async {
+    var stream = await widget.room.timeline();
```
gnunicorn, line 242: See remark above for `_getMessages` on why the backwards pagination isn't working as expected.

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+  //     margin: nextMessageInGroup
+  //         ? const BubbleEdges.symmetric(horizontal: 6)
+  //         : null,
+  //     nip: nextMessageInGroup
+  //         ? BubbleNip.no
+  //         : _user.id != message.author.id
+  //             ? BubbleNip.leftBottom
+  //             : BubbleNip.rightBottom,
+  //   );
+  // }
```
gnunicorn, line 275:
```suggestion
```

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -0,0 +1,435 @@
...
+                user: _user,
+                //custom avatar builder
+                avatarBuilder: (userId) {
+                  return Padding(
+                    padding: const EdgeInsets.only(right: 10),
+                    child: CircleAvatar(
+                      radius: 15,
+                      child: Text(getNameFromId(userId)[0].toUpperCase()),
+                    ),
+                  );
```
gnunicorn, line 407: We should reuse the (then new) avatar-component here. The user-name-to-icon should only be a fallback.

gnunicorn (commented):
Thanks heaps, this takes us a lot further @gtalha07 ! I have a few remarks, some general, some we should fix before the merge others, we want to convert into tickets and fix later.

General remarks: please stick to one atomic change at a time, drive-by-fixes increase the PR size make it a lot harder to comprehend what is going and why. Instead for things like the style-changes or the logo-update just quickly jump into `main` branch off, do the fix, raise the PR and merge that branch into yours. Git-merging allows for a quick and easy way to keep that fine and your work fast -  and this makes reviewing the distinct parts a lot easier. For this, let's keep it for now (aside from the stuff mentioned below) - but in the future let's stick to that style.

Please fix the following, then we can merge:
 - clean up the code to remove outcommented stuff - just doesn't belong into `main` ever (I've not remarked every instance, please go through and fix it yourself)
 - please revert the new-hamburger-logo-change: it's the wrong icon (and reopen the ticket belonging to it)
 - Clarify/remove the wrapping function calls
 - add a snackbar info to the user that the images/file they've added to the chat isn't actually send to the server right now
 
TThe following, we keep the code as is (maybe add FIXME's where I asked about that) but please convert the specific remarks into separate issues (all tagged `bug` and `s-chat`):
 - localisation
 - colors and theming
 - sending image messages
 - sending file messages
 - creating reusable avatar-component
 - showing the correct user-icon within the chat
 - fixing the backpaging in chat
 - user_id -> User.display_name()
 - replacing our own time calculation with `time`-module
 - recursion -> loop for listening to new messages
 - the send-receive-message-flow

app/lib/common/store/chatTheme.dart
```diff
@@ -0,0 +1,372 @@
...
+    Color backgroundColor = dark,
+    EdgeInsets dateDividerMargin = const EdgeInsets.only(
+      bottom: 32,
+      top: 16,
+    ),
+    TextStyle dateDividerTextStyle = const TextStyle(
+      color: neutral7,
+      fontSize: 12,
+      fontWeight: FontWeight.w800,
+      height: 1.333,
```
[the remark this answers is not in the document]
gtalha07: I'll remove these commented codes in next commit. Although yes, it is not very much important to keep them as they it can be scrapped again.

app/lib/common/widget/AppCommon.dart
```diff
@@ -1,13 +1,49 @@
...
-    style: GoogleFonts.montserrat(
+    style: const TextStyle(
       fontSize: 20,
       fontWeight: FontWeight.w700,
       color: Colors.white,
     ),
   );
 }
+
+String getNameFromId(String name) {
```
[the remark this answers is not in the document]
gtalha07: Noted.

app/lib/common/widget/AppCommon.dart
```diff
@@ -1,13 +1,49 @@
...
+  if (name.contains('@')) {
+    start = name.indexOf('@') + 1;
+  }
+  if (name.contains(':')) {
+    end = name.indexOf(':');
+  }
+  return name.substring(start, end);
+}
+
+String formatedTime(int value) {
```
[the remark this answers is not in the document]
gtalha07: I haven't checked this package but I initially used [duration](https://pub.dev/packages/duration) and it was not correctly parsing the seconds into hours (the calculated was 5 hours less than actual).  So I came up with own custom solution. I can open ticket for it if we want to use that instead of custom function.

app/lib/common/widget/ChatBottomWidget.dart
```diff
@@ -0,0 +1,32 @@
+// import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
gtalha07: Noted.

app/lib/common/widget/ChatOverview.dart
```diff
@@ -0,0 +1,149 @@
...
+            );
+          } else {
+            return CircleAvatar(
+              radius: 25,
+              backgroundColor: Colors.grey[700],
+              child: SvgPicture.asset('assets/images/people.svg'),
+            );
+          }
+        },
+      ),
```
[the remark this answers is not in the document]
gtalha07: Yes that would be good, I can make custom widget for it. Good catch!

app/lib/common/widget/emptyMessagesPlaceholder.dart
gtalha07: Yes regarding that we need to reopen ticket for it as some of the content has been changed since it was implemented. We need to rework on it.

app/lib/common/widget/NewsSideBar.dart
gtalha07: This relates to throwing bug if we don't use cached image in news feed avatar. Although its unrelated, I'll remove that.

app/lib/screens/HomeScreens/ChatList.dart
gtalha07: There are various ways of doing this, we can setup ThemeData for this that will control all the theme of the app, so we don't need to hard code values in every widget. You can take the example of textTheme which I have implemented. This can go in a seperate PR.

app/lib/screens/HomeScreens/ChatScreen.dart
gtalha07: I'll fix that here.

app/lib/common/widget/NewsSideBar.dart
gnunicorn: @gtalha07 leave it in, just a comment for later to keep things one-thing-per-PR.

app/lib/repository/client.dart
gtalha07: The purpose was to just have more readability, as it would be nice to have all sdk functions seperately in a dart file so it doesn't get complicated to look out for bugs in the future. But I think thats why ffi file serves that case which I couldn't understood it earlier. I'll update it.

PR #73, opened 2022-05-03
Build-in Google Fonts & add license

### LINKED ISSUE #24: Build-in Google Fonts & add license
Since #21, we are using Google Fonts. But the set up is incomplete:

- [ ] We can't (for legal reasons) have the App pull the fonts at startup but instead have to download the ones we use and have them compiled in - as per  https://pub.dev/packages/google_fonts
- [ ] we need to expose the license they are running under, too (as the docs say).

sachintha-s:
fixes #24


gnunicorn (approved):
LGTM

sachintha-s:
## Note:
when we run the app it displays the following error message.
online font error

#### Why this error? 
I disabled font loading online since we are using a downloaded font (Roboto). but still, the font named "Montserrat" was as previous in some files. we should replace that with Roboto but I didn't do it because It might be a cause for conflict.

gnunicorn:
> I disabled font loading online since we are using a downloaded font (Roboto). but still, the font named "Montserrat" was as previous in some files. we should replace that with Roboto but I didn't do it because It might be a cause for conflict.

@sachintha-s please just replace them in all files.

sachintha-s:
> > I disabled font loading online since we are using a downloaded font (Roboto). but still, the font named "Montserrat" was as previous in some files. we should replace that with Roboto but I didn't do it because It might be a cause for conflict.
> 
> @sachintha-s please just replace them in all files.

OK,
If you have, can I see the UI/UX design (Figma)? then I can use different font styles and sizes with respect to the design.

app/lib/common/store/Colors.dart
```diff
@@ -4,4 +4,5 @@ class AppColors {
   static const Color primaryColor = Color(0xffE9385D);
   static const Color textFieldColor = Color(0xff171717);
   static const Color backgroundColor = Color(0xff0A0A0A);
+  static const Color white = Colors.white;
 }
```
gtalha07, line 8: This needs to get updated as we don't want to define seperate classes for colors and styles. Rather have single class to handle theme overall. Already mentioned this in #79. For now its okay but open a seperate PR for this.

app/lib/common/store/textTheme.dart
```diff
@@ -0,0 +1,14 @@
...
+class CustomTextTheme {
+ static TextTheme textTheme = TextTheme(
+    bodyText1: GoogleFonts.roboto(
+      fontSize: 16,
+      color: AppColors.white,
+      fontWeight: FontWeight.w600,
+      letterSpacing: 0.2,
+    ),
+  );
+}
```
gtalha07, line 14: Same as commented for color class.

gtalha07 (approved):

gnunicorn:
@sachintha-s please resolve the merge conflict :) .

Issue #72, opened 2022-05-02
Custom Reusable Avatar Builder


gtalha07:
Custom reusable widget for avatar displaying needs to be implemented.

gnunicorn:
We want a reusable component, that given the size and `avatar()` and `display_name()` of a user or team creates the circle avatar for us.

PR #74, opened 2022-05-03
Custom Reusable Avatar

sachintha-s:
#72


app/lib/common/widget/customAvatar.dart
```diff
@@ -0,0 +1,69 @@
...
+            future: displayName,
+            builder: (
+              BuildContext context,
+              AsyncSnapshot<String> snapshot,
+            ) {
+              if (snapshot.hasData) {
+                return Container(
+                  margin: const EdgeInsets.all(10),
+                  child: CircleAvatar(
+                    backgroundColor: Colors.brown.shade800,
```
gnunicorn, line 47: we should do something more clever than just have all avatars be brown. e.g. what [colorized avatar](https://flutterkit.dev/flutter-plugin-that-allows-users-to-create-text-avatar-easily/) does...

app/lib/common/widget/customAvatar.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return FutureBuilder<List<int>>(
+      future: avatar,
+      builder: (
+        BuildContext context,
+        AsyncSnapshot<List<int>> snapshot,
+      ) {
+        if (snapshot.hasData) {
```
gnunicorn, line 24: the return value of `.avatar()` comes back with an option. Is this covered by this check, that it might come back without data?

gnunicorn (changes requested):
this holds a lot of unrelated changes (building on top of the other PR I assume, rather than switching back to main) and isn't actually using the new widget (e.g. in the news-section). Please also replace the existing uses of the avatar throughout the app with this one and fix the review comments made below.

app/lib/common/widget/customAvatar.dart
sachintha-s: Yes, shall I close this request and start a new branch related to the main branch and make these changes there?

app/lib/common/widget/customAvatar.dart
sachintha-s: **snapshot.hasData** checks whether this snapshot contains a non-null value. I changed the code as bellows. 
```suggestion
        if (snapshot.hasData && snapshot.requireData.isNotEmpty) {
```
here, **snapshot.requireData.isNotEmpty** checks whether the required data of this snapshot is not empty.
How about this check?

app/lib/common/widget/customAvatar.dart
gnunicorn: let's do that here first, please. there's still other pending things anyways before we can merge this.

app/lib/common/widget/customAvatar.dart
gtalha07: Have this code here to also handle what to do if its empty. This also fixes navigation crashes:
```
 if (snapshot.connectionState == ConnectionState.waiting) {
            return SizedBox(
              height: 20,
              width: 20,
              child: CircularProgressIndicator(color: AppColors.primaryColor),
            );
          } else {
            if (snapshot.hasData) {
              return CircleAvatar(
                radius: 25,
                backgroundImage:
                    MemoryImage(Uint8List.fromList(snapshot.requireData)),
              );
            } else {
              return CircleAvatar(
                radius: 25,
                backgroundColor: Colors.grey[700],
                child: SvgPicture.asset('assets/images/people.svg'),
              );
            }
          }
```

app/lib/common/widget/ChatOverview.dart
```diff
@@ -68,24 +68,12 @@ class ChatListItem extends StatelessWidget {
...
-          } else {
-            return CircleAvatar(
-              radius: 25,
-              backgroundColor: Colors.grey[700],
-              child: SvgPicture.asset('assets/images/people.svg'),
-            );
-          }
-        },
+      leading: CustomAvatar(
+        avatar: room.avatar().then((fb) => fb.asTypedList()),
```
gnunicorn, line 72: isn't this kind of always the pattern? Shouldn't we then put the `then( (fb) ..` part into the component rather than repeating it every time?

app/lib/common/widget/SideMenu.dart
```diff
@@ -88,40 +87,23 @@ class _SideDrawerState extends State<SideDrawer> {
...
-                                    radius: 24,
-                                  ),
-                                );
-                              }
-                            },
+                          Container(
+                            margin: EdgeInsets.all(10),
+                            child: CustomAvatar(
+                              radius: 24,
+                              avatar: avatar,
```
gnunicorn, line 102: here it is a bit more hidden as it happens above in L85, but it does it again.

More notably, the list conversion probably isn't even really needed but instead another feature might be implemented on the SDK to directly support the pattern more clean and in that case it'd be nice if we'd only have to change the component, rather than its interface.

app/lib/common/widget/customAvatar.dart
```diff
@@ -0,0 +1,95 @@
+import 'dart:typed_data';
+import 'package:colorize_text_avatar/colorize_text_avatar.dart';
+import 'package:flutter/material.dart';
+import 'package:effektio/common/store/Colors.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+
+class CustomAvatar extends StatelessWidget {
+  final Future<List<int>> avatar;
```
gnunicorn, line 8: as said above, let's fix this external interface to require the specific FFI-Type and do all native conversion within the component please.

app/lib/common/widget/customAvatar.dart
```diff
@@ -0,0 +1,95 @@
...
+                  builder: (
+                    BuildContext context,
+                    AsyncSnapshot<String> snapshot,
+                  ) {
+                    if (snapshot.hasData) {
+                      return TextAvatar(
+                        numberLetters: 2,
+                        shape: Shape.Circular,
+                        upperCase: true,
+                        text: snapshot.data ?? 'N',
```
gnunicorn, line 74:
```suggestion
                        text: snapshot.data ?? 'e',
```

Why n? Let's go for e (as `effektio`).  Though in reality we also have a case anyways (at least their username).

gnunicorn (approved):
thanks. just one more aspect: moving the repeating conversion into the component. Then you can just merge it (no need for another review) :) .

PR #84, opened 2022-05-05
Chat bug fixes 

### LINKED ISSUE #81: Chat-icon faulty zoom level
While the Chat icon is correct in the list view (see `OpenTechSchool`)
Screenshot_1651678482

It is not the right zoom level in the chat windows (see top right):

Screenshot_1651678475

gtalha07:


gnunicorn (approved):
lgtm

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -295,122 +295,79 @@ class _ChatScreenState extends State<ChatScreen> {
...
+            FutureBuilder<FfiListMember>(
+              future: widget.room.activeMembers(),
+              builder: (
+                BuildContext context,
+                AsyncSnapshot<FfiListMember> snapshot,
+              ) {
+                if (snapshot.hasData) {
+                  return Text(
+                    '${snapshot.requireData.length.toString()} Members',
+                    style: GoogleFonts.montserrat(
```
gnunicorn: I think we moved away from `montserrat` with the google licence PR...

app/lib/common/widget/ChatOverview.dart
```diff
@@ -69,32 +69,12 @@ class ChatListItem extends StatelessWidget {
...
-            } else {
-              return CircleAvatar(
-                radius: 25,
-                backgroundColor: Colors.grey[700],
-                child: SvgPicture.asset('assets/images/people.svg'),
-              );
-            }
-          }
-        },
+      leading: CustomAvatar(
```
gnunicorn: is this in anticipation? because the custom Avatar isn't in here and isn't merged yet (and the `.avatar()` interface isn't implemented yet)

app/lib/common/widget/AppCommon.dart
```diff
@@ -25,18 +26,14 @@ String getNameFromId(String name) {
   return name.substring(start, end);
 }
 
-String formatedTime(int value) {
-  int h, m;
+String? formatedTime(int value) {
+  final int h = (value.seconds.inHours + 5) % 24;
```
gnunicorn, line 30: why `+ 5`? Does this need some timezone handling?

app/lib/common/widget/AppCommon.dart
```diff
@@ -25,18 +26,14 @@ String getNameFromId(String name) {
   return name.substring(start, end);
 }
 
-String formatedTime(int value) {
-  int h, m;
+String? formatedTime(int value) {
```
gnunicorn, line 29: okay. help me out here. how is this not just a call to [`intl`](https://pub.dev/documentation/intl/latest/intl/DateFormat-class.html)s `DateFormat.hm().format(DateTime.fromMillisecondsSinceEpoch(value, isUtc:true))` ? 

I know I asked for using `time` but simply because I wasn't aware that `DateTime` is in core and formatting can easily be done with `intl`. Let's go this route instead please.

[1]: https://api.dart.dev/stable/2.16.2/dart-core/DateTime/DateTime.fromMillisecondsSinceEpoch.html

gnunicorn (changes requested):
Not sure how you want to go about with the custom-avatar until that PR has landed... Some other remarks as well.

app/lib/screens/HomeScreens/ChatScreen.dart
gtalha07: I was wondering why google fonts giving me errors. I thought its updated in merged PR

app/lib/common/widget/AppCommon.dart
gtalha07: Possibly as based on my location its 5 hours less.

app/lib/common/widget/ChatOverview.dart
gtalha07: This is outdated, I merged the PR earlier.

app/lib/common/widget/ChatOverview.dart
```diff
@@ -1,12 +1,11 @@
 // ignore_for_file: prefer_const_constructors, avoid_unnecessary_containers
 
-
 import 'package:effektio/common/widget/AppCommon.dart';
 import 'package:effektio/common/widget/customAvatar.dart';
 import 'package:effektio/screens/HomeScreens/ChatScreen.dart';
 import 'package:effektio_flutter_sdk/effektio_flutter_sdk_ffi.dart';
 import 'package:flutter/material.dart';
-
+import 'package:intl/intl.dart';
```
gnunicorn, line 8: don't we also have to add `intl` to the pubspec? Or is that always included?

gnunicorn (approved):
import question.

app/lib/common/widget/ChatOverview.dart
gtalha07: It is already there in pubspec, was added during localization work.

PR #113, opened 2022-05-23
Implement image sending from sender

bitfriend:
I tested image sending.
I confirmed message history appears in chat.
But image is not displayed.
Because the feature that fetches the received image is not implemented on flutter & rust.
Now implementing it.


bitfriend:
Reformatted by rust code

bitfriend:
Reformatted by rust code

native/effektio/Cargo.toml
```diff
@@ -51,6 +52,9 @@ features = [
     "anyhow",
 ]
 
+[dependencies.ruma-common]
+features = ["unstable-msc3552"]
```
gnunicorn, line 56: I appreciate you finding out there is an open msc... but there's also an already implemented version we should be doing instead...

gnunicorn (changes requested):
we don't want to send using the newly, yet to be approved MSC but instead use the existing "attachment" types.

native/effektio/Cargo.toml
bitfriend: Noted. Will check it.

native/effektio/src/api/room.rs
```diff
@@ -194,6 +200,44 @@ impl Room {
...
+            .spawn(async move {
+                let url = OwnedMxcUri::from(uri);
+                let mut info = FileContentInfo::default();
+                info.name = name;
+                info.mimetype = mimetype;
+                info.size = match size {
+                    Some(val) => UInt::new(val),
+                    None => None,
+                };
+                let file = FileContent::plain(url, Some(Box::new(info)));
```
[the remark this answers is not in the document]
bitfriend: Noted. Will check it.

native/effektio/Cargo.toml
bitfriend: Just pushed #0eb3936393570afddc6ba244aaa9ab4ba7809277 (Replace msc3552 with standard on image send)

native/effektio/src/api/room.rs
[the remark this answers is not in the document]
bitfriend: Just pushed #0eb3936393570afddc6ba244aaa9ab4ba7809277 (Replace msc3552 with standard on image send)

native/effektio/src/api/room.rs
```diff
@@ -205,9 +203,11 @@ impl Room {
         &self,
         message: String,
```
gnunicorn, line 204: message isn't used anymore.

```suggestion
```

native/effektio/api.rsh
```diff
@@ -139,7 +139,7 @@ object Conversation {
     /// received over timeline().next()
     fn send_plain_message(text_message: string) -> Future<Result<string>>;
 
-    fn send_image_message(message: string, uri: string, name: Option<string>, mimetype: Option<string>, size: Option<u64>) -> Future<Result<string>>;
+    fn send_image_message(message: string, uri: string, name: string, mimetype: string, size: u32, width: u32, height: u32) -> Future<Result<string>>;
```
gnunicorn, line 142: `message` isn't used anymore.

gnunicorn (approved):
code looks fine, and seems to work, but the UI shows the now-outdated error message still:
Screenshot_1653421761

bitfriend:
The `get_media_content` function of `matrix sdk` returns image binary not file URL.
So I couldn't display image with URL on flutter side.
I implemented custom image message builder using `Image.memory`.

The user credential is needed before encrypted image can be decrypted.
So I implemented `CURRENT_CLIENT` variable globally, it keeps active `matrix client` when user does login.

And on rust side, I implemented the class that keeps image-specific info, excluding standard message info.
It is `ImageDescription`.

I used `string_validator` library to check if image URL is valid on flutter side.
Because the image that user just sent is displayed from local not remote.

bitfriend:
`RUNTIME.block_on` must be replaced with `RUNTIME.spawn`, before we can avoid deadlock when opening `ChatScreen` page.
I replaced it in `bin_data` of `ImageDescription`.
But I didn't replace in `image_description` of `RoomMessage`.
Now I can see bug that image is not displayed if above function replaced.
This is problem that should be fixed as soon.

native/effektio/src/api/client.rs
```diff
@@ -47,6 +47,8 @@ impl std::ops::Deref for Client {
     }
 }
 
+static mut CURRENT_CLIENT: Option<MatrixClient> = None;
```
gnunicorn, line 50: we can't do that, sorry. The client is hold by the surrounding SDK and the code may not assume that there is a global to store the current client. Why? Because the SDK is meant to allow for multi-client support (even if the UI doesn't implement it yet).

native/effektio/src/api/messages.rs
```diff
@@ -33,6 +39,91 @@ impl RoomMessage {
...
+    }
+}
+
+pub struct ImageDescription {
+    _bin_data: Vec<u8>,
+    _name: String,
+    _mimetype: Option<String>,
+    _size: u64,
+    _width: Option<u64>,
+    _height: Option<u64>,
```
gnunicorn, line 96: in rust all fields are private, unless prefixed with `pub`. `_`-prefixes are only used for internal variables, that aren't actually needed. Let's stick to that model and stay with regular names instead.

```suggestion
    bin_data: Vec<u8>,
    name: String,
    mimetype: Option<String>,
    size: u64,
    width: Option<u64>,
    height: Option<u64>,
```

native/effektio/src/api/messages.rs
```diff
@@ -33,6 +39,91 @@ impl RoomMessage {
...
+}
+
+impl ImageDescription {
+    pub async fn bin_data(&self) -> Result<api::FfiBuffer<u8>> {
```
gnunicorn, line 100: this shouldn't copy here but only on request actually call to client and fetch the image and return it.

native/effektio/src/api/messages.rs
```diff
@@ -33,6 +39,91 @@ impl RoomMessage {
...
+                    Ok(description)
+                })
+            }
+            _ => bail!("Invalid file format"),
+        }
+    }
+}
+
+pub struct ImageDescription {
+    _bin_data: Vec<u8>,
```
gnunicorn, line 91: We shouldn't hold this constantly in memory here.

native/effektio/src/api/messages.rs
```diff
@@ -33,6 +39,91 @@ impl RoomMessage {
...
+                RUNTIME.block_on(async move {
+                    let bin_data = client
+                        .get_media_content(
+                            &MediaRequest {
+                                source: content.source.clone(),
+                                format: MediaFormat::File,
+                            },
+                            false,
+                        )
+                        .await?;
```
gnunicorn, line 61: we shouldn't do this here but only upon request of the actual function call.

native/effektio/src/api/messages.rs
```diff
@@ -33,6 +39,91 @@ impl RoomMessage {
...
     }
+
+    pub fn msgtype(&self) -> String {
+        self.inner.content.msgtype().to_string()
+    }
+
+    pub async fn image_description(&self) -> Result<ImageDescription> {
+        match &self.inner.content.msgtype {
+            MessageType::Image(content) => {
+                let client = Client::current_client().unwrap();
```
gnunicorn, line 50: rather than assuming a global, we should put `client` onto the room-message (and the image description), so we can do:

   ```suggestion
                let client = self.client.clone();
```

File: native/effektio/api.rsh [hunk 11]
```diff
@@ -94,6 +94,25 @@ object RoomMessage {
...
     fn origin_server_ts() -> u64;
+
+    fn msgtype() -> string;
+
+    fn image_description() -> Future<Result<ImageDescription>>;
+}
+
+object ImageDescription {
+
+    fn bin_data() -> Future<Result<buffer<u8>>>;
```
gnunicorn, line 105: can we have a better name for this function? How about `image_binary` or something?

native/effektio/api.rsh
```diff
@@ -94,6 +94,25 @@ object RoomMessage {
...
+
+    fn name() -> string;
+
+    fn mimetype() -> Option<string>;
+
+    fn size() -> u64;
+
+    fn width() -> Option<u64>;
+
+    fn height() -> Option<u64>;
```
gnunicorn, line 115: Please add doc comments to all new functions in this file.

gnunicorn (changes requested):
Please revert from using a global client and instead pass a `client: self.client.clone()` when creating messages from the room.

Further comments inline.

native/effektio/src/api/messages.rs
bitfriend: Yes, rust recognize `_` as private property in struct.
By the way, `ImageDescription` must keep the getter functions about these fields.
So I appended the prefix `_` to properties.

native/effektio/src/api/messages.rs
bitfriend: Will move fetching function from message to client?

native/effektio/api.rsh
bitfriend: Great idea

native/effektio/src/api/messages.rs
gnunicorn: Sure, `_`-prefixes are acceptable, but are unnecessary (and make it needlessly harder to read), so I'd prefer we dropped them. Of course that doesn't mean we don't need the getter functions, that's totally okay. Was just talking about style...

native/effektio/src/api/client.rs
bitfriend: Yes, correct.
It doesn't look smart for me too.
But I guess on rust side there is no way to manage the multiple clients and set active to some client.
Alternatively on flutter side there is `currentClient` of `EffektioSdk`.
If we go and back between rust and flutter, we can get delay.
Before flutter can deliver current client to rust, flutter must pass token to rust and rust must login with it, so it takes a bit of time to perform this login, because login is operation via network.
Really I tested it and confirmed delay when opening `ChatScreen` page.
So I just wanted to finish all operations in rust, without go and back between rust and flutter.

Okay.
Will work from `currentClient` of `EffektioSdk` of flutter side.

native/effektio/src/api/messages.rs
bitfriend: Understood.
Will change property name and keep getter function name.

native/effektio/src/api/client.rs
gtalha07: Did you tested it in '''release''' build?

bitfriend: No, I tested in debug build

bitfriend:
1. In case that `_insertMessage` is wrapped into `setState`.
What if the 1st image binary reaches before `initState` completes in `ChatScreen` page?

2. In case that `_insertMessage` is not wrapped into `setState`.
What if the fetching image binary reaches after `initState` completed in `ChatScreen` page?

Without mutex, it is impossible to solve these problems atomically.
So I added `mutex` to dependency and implemented mutex for state updating.

PS:
As you see, now this app is handling the image caching in low level directly.
So mutex and image custom builder was introduced.
If it is possible to use remote URL (that means image is not encrypted), we could remove all these complicated things.
But we are targeting the encrypted images, so we must handle the image caching directly.

bitfriend:
Hi, Ben.
Your commit (https://github.com/effektio/effektio/pull/113/commits/211b4f53de77df15387ef5ade22e285b16c14562) to my branch is working well.
Thank you for great help again.

bitfriend:
I used `path.join` to prepare file download path.
This function depends on `libandroidicu.so` and I included this library into `jniLibs`.
I could get this library for only `x86_64`, not `x86`/`arm64-v8a`/`armeabi-v7a`.

I used `fluttertoast` to show error in case that user clicked the downloaded file and there is no associated with that file.

I moved the image/file reading from message to room.
Because room on rust side keeps client and `ChatScreen.dart` contains room object.
So we can keep `messages.rs` simple as original (I had introduced matrix client to read image/file in `messages.rs`).

gnunicorn:
> I used fluttertoast to show error in case that user clicked the downloaded file and there is no associated with that file.

we've just removed fluttertoast (see #119 ), please use the default Material Snackbar.

> I used path.join to prepare file download path on flutter side.
> This function depends on libandroidicu.so and I included this library into jniLibs.

Let's not. Please use rust's `path`-stuff for anything localised - just add a new method.

bitfriend:
> > I used fluttertoast to show error in case that user clicked the downloaded file and there is no associated with that file.
> 
> we've just removed fluttertoast (see #119 ), please use the default Material Snackbar.
> 
> > I used path.join to prepare file download path on flutter side.
> > This function depends on libandroidicu.so and I included this library into jniLibs.
> 
> Let's not. Please use rust's `path`-stuff for anything localised - just add a new method.

Great idea.
Will do.

bitfriend:
`ubuntu-latest` was upgraded from `20220529` to `20220605`.
Now we are getting the error [unable to find library -lgcc](https://github.com/bbqsrc/cargo-ndk/issues/22).
2 changes made to fix this bug.
1. Downgrade `cargo-ndk`. Replace `cargo install cargo-ndk` with `cargo install cargo-ndk --version 2.6.0` in `Makefile.toml` of root directory.
2. Add this step prior to `cargo make --profile release android-arm`
```
      - name: Copy ndk utilities for cargo make
        run: |
          rmdir -r /usr/local/lib/android/sdk/ndk-bundle/toolchains
          mkdir -p /usr/local/lib/android/sdk/ndk-bundle/toolchains
          cp -R ${{ steps.install-ndk-22.outputs.ndk-path }}/toolchains/llvm /usr/local/lib/android/sdk/ndk-bundle/toolchains
```

native/effektio/api.rsh
```diff
@@ -94,6 +94,45 @@ object RoomMessage {
 
     /// the server receiving timestamp
     fn origin_server_ts() -> u64;
+
+    /// m.audio, m.emote, m.file, m.image, m.location, m.service_notice, m.text, m.video or m.key.verification.request
+    fn msgtype() -> string;
```
gnunicorn, line 99: we might want to consider switching this into an `Enum`-type... but later...

native/effektio/api.rsh
```diff
@@ -94,6 +94,45 @@ object RoomMessage {
...
+
+    /// contains source data, name, mimetype, size, width and height
+    fn image_description() -> Result<ImageDescription>;
+
+    /// contains source data, name, mimetype and size
+    fn file_description() -> Result<FileDescription>;
```
gnunicorn, line 105: What if there is no image/file? I assume this is then `None`? Shouldn't this be `Option` rather than `Result` then?

native/effektio/api.rsh
```diff
@@ -94,6 +94,45 @@ object RoomMessage {
...
+
+object FileDescription {
+
+    /// file name
+    fn name() -> string;
+
+    /// MIME
+    fn mimetype() -> Option<string>;
+
+    /// file size
```
gnunicorn, line 134:
```suggestion
    /// file size in byte
```

native/effektio/api.rsh
```diff
@@ -138,6 +177,19 @@ object Conversation {
     /// returns the event_id as given by the server of the event soon after
     /// received over timeline().next()
     fn send_plain_message(text_message: string) -> Future<Result<string>>;
+
+    fn send_image_message(uri: string, name: string, mimetype: string, size: u32, width: u32, height: u32) -> Future<Result<string>>;
+
+    /// decrypted image file data
+    fn image_binary(event_id: string) -> Future<Result<buffer<u8>>>;
```
gnunicorn, line 184: why is this function not on `image` or `RoomMessage`?

native/effektio/src/api/messages.rs
```diff
@@ -33,14 +35,107 @@ impl RoomMessage {
...
+        match &self.inner.content.msgtype {
+            MessageType::Image(content) => {
+                let info = content.info.as_ref().unwrap();
+                let description = ImageDescription {
+                    img_name: content.body.clone(),
+                    img_mimetype: info.mimetype.clone(),
+                    img_size: match info.size {
+                        Some(value) => u64::from(value),
+                        None => 0,
+                    },
```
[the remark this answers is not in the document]
gnunicorn: But more interestingly... if `size` might be optional... why not just make it optional in the API, too ... same as width and height... size `0` is not the same as an unknown size, after all...

native/effektio/src/api/messages.rs
```diff
@@ -33,14 +35,107 @@ impl RoomMessage {
...
+                    img_height: info.height.map(u64::from),
+                };
+                Ok(description)
+            }
+            _ => bail!("Invalid file format"),
+        }
+    }
+
+    pub fn file_description(&self) -> Result<FileDescription> {
+        match &self.inner.content.msgtype {
```
gnunicorn, line 64: same as above re if-let

native/effektio/src/api/room.rs
```diff
@@ -194,6 +206,190 @@ impl Room {
...
+                                    },
+                                    false,
+                                )
+                                .await?;
+                            file.write_all(&data)?;
+                            let key =
+                                [room.room_id().as_str().as_bytes(), event_id.as_bytes()].concat();
+                            client
+                                .store()
+                                .set_custom_value(&key, path.to_str().unwrap().as_bytes().to_vec())
```
gnunicorn, line 344: please don't `unwrap` in actual code (tests are fine). Instead always use `.expect(MSG)` with a message stating, why we can be sure this never panics. e.g. : 

```suggestion
                                .set_custom_value(&key, path.to_str().expect("Path was generated from strings. Must be string").as_bytes().to_vec())
```

native/effektio/src/api/room.rs
```diff
@@ -194,6 +206,190 @@ impl Room {
...
+                        MessageType::File(content) => {
+                            let key = [
+                                room.room_id().as_str().as_bytes(),
+                                event_id.as_str().as_bytes(),
+                            ]
+                            .concat();
+                            let path = client.store().get_custom_value(&key).await?;
+                            match path {
+                                Some(value) => {
+                                    let text = std::str::from_utf8(&value).unwrap();
```
gnunicorn, line 380: same as above. please don't unwrap. always `expect`.

native/effektio/src/api/room.rs
```diff
@@ -194,6 +206,190 @@ impl Room {
...
+                                    Ok(text.to_owned())
+                                }
+                                None => Ok("".to_owned()),
+                            }
+                        }
+                        _ => Ok("".to_owned()),
+                    },
+                    _ => Ok("".to_owned()),
```
gnunicorn, line 388: Pretty sure, these should all be `None` rather than `Ok` of an empty string.

gnunicorn (approved):
There's a bunch of minor API design and internal style remarks. Please fix at your own peril and merge.

native/effektio/api.rsh
gnunicorn: let's make that into an issue for later, once the PR is merged.

native/effektio/api.rsh
bitfriend: In fact, I was planning to user these functions for only file/image.
Will fix.

native/effektio/api.rsh
```diff
@@ -94,6 +94,45 @@ object RoomMessage {
...
+object ImageDescription {
+
+    /// file name
+    fn name() -> string;
+
+    /// MIME
+    fn mimetype() -> Option<string>;
+
+    /// file size
+    fn size() -> u64;
```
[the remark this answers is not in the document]
bitfriend: Yes, you are correct.

native/effektio/api.rsh
bitfriend: Yes, you are correct

native/effektio/api.rsh
bitfriend: I wanted to abstract image description and file description into a single struct.
But they have different fields, so I couldn't abstract them.

native/effektio/api.rsh
bitfriend: Flutter keeps room as member variable in UI.
Flutter doesn't keep room message as variable, because it may have to load so many messages in screen.
So I wanted to get image binary data from member variable.
Getting temporary variable for each message may be slow.

gnunicorn: yeah, fair enough then. keep it the way it is.

native/effektio/src/api/messages.rs
```diff
@@ -33,14 +35,107 @@ impl RoomMessage {
...
         self.inner.origin_server_ts.as_secs().into()
     }
+
+    pub fn msgtype(&self) -> String {
+        self.inner.content.msgtype().to_string()
+    }
+
+    pub fn image_description(&self) -> Result<ImageDescription> {
+        match &self.inner.content.msgtype {
+            MessageType::Image(content) => {
```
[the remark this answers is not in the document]
bitfriend: Cool

native/effektio/src/api/messages.rs
bitfriend: Cool

native/effektio/src/api/room.rs
bitfriend: Will fix

PR #123, opened 2022-05-29
Like animation for news items

### LINKED ISSUE #37: Like animation for news items
allow easy double tap to like a news item.

sachintha-s:
fix #37: This PR contains like-button animation.
Design: https://www.figma.com/file/Bua5ifrK6H3LFYBXVFPQ5w/eFFEKTIO?node-id=0%3A1


app/lib/blocs/like_animation.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/cupertino.dart';
+
+class LikeAnimation {
```
gtalha07, line 3: I don't think it goes in bloc directory as this animation doesn't uses bloc styles. Rather have an ```animations``` folder in ```common``` where we can put animations of all widgets.

app/lib/common/widget/like_button.dart
```diff
@@ -0,0 +1,176 @@
+import 'package:effektio/blocs/like_animation.dart';
+import 'package:effektio/common/store/separatedThemes.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+
+class LikeButton extends StatefulWidget {
```
gtalha07, line 6: I would rather like to use packages for achieving animations in this regard but its okay if it doesn't fulfill the required one.

gtalha07 (approved):

app/lib/common/widget/like_button.dart
sachintha-s: I didn't find such a package for this animation so I built this.

Issue #119, opened 2022-05-26
Remove Fluttertoast in favor of materials Snackbar


gnunicorn:
let's remove the additional dependency on `fluttertoast` and use Material Design's [`Snackbar`](https://api.flutter.dev/flutter/material/SnackBar-class.html)s instead.

PR #124, opened 2022-05-29
Remove fluttertoast in favor of Material Snackbar

gtalha07:
This PR includes changes as per #119.


gtalha07:
To not pollute git history since main is not pushed with fix #121, closing this PR and will open new after #121 merge

Issue #63, opened 2022-04-24
Newsfeed comment replies


gnunicorn:
![grafik](https://user-images.githubusercontent.com/40496/164989759-191076c4-deb6-4d75-8240-291d9d7d2001.png)
![grafik](https://user-images.githubusercontent.com/40496/164989780-00499f4c-2c69-49e8-ac17-adfcc1122cea.png)
![grafik](https://user-images.githubusercontent.com/40496/164989786-ba341d1d-6d47-4bf4-b851-7d99b9ad3367.png)
![grafik](https://user-images.githubusercontent.com/40496/164989798-0e2407b0-1787-459a-bb55-a95ead89585f.png)

Issue #62, opened 2022-04-24
Comments on News


gnunicorn:
Designs:

![grafik](https://user-images.githubusercontent.com/40496/164989623-a9f8d76b-6c5a-4aeb-85e4-456fb6db7dd0.png) 
![grafik](https://user-images.githubusercontent.com/40496/164989637-5773f857-825a-414f-805f-3639e941c81b.png) 
![grafik](https://user-images.githubusercontent.com/40496/164989652-00a5efc2-8568-4ee2-8790-ebe345b7d18a.png)
![grafik](https://user-images.githubusercontent.com/40496/164989657-662ada45-1e8c-4376-bd36-91a508afda4b.png)
![grafik](https://user-images.githubusercontent.com/40496/164989675-2c9b4fa3-958f-468c-a0ce-70dca20b4493.png)
![grafik](https://user-images.githubusercontent.com/40496/164989685-a8183c6e-c3a7-412f-bfe5-b18d291ac261.png)

PR #126, opened 2022-05-31
News comment UI as per issues #62 and #63

inderjeet-developer7:
// To do
- Adapt mocked data from API when implemented instead of static one's
- Open keyboard on Reply text tap and swipe action


gnunicorn:
@inderjeet-developer7 please fix the remaining [style error `flutter analyze` is complaining about in the CI job](https://github.com/effektio/effektio/runs/6702195351?check_suite_focus=true#step:6:12)

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -115,7 +115,7 @@ class _ChatScreenState extends State<ChatScreen> {
     final index = _messages.indexWhere((element) => element.id == message.id);
     final updatedMessage = _messages[index].copyWith(previewData: previewData);
 
-    WidgetsBinding.instance.addPostFrameCallback((_) {
```
gtalha07, line 118: Please update your flutter SDK to 3.0 as this is not the error and revert this change in new commit. This is deprecated and it isn't required to make ```WidgetBindings``` nullable now.

inderjeet-developer7: okay

PR #131, opened 2022-06-06
fix chat list scrolling, fix text below room title

gtalha07:
Chat list now gets scrolled through expected gesture area. Misplaced string is now corrected in room subtitle.


bitfriend:
Thank you for your quick change

PR #137, opened 2022-06-12
Display text messages in real-time in a room

### LINKED ISSUE #130: Incoming text messages not showing in real-time
Chat should update messages in real-time in a room.

gtalha07:
This PR fixes the automatic updation of chat room whenever new message arrives rather than doing manually.


gtalha07:
PR #142 is handling both text and file messages updation so this is useless now. Closing

PR #143, opened 2022-06-17
extending getX use to login/signup, remove bloc dependency

gtalha07:
PR includes following changes:
1. Add controllers for handling authentication/registration UI.
2. Remove bloc depedency and bloc files.
3. Remove splash screen UI which was using bloc.


bitfriend (approved):
Great work.
BTW too many line wrappings may make developer's thinking distributed.
Improved lint rule may be better for code readability.

gtalha07:
> Great work. BTW too many line wrappings may make developer's thinking distributed. Improved lint rule may be better for code readability.

Yes, It sometimes becomes confusing. It is also because we are using stricter rule of trailing commas which forces line wraps on ending widget parameters too. Although I'm used to it :p

gtalha07:
> Great work. BTW too many line wrappings may make developer's thinking distributed. Improved lint rule may be better for code readability.

We can solve something for it.

PR #146, opened 2022-06-21
Issue127 invitation

bitfriend:
Implemented the fetch of past invitation event

PR #148, opened 2022-06-23
Issue127 invitation active members

bitfriend:
I implemented `get_state_events` for all types, but not found the past invitation from `@bitfriend`


bitfriend:
Fixed by `client.register_event_handler(|ev: SyncRoomMemberEvent|)` of PR #146 .

PR #162, opened 2022-07-13
Cross signing UI-issue76

gtalha07:
PR implements the cross-signing UI which the user will get shown on session verification request.


bitfriend (approved):
Great work about splitting cross signing into separate class.
This class may need another name (ex: Emoji verification), when we implement QR code verification.

gtalha07:
> Great work about splitting cross signing into separate class. This class may need another name (ex: Emoji verification), when we implement QR code verification.

Good Idea, I guess we'll look at that later stage.

PR #163, opened 2022-07-13
Get iOS to start up in the simulator

gnunicorn:
it builds and starts (I even see the logline), but then it crashes. Updated the dependencies and it works:

Screenshot 2022-07-13 at 22 36 06

Simulator Screen Shot - iPhone 12 mini - 2022-07-13 at 22 35 36

Simulator Screen Shot - iPhone 12 mini - 2022-07-13 at 22 35 31


bitfriend (approved):
Great work

bitfriend (approved):

bitfriend (approved):
We have to confirm this change should not interfere with android build

gnunicorn:
will merge as is, even though the tests don't pass yet (which is fine, they are not required), but we need the sdk-upgrade in main...

PR #170, opened 2022-07-15
Typing notification event from peer

bitfriend:


bitfriend:
Now we can get event receiver using `SyncState::getEphemeralEventRx` on flutter side

app/lib/common/widget/CrossSigning.dart
```diff
@@ -5,7 +5,7 @@ import 'dart:async';
 import 'package:effektio/common/store/separatedThemes.dart';
 import 'package:effektio/common/widget/AppCommon.dart';
 import 'package:effektio_flutter_sdk/effektio_flutter_sdk_ffi.dart'
-    show SyncState, CrossSigningEvent, Client, FfiListEmojiUnit;
+    show Client, FfiListEmojiUnit, GenericEvent, SyncState;
```
gtalha07: Using the different class here. Does this handle both emoji and other events? 
Note: Just for my reminder, the PR push can break the implemented cross-signing UI and I need to fix it.

gtalha07:
> @bitfriend I am confused. this seems to contain a lot of unrelated changes. what does the emoji verification have to do with typing notification? please clarify...

Seems like the CrossSigningEvent handler is changed into GenericEvent handler, which has support of listening both cross signing and invitation events?. Please correct me if I'm wrong @bitfriend

bitfriend:
@gtalha07 I guess it is better to change `startCrossSigning` to `install`.
Later we may have to implement `uninstall` function too.

native/cli/Cargo.toml
```diff
@@ -21,6 +21,8 @@ term-table = "1.3.2"
 
 [dependencies.effektio-core]
 path = "../core"
+default-features = false
```
gnunicorn, line 24: Are you sure we want these?

native/effektio-test/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #[cfg(test)]
 mod auth;
-#[cfg(test)]
-mod emoji_verification;
+// #[cfg(test)]
```
gnunicorn, line 3: What's the point of this? Doesn't the test work anymore? Can we please leave is enabled led?

gnunicorn (approved):
Let's fix my small comments a d merge this. But let's rethink that giant sync-callback. I think we probably should switch to type-specific event-handlers instead (after this PR).

native/cli/Cargo.toml
bitfriend: image
I inserted that code to reset database.
I think anyone would need mock data when he wants to reset his local synapse server.
What do you think?

native/effektio-test/src/lib.rs
bitfriend: As I said, when `emoji_verification` test is enabled, `typing_notification` test is not working.
I guess frequent usage of registered accounts of synapse server has some problem.
When performing `typing_notification` test, should I make new account to synapse server?

gnunicorn: That's weird. Let's merge as is, but file a bug - this needs fixing, synapse should be able to deal with both.

native/cli/Cargo.toml
gnunicorn: I don't understand. Why did you disabled default features on this dependency?

bitfriend: Fixed by https://github.com/effektio/effektio/pull/170/commits/5cad8afaf03ceaadbf8e76c35ab1e5c0257cbc73

native/effektio-test/src/lib.rs
bitfriend: Just opened #202 and assigned it to you and me, because it is rust bug.

PR #177, opened 2022-07-20
Issue76 cross signing 2

bitfriend:
We have to implement more events like typing notification or read notification event from peer.
The CrossSigningEvent that was already implemented may not cover its appropriate meaning or feature.
We will rearrange event part of api so that it can hold more events in the future.


bitfriend:
And we should implement the integration test about emoji verification.

.github/workflows/app.yml
```diff
@@ -34,14 +34,18 @@ jobs:
       - uses: Swatinem/rust-cache@v1
 
       - uses: nttld/setup-ndk@v1
+        id: setup-ndk
         with:
           ndk-version: r23c
+          add-to-path: false
```
gnunicorn, line 40: why not adding it to the path?

native/effektio-test/Cargo.toml
```diff
@@ -14,8 +14,8 @@ features = ["with-mocks"]
 
 [dependencies]
 anyhow = "1"
-env_logger = "0.9.0"
+env_logger = "0.8"
```
gnunicorn, line 17: why are you dropping to a lower version?

```suggestion
env_logger = "0.9.0"
```

native/effektio-test/Cargo.toml
```diff
@@ -14,8 +14,8 @@ features = ["with-mocks"]
...
 futures = "0.3.17"
+log = "0.4"
 tempfile = "3.3.0"
-tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
-zenv = "0.8.0"
\ No newline at end of file
+tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
```
gnunicorn, line 21: Missing final newline
```suggestion
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }

```

.github/workflows/app.yml
gnunicorn, line 40: why not adding it to the path?

native/effektio-test/src/emoji_verification.rs
```diff
@@ -0,0 +1,196 @@
...
+    .await?;
+
+    let bob_device_id = bob.device_id().await.expect("bob should get device id");
+    info!("bob device id: {}", bob_device_id);
+    // we have two devices logged in
+
+    // sync both up to ensure they've seen the other device
+    let syncer = alice.start_sync();
+    let mut first_synced = syncer.get_first_synced_rx().expect("not yet read");
+    while first_synced.next().await != Some(true) {} // let's wait for it to ha
```
gnunicorn, line 50:
```suggestion
    while first_synced.next().await != Some(true) {} // let's wait for it have synced
```

native/effektio/src/events/emoji_verification.rs
```diff
@@ -0,0 +1,577 @@
...
+use super::RUNTIME;
+
+#[derive(Clone, Debug)]
+pub struct EmojiVerificationEvent {
+    client: Client,
+    event_name: String,
+    txn_id: String,
+    sender: String,
+}
+
```
gnunicorn, line 26: some docs on how to use that would be nice, as this seems to be a thing that quickly becomes hard to follow again... Mind adding some inline docs here, describing a flow similar to the integration tests? Or link to some existing docs (matrix sdk example or whatever describes this nicely).

native/effektio/src/events/mod.rs
```diff
@@ -0,0 +1,13 @@
+use lazy_static::lazy_static;
+use tokio::runtime;
+
+lazy_static! {
+    static ref RUNTIME: runtime::Runtime =
+        runtime::Runtime::new().expect("Can't start Tokio runtime");
+}
```
gnunicorn, line 7: ??? Why are you adding a second runtime here? Please don't do that, use the one from `api.rs`.
```suggestion
```

gnunicorn (approved):
Some minor remarks on style, and a big one: please don't create a second runtime. Once all comments are addressed, this should be good to go though. Merge at your digression.

.github/workflows/app.yml
bitfriend: Before I touched this part, `ANDROID_NDK_HOME` env variable was not working.
So I thought it is reasonable to delete this variable, because it is not working.
And [the tutorial that I followed](https://github.com/Haivision/srt/actions/runs/1312237281/workflow) is using `add-to-path: false` too.
If you want, I will try to delete `add-to-path: false`.

native/effektio-test/Cargo.toml
bitfriend: `android_logger` is based on `env_logger 0.8`.
If we have 2 versions of `env_logger`, library size would be larger.
I wanted to reduce library size.

bitfriend: `android_logger 0.11` is based on `env_logger 0.9`

native/effektio/src/events/emoji_verification.rs
bitfriend: Do you want something like this?
```rust
    /// Get information of all our own devices.
    ///
    /// # Examples
    ///
    /// ```no_run
    /// # use matrix_sdk::{Client, config::SyncSettings};
    /// # use futures::executor::block_on;
    /// # use url::Url;
    /// # use std::convert::TryFrom;
    /// # block_on(async {
    /// # let homeserver = Url::parse("http://localhost:8080")?;
    /// # let mut client = Client::new(homeserver).await?;
    /// let response = client.devices().await?;
    ///
    /// for device in response.devices {
    ///     println!(
    ///         "Device: {} {}",
    ///         device.device_id,
    ///         device.display_name.as_deref().unwrap_or("")
    ///     );
    /// }
    /// # anyhow::Ok(()) });
    /// ```
    pub async fn devices(&self) -> HttpResult<get_devices::v3::Response> {
        let request = get_devices::v3::Request::new();

        self.send(request, None).await
    }
```

native/effektio/src/events/mod.rs
bitfriend: Fixed.
https://github.com/effektio/effektio/pull/177/commits/a399579d28b8e10df1d8f1af4e438c173d57e425

.github/workflows/app.yml
gnunicorn: the example you are linking at has an error (as github says), but if it works, fine with me. was just wondering if that had a special reason... as it means we don't have `adb` and stuff in `$PATH` and that makes other things harder usually..

native/effektio-test/Cargo.toml
gnunicorn: yeah, we should upgrade `android_logger` in that case. ugprading should always be the first choice.

native/effektio/src/events/emoji_verification.rs
gnunicorn: yes, but you say the SDK already has those? Then refer to that in a doc-comment. Just having that without any docs feels bad ...

.github/workflows/app.yml
bitfriend: When ndk r23 is updated to fix this error in the future and someone is about to remove this limit, he could notice env variable problem with seeing `add-to-path: false` and `ANDROID_NDK_HOME`.

native/effektio-test/Cargo.toml
bitfriend: In fact, I thought version change of dependency may affect the stability of library.
I just tested `android_logger 0.11` and confirmed it is working well.
https://github.com/effektio/effektio/pull/177/commits/3505714903c9d3698faf299b3d0154f6ce8e479a

.github/workflows/app.yml
bitfriend: Explained in https://github.com/effektio/effektio/pull/177#discussion_r930449598

native/effektio-test/src/emoji_verification.rs
bitfriend: Fixed

native/effektio/src/events/emoji_verification.rs
bitfriend: Is that okay to clone code snippet from integration test here?

gnunicorn: the integration tests aren't complaining about the emoji-verfication but about permission problems (addressed in #174 ). so just merge, ignoring that test for now. all non-required tests aren't in a state to be enforced yet.

gnunicorn (approved):
lgtm

bitfriend:
Bob starts verification flow from `m.key.verification.request`.
Alice starts verification flow from `m.key.verification.ready`.
Whether to verify this device or other device will be based on where to start verification flow.
I introduced `verifyingThisDev` field into `VerifEvent` class.
Based on `verifyingThisDev`, dialog content will be differentiated.
For example:
verifyingThisDev == true -> Title will be `Verify this device`.

bitfriend:
This solution has limitation on flutter side.
It is `Get.back()`.
For example, suppose when verification flow goes from `m.key.verification.accept` to `m.key.verification.key`.
Accept dialog has only "Waiting for" text and it has no any button for next step.
It can move forward only when `m.key.verification.key` event arrives.
`m.key.verification.key` uses `Get.back()` to close this "Waiting for" dialog, because we are based on `GetX` library.
The problem is `Get.back()` is for multi purposes.
`Get.back()` is used to close all popup or dialogs.
Thus frontend developer must not call `Get.back()` to close non-verification dialog during verification flow.

We will keep `GetX` library or find alternative library?
In this case, `GetX` has disadvantage that `Get.back()` cannot close the specified dialog and it can close only topmost dialog.
We are not sure the frontend developer would not call `Get.back()` to close non-verification dialog during verification flow.

gtalha07:
> This solution has limitation on flutter side. It is `Get.back()`. For example, suppose when verification flow goes from `m.key.verification.accept` to `m.key.verification.key`. Accept dialog has only "Waiting for" text and it has no any button for next step. It can move forward only when `m.key.verification.key` event arrives. `m.key.verification.key` uses `Get.back()` to close this "Waiting for" dialog, because we are based on `GetX` library. The problem is `Get.back()` is for multi purposes. `Get.back()` is used to close all popup or dialogs. Thus frontend developer must not call `Get.back()` to close non-verification dialog during verification flow.
> 
> We will keep `GetX` library or find alternative library? In this case, `GetX` has disadvantage that `Get.back()` cannot close the specified dialog and it can close only topmost dialog. We are not sure the frontend developer would not call `Get.back()` to close non-verification dialog during verification flow.

How about we just don't show 'waiting for @username' dialog which appears only for short second at app, and only show emoji verification screen after it. Is it possible? We dont need bottom sheet there and could it solve the issue?

gtalha07:
Overall it looks much better than previous version. Great! You can merge your work. We can deal with UI seperately. The functionality is solid.

bitfriend:
> > This solution has limitation on flutter side. It is `Get.back()`. For example, suppose when verification flow goes from `m.key.verification.accept` to `m.key.verification.key`. Accept dialog has only "Waiting for" text and it has no any button for next step. It can move forward only when `m.key.verification.key` event arrives. `m.key.verification.key` uses `Get.back()` to close this "Waiting for" dialog, because we are based on `GetX` library. The problem is `Get.back()` is for multi purposes. `Get.back()` is used to close all popup or dialogs. Thus frontend developer must not call `Get.back()` to close non-verification dialog during verification flow.
> > We will keep `GetX` library or find alternative library? In this case, `GetX` has disadvantage that `Get.back()` cannot close the specified dialog and it can close only topmost dialog. We are not sure the frontend developer would not call `Get.back()` to close non-verification dialog during verification flow.
> 
> How about we just don't show 'waiting for @username' dialog which appears only for short second at app, and only show emoji verification screen after it. Is it possible? We dont need bottom sheet there and could it solve the issue?

Of course, it is possible.
In fact, I referenced **element app** during implementation of emoji verification.
So I thought it is reasonable to show what this user is waiting for.
If not needed, I will remove it.

bitfriend:
> Overall it looks much better than previous version. Great! You can merge your work. We can deal with UI seperately. The functionality is solid.

Yes, this feature was tested thoroughly on rust side.
It took a long time to implement flutter UI because of complicated logic.

bitfriend:
The following is the finished demo.

This is verification on bob side.
Element app requested verification of emulator to our app.

https://user-images.githubusercontent.com/48786056/183156235-7b84890e-a381-4280-9fe8-7ac817a5b292.mp4

This is verification on alice side.
Our app requested verification of emulator to Element app.

https://user-images.githubusercontent.com/48786056/183156304-739d1bc9-6358-4773-93bb-0407d7f96542.mp4

Issue #134, opened 2022-06-09
Support markup in ChatList page


bitfriend:
`ChatScreen` is main text editing page so it supports markup.
But now `ChatList` page doesn't support markup.
![image](https://user-images.githubusercontent.com/48786056/172851211-291845c8-ec0d-41bd-b6bc-36fbcce27e83.png)

gtalha07:
We could probably use this: https://pub.dev/packages/markup_text

bitfriend:
Great

gtalha07:
Closing per #191

PR #191, opened 2022-08-02
Markdown support for chat list latest messages.

gtalha07:
PR for #134.


bitfriend (approved):
Great work.
I think it would support more markdown features using pattern.

gtalha07:
> Great work. I think it would support more markdown features using pattern.

Yes, for more features, we just need regex patterns. Although ```flutter_chat_ui``` only supports some of them.

PR #197, opened 2022-08-05
Fix issue in library path on android api 23

bitfriend:
`dlopen` fails on android api 23.
https://github.com/effektio/effektio/issues/122#issuecomment-1206237934


gnunicorn (approved):
looks good.

Is this something we can test in the CI somehow? This only shows up when trying to run the App, right?

bitfriend:
This error was runtime error.
It can be displayed during runtime.
image

PR #203, opened 2022-08-07
Issue173 read notification from peer

bitfriend:
This is solution about https://github.com/effektio/effektio/issues/173


bitfriend:
Sometimes getting this error.
image
Because receipt test is interfered by typing test.
The simplest solution is to turn off typing test when focusing on receipt test.
`native\effektio-test\src\lib.rs`

gnunicorn (approved):
lgtm

PR #205, opened 2022-08-08
issue 190 fixed

### LINKED ISSUE #190: Unnecessarily Chat widget update when opening and closing keyboard.
While typing in text field of chat room, the Chat widget re-updates on toggling and closing keyboard in input textfield.

sachintha-s:
#190 
Now the chat screen won't rebuild (see invite widget and room avatar)

https://user-images.githubusercontent.com/65540448/183393457-eb31606f-8a49-4686-a6fb-a8eb4725b5e4.mp4

but you can see that the chat avatars are refreshing, I went through the **flutter_chat_ui** package and found the problem.
The problem is that the flutter_chat_ui package uses media queries, But I think this is not a big problem.

image


bitfriend:
Great work for usage of `LayoutBuilder`.

bitfriend (approved):
Great

bitfriend:
That code is working well on rotation?

sachintha-s:
> That code is working well on rotation?

Yes, but in landscape mode, the keyboard takes up about 80% of the screen height, so we can experience the bottom overflow problem.

image

I think we should disappear the app bar when the app is in landscape mode to solve this problem. Or do you have other ideas?

gtalha07:
> > That code is working well on rotation?
> 
> Yes, but in landscape mode, the keyboard takes up about 70% of the screen height, so we can experience the bottom overflow problem.
> 
> image
> 
> I think we should disappear the app bar when the app is in landscape mode to solve this problem. Or do you have other ideas?

It seems like the keyboard is taking too much height on landscape, and with this overflow margin increases when invite widget is shown. With the current constraints of keyboard, nothing is visible on the chat screen. Might be adjusting keyboard height or making it smaller when orientation changes?

gtalha07 (changes requested):
I think the only way of avoiding overflow is to check orientation at build time and set ```resizeToAvoidBottomInset``` to false in ```Scaffold``` if landscape mode. However this will cause keyboard to overlay on input textfield which is fine I guess. What do you think @sachintha-s ?

sachintha-s:
> I think the only way of avoiding overflow is to check orientation at build time and set `resizeToAvoidBottomInset` to false in `Scaffold` if landscape mode. However this will cause keyboard to overlay on input textfield which is fine I guess. What do you think @sachintha-s ?

@gtalha07 Since we can't resize the keyboard, I think your way might be the solution. But with this, the text field is not visible when the user types something.  However, I will update the code as you mentioned.

gtalha07 (approved):

PR #206, opened 2022-08-08
fix chat room user avatar

gtalha07:
Fixes the chat room avatars showing as room avatar on all recieved messages.

### DIFF
--- a/app/lib/controllers/chat_controller.dart
+++ b/app/lib/controllers/chat_controller.dart
@@ -323,7 +323,10 @@ class ChatController extends GetxController {
       ImageDescription? description = message.imageDescription();
       if (description != null) {
         types.ImageMessage m = types.ImageMessage(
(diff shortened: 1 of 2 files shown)


app/lib/controllers/chat_controller.dart
```diff
@@ -303,7 +303,10 @@ class ChatController extends GetxController {
       FileDescription? description = message.fileDescription();
       if (description != null) {
         types.FileMessage m = types.FileMessage(
-          author: types.User(id: message.sender()),
+          author: types.User(
+            id: message.sender(),
+            firstName: getNameFromId(message.sender()),
+          ),
```
sachintha-s, line 309: I think we should do the same for pictures. Because when sending a picture, we don't see the avatar.

sachintha-s (approved):
Great

bitfriend (approved):
Great

PR #210, opened 2022-08-11
Refactor sync_with_callback with register_event_handler

bitfriend:
Will refactor cross-signing as 1st stage of sync_with_callback.


app/lib/common/widget/CrossSigning.dart
```diff
@@ -22,19 +22,19 @@ class VerifEvent {
 }
 
 class CrossSigning {
```
gtalha07, line 24: Please update the UI code here as from main. :)

gtalha07 (approved):
LGTM.

gnunicorn (approved):
only looked over super superficially.

PR #211, opened 2022-08-11
Improve code about user avatar

### LINKED ISSUE #207: Fetch user avatars images from rust.
We would need to fetch user avatars from backend to render on flutter chat ui via the ```RoomMessage``` API.

bitfriend:
This PR fixes #207.

`UserId` can be accessible on rust side, not flutter side.
So I replaced `UserId` with `String` in parameter of `room.get_member()`, because `String` is accessible from flutter side.
And implemented sample code about chat avatar.


app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -118,6 +119,29 @@ class _ChatScreenState extends State<ChatScreen> {
     );
   }
 
+  Future<FfiBufferUint8> _userAvatar(String userId) async {
+    final member = await widget.room.getMember(userId);
+    return member.avatar();
+  }
+
+  Widget _avatarBuilder(String userId) {
```
gtalha07, line 127: I think I have removed this as we don't need this custom method to apply avatars. The package already supports it and also have algorithm to change user avatar background colors based on ```userID```.  Or this is something we needed here?

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -118,6 +119,29 @@ class _ChatScreenState extends State<ChatScreen> {
...
+    return GetBuilder<ChatController>(
+      id: 'Avatar',
+      builder: (ChatController controller) {
+        return Padding(
+          padding: const EdgeInsets.only(right: 10),
+          child: CustomAvatar(
+            avatar: _userAvatar(userId),
+            displayName: null,
+            radius: 15,
```
gtalha07, line 136: Also the image avatar radius appear smaller and text avatar appears bigger. Is there a way to make it even to image avatar size?

app/lib/screens/HomeScreens/ChatScreen.dart
bitfriend: This function is needed to get avatar from user id.

app/lib/screens/HomeScreens/ChatScreen.dart
bitfriend: What if this code should work on desktop or web, not just mobile?

app/lib/screens/HomeScreens/ChatScreen.dart
gtalha07: Then its okay to use. Another problem with this is that text avatars appear bigger than usual.

app/lib/screens/HomeScreens/ChatScreen.dart
gtalha07: On mobile, it looks weirdly. Like images avatars appear smaller and text one bigger. Not a good looking conversation UI. :)

gtalha07: Anyways we should do something about text avatars size. They should be even depending upon platforms.

bitfriend: Is it impossible to resize image on flutter side?

gnunicorn (approved):

app/lib/screens/HomeScreens/ChatScreen.dart
bitfriend: Of course, I can add `width` & `height` parameters to `Member::avatar()` function.
But I think it is efficient to resize with Image class method on flutter.
Is it impossible to resize with Image class method on flutter?
We have to think about web or desktop too.

gtalha07 (approved):

PR #217, opened 2022-08-19
Issue130 realtime chat message updates.

gtalha07:
This PR includes fixes to real-time message updates both on Chat List and Chat Screen.

### DIFF
--- a/app/lib/common/widget/ChatOverview.dart
+++ b/app/lib/common/widget/ChatOverview.dart
@@ -24,24 +24,20 @@ class ChatOverview extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return ListView.builder(
+    return ListView.separated(
       physics: NeverScrollableScrollPhysics(),
       shrinkWrap: true,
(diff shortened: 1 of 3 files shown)


bitfriend (approved):
Great work

sachintha-s:
Great, **One improvement:** when the app receives a message it should move to the top of the chat list with respect to the time when the message was sent.

gtalha07:
> Great, **One improvement:** when the app receives a message it should move to the top of the chat list with respect to the time when the message was sent.

I think we should implement this sorting and comparing date time objects in seperate PR, what do you think? @sachintha-s

sachintha-s:
> > Great, **One improvement:** when the app receives a message it should move to the top of the chat list with respect to the time when the message was sent.
> 
> I think we should implement this sorting and comparing date time objects in seperate PR, what do you think? @sachintha-s

Ok, Let's do it in separate PR. now I will approve this.

sachintha-s (approved):

PR #222, opened 2022-08-22
issue136_Typing-Indicator

gtalha07:


app/lib/main.dart
```diff
@@ -129,14 +128,16 @@ class _EffektioHomeState extends State<EffektioHome>
     //Start listening for cross signing events
     crossSigning.installDeviceChangedEvent(dlc.getChangedEventRx()!);
     crossSigning.installSessionVerificationEvent(svc.getEventRx()!);
-    tnc = await client.getTypingNotificationController();
-    tnc.getEventRx()!.listen((event) {
+    TypingNotificationController tnc =
```
gtalha07, line 131: Is it really necessary to initialize controller here? As its usage only comprehends to the chat. It would seem more sensible to have it when chat controller initializes or when we are at the chatlist screen (if we want to show typing event at the the place of latest message). If its for testing and not final, then you can ignore this.

app/lib/screens/HomeScreens/ChatScreen.dart
```diff
@@ -25,10 +31,15 @@ import 'package:themed/themed.dart';
 import 'package:transparent_image/transparent_image.dart';
 
 class ChatScreen extends StatefulWidget {
+  final Client client;
```
gtalha07, line 34: I think another approach which could seem more handy would be to pass typing controller instead of whole client object. Although that's pretty subjective as we might need client eventually when more features are added.

gtalha07:
Discarded as outdated. Will open new PR!

PR #226, opened 2022-08-30
Fix logging

gnunicorn:
on desktop and android. See:

grafik


bitfriend (approved):
Now we can see rust log in flutter console too.
Great work.

PR #237, opened 2022-09-17
Mistakes corrected

HatimWayne:
I have tried to correct all the mistakes I could spot, from regular translation errors, to confusion between German and Dutch, hope this helps!


gnunicorn (commented):
Thanks for the fixes!

It didn't catch it all, still requires some review and overall refinement, but it def. catches the biggest blunders ;)

PR #246, opened 2022-09-20
State handler chat list

gtalha07:
**Description**
1. To avoid frequent recallings of ```FutureBuilder```, async calls have been put in ```initState()``` so it should only fetch once.
2. Added ```ChatListController``` to handle latest messages state.
3. Some minor improvements.


gtalha07:
Discarded in favor of PR #248

PR #248, opened 2022-09-21
Issue230 status info for room message

bitfriend:
**Description**

1. To avoid frequent recallings of FutureBuilder, async calls have been put in initState() so it should only fetch once.
2. Added ChatListController to handle latest messages state.
3. Some minor improvements.


gtalha07:
Discarded and will make new PR.

app/lib/controllers/chat_room_controller.dart
```diff
@@ -75,10 +113,10 @@ class ChatRoomController extends GetxController {
   Future<void> handleSendPressed(String message) async {
     // image or video is sent automatically
     // user will click "send" button explicitly for text only
-    await room.typingNotice(false);
-    var eventId = await room.sendPlainMessage(message);
+    await _room!.typingNotice(false);
```
gtalha07, line 116: Don't think we need this now. As typing events are handled with stream.

app/lib/controllers/chat_room_controller.dart
```diff
@@ -341,24 +387,12 @@ class ChatRoomController extends GetxController {
...
     isSendButtonVisible = textEditingController.text.trim().isNotEmpty;
     update();
   }
 
-  @override
-  void onClose() {
-    super.onClose();
-    textEditingController.dispose();
-    focusNode.removeListener(() {});
+  Future<bool> typingNotice(bool typing) async {
```
gtalha07, line 395: Here too.

app/lib/controllers/chat_room_controller.dart
```diff
@@ -22,38 +25,73 @@ import 'package:path_provider/path_provider.dart';
 import 'package:permission_handler/permission_handler.dart';
 
 class ChatRoomController extends GetxController {
-  List<types.Message> messages = [];
+  Client client;
+  String userId;
+  RxList<types.Message> messages = RxList([]);
```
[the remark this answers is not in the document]
bitfriend: Ah, I used it due to bug that initial loading of list missed some items.
It is not effective now, so I will revert `RxList` to `List`.

app/lib/controllers/chat_room_controller.dart
bitfriend: Yeah, it seemed to be strange for me too.
I wanted to keep original logic as much as possible.
If you don't need that code, I will remove it.

gtalha07: Yes, please remove this.

gtalha07: This code sends the typing notification back to server. But we don't need this now. The indicator will disappear automatically if the list is empty. Also this applies to chatlist.

gtalha07: The list gets empty after some duration (~4 sec). The code is in typing indicator PR.

app/lib/controllers/chat_room_controller.dart
bitfriend: `typingNotice()` is for sending notification about my typing.
We have to implement sending of typing event also, so that peer can know if I am typing.

PR #252, opened 2022-09-27
Remove twice deallocation of string

bitfriend:
Applied https://github.com/effektio/ffi-gen/pull/9.
When it is built, the following code is removed from `effektio_flutter_sdk_ffi.dart`.
```dart
    if (parts.capacity > 0) {
      final ffi.Pointer<ffi.Void> tmp2_0;
      tmp2_0 = ffi.Pointer.fromAddress(parts.addr);
      _api.__deallocate(tmp2_0, parts.capacity * 1, 1);
    }
```
This code used to deallocate string twice and it caused error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1023,7 +1023,7 @@ dependencies = [
 [[package]]
 name = "ffi-gen"
 version = "0.1.13"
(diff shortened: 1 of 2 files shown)


gnunicorn (approved):

bitfriend:
Solved https://github.com/effektio/effektio/issues/251

PR #257, opened 2022-09-28
Issue136 typing indicator feature

gtalha07:
### Description:
This PR features typing indicator addition in the chat screen and chat list in textual form and animated widget.
Screenshot_1665161575


bitfriend (approved):
Great work.
Actually I replaced `String?userId` with `String userId` because `client.userId()` is async no more.
I found you reverted it.
And I think it is better to not use `// ignore` directive in flutter code.
I will fix them by myself later.

bitfriend:
It is not required to place room/list controller as member.
Especially it may occur conflict on initState(), because room controller is initializing but list controller may be instantiated.
We can get it temporarily to use in event.
image

bitfriend:
Typing animation is displayed if emulator height is large.
But it is not displayed in Pixel 2 screen size.
I guess they have to be modified.
image
I think current layout is not flexible.
It may be required to change all occurrences of `MediaQuery.of(context).size.height`.

PR #276, opened 2022-10-14
Issue198 cleanup code periodically

bitfriend:


sachintha-s (commented):
@bitfriend  What is the reason for removing the private variables?

app/lib/main.dart
```diff
@@ -89,31 +89,28 @@ class EffektioHome extends StatefulWidget {
...
 
   @override
   void dispose() {
-    super.dispose();
-    _crossSigning?.dispose();
+    crossSigning?.dispose();
     Get.delete<ChatListController>();
     Get.delete<ChatRoomController>();
     Get.delete<ReceiptController>();
+    super.dispose();
```
gtalha07, line 113: It is indeed correct way to order. So it can dispose all resources before calling widget lifecycle end state.

app/lib/widgets/ChatListItem.dart
```diff
@@ -99,91 +99,82 @@ class _ChatListItemState extends State<ChatListItem> {
...
-          : Text(
-              _multiUserTextBuilder(widget.typingUsers),
-              style: ChatTheme01.latestChatStyle.copyWith(
-                fontStyle: FontStyle.italic,
-              ),
-            );
+      if (widget.typingUsers.isEmpty) {
+        return const SizedBox();
+      }
+      return Text(
```
gtalha07, line 105: Won't this run regardless of ```if``` above is true or false? I suppose the conditional operator removed works like ```if```` ````else```.

app/lib/widgets/CrossSigning.dart
```diff
@@ -734,61 +739,74 @@ class CrossSigning {
...
+    );
+  }
+
+  Widget _buildBodyOnKey(
+    BuildContext context,
+    VerificationEvent event,
+    Function setState,
+  ) {
+    if (waitForMatch) {
+      String waitingFor = sprintf(
```
gtalha07, line 753: Is the string interpolation necessary here?

app/lib/widgets/CustomAvatar.dart
```diff
@@ -35,61 +35,53 @@ class CustomAvatar extends StatelessWidget {
...
-          } else {
-            if (isGroup) {
-              return CircleAvatar(
-                radius: radius,
-                backgroundColor: Colors.grey[700],
-                child: SvgPicture.asset('assets/images/people.svg'),
+        }
+        if (snapshot.hasData && snapshot.requireData.isNotEmpty) {
+          return CircleAvatar(
+            backgroundImage: MemoryImage(
```
gtalha07, line 41: This should be replaced with ```CachedMemoryImage```.

app/lib/widgets/InviteInfoWidget.dart
```diff
@@ -30,61 +30,77 @@ class InviteInfoWidget extends StatelessWidget {
...
+    );
+  }
+
+  Widget _buildAcceptButton(BuildContext context) {
+    return SizedBox(
+      width: MediaQuery.of(context).size.width * 0.46,
+      child: elevatedButton(
+        AppLocalizations.of(context)!.accept,
+        AppCommonTheme.greenButtonColor,
+        () async {
```
gtalha07, line 83: We could remove braces and convert it into single expression function here.

app/lib/widgets/InviteInfoWidget.dart
```diff
@@ -30,61 +30,77 @@ class InviteInfoWidget extends StatelessWidget {
...
+    );
+  }
+
+  Widget _buildRejectButton(BuildContext context) {
+    return SizedBox(
+      width: MediaQuery.of(context).size.width * 0.46,
+      child: elevatedButton(
+        AppLocalizations.of(context)!.decline,
+        AppCommonTheme.primaryColor,
+        () async {
```
gtalha07, line 98: Same as above.

app/lib/widgets/NewsItem.dart
```diff
@@ -111,4 +71,54 @@ class NewsItem extends StatelessWidget {
       ],
     );
   }
+
+  Widget? _buildImage() {
+    var image = news.image();
+    if (image == null) {
+      return null;
+    }
+    return Image.memory(Uint8List.fromList(image), fit: BoxFit.cover);
```
gtalha07, line 80: Also could use ```CachedMemoryImage``` here.

app/lib/widgets/ToDoTaskItem.dart
```diff
@@ -49,11 +49,9 @@ class _ToDoTaskItemState extends State<ToDoTaskItem> {
   @override
   void initState() {
     super.initState();
-    _avatars = _getMockAvatars(countPeople);
+    avatars = getMockAvatars(countPeople);
     if (widget.dateTime.contains('All Day')) {
-      setState(() {
```
gtalha07, line 54: How are you updating the state here by removing ```setState(()){}```

app/lib/widgets/TypeIndicator.dart
```diff
@@ -56,189 +56,192 @@ class TypeIndicator extends StatefulWidget {
 class _TypeIndicatorState extends State<TypeIndicator>
     with TickerProviderStateMixin {
   late double stackingWidth;
-  late AnimationController _appearanceController;
-  late AnimationController _animatedCirclesController;
-  late Animation<double> _indicatorSpaceAnimation;
-  late Animation<Offset> _firstCircleOffsetAnimation;
-  late Animation<Offset> _secondCircleOffsetAnimation;
-  late Animation<Offset> _thirdCircleOffsetAnimation;
+  late AnimationController appearanceController;
```
gtalha07, line 59: Imo this should be declared private. We would want restricted access here.

gtalha07 (commented):
Please check my comments for feedback. Overall LGTM!

bitfriend:
> @bitfriend What is the reason for replacing the private variables?

```dart
class EffektioHome extends StatefulWidget {
  const EffektioHome({Key? key}) : super(key: key);

  @override
  _EffektioHomeState createState() => _EffektioHomeState();
}

class _EffektioHomeState extends State<EffektioHome>
    with SingleTickerProviderStateMixin {
  late Future<Client> client;
  int tabIndex = 0;
  late TabController tabController;
  CrossSigning? crossSigning;
}
```
As we know, `_EffektioHomeState` is used for `EffektioHome`.
So it doesn't make sense to make member variables of `_EffektioHomeState` as private.
But the following class is used in any class, so its member variables must be as private.
```dart
class ChatListController extends GetxController {
  Client client;
  List<JoinedRoom> joinedRooms = [];
  List<Invitation> invitations = [];
  bool initialLoaded = false;
  StreamSubscription<FfiListConversation>? _convosSubscription;
  StreamSubscription<FfiListInvitation>? _invitesSubscription;
  StreamSubscription<TypingEvent>? _typingSubscription;

  ChatListController({required this.client}) : super();
}
```
I think it makes sense to use the public variable in appropriate location.

app/lib/main.dart
bitfriend: Yeah, I met debugging error there so I fixed it.

app/lib/widgets/CrossSigning.dart
bitfriend: Fixed it

app/lib/widgets/CustomAvatar.dart
bitfriend: Fixed it, but it seems to be not cached.
You can run app to confirm it.

app/lib/widgets/InviteInfoWidget.dart
bitfriend: Do you want this code?
```dart
  Widget _buildAcceptButton(BuildContext context) {
    return SizedBox(
      width: MediaQuery.of(context).size.width * 0.46,
      child: elevatedButton(
        AppLocalizations.of(context)!.accept,
        AppCommonTheme.greenButtonColor,
        () async => await client.acceptInvitation(groupId),
        AppCommonTheme.appBarTitleStyle.copyWith(
          fontSize: 14,
          fontWeight: FontWeight.w500,
        ),
      ),
    );
  }
```
I guess it looks strange somewhat.

app/lib/widgets/InviteInfoWidget.dart
bitfriend: This code looks strange somewhat.
`() async => await client.acceptInvitation(groupId)`

app/lib/widgets/NewsItem.dart
bitfriend: Fixed it

app/lib/widgets/ToDoTaskItem.dart
bitfriend: Please read this blog.
https://levelup.gitconnected.com/6-easy-tips-when-working-with-setstate-in-a-flutter-application-3d629bc77a16

Tip 3: Don’t call setState in initState methods

`initState` will trigger a rebuild after completion so it’s not necessary to call `setState` inside this method. This method is intended to initialize state-relevant properties like setting default values or subscribing to streams. Don’t do anything else here!

app/lib/widgets/TypeIndicator.dart
bitfriend: Please read this comment.
https://github.com/effektio/effektio/pull/276#issuecomment-1279790599

app/lib/widgets/InviteInfoWidget.dart
gtalha07: It just gives more readability ;)

bitfriend:
Also Ben said about underlined variable in flutter code before.
He mentioned frequent usage of underlined variables may cause confusing or decrease readability of flutter code.
If we use public variable in appropriate location, we can increase readability of flutter code.

app/lib/widgets/CustomAvatar.dart
gtalha07: Yeah caching doesn't seems to be working properly with ```CachedMemoryImage``` too. In 23 API env this bug is not there, only occurs in 32+ above APIs.  We have to dig deeper into this.

app/lib/widgets/InviteInfoWidget.dart
gtalha07: Looks fine to me, also its recommendation from flutter.

app/lib/widgets/NewsItem.dart
gtalha07: Need to find alternative solutions for caching.

app/lib/widgets/CustomAvatar.dart
bitfriend: No, I tested on api 23 and found caching is not working.

gtalha07: Strange, I didn't found issues. Anyways, the bug is confirmed and also causes native crashes. I will open ticket for this.

bitfriend: My side had no native crash.

Issue #212, opened 2022-08-12
Mention ```@``` and highlight message support in chat.


gtalha07:
Chat should have the support of mentioning someone via ```@``` and highlighting the message. A regex pattern would be required and suitable for it. 

Note: Please refer to [effektio docs](https://docs.effektio.org/api/main/dart-sdk/effektio/Client-class.html) for relevant APIs and file an issue if needs implementing or missing.

sachintha-s:
fixed #277

PR #277, opened 2022-10-16
mentioning room members

sachintha-s:
**This PR includes mentioning room members and showing mentioned members.**

Untitled
#212


app/pubspec.yaml
```diff
@@ -70,6 +70,8 @@ dependencies:
   get_time_ago: ^1.1.5
   implicitly_animated_reorderable_list: ^0.4.2
   cached_memory_image: ^1.3.1
+  flutter_mentions: ^2.0.1
+  flutter_html: ^3.0.0-alpha.5
```
gtalha07, line 74: @gnunicorn @sachintha-s  Okay I've checked their compatibility issues with ```flutter_svg```. Apparently the dev is still using old version library of ```flutter_svg``` in the package which kinda sucks despite of many community requests. We can go with it but its alpha stage and many features missing and its not recommended for production. The workaround is to downgrade ```flutter_svg``` which also I would not recommend doing it.

gtalha07: Just for heads-up as we will upgrade this soon as they add latest version of ```flutter_svg```.

gtalha07:
Screenshot_1666010239
We need a bit of state handling here. It shows URL instead of markup text on sending message from effektio. But if I re-open the chat, it shows correctly.

sachintha-s:
> Thanks for this @sachintha-s. But we would need this mentioned markup display in chat list item too. I suppose you planned to implement this there too?

I think it is already shown as it is in the Figma design.
image
What should be improved? Color?

gtalha07:
> > Thanks for this @sachintha-s. But we would need this mentioned markup display in chat list item too. I suppose you planned to implement this there too?
> 
> I think it is already shown as it is in the Figma design. image What should be improved? Color?

Could be color with highlighted text OR something which is more prominent.

sachintha-s:
> Also I would like mention display a bit more appealing. Is the mention widget customizable?

Yes, it is possible, if you have any design ideas please mention them.

gnunicorn:
maybe @kokodavid has some ideas re design.

bitfriend (approved):
Looks good

app/pubspec.yaml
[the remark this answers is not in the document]
gtalha07: Already discussed [here](https://github.com/effektio/effektio/pull/277#discussion_r996994677)

kokodavid:
Let me look into how we can do some designs on that.

gtalha07:
> @gtalha07 should we be merging this as is for now and put up tickets for the remaining things, like a prettier design, to do for later? What do you think?

Yes, @sachintha-s will get this merged after applying review fixes.

PR #294, opened 2022-10-21
Issue198 cleanup code periodically

bitfriend:


app/lib/screens/HomeScreens/chat/ChatProfile.dart
```diff
@@ -27,10 +27,19 @@ class ChatProfileScreen extends StatefulWidget {
...
-  String chatName = '';
+  String? chatName;
   String chatDesc =
       'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam nec aliquam ex. Nam bibendum scelerisque placerat.';
 
+  @override
+  void initState() {
+    super.initState();
+
+    widget.room.displayName().then((value) {
```
gtalha07, line 38: This call should be wrapped with ```if(mounted)``` condition as a safe-check to avoid uncertain crash and actually checking first if widget exists in tree.

gtalha07 (approved):
LGTM except you need to add safe check which I've commented. :)

app/lib/screens/HomeScreens/chat/ChatProfile.dart
bitfriend: Understood.
I replaced all occurrences.

gtalha07: This is only required if you use ```setState(){}``` in ```initState(){}```. Not needed everywhere though :)

PR #295, opened 2022-10-22
Issue291 logout

bitfriend:


app/lib/widgets/SideMenu.dart
```diff
@@ -312,7 +316,14 @@ class _SideDrawerState extends State<SideDrawer> {
         AppLocalizations.of(context)!.logOut,
         style: SideMenuAndProfileTheme.signOutText,
       ),
-      onTap: () {
+      onTap: () async {
+        if (Get.isRegistered<CrossSigning>()) {
```
gtalha07, line 320: What does this have to do with CrossSigning? Am I missing something here?

bitfriend: We have to remove all event listeners prior to logout.
If not, event listening is kept after logout and we should get error that it will access invalid objects.

gtalha07 (approved):
LGTM!

app/lib/widgets/SideMenu.dart
bitfriend: Actually, typing listener or receipt listener is in chat list level and they are deleted when chat list disappears.
But cross signing is in global level so its allocation and deallocation should be dealt globally.
So I allocated it in `main.dart` and deallocated it in logout click.

PR #298, opened 2022-10-25
improve avatars render

gtalha07:


bitfriend (approved):
Looks good.
Anyway we have to cache avatar to avoid re-rendering of avatar whenever scrolling or other acting.
After timeline upgrade, I will be there.

PR #299, opened 2022-10-25
Issue287 members avatars and names display in chat group description

gtalha07:
**Description:**

- [x] Show active members of room. (with avatar and display name)
- [x] Use batch fetching for mentions, members and chat avatars.


bitfriend (approved):
Looks good to me.
I think it may be nice option to convert all `buildXxx()` functions into stateless widget later.

PR #321, opened 2022-11-01
Fix bug that event handlers are not working

bitfriend:
Rust updated latest message, but flutter doesn't recognize that change.
image
On flutter side, if `fetched` is `true`, `msg` should be not `null`.
But `msg` is always `null` in fact.
I don't know that reason.


bitfriend:
When I placed the fetching of latest message in `client.conversationsRx().listen()`, the recursive event callback occurred.
In order to avoid this issue, I moved fetching of latest message back to `ConversationController::load_rooms()`.

native/effektio/src/api/conversation.rs
```diff
@@ -52,66 +53,30 @@ impl Conversation {
...
 
-    pub(crate) fn set_latest_message(&self, msg: RoomMessage) {
-        self.latest_message.set(Some(msg));
+    fn set_latest_message(&mut self, msg: RoomMessage) {
+        self.latest_message = Some(msg);
     }
 
     pub fn latest_message(&self) -> Option<RoomMessage> {
-        self.latest_message.lock_mut().take()
+        self.latest_message.clone()
```
gnunicorn, line 79: 👍

native/effektio/src/api/conversation.rs
```diff
@@ -206,11 +168,14 @@ impl ConversationController {
...
     }
 
-    pub fn load_rooms(&self, convos: &Vec<Conversation>) {
-        for convo in convos.iter() {
-            convo.load_latest_message();
+    pub async fn load_rooms(&mut self, convos: &Vec<Conversation>) {
+        let mut conversations: Vec<Conversation> = vec![];
+        for convo in convos {
+            let mut conversation = convo.clone();
+            conversation.fetch_latest_message().await;
```
gnunicorn, line 175: should we really delay the returning of the entire list just to have the latest message available on each? I'd rather we moved this off into a background task - but then it can't be `mut` anymore. so, for now, I am fine with this, but we might want to put down a ticket that we want to optimise this for first-view-speed.

native/effektio/src/api/invitation.rs
```diff
@@ -144,34 +142,29 @@ impl InvitationController {
...
         client.add_event_handler_context(me);
         let handle = client.add_event_handler(
-            |ev: OriginalSyncRoomMemberEvent,
+            |ev: SyncRoomMemberEvent,
              room: MatrixRoom,
-             Ctx(client): Ctx<MatrixClient>,
-             Ctx(me): Ctx<InvitationController>,
-             handle: EventHandlerHandle| async move {
+             c: MatrixClient,
+             Ctx(me): Ctx<InvitationController>| async move {
```
gnunicorn, line 165: if `Ctx(client): Ctx<MatrixClient>` can be written as `c: MatrixClient,` now, can't `Ctx(me): Ctx<InvitationController>` also be written as `me: InvitationController` now? (same for the change above)

native/effektio/src/api/receipt.rs
```diff
@@ -111,10 +111,7 @@ impl ReceiptController {
         let me = self.clone();
         client.add_event_handler_context(me);
         let handle = client.add_event_handler(
-            |ev: SyncEphemeralRoomEvent<ReceiptEventContent>,
-             room: MatrixRoom,
-             Ctx(me): Ctx<ReceiptController>,
-             handle: EventHandlerHandle| async move {
+            |ev: SyncReceiptEvent, room: MatrixRoom, Ctx(me): Ctx<ReceiptController>| async move {
```
gnunicorn, line 114: same question about `Ctx(me)` here.

gnunicorn (approved):
Looking good. Just some remarks on style and potential future optimizations - CI also complains about the style, that needs to be fixed, too.

native/effektio/src/api/conversation.rs
bitfriend: I implemented flutter fetch function that is called from conversation list receiver, but it occurred recursive event callback.
I couldn't find that reason.
So I removed flutter fetch function and processed fetching in rust `ConversationController::load_rooms()` function.
You can see it here: https://github.com/effektio/effektio/pull/321/commits/b1f83e2829af4c20c28daeb0c1fbe7d660e740bf

native/effektio/src/api/invitation.rs
bitfriend: `client` is provided as default, but controller is provided by custom parameter.
Please read this code snippet: https://github.com/matrix-org/matrix-rust-sdk/blob/f8b02d1a11d07d28ce87132b93d30d6b03d76b5a/crates/matrix-sdk/src/event_handler/mod.rs#L422

native/effektio/src/api/receipt.rs
bitfriend: Please read this code snippet: https://github.com/matrix-org/matrix-rust-sdk/blob/f8b02d1a11d07d28ce87132b93d30d6b03d76b5a/crates/matrix-sdk/src/event_handler/mod.rs#L422

native/effektio/src/api/conversation.rs
bitfriend: I will remember and deal with this problem again later.

Issue #310, opened 2022-10-26
Blockout all non-implemented features


gnunicorn:
We currently show a lot features in the UI, but many aren't actually implemented or render wrong views. That is confusing. Instead, we want to upgrade the feature to use a muted color ("grayed out") on the link or button and instead of moving anywhere, it should show a [snackbar message](https://docs.flutter.dev/cookbook/design/snackbars) on the bottom that this feature isn't yet implemented. It should be build as a reusable component we can easily add or remove.

This should be applied to _every_ button we don't have a route for yet.

In particular:
 - [ ] all non-implemented sections in the bar on the left
 - [ ] the profile-icon in the sidebar
 - [ ] the profile-icon in chat

PR #323, opened 2022-11-01
Restructure bottom bar, clean up tasks and pins

### LINKED ISSUE #309: Reorganise Bottom bar
The bottom bar for the current version should have the following icons (and only those):
 - News
 - Pins
 - Tasks
 - Chat

gnunicorn:
This cleans up the bottom bar (in accordance with #309) and cleans up the tasks, pins and news screens:
 - [x] tasks: remove top navigation
 - [x] tasks: unimplemented features have snackbar, refs #310
 - [x] news: unimplemented features have snackbar, refs #310
 - [x] Pins: refactor for prettier screen

Fixes #309

### DIFF
--- a/app/lib/screens/SideMenuScreens/ToDo.dart
+++ b/app/lib/screens/SideMenuScreens/ToDo.dart
@@ -2,6 +2,7 @@ import 'package:effektio/common/store/MockData.dart';
 import 'package:effektio/common/store/themes/SeperatedThemes.dart';
 import 'package:effektio/controllers/todo_controller.dart';
 import 'package:flutter/material.dart';
+import 'package:effektio/widgets/AppCommon.dart';
 import 'package:get/get.dart';
 
 class ToDoScreen extends StatefulWidget {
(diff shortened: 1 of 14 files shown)


app/lib/screens/SideMenuScreens/ToDo.dart
```diff
@@ -38,7 +37,16 @@ class _ToDoScreenState extends State<ToDoScreen> {
       floatingActionButton: FloatingActionButton(
         backgroundColor: ToDoTheme.floatingABColor,
         onPressed: () {
-          Navigator.pushNamed(context, '/addTodo');
+          ScaffoldMessenger.of(context).showSnackBar(
+            SnackBar(
```
gtalha07, line 41: Since this will be frequently occuring widget, we can convert this into reusable method and can pass generic strings.

app/lib/screens/SideMenuScreens/ToDo.dart
```diff
@@ -106,6 +97,17 @@ class _ToDoScreenState extends State<ToDoScreen> {
       splashColor: ToDoTheme.primaryTextColor,
       onTap: () {
         todoController.updateIndex(index);
+        ScaffoldMessenger.of(context).clearSnackBars();
+        ScaffoldMessenger.of(context).showSnackBar(
+          SnackBar(
```
gtalha07, line 102: Could use reusable widget here.

gtalha07 (approved):
Overall looks good except we could use reusable pattern for snackbars.

app/lib/screens/SideMenuScreens/ToDo.dart
gnunicorn: yeah, you are right. I was just wondering if that's really worth is considering this is a pretty short piece of code. But once I started playing with the clear and adapting the UI, it kinda started to make sense. Will create this as an external widget.

gnunicorn: coming to think of it, this isn't really suitable as a widget, as it doesn't really render anything but would only show once-off message and I don't want it to wrap around other widgets to manage the `onPressed` side. I'll just convert this into a helper function.

PR #326, opened 2022-11-02
Disable unimplemented

### LINKED ISSUE #310: Blockout all non-implemented features

gnunicorn:
fixes #310 .


app/lib/common/store/themes/SeperatedThemes.dart
```diff
@@ -26,6 +26,8 @@ class AppCommonTheme {
 
 class SideMenuAndProfileTheme {
   static const sideMenuTextColor = ColorRef(Color.fromRGBO(196, 196, 196, 1));
+  static const sideMenuTextColorDisabled =
+      ColorRef(Color.fromRGBO(196, 196, 196, 0.4));
```
gtalha07, line 30: Reminds me if const value of color is better than calling static method of conversion of color ?  Although if you hover over this line, flutter tools will give option to convert this into hex value.

gtalha07 (approved):
LGTM.

app/lib/common/store/themes/SeperatedThemes.dart
gnunicorn: I have no idea :D ... but this isn't an optimization that matters to us at the moment..

PR #329, opened 2022-11-03
[Patch] allow async fetch once for client info

gtalha07:
### Changes:
Everytime we open the sidemenu, it triggers async fetch of client avatars and name. This has been changed to fetch only once the client is logged in. So it doesn't show re-rendering animations.


bitfriend (approved):
Looks good to me

PR #332, opened 2022-11-04
Ben fixing mac nightly build

gnunicorn:
- [x] update launcher icon
- [x] localised library name fixes for macosx dylib - CI tests are still running.
   I think I've finally found the issue for mac OSx builds are not portable (nightly builds don't run locally, copying apps to other machines generally fail to start): the path to the original dylib is hardcoded into the final fat-dylib, which is used to load them. That failed as they are not at the same location on all machines. After a lot of back-and-forth, I've found a way to replace the internal name with a bundle-specific portable name using `install_name_tool`. This PR uses that to overwrite the internal lib name with the one we need to have a portable lib.


gnunicorn:
It's finally working!

gnunicorn:
we can confirm this properly fixes it.

PR #336, opened 2022-11-04
fix news bar alignment

gtalha07:
### Description:
Fixed the like icon alignment.


sachintha-s (changes requested):
Please tap on the like button and check it out. When it animates it goes to the edge of the device. I think the alignment should be improved when it is animated.

gtalha07:
> Please tap on the like button and check it out. When it animates it goes to the edge of the device. I think the alignment should be improved when it is animated.

Will fix it.

app/lib/widgets/NewsItem.dart
```diff
@@ -37,37 +39,41 @@ class NewsItem extends StatelessWidget {
...
-                      _buildTitle(bgColor, fgColor),
-                      const SizedBox(height: 10),
-                      _buildSubtitle(bgColor, fgColor),
-                      const SizedBox(height: 10),
-                    ],
+        LayoutBuilder(
+          builder: (context, constraints) {
+            return SizedBox(
+              height: constraints.maxWidth >= 600
+                  ? isDesktop
```
bitfriend, line 46: Yeah, desktop layout should be different somewhat from mobile desktop.

bitfriend (approved):
Looks good to me

PR #339, opened 2022-11-07
Issue312 fetch users out of current room

bitfriend:


app/lib/controllers/chat_room_controller.dart
```diff
@@ -103,22 +103,22 @@ class ChatRoomController extends GetxController {
...
-      _stream = _currentRoom!.timeline();
+      _stream = convoRoom.timeline();
       // i am fetching messages from remote
       if (_currentRoom == null) {
         // user may close chat screen before long loading completed
         isLoading.value = false;
         return;
       }
-      var msgs = await _stream!.paginateBackwards(10);
+      var msgs = await _stream!.paginateBackwards(5);
```
gtalha07, line 121: Why 5? I think 10 is optimal number for fetch. Besides, scrolling is disabled if small number of messages are there initially. Weird though.

gtalha07 (approved):
LGTM!

app/lib/controllers/chat_room_controller.dart
bitfriend: 10 is too large to fetch successful decrypted messages under encryption module.
So I was testing with 5.

bitfriend: Reverted to 10

PR #346, opened 2022-11-09
Idenitfy Effektio via device_name

gnunicorn:
- [x] expose device_name on native lib
- [x] calculate device name from flutter platform
- [x] identifier for versioned releases


gnunicorn:
See, what I get in my fluffy chat now:

Screenshot 2022-11-09 at 18 58 32

PR #347, opened 2022-11-10
Upgrade verification to v0.6

bitfriend:
`matrix-sdk` has some changes when comparing `v0.6` with `v0.5`.
https://github.com/matrix-org/matrix-rust-sdk/blob/main/UPGRADING-0.5-to-0.6.md
This PR will review and fix bug about verification feature on `v0.6`.

### DIFF
--- a/app/analysis_options.yaml
+++ b/app/analysis_options.yaml
@@ -7,3 +7,7 @@ linter:
     always_declare_return_types: true
     always_use_package_imports: true
     file_names: false
+
+analyzer:
+  errors:
+    unused_import: warning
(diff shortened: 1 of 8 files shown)


bitfriend:
I was getting this error again.
https://github.com/matrix-org/matrix-rust-sdk/issues/117
I logged out and retried, then last mismatch disappeared.
I guess session storage should be reset if rust engine has some changes.
That is, logout/login is needed.

bitfriend:
Now I am getting new device prompt when verification done.
I will fix that issue in other PR.

bitfriend:
When trying verification without restarting app after login, verification works well without mismatch.
When trying verification with restarting app after login, verification shows mismatch error of emoji.
If user closes and opens app, session storage seems to be immutable.
I tested app on Windows 10.

PR #366, opened 2022-11-16
fix verification sheet layout

gtalha07:
- [x]  Fixed the verification emoji appearance in desktop version.


bitfriend (approved):
Just tested.
Nice layout.

gtalha07:
> Just tested. Nice layout. I found desktop shows emoji letters in a line. As we know, mobile shows emoji letters in 2 lines.

Yeah actually on desktop it could show in 2 lines but due to ```GridView``` being scrollable it appeared like it only shows the first line. Whereas you could actually scroll to see other line. This is confusing user-wise so I put all of them on one line.

gtalha07:
Also adding flexible doesn't throws overflow errors while resizing the app (unless at a point just breaks the whole UI), so I guess its good enough.

bitfriend:
Cool

PR #371, opened 2022-11-18
Issue305 - Double tap on news comment

inderjeet-developer7:


app/lib/widgets/CommentView.dart
```diff
@@ -22,162 +23,148 @@ class CommentView extends StatefulWidget {
...
                         ),
-                      ),
-                      Text(
-                        widget.comment,
-                        style: const TextStyle(
-                          color: Colors.white,
-                          fontSize: 16,
+                        Text(
+                          widget.commentModel.comment,
+                          style: const TextStyle(
```
gtalha07, line 75: Please add theme constants in ```seperatedThemes.dart``` if reusable.

app/lib/widgets/ReplyView.dart
```diff
@@ -37,11 +39,12 @@ class ReplyViewState extends State<ReplyView> {
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
                   Text(
-                    widget.name,
-                    style: TextStyle(color: widget.titleColor, fontSize: 16),
+                    widget.replyModel.name,
+                    style: TextStyle(
+                        color: widget.replyModel.titleColor, fontSize: 16),
```
gtalha07, line 44: Same here.

gtalha07 (approved):
LGTM!

app/lib/widgets/CommentView.dart
inderjeet-developer77: we cannot as right now I am passing color from the previous screen.. once API is integrated... we will set colors randomly

app/lib/widgets/ReplyView.dart
inderjeet-developer77: same as above

PR #373, opened 2022-11-19
Fix event decryption error

bitfriend:
After upgrade to `matrix-sdk v0.6`, some event decryption failed.
Applied to `retry-decrypt` of jplatte, but still some rooms don't show messages.
And under `retry-decrypt`, any room is not shown in room list after restart.
That bug must be fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,9 +201,9 @@ dependencies = [
 
 [[package]]
 name = "base64"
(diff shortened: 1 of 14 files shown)


bitfriend:
This PR was closed because I renamed `issue370_event-decryption-v0.6` to `issue370_retry-decrypt-v0.6`.

PR #380, opened 2022-11-23
optimize decoded images

gtalha07:
- [x] Optimize fetched decoded images.
- [x] Optimize fetched decoded avatars.
- [x] Discard ```CachedMemoryImage``` package.

### DIFF
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -56,6 +56,7 @@ class Effektio extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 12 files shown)


app/lib/main.dart
```diff
@@ -56,6 +56,7 @@ class Effektio extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    debugInvertOversizedImages = true;
```
bitfriend, line 59: I am happy if you leave some comment here.

Issue #337, opened 2022-11-07
internet connectivity state in app


gtalha07:
We should add network aware handlers/widgets in the application which would inform the internet connectivity state of the application. These packages seems to be helpful:
https://pub.dev/packages/connectivity_plus
https://pub.dev/packages/connectivity_wrapper

kokodavid:
![image](https://user-images.githubusercontent.com/47295684/210715399-aba880eb-586c-4e32-a932-8b1e90923126.png)

gtalha07:
Closed as #432 merged.

PR #381, opened 2022-11-24
Internet State Management #337

inderjeet-developer7:


app/lib/screens/HomeScreens/chat/ChatScreen.dart
```diff
@@ -262,7 +266,25 @@ class _ChatScreenState extends State<ChatScreen>
           body: Obx(
             () => SafeArea(
               bottom: false,
-              child: buildBody(context),
+              child: networkController.connectionType.value == '0'
+                  ? Center(
```
gtalha07, line 270: Is this the same widget used everywhere? Maybe define it one place and calling it wherever needed? :)

app/lib/screens/HomeScreens/news/News.dart
```diff
@@ -40,121 +43,145 @@ class _NewsScreenState extends State<NewsScreen>
...
-                      color: Colors.white,
-                      fontSize: 14,
-                      shadows: <Shadow>[
-                        Shadow(
-                          blurRadius: 1.0,
-                          color: Colors.black,
+        return Obx(
+          () => Container(
+            child: networkController.connectionType.value == '0'
+                ? Center(
```
gtalha07, line 49: Think so, just write it in as seperate widget and instead call it here.

gtalha07 (approved):

PR #384, opened 2022-11-25
prevent sending message event twice

gtalha07:
This patch fixes the sending of event messages twice when send button is pressed.


gtalha07:
> When I was writing msg send code, I just wanted to let user show his message immediately without waiting for it to be sent. If we wish to avoid complexity, I think your code is good.
I know and actually the old code worked when we were using v0.5. Now we have VecDiff stream doing this work and automatically adds events. So this code was rendering event twice while still sending one to server.

gnunicorn:
> I know and actually the old code worked when we were using v0.5. Now we have VecDiff stream doing this work and automatically adds events. So this code was rendering event twice while still sending one to server. :)

Okay, this might fixes a twice-rendering issue, but I still have the issue that if I don't have any visual feedback, and it takes a while before the message sending finishes, I am sending it more than once. The other day, I accidentially sent a message 5 times because of that.

I think we want to disable the text-field and sent-button until we have seen the event in the vecdiff (as not-yet-sent) or - if that is too complicated for now - until we have confirmed it being sent....

I also think we might have a new, lighter send method on the timeline object now, that is already managing draft messages ...

gnunicorn (approved):
looks good to me.

gtalha07:
> > I know and actually the old code worked when we were using v0.5. Now we have VecDiff stream doing this work and automatically adds events. So this code was rendering event twice while still sending one to server. :)
> 
> Okay, this might fixes a twice-rendering issue, but I still have the issue that if I don't have any visual feedback, and it takes a while before the message sending finishes, I am sending it more than once. The other day, I accidentially sent a message 5 times because of that.
> 
> I think we want to disable the text-field and sent-button until we have seen the event in the vecdiff (as not-yet-sent) or - if that is too complicated for now - until we have confirmed it being sent....
> 
> I also think we might have a new, lighter send method on the timeline object now, that is already managing draft messages ...

Okay I will look into it and push that fix here too.

PR #387, opened 2022-11-27
Emoji reaction

bitfriend:
Will solve https://github.com/effektio/effektio/issues/139

### DIFF
--- a/app/lib/screens/HomeScreens/chat/ChatScreen.dart
+++ b/app/lib/screens/HomeScreens/chat/ChatScreen.dart
@@ -179,13 +179,18 @@ class _ChatScreenState extends State<ChatScreen> {
         borderRadius: BorderRadius.circular(15),
         child: Image.memory(
           base64Decode(imageMessage.metadata?['base64']),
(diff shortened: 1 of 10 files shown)


app/lib/screens/HomeScreens/chat/ChatScreen.dart
```diff
@@ -227,14 +232,25 @@ class _ChatScreenState extends State<ChatScreen> {
...
         ),
       );
     }
   }
 
+  Widget customMessageBuilder(
+    types.CustomMessage customMessage, {
+    required int messageWidth,
+  }) {
+    return const SizedBox();
```
gtalha07, line 251: @bitfriend Try replacing with Placeholder text.

gtalha07: Also we want to hide receipts but not bubble. Does unsupported message in matrix terms returns something? Decryption failed error?

bitfriend: I don't know, but we can use the same text as element app.

PR #389, opened 2022-11-28
add notification overlay for new session alert

gtalha07:
### Description:

- [x] Added notification overlay for session alert pop-up.

**Current Issue:** ```DeviceChangeRx()``` is constantly emitting device ids which is triggering pop-ups infinitely. Notification alert has been set up to auto-dismiss with 1 sec interval to prevent stacking up. This needs an fix on rust-side.

image
image


bitfriend (approved):
Looks good to me

PR #390, opened 2022-11-28
encode and decode image data

gtalha07:
- [ ] Encode image to base64 and decode to binary at render.


bitfriend (approved):
Looks good to me

PR #393, opened 2022-12-01
Chat reply

inderjeet-developer7:


gtalha07 (approved):
Nicely done. One thing I would like to ask what gesture you used for reply? Swipe?

inderjeet-developer77:
> Nicely done. One thing I would like to ask what gesture you used for reply? Swipe?

Nope... did not use any gestures as the video didn't have them... Was thinking to add Swipe to reply but then dropped the idea

PR #394, opened 2022-12-01
Use unsupported message to keep track of index in diff receiver

bitfriend:
Will fix issue that is related with wrong index of timeline event.
https://github.com/effektio/effektio/issues/382


gtalha07 (approved):
LGTM

PR #397, opened 2022-12-03
Issue370 event decryption v0.6

bitfriend:
Append decrypt-failed event as unsupported message.
It will solve the issue that index tracking in diff receiver of conversation gets incorrect sometimes.
It should return message structure with event id, sender, room id and timestamp, even though decrypt failed.


gtalha07 (approved):
LGTM. 
Note: Some of decrypt failed events appear in non-encrypted rooms. This needs to be looked as seperate bug as it might needs internal fixes in the API.

bitfriend:
`TimelineItemContent::RedactedMessage`, `TimelineItemContent::FailedToParseMessageLike` and `TimelineItemContent::FailedToParseState` exist as event content that must not be placed in msg list, in addition to `TimelineItemContent::UnableToDecrypt`.
I will implement the error handler for those 3 event contents in more details.

PR #402, opened 2022-12-08
[TimelineItem] filter null events and call pagination till scroll trigger

gtalha07:
- [x] Filter null events so it don't gets rendered in messages.
- [x]  Call pagination API till initial maintain count threshold is met.


gtalha07:
Closed as fixed via #397.

PR #422, opened 2023-01-03
Issue383 chat markdown support improvement

gtalha07:
- [x] use matrix html code language to parse content.
- [x] remove mx-reply tags using ```simplify_body()``` from @bitfriend code.
- [x] link preview without matrix identifiers.
- [x] Emoji enlargement behavior when text only contains emojis.


bitfriend (approved):
Looks good to me.
It was great choice that you used `flutter_matrix_html`, not `flutter_html`.
Needless to say, `flutter_matrix_html` will do many things for us that are related to `matrix`.

PR #427, opened 2023-01-06
Fixing mock command

gnunicorn:
This fixes the unwrap failure in the `cli mock` command.


bitfriend:
When executing `cargo run -p effektio-cli -- mock http://192.168.142.130:8008`, I got stack overflow error in Windows.
But Ubuntu doesn't show that error.
I guess Windows has issue about stack overflow.

bitfriend:
This is screenshot of integration in my ubuntu vm.
image
As you see, this test didn't show `Room with alias not found`.
That is, local test is not exactly same as github test.
We decided to fix this issue later, because it may take a long time and we have urgent issues.

PR #444, opened 2023-01-13
Model polling via wait_for

gnunicorn:
We sometimes (like everytime we `sent` for one), don't directly get notified about the model, as the sync might bring it down the wire later. With the new `wait_for_TYPE`-functions on client, you can now `await` with just the `event_id` received as a result from `send?`. Pretty nifty.


native/effektio/api.rsh
```diff
@@ -16,6 +16,12 @@ fn register_with_registration_token(basepath: string, username: string, password
...
 fn gen_mock_news() -> Vec<News>;
 
+/// Representing a time frame
+object Duration {}
+
+fn duration_from_secs(secs: u64) -> Duration;
+
+
 /// Representing a color
 object Color {
```
gtalha07, line 26: I saw that Color class has been implemented in rust which makes Flutter confuse as to whom it is referencing when mapping the object to Flutter class ```ToDoList```. ```Color``` is also present in the Flutter core ```painting.dart```.

gtalha07 (approved):
LGTM!

native/effektio/api.rsh
gnunicorn: nothing new in this one, but we can rename it here if you like. `EfkColor` or `MatrixColor` what do you prefer?

PR #446, opened 2023-01-16
Beamer Integration

inderjeet-developer7:


app/lib/main.dart
```diff
@@ -55,19 +53,23 @@ Future<void> startApp() async {
...
+    locationBuilder: Routes.getRoutes(),
+  );
 
   @override
   Widget build(BuildContext context) {
     debugInvertOversizedImages = true; // detect non-optimized images
     return Portal(
       child: Themed(
         child: OverlaySupport.global(
-          child: GetMaterialApp(
```
gtalha07, line 70: I think not having this wrap or replacing would not trigger the cross-signing overlay as its built around ```GetX``` components. Can you try wrapping ```MaterialApp.router``` with parent ```GetMaterial``` or maybe we have to use ```OverlaySupport``` components.

app/lib/models/EditGroupInfoModel.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:effektio_flutter_sdk/effektio_flutter_sdk_ffi.dart';
+
+class EditGroupInfoModel {
```
gtalha07, line 3: Are these models needed for ```beamer``` functionality?

gtalha07 (commented):
LGTM!

app/lib/main.dart
inderjeet-developer7: I am using same overlay for internet connection handling and it works fine, You can test once and if it doesn't works, we can look into it. Only a minor change would be needed as the MaterialApp.router is wrapped with OverlaySupport.global itself.

app/lib/models/EditGroupInfoModel.dart
inderjeet-developer7: Yes, to pass data into beamer we have to pass it in data key which takes "data as MyObject" where MyObject is a class. So these models are used there, Regardless wherever there is only one parameter to pass, I passed it directly without model class

PR #478, opened 2023-02-06
Improve Flutter code 

gtalha07:
### Changes:
- [x] extract widgets from methods to keep performance pattern.
- [x] Rename files according to flutter convention.


bitfriend (approved):
I found you changed all rendering functions of child components to class.
Nice work.

PR #483, opened 2023-02-13
render non-user events

gtalha07:
### Description:

- [x] Render ```m.notice``` and ```m.server_notice``` events.
- [x] Render ```m.room.member``` events.
- [x] Mute non-user events using less prominent styling.


bitfriend (approved):
Great work.

PR #484, opened 2023-02-13
Sometimes update event is fetched prior to insert event

bitfriend:
This case occurs because of network delay.


gtalha07 (approved):
LGTM!

PR #488, opened 2023-02-14
Issue461 bug report feature

bitfriend:
Will implement console & file logging.
User will be able to submit log file when he meets any bug.


bitfriend:
The path of log file is as following:
windows
`C:\Users\xxx\Documents\app_2023-02-15_01-37-35.log`
android
`/data/user/0/org.effektio.app/app_flutter/app_2023-02-14_18-05-51.log`

app/packages/rust_sdk/lib/effektio_flutter_sdk.dart
```diff
@@ -233,4 +257,37 @@ class EffektioSdk {
...
+      defaultValue: 'https://rageshake.acter.global/api/submit',
+    );
+    const username = String.fromEnvironment(
+      'RAGESHAKE_USERNAME',
+      defaultValue: 'acter',
+    );
+    const password = String.fromEnvironment(
+      'RAGESHAKE_PASSWORD',
+      defaultValue: 'global',
+    );
```
gnunicorn, line 273: Please use defaults for a localhost setup, not the ones we were to use in the actual setup (and I know have to change, as they've been leaked ;P )

native/effektio/Cargo.toml
```diff
@@ -77,9 +80,9 @@ android_logger = "0.12"
 tokio = { version = "1", features = ["rt-multi-thread", "parking_lot"] }
 tracing-android = "0.2"
 
-#   ----   DESKTOP
-[target.'cfg(all(not(target_os = "android"), not(target_arch = "wasm32"), not(target_os = "ios")))'.dependencies]
-flexi_logger = "0.24.0"
```
gnunicorn, line 82: Why did you remove flexilogger for desktop setups? The setup below is only tied to macos desktops. What about Linux and Windows desktops?

native/effektio/src/platform/android.rs
```diff
@@ -12,19 +23,137 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
...
+    let log_level = match filter {
+        Some(ref filter) => FilterBuilder::new().parse(&filter).build(),
+        None => FilterBuilder::new().build(),
+    };
+
     let mut log_config = Config::default()
-        .with_min_level(Level::Trace)
-        .with_tag("effektio-sdk");
+        .with_max_level(LevelFilter::Trace)
+        .with_tag(app_name.as_str());
```
gnunicorn, line 39: isn't the `app_name` known at compile time? Why is this a runtime parameter ...

native/effektio/src/platform/android.rs
```diff
@@ -12,19 +23,137 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
...
+    with_log: bool,
+) -> Result<bool> {
+    let mut form = Form::new()
+        .text("text", text)
+        .text("user_agent", "Mozilla/0.9")
+        .text("app", app_name)
+        .text("version", version)
+        .text("label", label);
+    if with_log {
+        unsafe {
```
gnunicorn, line 126: why does this have an `unsafe` block? This shouldn't be unsafe.

native/effektio/src/platform/android.rs
```diff
@@ -12,19 +23,137 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
...
+    if level == log::LevelFilter::Off {
+        log::set_boxed_logger(Box::new(native::NopLogger)).unwrap();
+    } else {
+        log::set_boxed_logger(Box::new(dispatch.clone())).unwrap();
+    }
+    log::set_max_level(level);
+
+    unsafe {
+        FILE_LOGGER = Some(dispatch);
+    }
```
gnunicorn, line 75: Please use a global mutex and/or lazy, rather than `unsafe` here, please.

native/effektio/src/platform/android.rs
```diff
@@ -12,19 +23,137 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
...
+) -> Result<bool> {
+    let mut form = Form::new()
+        .text("text", text)
+        .text("user_agent", "Mozilla/0.9")
+        .text("app", app_name)
+        .text("version", version)
+        .text("label", label);
+    if with_log {
+        unsafe {
+            if let Some(dispatch) = &FILE_LOGGER {
```
gnunicorn, line 127: What if this isn't set? We should probably `Err` here, rather than hiding the bug that our bug reporting is broken.

native/effektio/src/platform/desktop.rs
```diff
@@ -18,17 +23,102 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
     Ok(builder)
 }
 
-pub fn init_logging(filter: Option<String>) -> Result<()> {
+static mut FILE_LOGGER: Option<Arc<fern::ImplDispatch>> = None;
```
gnunicorn, line 26: same as above, please use a lazy-static and/or mutex here for a global not `unsafe`.

native/effektio/src/platform/desktop.rs
```diff
@@ -18,17 +23,102 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
...
+    log::set_max_level(level);
+
+    unsafe {
+        FILE_LOGGER = Some(dispatch);
+    }
+
     Ok(())
 }
+
+pub async fn report_bug(
```
gnunicorn, line 74: this is the same as for `android` and below, right? Please put it into a common place then as one function and import it.  Having three copies will lead to the code diverging...

gnunicorn (changes requested):

app/packages/rust_sdk/lib/effektio_flutter_sdk.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/488/commits/b14d323fbe0bcf467b22b3be661ca795f569104a

native/effektio/Cargo.toml
bitfriend: I confirmed `fern` instead of `flexi_logger` is working on windows and macos.
It is needed to reduce unnecessary libraries, because of ruby-macho in ios build.

app/lib/common/widgets/side_menu.dart
```diff
@@ -53,6 +53,7 @@ class SideDrawer extends StatelessWidget {
               const _GroupBudgetingItem(),
               const _SharedDocumentsItem(),
               const _GalleryItem(),
+              const _BugReportItem(),
```
gtalha07, line 56: Ideally I would like this button to be placed on the screen with overlay context, so it should be accessible globally across all screens. You can use that context further to pop off once user is done reporting.

gtalha07 (commented):
Looks good on flutter side.

native/effektio/src/platform/android.rs
bitfriend: I thought it is good because it works.
Well, I was wondering app crashes sometimes.
That was because of `unsafe` block.
Fixed by https://github.com/acterglobal/a3/pull/488/commits/91869ec5f90d3912b76ec5034ff404afb4b42127

native/effektio/src/platform/android.rs
bitfriend: Yeah, I fixed that by `lazy_static`.
Please reference https://github.com/acterglobal/a3/pull/488/commits/91869ec5f90d3912b76ec5034ff404afb4b42127

app/lib/common/widgets/side_menu.dart
gnunicorn: I think this is fine fore now. In a later iteration we might overlay it everywhere or put it under the settings to make this cleaner...

native/effektio/src/platform/android.rs
bitfriend: If we use `Err`, app will crash.
I think it is preferable to use `Ok(false)` than `Err`.
```rust
    if with_log {
        match &*FILE_LOGGER.lock().unwrap() {
            Some(dispatch) => {
                let res = dispatch.rotate();
                for output in res.iter() {
                    match output {
                        Some((old_path, new_path)) => {
                            let log_path = old_path.canonicalize()?.to_string_lossy().to_string();
                            let file = fs::read(log_path)?;
                            let filename =
                                old_path.file_name().unwrap().to_string_lossy().to_string();
                            let file_part = Part::bytes(file)
                                .file_name(filename)
                                .mime_str("text/plain")?;
                            form = form.part("log", file_part);
                            break;
                        }
                        None => {}
                    }
                }
            }
            None => {
                return Ok(false);
            }
        }
    }
```

native/effektio/src/platform/desktop.rs
bitfriend: Yeah, I fixed that by `lazy_static`.
Please reference https://github.com/acterglobal/a3/pull/488/commits/91869ec5f90d3912b76ec5034ff404afb4b42127

native/effektio/src/platform/android.rs
gnunicorn: Well, it crashed because the outer layer isn't handling the `Err` correctly, isn't it? It should probably `catch`, while we here still `Err` (which shouldn't be the case usually anyways).

bitfriend: What about this code?
I replaced `Ok(false)` with `bail!()`.
```rust
    if with_log {
        match &*FILE_LOGGER.lock().unwrap() {
            Some(dispatch) => {
                let res = dispatch.rotate();
                for output in res.iter() {
                    match output {
                        Some((old_path, new_path)) => {
                            let log_path = old_path.canonicalize()?.to_string_lossy().to_string();
                            let file = fs::read(log_path)?;
                            let filename =
                                old_path.file_name().unwrap().to_string_lossy().to_string();
                            let file_part = Part::bytes(file)
                                .file_name(filename)
                                .mime_str("text/plain")?;
                            form = form.part("log", file_part);
                            break;
                        }
                        None => {}
                    }
                }
            }
            None => {
                return bail!("You didn't set up file logger.");
            }
        }
    }
```

bitfriend: Fixed by https://github.com/acterglobal/a3/pull/488/commits/586bec5b92bf6d22a591b1a77bfcc25062b527b7

native/effektio/src/platform/desktop.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/488/commits/3ee79ed77286ca599993d15ba9df4e092e7f77f6

app/lib/common/widgets/side_menu.dart
bitfriend: Will ignore for now.

native/effektio/src/api.rs
```diff
@@ -85,7 +85,22 @@ mod api {
...
 }
 
-fn init_logging(filter: Option<String>) -> Result<()> {
-    platform::init_logging(filter)?;
-    Ok(())
+fn init_logging(log_dir: String, filter: Option<String>) -> Result<()> {
+    platform::init_logging(log_dir, filter)
+}
+
+async fn report_bug(
```
gnunicorn, line 92: clippy is complaining about this one, too

native/effektio/src/platform/android.rs
```diff
@@ -1,30 +1,101 @@
...
+    let log_level = match filter {
+        Some(ref filter) => FilterBuilder::new().parse(&filter).build(),
+        None => FilterBuilder::new().build(),
+    };
+
     let mut log_config = Config::default()
-        .with_min_level(Level::Trace)
-        .with_tag("effektio-sdk");
+        .with_max_level(LevelFilter::Trace)
+        .with_tag("com.example.effektio");
```
gnunicorn, line 33: we should probably move that into a const somewhere, no?

native/effektio/src/platform/desktop.rs
```diff
@@ -18,17 +23,48 @@ pub async fn new_client_config(base_path: String, home: String) -> Result<Client
     Ok(builder)
 }
 
-pub fn init_logging(filter: Option<String>) -> Result<()> {
+// this excludes macos, because macos and ios is very much alike in logging
+
+pub fn init_logging(log_dir: String, filter: Option<String>) -> Result<()> {
```
gnunicorn, line 28: we are still having thee almost identical copies of the same code here, don't we? please unify that into one fn that is called from within these...

gnunicorn (changes requested):

gnunicorn:
Optional re:

it is called a rage-shake... how about [we use the rage detection](https://pub.dev/packages/shake) to allow to trigger it anywhere (rather than an overlay button)? Doesn't have to be in this PR though.

app/lib/main.dart
```diff
@@ -365,6 +371,7 @@ class _EffektioHomeState extends State<EffektioHome>
             ),
             floatingActionButton: FloatingActionButton(
               onPressed: () async {
+                throw Exception('force exception');
```
gnunicorn, line 374: this should probably not be here.

native/effektio/src/platform/desktop.rs
bitfriend: In fact, `init_logging()` has the specified implementation by platform, like `new_client_config()`.
So I split it into `android`, `ios` and `desktop`.
Will I unify `new_client_config()` too?

app/lib/main.dart
bitfriend: Sorry, my bad.
It was my test code.
Removed.

bitfriend:
> Optional re:
> 
> it is called a rage-shake... how about [we use the rage detection](https://pub.dev/packages/shake) to allow to trigger it anywhere (rather than an overlay button)? Doesn't have to be in this PR though.

It is possible in mobile.
What will we do in desktop?

native/effektio/src/api.rs
bitfriend: Fixed

bitfriend:
Implemented shake in only mobile, because it is not available in desktop.

native/effektio/src/platform/android.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/488/commits/be70dcb45f7585bfa05e325db7e4615e2ea72984

app/lib/features/bug_report/controllers/bug_report_controller.dart
```diff
@@ -1,8 +1,37 @@
...
+
+const rageshakeUsername = String.fromEnvironment(
+  'RAGESHAKE_USERNAME',
+  defaultValue: 'alice',
+);
+
+const rageshakePassword = String.fromEnvironment(
+  'RAGESHAKE_PASSWORD',
+  defaultValue: 'secret',
+);
```
gnunicorn, line 24: Did we try if these are actually needed? The config docs imply this is only for the other `api/listings` API endpoint.

app/lib/features/bug_report/controllers/bug_report_controller.dart
```diff
@@ -1,8 +1,37 @@
...
+
+const issueAppName = String.fromEnvironment(
+  'ISSUE_APP_NAME',
+  defaultValue: 'a3-nightly',
+);
+
+const issueAppVersion = String.fromEnvironment(
+  'ISSUE_APP_VERSION',
+  defaultValue: '1.0.0',
+);
```
gnunicorn: Let's prefix them all with the same `RAGESHAKE_` please ...

```suggestion
const issueAppName = String.fromEnvironment(
  'RAGESHAKE_APP_NAME',
  defaultValue: 'a3-nightly',
);

const issueAppVersion = String.fromEnvironment(
  'RAGESHAKE_VERSION',
  defaultValue: '1.0.0',
);
```

native/effektio/src/platform/desktop.rs
gnunicorn: How do these implementations actually differ? To me these look too similar to not unify them in one place. Same as for `new_client_config`, just reuse a `native` implementation:

```rust
pub async fn init_logging(log_dir: String, filter: Option<String) -> Result<()> {
    Ok(native::init_logging(log_dir, filter).await?)
}
```

gnunicorn (approved):
Still some remarks, but once done feel free to merge them.

app/lib/features/bug_report/controllers/bug_report_controller.dart
bitfriend: You were correct.
Fixed by https://github.com/acterglobal/a3/pull/488/commits/e5d210a86950453c80bfbfbea240e79f2279f62f

app/lib/features/bug_report/controllers/bug_report_controller.dart
bitfriend: Which one is correct?
image

gnunicorn: `a3-nightly`

I was just referring to the name of the env-var.

bitfriend: Appended `RAGESHAKE_` prefix.

PR #505, opened 2023-03-04
Adaptive layout.

gtalha07:
**Description:**
- adaptive layout for the core UI to be compliable across platforms. 
- In desktop, user can adjust windows resolution:
  1.  Small res: dashboard and navigation rail will get adapted to bottom navigation.
 - In mobile: 
  1. Small and Up Res: Bottom navigation will be shown regardless screen size changes. 
  Moreover, additional customization to UI behavior is possible now thanks to ```AdaptiveLayout```  widget  according to user interface requirements.


bitfriend (approved):
Looks good to me.
It is lucky that material flutter has the concept of breakpoint like material javascript.

PR #522, opened 2023-03-09
Fix bug that screenshot for bug report doesn't work

bitfriend:
We must allocate a `ScreenshotController` by every screen that needs screenshot.
And must wrap every screen that needs screenshot by `Screenshot`.


bitfriend:
The following code performs the heavy operation in `setState()`.
```dart
setState(() {
  _selectedIndex = index;
  pageController.jumpPage(_selectedIndex);
});
```
Heavy `setState()` may affect the app's total performance.
We can move the meaningless part to outside of `setState()`.
```dart
setState(() => _selectedIndex = index);
pageController.jumpPage(index);
```

app/lib/features/home/pages/home_page.dart
```diff
@@ -40,6 +41,7 @@ class _HomePageState extends ConsumerState<HomePage> {
   void initState() {
     super.initState();
     pageController = PageController(initialPage: _selectedIndex);
+    screenshotController = ScreenshotController();
```
gtalha07, line 44: Is this really needed to be constructed here? ```late``` is bad if for some reason widget doesn't get inserted it would throw ```InitializationError```.
How about having it initialised above?

gtalha07 (approved):
LGTM but please address the concern :)

app/lib/features/home/pages/home_page.dart
bitfriend: I thought it is reasonable to initialize `pageController` and `screenshotController` at the same place.
Your idea is great, because `screenshotController` needs no arguments.

bitfriend: Fixed by https://github.com/acterglobal/a3/pull/522/commits/0f4fee6a5f70b48381102fcf28602f0e690e86d1

PR #567, opened 2023-03-20
update and unify theme

gtalha07:
### Description:
- Merged seperated themes into one ```ThemeData```  globally accessible across the app.
- Updated feature themes with new colors and styles.
- Improved navigation rail style.


bitfriend (approved):
Great work.
I found you made color as variable and used this color palette for all colors.
When design team changes color palette, we can see that change easily by applying it to flutter color table.
I like that you use built-in architecture that material provides.

PR #584, opened 2023-04-14
disable android build workflow

gtalha07:


gnunicorn:
I've just removed the file ;) .

PR #587, opened 2023-04-19
News missing features and UI updation.

gtalha07:


gnunicorn (approved):

app/pubspec.yaml
```diff
@@ -46,6 +46,7 @@ dependencies:
   google_fonts: ^2.3.1
   flutter_link_previewer: ^3.0.1
   bubble: ^1.2.1
+  colorize_text_avatar: ^1.0.2
```
gnunicorn, line 49: this package is unfortunately insufficient for our use case, for two reasons:
1) it doesn't allow for different text inputs to display and to calculate the color
2) the color is just picked based of the starting letter [see here](https://pub.dev/documentation/colorize_text_avatar/latest/colorize_text_avatar/colorData.html) - rather than the entire input. this will not make distinct enough colors (quite the opposite things starting with the name letter will even have the exact same letter & color making it even harder to distinguish them).

it's fine to go for this for now, but we have to replace it rather soon...

bitfriend:
Because `ffi-gen` doesn't support vector of non-primitive, we can't use `Vec<NewsSlide>` as function parameter.
So I split slide vec into single slide parameter.

PR #591, opened 2023-05-03
Switching to multiavatar for user icon fallback

gnunicorn:
Look:
grafik


gtalha07 (approved):
Looks good to me. Nice work!

PR #594, opened 2023-05-05
rename custom-avatar to user_avatar

gnunicorn:


gtalha07 (approved):
LGTM. I think this is better rather than having multiple avatar widgets. Please remove the avatar widget from features folder if its present.

gnunicorn:
superseeded by #602

PR #597, opened 2023-05-10
Fix the old issues in github actions runner

bitfriend:
Will fix the old issues that occurred in the past PRs but not fixed.


bitfriend:
`intl` version (0.17) conflicted with flutter version in github runner.
`intl 0.18` was needed to fix that issue.
So I did run this command `flutter update-packages --force-upgrade` in according to [this comment](https://github.com/flutter/flutter/issues/117163#issuecomment-1450806058).

bitfriend:
I hope this PR would fix that the last nightly build failed and desktop build was missed.

PR #599, opened 2023-05-12
Space widgets

gtalha07:


app/lib/common/widgets/square_avatar.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:flutter/material.dart';
+
+class SquareAvatar extends StatelessWidget {
```
gtalha07, line 3: I'm seeing the possibility to get it integrated in custom avatar we have. By introducing ```isSpace``` flag? What do you think?

kokodavid: @gtalha07 That would work , what about the shape of the avatars ? I presume the custom avatar is circular while Spaces need square ones .

PR #604, opened 2023-05-16
Append flutter 3.10 issue to dev doc

bitfriend:
Append something to dev doc.

### DIFF
--- a/docs/content/docs/getting-started/troubleshooting.md
+++ b/docs/content/docs/getting-started/troubleshooting.md
@@ -85,4 +85,4 @@ Many packages (incl. `intl`) are not ready for `3.10`, so we can't use `flutter 
 We continue to use `flutter 3.7.12` unless they are not ready.
 
 Please run `flutter --version` and if current version is greater than `3.7`, reinstall `3.7.12`.
-Flutter doesn't support downgrade, so you have to delete `3.10` and install `3.7` newly.
+Try running `flutter downgrade 3.7.12`. That, however, seems to not be supported for all platforms. So, if it fails, you have to delete `3.10` and install `3.7` newly.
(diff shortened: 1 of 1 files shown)


docs/content/docs/getting-started/troubleshooting.md
```diff
@@ -65,3 +65,13 @@ Release build: 223MB
 When using release build of `libacter.a`, the error `Invalid argument @ io_fread` disappeared.
 The compile command of release build is the following:
 `cargo make --profile release ios`
+
```
gnunicorn, line 68: maybe let's add the error case people would, see, too?

`````suggestion
## Resolving package `intl 0.18` fails

If you see
```
Resolving dependencies...
Because acter depends on flutter_localizations from sdk which depends on intl 0.18.0, intl 0.18.0 is required.
So, because acter depends on intl 0.17.0, version solving failed.
Exited (1)
```

You are using Flutter 3.10 while we only support 3.7 so far. See [Flutter 3.10 issue](#flutter-3-10-issue).

`````

File: docs/content/docs/getting-started/troubleshooting.md [hunk 2]
```diff
@@ -65,3 +65,13 @@ Release build: 223MB
...
+## Flutter 3.10 issue
+
+`Github` is using the latest version of flutter (`3.10` as of 5/11/2023).
+Many packages (incl. `intl`) are not ready for `3.10`, so we can't use `flutter 3.10`.
+(Our project is using `flutter_localizations` and it is using `intl`.)
+We continue to use `flutter 3.7.12` unless they are not ready.
+
+Please run `flutter --version` and if current version is greater than `3.7`, reinstall `3.7.12`.
+Flutter doesn't support downgrade, so you have to delete `3.10` and install `3.7` newly.
```
gnunicorn, line 77: what? Here on Linux I can do `flutter downgrade 3.7.12` and it does that...

```suggestion
Try running `flutter downgrade 3.7.12`. That, however, seems to not be supported for all platforms. So, if it fails, you have to delete `3.10` and install `3.7` newly.
```

gnunicorn (approved):
Thanks, mate. Just some minor additions.

PR #609, opened 2023-05-17
New bottombar, sidebar, routing, quickjump & more stubs

gnunicorn:
Brings the overhall of the full routing system and UX around that with:
- [x] newly defined 5-items bottom bar
- [x] location-driven state changes of navigation items
   - [x] rebased on riverpod for state management 
 - [x] quick-jump (desktop only)
     - [x] with `cmd+k` / `ctrl+k` shortcut
- [x] avatar-icons leading to profile-page
- [x] enum-based routing for better typed routing needs
- [x] overlay routing
    - [x] modals (used for `bug-report` and `quickjump`)
    - [x] sidebar actions (example: quick-action -> add task -> sidebar)
- [x] default page header with expanding sliver and background gradient
- [x] 404 error page  
- [x] new stub pages for
   - [x] search page on mobile
   - [x] activities page 

In action it looks like this:

https://github.com/acterglobal/a3/assets/40496/ccf6feb3-a8d4-473a-a6b2-c82f3310ad5c

https://github.com/acterglobal/a3/assets/40496/85a740cf-6da1-48a3-9fb0-7720d0b23d9b


app/lib/features/home/providers/navigation.dart
```diff
@@ -0,0 +1,229 @@
...
+          error.toString(),
+          style: Theme.of(context).textTheme.labelSmall,
+          softWrap: false,
+        ),
+        location: null,
+      )
+    ],
+    data: (spaces) {
+      spaces.sort((a, b) {
+        // FIXME probably not the way we want to sort
```
gtalha07, line 44: The most traditional way to interact with data is to be done using ```StateNotifierProvider``` with State object (Space) as a param. But wit async crud, this usually won't be necessary.

app/lib/features/home/providers/navigation.dart
```diff
@@ -0,0 +1,229 @@
...
+    data: (spaces) {
+      if (spaces.isEmpty) {
+        return features;
+      }
+      return [...features, ...spaces];
+    },
+  );
+});
+
+final currentSelectedSidebarIndexProvider =
```
gtalha07, line 160: A lot cleaner way in my mind would go something like having ```BottomBarNavigationItem``` as immutable state object (achievable) with freezed and doing interaction calls inside ```StateNotifierProvider<BottomBarNavigationItem>```.

app/lib/main/routing/routing.dart
```diff
@@ -0,0 +1,252 @@
...
+import 'package:acter/common/dialogs/error.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+// ignore: implementation_imports
+import 'package:go_router/src/information_provider.dart';
+
+import 'package:acter/main/routing/routes.dart';
+
+final desktopPlatforms = [
```
gtalha07, line 25: could remove this and import from utils constant (already defined) as this is common use-case for us.

gtalha07 (approved):
This brings many good and major changes. Great work overall!

app/lib/features/home/providers/navigation.dart
gnunicorn: the currentlocation notifications was quite a tricky part and I am not sure I get - entirely - what you are proposing but I am open to trying it.

gtalha07: See the ```bug_report_state.dart``` where we have ```BugReport``` freezed class and StateNotifier controlling state object.

gnunicorn: I see, but the problem is that we don't have an interaction to update the index but instead are basing that off `ref.watch(currentLocationProvider)` ....

not sure how you want to use a freezed class here or what benefit that has ...?

gtalha07: Oh...I was thinking maybe location updates based on user navigation.

PR #616, opened 2023-05-22
Nightly build crashing on window resize event in macos

bitfriend:
Will fix [issue613](https://github.com/acterglobal/a3/issues/613)


gnunicorn:
@bitfriend how does this fix the issue? Can we have a link to something describing the problem and put a comment on the line in the code so it doesn't get removed and cause the problem to re-occur in the future!?!

gnunicorn (approved):
looks fine.

gnunicorn:
thanks!

bitfriend:
2 videos that Emil uploaded was helpful to me.

PR #618, opened 2023-05-23
Settings and persistant labs flags

gnunicorn:
We now have some general settings (thanks to `settings-ui`), including basic build and license info screens. And we have the new `Labs`-area in there, too. Allowing the user to toggle on/off experimental feature and us to guard things behind them for testing.  It even allows for a set of default-features that are active unless the user didn't turn them off at some point prior. Allowing for the nice process of:
1. add a new feature with behind a feature flag
2. have user test it
3. put the feature flag on by default
4. remove the feature flag and just have the feature.

It uses riverpods and saves the settings to the sharedPreferences, so it is _live_ (where possible) and persistent between restart (even installs).

See it in action:

lab-settings


gtalha07 (approved):
LGTM and hands-on for huge work!

PR #619, opened 2023-05-24
Nightly build crashing on window resize event

bitfriend:
`hasShadow` didn't fix crash on mac.
And @gtalha07 reported that this crash got on not only mac but also windows.
This PR is subsequent to [PR#616](https://github.com/acterglobal/a3/pull/616).


gnunicorn (approved):
go for it.

PR #621, opened 2023-05-25
project lib restructure

gtalha07:
- Keep providers folder common for features with state notifiers residing in it to make better async pattern overall.
- correct some code files directory and typos.


bitfriend (approved):
Looks good to me

app/lib/common/providers/common_providers.dart
```diff
@@ -0,0 +1,81 @@
+import 'package:acter/common/models/profile_data.dart';
```
gnunicorn, line 1: isn't this filename a bit too much of a catch-all? like account, network state and loading, chat and profile data are clearly not the same category of things ...

app/lib/router/router.dart
```diff
@@ -1,8 +1,7 @@
 import 'package:acter/common/dialogs/dialog_page.dart';
```
gnunicorn, line 1: I think this file-rename is also confusing. if anything it should be `router/routes.dart`, no?

gnunicorn (approved):
Overall fine with me, but two things I think we should change still. Can be done either in this PR or afterwards, though.

app/lib/common/providers/common_providers.dart
gtalha07: Yes, I agree with not belonging to same categories but I named it as in the specifics of containing the providers which are global in application layout and is intended to be used in more than one features.

app/lib/router/router.dart
gtalha07: Hmm, I have created another file with similar name which includes our route definitions in ```utils``` directory. This is more like our navigation function logic. How does that sounds?

PR #622, opened 2023-05-30
Replace signup with register

bitfriend:
Will replace `signin/signout/signup` with `login/logout/register`.


bitfriend:
I replaced `signUp` with `register` in `app/lib/l10n/app_en.arb` and `app/lib/l10n/app_de.arb`.
If you get parsing error about `AppLocalizations.of(context)!.register`, please run `flutter gen-l10n` in `/app` directory.

PR #630, opened 2023-06-01
Some short fixes before GetX refactor

gtalha07:
- Correct file namings and their semantics as most files seems to have same names due to different merges.
- Remove duplicate widgets and providers which were performing same goals.
- Make Top Nav bar in ```SpaceShell``` responsive (now adapts to mobile view).


bitfriend (approved):
Looks good to me

PR #632, opened 2023-06-03
correct acter avatars implementation

gtalha07:


app/lib/features/chat/widgets/pending_req_list_view.dart
```diff
@@ -23,12 +22,6 @@ class PendingReqListView extends StatelessWidget {
         leading: ActerAvatar(
           mode: DisplayMode.User,
           uniqueId: userId,
-          avatarProviderFuture: avatar != null
```
gnunicorn, line 26: like here as well... Why is it gone without replacement? Some info in the PR description would have been good.

app/lib/features/chat/widgets/list_item.dart
```diff
@@ -53,12 +52,7 @@ class _ChatListItemState extends State<ListItem> {
           leading: ActerAvatar(
             mode: DisplayMode.GroupChat, // FIXME: checking for DM somehow?
             uniqueId: roomId,
-            avatarProviderFuture: widget.room.avatar != null
```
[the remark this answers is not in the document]
gtalha07: Yeah, I did some changes in the acter avatar code, so I didn't do correct cast types there. It's not accepting parameter here, I disabled the avatars for chat now (will fix them during refactoring of chat).

app/lib/features/chat/widgets/pending_req_list_view.dart
gtalha07: Yeah, PR description is forgotten. Apologies, I'll add it right away.

PR #635, opened 2023-06-06
Support registration-token driven UI-test

gnunicorn:
With this patch merged, we can run the flutter UI integration tests against the publicly hosted synapse-matrix instances for testing. The server `m-1.acter.global` has the appropriate mock data and accounts set up and can be used from anywhere. Note that, in order to prevent spam and abuse, you need a registration token to register, which is also prefixed to all mock-account passwords - thus is also needed for each test run. See the docs for details.

Specifically this contains:

- Refactor the auth module for better registration and registration-token-driven registration support, make `testing` module reuse as much code as possible to ensure we are not diverging
- mock-cli update for registration token
- flutter integration tests update for registration-token support
- docs explaining it all.


gtalha07 (approved):
LGTM!

PR #642, opened 2023-06-13
String deallocation fix

bitfriend:
Clone string item using `String::from()` to deallocate string with `toDartString()`.
Among the functions that returns `Vec<String>`, will apply this way for only functions that calculates string vector indirectly.
Will not touch the functions that fetches from direct property of crate.
image


app/packages/rust_sdk/lib/acter_flutter_sdk_ffi.dart
```diff
@@ -421,6 +421,11 @@ class FfiString {
     final parts = _api._ffiStringIntoParts(_box.borrow());
     final ffi.Pointer<ffi.Uint8> tmp2_0 = ffi.Pointer.fromAddress(parts.addr);
     final tmp1 = utf8.decode(tmp2_0.asTypedList(parts.len));
+    if (parts.capacity > 0) {
```
gnunicorn, line 424: why has this been re-introduced? We had confirmed this causes crashes.

bitfriend:
But github runner lint considers `String::from()` as unnecessary function.
https://github.com/acterglobal/a3/actions/runs/5255011628/jobs/9494356327
So I deleted `String::from()`.

Issue #579, opened 2023-04-06
App crashes when any space is selected from navigation rail. 


gtalha07:
Issue appears to be coming in windows desktop build (debug).
**Logs** from ```flutter --verbose``` just before moment of crashing:
```
[ +995 ms] Service protocol connection closed.
[   +2 ms] Lost connection to device.
[   +2 ms] DevFS: Deleting filesystem on the device (file:///C:/Users/Talha/AppData/Local/Temp/app26c15bff/app/)
[ +258 ms] Ignored error while cleaning up DevFS: TimeoutException after 0:00:00.250000: Future not completed
[   +2 ms] "flutter run" took 28,574ms.
[ +261 ms] ensureAnalyticsSent: 258ms
[        ] Running 1 shutdown hook
[  +13 ms] Shutdown hooks complete
[        ] exiting with code 0

```

gnunicorn:
is this still happening?

bitfriend:
Nested `ref.watch()` doesn't seem to work in flutter release mode.
For now it looks like the best solution to replace nested `ref.watch()` with simple `ref.watch()`.

PR #643, opened 2023-06-13
Fix space navigation crash

bitfriend:
Will fix issue#579.
https://github.com/acterglobal/a3/issues/579
I am sure that space navigation crash comes from nested `ref.watch()`.
Nested `ref.watch()` works in debug build.
But it seems to request big memory in release build.


bitfriend:
BTW rust sync callback seems to not work on release runtime yet.

gtalha07 (approved):
LGTM. But it didn't solved crash issue.

gnunicorn:
@bitfriend is this still relevant?

bitfriend:
Main idea of this PR was to replace nested `ref.watch()` with simple `ref.watch()`.
Now `main` branch works well without it, so we can discard this PR.

PR #644, opened 2023-06-13
replace ChatRoomController with notifiers

gtalha07:


gtalha07:
Discarding as an attempt to rewrite chat room with riverpod states.

PR #652, opened 2023-06-20
Spaces CRUD 

gtalha07:


bitfriend:
When creating space, we will call `setTopic()` after `createActerSpace()`.
image
Ironically the power level of creator is 0 not 100 after room is created.
image
To avoid failure of permission checking, Ben recommended to extend `CreateSpaceSettings`.

PR #654, opened 2023-06-21
Update NDK to r25, cargo-ndk to 3.2, unbreak Android

gnunicorn:
Since nightly-2023-06-10 the built for Android wasn't properly working, caused by us dropping the NDK version to below r23c. This updates to latest NDK 25 (so you might have to install that via the Android platform tools) and uses the cargo-ndk 3.2 series (which you might want to install, too with `cargo make setup-android`) to built the library again.

To fix the `-lgcc` problem that came from the previous update including the new `sqlite`-feature, we had to drop the `-Zbuild-std` from `cargo-ndk` and instead include the prebuilt android SDK so that libsqlite can piggy back off that. Unfortunately, `-Zbuild-std` was a useful optimisation which decreased the built size a lot (around 40mb for the entire App before is now 111mb) , so we want to get back to it when we can. But for now it is more important to get android nightly to work again.

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -34,7 +34,7 @@ jobs:
           #   os: ubuntu-latest
           #   cargo_make_setup: setup-android
           #   cargo_make_args: android-arm
-          #   with_ndk_version: r22b
+          #   with_ndk_version: r23c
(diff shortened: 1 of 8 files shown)


bitfriend (approved):
Looks good to me

PR #656, opened 2023-06-21
Trigger first-sync before refreshing all history

gnunicorn:
This revamps the first-sync-state manager in two significant ways:
1. it is its own StateNotifier with proper observability now
2. the firstSyncedRx doesn't wait for the (potentially never ending) refresh_history_on_start before triggering but happens before now.


bitfriend (approved):
Great work

PR #658, opened 2023-06-23
Issue524 room permission

bitfriend:
For [issue#524](https://github.com/acterglobal/a3/issues/524), Implement permission checking prior to all write operations.


gnunicorn (changes requested):
there appears to be a lot of unrelated code in the PR. at least going from the description that is only talking about space-permissions (like all the part about VS but also the event-handler). You want to review that again first and either limit the scope to the specific part or expand the description what all this is about, please?!?

bitfriend:
Reduced many changes.
Please review again.

native/acter/src/api/room.rs
```diff
@@ -147,8 +148,21 @@ impl Room {
...
+        let my_id = room
+            .client()
+            .user_id()
+            .context("User not found")?
+            .to_owned();
+
         RUNTIME
             .spawn(async move {
+                let member = room
+                    .get_member(&my_id)
```
gnunicorn, line 160: if we need to fetch `member` very often here, how about we just store it right in the `Room` object? It might have to be an `Option` as we could be facing looking at rooms, we aren't part of, but in general, that could save us quite a bit of noise and db-fetching in between...

native/acter/src/api/spaces.rs
```diff
@@ -82,182 +109,245 @@ impl Space {
...
         // Tasks
-        self.room.add_event_handler(
-            move |ev: SyncTaskListEvent,
-                  client: SdkClient,
-                  //  room: Room,
-                  encryption_info: Option<EncryptionInfo>,
-                  Ctx(executor): Ctx<Executor>| {
-                let room_id = room_id.clone();
-                async move {
+        let handle =
```
gnunicorn, line 120: yeah, I still don't have any explanation on why we have to store room handlers now ... Care to give one?

I fear that this isn't a proper way to do it, as we don't necessarily keep those spaces around forever either - we pull them in and destroy them and pull them in again at a whim... if we need to store these event-handlers for longevity, this isn't the place ...

gnunicorn (commented):
I have still open questions about this change.

PR #661, opened 2023-06-24
Fix issue in context()

bitfriend:
- Keep `.context()` for only `Option<xxx>`, because original exception is hidden by it on failure. We need original exception text on failure.


native/acter/src/api/account.rs
```diff
@@ -68,7 +65,7 @@ impl Account {
                 let buf = account
                     .get_avatar(MediaFormat::File)
                     .await?
-                    .context("No avatar URL given")?;
+                    .unwrap_or_default();
```
gnunicorn, line 68: what is `default` here? An empty buffer? Then we do not want to return that. We should not be returning empty values of things that are clearly `None`.

bitfriend: Fixed

gnunicorn (approved):
Looking good. Thanks for taking care of this!

PR #662, opened 2023-06-24
Filter out left spaces

gnunicorn:
Before this change we'd hand out all spaces, regardless of whether the underlying room was one we had joined, left or just received an invite for. The splitting code now takes a filter parameter that allows for filtering only certain types of spaces, which is applied to `client.spaces()` to filter out any spaces we had left (or were kicked off).

An integration test shows that this works as expect _and_ that the information subscription system works as expected, too. Including to inform you the space you listened to had been removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,7 @@ dependencies = [
  "acter",
  "acter-core",
  "anyhow",
+ "async-broadcast",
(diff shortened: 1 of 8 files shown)


gtalha07 (approved):
LGTM!

PR #663, opened 2023-06-26
Replace log with tracing

bitfriend:
`tracing` is more detailed than `log` and allows for a nested information tree and performance tracking.
Will encapsulate `log` and replace `log` macros with `tracing` macros.


gnunicorn (approved):
assuming that this is the same replace as the 10files I looked at, all fine :)

PR #678, opened 2023-06-27
Parse meta info of media msg on rust side

bitfriend:
- Extract mime type & file size on rust side
- Extract image dimensions on rust side (using `imagesize` crate)
- Extract audio duration on rust side (using `lofty 0.11` and `lofty_attr 0.6)
- Extract video duration & dimensions on rust side (using `metadata` or `video_ludo` crate)


bitfriend:
Ben said that mobile is more important than desktop.
He recommended to make [android meta parsing](https://docs.rs/jni-android-sys/0.0.10/jni_android_sys/android/media/struct.MediaMetadataRetriever.html) and [ios meta parsing](https://developer.apple.com/documentation/avfoundation/media_assets/retrieving_media_metadata) by platform.
Will replace the current implementation with them later.

native/acter/src/api/account.rs
```diff
@@ -72,6 +72,11 @@ impl Account {
     pub async fn upload_avatar(&self, uri: String) -> Result<OwnedMxcUri> {
         let account = self.account.clone();
         let path = PathBuf::from(uri);
+        let guess = mime_guess::from_path(path.clone());
+        let content_type = guess.first().context("MIME type should be given")?;
```
gnunicorn, line 76: This would be the user facing error, probably want to do something a bit more user friendly:

```suggestion
        let content_type = guess.first().context("Could not detect mime type. Are you sure this is an image?")?;
```

native/acter/src/api/attachments.rs
```diff
@@ -238,123 +238,233 @@ impl AttachmentsManager {
...
-            mimetype,
-            size: size.and_then(UInt::new),
-            blurhash,
-        });
-        let url = Box::<MxcUri>::from(url.as_str());
-        let mut builder = self.inner.draft_builder();
+        let r = room.clone();
+        let image_content = RUNTIME
+            .spawn(async move {
+                let url = Box::<MxcUri>::from(url_or_path.as_str()); // http not allowed for remote url
```
gnunicorn, line 257: I am not convinced about the change of to `url_or_path` ... I don't see the usecase for us in the UI of allowing the user of using an already existing (public) MxcUri - is that really being used? I'd go for only accepting path (for now ... for webassembly later we might have to offer other ways of doing this).

native/acter/src/api/attachments.rs
```diff
@@ -238,123 +238,233 @@ impl AttachmentsManager {
...
+                if url.is_valid() {
+                    return anyhow::Ok(ImageMessageEventContent::plain(body, url.into()));
+                }
+                let path = PathBuf::from(url_or_path.clone());
+                let guess = mime_guess::from_path(path.clone());
+                let content_type = guess.first().context("No MIME type")?;
```
gnunicorn, line 263: same as above, let's give the user a bit more information:
```suggestion
                let content_type = guess.first().context("Detecting mime type failed. Are you sure this is an Image?")?;
```

Applies to the other functions that follows as well, of course.

native/acter/src/api/news.rs
```diff
@@ -377,21 +377,24 @@ pub struct NewsEntryDraft {
...
+        let text_content = TextMessageEventContent::markdown(body);
+
         self.slides.push(NewsSlide {
             client: self.client.clone(),
             room: self.room.clone().into(),
-            inner: news::NewsSlide::new_text(body),
+            inner: news::NewsSlide {
+                content: news::NewsContent::Text(text_content),
+                references: Default::default(),
+            },
```
gnunicorn, line 390: Why do you move this NewsSlide creation part out of the `core` (which can be used without the dart) to the `api`? I think the internal format of `news::NewsSlide` for most of `api` should stay rather opaque and just use functions rather than know off and create it with all its fields...  Makes it a lot easier to change the internal structure of the core-types, if their primary way of being instantiated is through function calls..

native/core/src/events/news.rs
```diff
@@ -101,50 +98,6 @@ pub struct NewsSlide {
...
-        }
-    }
-
-    pub fn new_location(body: String, geo_uri: String) -> Self {
-        NewsSlide {
-            content: NewsContent::Location(LocationMessageEventContent::new(body, geo_uri)),
-            references: vec![],
-        }
-    }
-}
```
gnunicorn, line 146: as said above, I do not agree that this should all be moved from here to the `api`-crate. Quiet the opposite, I think it would be better to change the interface to taking the specific event-content as the param but keep this API.

Like:
```rust
  pub fn new_text(content: TextMessageEventContent) -> Self {
        NewsSlide {
            content: NewsContent::Text(content),
            references: vec![],
        }
    }

    pub fn new_image(content: ImageMessageEventContent) -> Self {
        NewsSlide {
            content: NewsContent::Image(content),
            references: vec![],
        }
    }
```

gnunicorn (changes requested):
I am really concerned with the lack of tests for this. The API changes I do not fully agree with (see specific comments), but the internal code looks fine. However I only glossed over most of it, as they seem very similar and it is hard to judge if that actually does what is expected. I'd very much like to see the integration tests suite for rust sending all these types using free-to-use-sample data to make sure that the API is detecting and sending correctly (with a second account receiving, decrypting and checking them against the fixture). That would really increase my confidence, also that we do not accidentally break them in the future.

Can I haz tests, pleaz!?!

gnunicorn:
outdated.

PR #683, opened 2023-06-29
New onboarding

kokodavid:


gnunicorn (approved):
Only glanced through it, seem fine enough. But there are some tests failing...

PR #687, opened 2023-06-29
Fix space tab navigation

gnunicorn:
Moves the space/* tabbar navigation management over into riverpod. Trick is not only use an external notifier we can ping from the routing, but also to delay the actual update in the notifier via `Future(() {})` or it fails terribly (with a notification that you can't change state while it renders).

Either way, before we had the navbar on each subview and re-rendering it lead to weird errors around ticks. Now there is a single one on the outer shell and it updates whenever we route to subpage, regardless of whether that was triggered by the tabbar or through any other click.


gnunicorn:
fixed in #687

PR #707, opened 2023-07-04
[Space Edit] allow setting name field and permission check before sending update

gtalha07:


app/lib/features/space/dialogs/edit_space_sheet.dart
```diff
@@ -291,6 +314,16 @@ class _EditSpacePageConsumerState extends ConsumerState<EditSpacePage> {
     }
   }
 
+  // permission check
+  Future<bool> permissionCheck() async {
+    var member = await widget.space.getMyMembership();
+    bool p1 = member.canString('CanSetTopic');
+    bool p2 = member.canString('CanUpdateAvatar');
```
gnunicorn, line 321: actually checking for one is enough: if it doesn't have the one, it doesn't have the other either.

bitfriend (approved):
Looks good to me

app/lib/features/space/dialogs/edit_space_sheet.dart
```diff
@@ -291,6 +314,16 @@ class _EditSpacePageConsumerState extends ConsumerState<EditSpacePage> {
     }
   }
 
+  // permission check
+  Future<bool> permissionCheck() async {
+    var member = await widget.space.getMyMembership();
```
gnunicorn, line 318: there is a potential race here, as the usage of `space` here doesn't bind to the riverpod notifier that updates it whenever the data comes down the pipe. Thus this isn't properly updating if e.g. you are on the screen and someone gives you admin priveleges - you'd have to reopen before this udpates. I don't think that is a critical problem here right now, but something to generally avoid when we have the chance ... better to use the riverpod ref.watch than handing around the object - riverpod.watch does a decent job of caching it for us ...

gnunicorn (approved):
general note (not critical), and a minor change request, otherwise looks good code-wise.

app/lib/features/space/dialogs/edit_space_sheet.dart
gtalha07: Yes I know, just for safety purposes. If the spec updates in later time, this could get potential of bug and weird scenarios. I can change if its important.

app/lib/features/space/dialogs/edit_space_sheet.dart
```diff
@@ -241,11 +241,33 @@ class _EditSpacePageConsumerState extends ConsumerState<EditSpacePage> {
...
+                btn2Color: Theme.of(context).colorScheme.success,
+              );
+              return;
+            }
             final roomId = await _handleUpdateSpace(context);
             debugPrint('Space Updated: $roomId');
             // refresh spaces and side bar
             ref.invalidate(spacesProvider);
             ref.invalidate(nav.spaceItemsProvider);
+            ref.invalidate(spaceProvider);
```
gtalha07, line 269: This will forcefully refresh the providers until the next frame loads .i.e. after navigation

app/lib/features/space/dialogs/edit_space_sheet.dart
gtalha07: Yes, that's a good take as we want to make things reactive as possible. My one approach to that for now is to just invalidate the providers to take on new data. Like this:https://github.com/acterglobal/a3/pull/707/files#r1251978913

app/lib/features/space/pages/shell_page.dart
```diff
@@ -113,7 +113,7 @@ class _ShellToolbar extends ConsumerWidget {
                 onTap: () => context.pushNamed(
                   Routes.editSpace.name,
                   pathParameters: {'spaceId': spaceId},
-                  extra: space,
+                  extra: spaceId,
```
gnunicorn, line 116: why have that twice?

app/lib/router/router.dart
```diff
@@ -224,7 +223,7 @@ List<RouteBase> makeRoutes(Ref ref) => [
               );
             },
             child: EditSpacePage(
-              space: state.extra as Space,
+              spaceId: state.extra as String,
```
gnunicorn, line 226: same here... why not:
`spaceId: state.queryParameters['spaceId'],` ?

app/lib/features/space/pages/shell_page.dart
gtalha07: The ```extra``` object as ```id```` is passed as data on the edit page to fetch the space. ```pathParameters``` is just for routing to correct space.

app/lib/router/router.dart
gtalha07: I will change to ```queryParameters```.

PR #708, opened 2023-07-04
Joining spaces

gnunicorn:
- [x] backend to search for spaces
- [x] join by searching public spaces
- [x] show details of room
    - [x] show joined status
- [x] endless scrolling on results 
- [x] allow to search for text
- [x] allow to select target servers
   - [x] custom set of target servers
- [x] custom target server
- [x] custom room name not in list 
   - [ ] and allow to join button
- [x] actual join attempts.
- [x] externalise public search into reusable widget. 

join_public_rooms

---

Known issues:
 - When switching to the newly joined space, the data is a bit awkward, like parent-space might be unknown and loading the name isn't fully working yet. Handing over the existing space-data as fallback would be nice
 - the ActerAvatar sometimes doesn't load the Avatars for us or fails when setting them too late
 - We can't actually tell the server to only give us _acter_ spaces, we get all Spaces (and sometimes even rooms?!?). Without pre-emptively loading extra data (which we can only do on public rooms that allow us to peak in it anyways), we can't really know whether these are just regular matrix-spaces or acter spaces in the listing unfortunately :( ....


app/lib/common/providers/space_providers.dart
```diff
@@ -21,12 +21,33 @@ final spaceProvider =
   () => AsyncSpaceNotifier(),
 );
 
+final maybeSpaceProvider = AsyncNotifierProvider.autoDispose
```
gtalha07, line 24: could you add brief word about the purpose of provider ? As its gonna keep increasing so having them commented is better :)

app/lib/common/utils/constants.dart
```diff
@@ -57,3 +64,25 @@ const List<TargetPlatform> desktopPlatforms = [
 
 // hide bottom nav at locations.
 const List<String> hideNavLocations = ['/updates/post', '/updates/edit'];
```
gtalha07, line 66: Not related to this PR but this constant needs to be removed. Nav bar can be hidden by using ```rootNavigatorKey``` (was not in knowledge at the time)

gtalha07 (approved):
Overall looking good. Nice!

bitfriend (approved):
Great work

PR #710, opened 2023-07-04
Append CanSetName for permission check

bitfriend:
- Append CanSetName for permission check
- Improve code
- Change error log
- Fix typo

### DIFF
--- a/native/acter/src/api/room.rs
+++ b/native/acter/src/api/room.rs
@@ -346,7 +346,10 @@ impl Room {
 
         RUNTIME
             .spawn(async move {
-                let member = room.get_member(&uid).await?.context("User not found")?;
+                let member = room
+                    .get_member(&uid)
+                    .await?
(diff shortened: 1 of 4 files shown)


native/acter/src/api/room.rs
```diff
@@ -149,12 +151,15 @@ impl Room {
     }
 
     pub async fn get_my_membership(&self) -> Result<Member> {
-        let SdkRoom::Joined(joined) = &self.room else {
+        let room = if let SdkRoom::Joined(r) = &self.room {
```
gnunicorn, line 154: any particular reason you prefer this over the other style?

gnunicorn (approved):
lgtm

native/acter/src/api/room.rs
bitfriend: Newer style was introduced by you before.
It looked better to me and I kept that style as much as possible.
That's all.

PR #711, opened 2023-07-04
Fix bug in list index of my spaces section

bitfriend:
Will fix the bug that my spaces list in right side is not displayed properly.
image

### DIFF
--- a/app/lib/features/home/widgets/my_spaces_section.dart
+++ b/app/lib/features/home/widgets/my_spaces_section.dart
@@ -1,16 +1,17 @@
 import 'dart:core';
 
+import 'package:acter/common/providers/space_providers.dart';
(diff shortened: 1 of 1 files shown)


bitfriend:
This is fixed screenshot.
image

app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -115,6 +75,47 @@ class MySpacesSection extends ConsumerWidget {
       ),
     );
   }
+
+  Widget? _spaceBuilder(
```
gtalha07, line 79: Widget method patterns aren't performance friendly, so try to have it minimised whenever possible and instead extract widgets or use stateless widget patterns.

PR #712, opened 2023-07-05
Restore verification that was deactivated when switching from GetX to Riverpod

bitfriend:
Verification was deactivated when switching from `GetX` to `Riverpod`. This switches the cross-signing verification to riverpods, too and thus reactivate the feature.


gnunicorn (approved):
oh, exciting!

gtalha07 (approved):
Nice work!

bitfriend:
This PR contains `tokio::spawn()`, so I tested this PR in release build too.

PR #717, opened 2023-07-05
Fix news posting

gnunicorn:
- [x] simplified news posting UI 
- [ ] ~~figure out why sync events aren't processed~~ refs #719 
- [x] hide previous stories-editor-based-ui
- [x] caching downloaded media makes the UI load them a lot faster on subsequent loads
- [x] post don't show up - see 1
- [x] show markdown text news - see 2
- [x] limiting on permissions - see 3

1: Prior to this change, we didn't actually fill the image-news-slide properly in the sending UI -- it never worked. Now we are actually adhering to whether the room/space is encrypted and encrypt the image before uploading and then use the following data to send a proper `ImageEventContent` (either encrypted or not).  
2: This introduces a new util widget `RenderHtml` that wrapps the `flutter_matrix_html` with additional support for internal links. This, in combination, with the activated `markdown`-feature on the core (and newly exposed methods for it), allows us to have html-based News entries with internal and external links.
3: newly added permissions checking and a provider for it make it easy for any feature to be checked for, and only show button when the user has at least one space with permissions of it. 

---

New creator UI:

grafik

Showing news:

grafik


native/acter/src/api/news.rs
```diff
@@ -360,31 +372,62 @@ impl NewsEntryDraft {
...
+        let is_encrypted = self.room.is_encrypted().await?;
+
+        let path = PathBuf::from(uri);
+        let mime_type = mimetype.parse::<mime::Mime>()?;
+        let mut image_content = if is_encrypted {
+            let encrypted_file = RUNTIME
+                .spawn(async move {
+                    let mut reader = std::fs::File::open(path)?;
+                    let encrypted_file = client
+                        .prepare_encrypted_file(&mime_type, &mut reader)
```
bitfriend, line 395: Nice bugfix

bitfriend (approved):
Great work, especially for encrypted upload.

PR #720, opened 2023-07-07
Fix logout/login views

gnunicorn:
- [x] move logout button to my-profile-page
- [x] clear/refresh UI when we logged out
- [x] logout when we get 'Invalid Access Token' from the server
- [x] back in login-page, registration-page
- [x] snackbar crash in login-page  
- [x] snackbar not visible on failed login

Also contains a fix for the async-triggers not firing. See how the spaces are refreshed now after login:

spaces-appaearing


app/packages/rust_sdk/lib/acter_flutter_sdk.dart
```diff
@@ -317,10 +318,11 @@ class ActerSdk {
...
   }
 
-  Future<void> logout() async {
+  Future<bool> logout() async {
     // remove current client from list
-    var client = _clients.removeAt(_index);
+    final client = _clients.removeAt(_index);
     _index = _index > 0 ? _index - 1 : 0;
+    print("Remainig clients $_clients");
     await _persistSessions();
```
gnunicorn, line 326: oddly this seems to write the wrong sessions... despite us having removed the item :( ...

gnunicorn: appears to have been a problem of having more than one client for some reason.

PR #732, opened 2023-07-10
Replace description_text with description

bitfriend:
Replace `description_text()` with `description()` in task and calendar event


gnunicorn:
Any particular reason to drop the `_text` suffix? The idea here was to also allow for a `_formatted` and `_markdown` variants and have the difference clear of what to expect.

bitfriend:
It was confused that we are using both `description_text()` and `description()`.

PR #772, opened 2023-07-19
Notifications, the prolog

gnunicorn:
Starting the notifications infrastructure:
- [x] load remote notifications from server
- [x] show notifications history in activities stream, as endless loader
- [x] read and issue notifications from the sync-loop
- [x] show simple local desktop notification for new notifications received
- [x] show actual content of notification in activity stream & notification
- [x] make it available behind a feature-flag (also cleans up the feature-flag section a bit)
- [x] ensure it works outside the activities-stream
- [ ] check on platforms
   - [x] linux
   - [ ] windows
   - [ ] mac
   - [x] iOS
   - [x] Android

Notes:
 1. Skipping desktop for now. Needs to be implement in our fork of awesome_notifications.
 2. This is not yet fully hooked up for push notifications, but will only work if the app is running
 3. notifications content, especially for chat messages in encrypted rooms, is very barebones

notification


gnunicorn:
on mobile currently seeing two notifications.

gnunicorn:
Upgrading to latest flutter 3.10 leads to quick-notify, which we used on desktop, to not work. I ripped it out and cleaned up the other parts so slightly. But it means desktop notifications do not yet work (other than for linux, where I did implement it in a fork of awesome_notifications we are using).

Additionally, the routing is broken for chat-messages, too..

app/lib/common/notifications/models.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:acter/common/utils/routes.dart';
+import 'package:acter/common/utils/utils.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' as ffi;
+
+class NotificationBrief {
```
gtalha07, line 5: Why not just compose it with ```freezed``` class? Makes it easier for handling states in ```StateNotifier``` and for immutability?

app/lib/features/activities/pages/activities_page.dart
```diff
@@ -1,12 +1,141 @@
...
+              data: (value) => Text(value.displayName ?? roomId),
+              error: (error, stackTrace) =>
+                  Text('Failed to load room due to $error'),
+              loading: () => Text(roomId),
+            );
+          },
+        );
+      }
+    } else {
+      avatar = null;
```
gtalha07, line 111: This would have a potential of throwing exception in widget building as it can't render ```NULL```. Make it ```const SizedBox.shrink()```

app/lib/features/activities/providers/notifiers/notifications_list_notifier.dart
```diff
@@ -0,0 +1,128 @@
...
+  NotificationListState addNotification(ffi.Notification n) {
+    final r = records ?? [];
+    r.insert(0, n);
+    return copyWith(
+      records: r,
+    );
+  }
+
+  @override
+  NotificationListState copyWith({
```
gtalha07, line 38: Do you think we could achieve this by ```freezed``` with minimal code?

app/lib/features/home/providers/notifiers/sync_notifier.dart
```diff
@@ -38,6 +41,8 @@ class SyncNotifier extends StateNotifier<LocalSyncState> {
       }
     });
 
+    // final notificationsPoller = syncState.notificationsStream();
```
gtalha07, line 44: Remove commented lines on finalising :)

app/lib/main.dart
```diff
@@ -34,15 +35,31 @@ Future<void> startAppInner() async {
...
+  const Acter({super.key});
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() => _ActerState();
+}
+
+class _ActerState extends ConsumerState<Acter> {
+  @override
+  void initState() {
+    setupNotificationsListeners();
```
gtalha07, line 55: The most recommended way is to have this line below ```super.initState()``` as we should have everything else after framework builds widget and vice versa in case of ```super.dispose()```

gtalha07 (commented):
Overall looks good. Just requested minor changes and some  pattern improvements (which aren't necessary at the moment)!  🚀

app/lib/features/activities/pages/activities_page.dart
gnunicorn: but `Null` is a valid `child`-value for `leading: SizedBox(height: 50, width: 50, child: avatar),`. I'd rather keep the tree smaller and use `null` when things are `null` (and `null` is supported).

app/lib/features/activities/providers/notifiers/notifications_list_notifier.dart
gnunicorn: Maybe, I don't know. I don't know `freezed` enough and was just going from other examples provided ...

gtalha07: If I understand correctly, you're overriding ```copyWith```. The ```NotificationListState``` can be simplified with freezed like class with generated constructor and ```copyWith``` method.

app/lib/common/notifications/models.dart
gtalha07: By converting it into freezed, it automatically generates class constructor and ```copyWith``` methods with minimal code sample. If you think that won't benefit simplicity here, you can ignore it :)

app/lib/features/activities/pages/activities_page.dart
gtalha07: I'm not sure reaching this state is very low probability, and to avoid UI exceptions for ```build()``` as what it would render in ```NULL``` case ?

gtalha07: If this is tested, then we can safely ignore it 👍🏻

PR #814, opened 2023-08-01
improve reply view and and implement swipe to reply

gtalha07:
Screenshot 2023-08-02 at 1 11 39 AM
Screenshot 2023-08-02 at 1 12 01 AM


gnunicorn (approved):
Looking good!

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -154,8 +126,77 @@ class _ChatBubble extends ConsumerWidget {
...
+                                            .copyWith(
+                                              color: Theme.of(context)
+                                                  .colorScheme
+                                                  .tertiary,
+                                            ),
+                                      ),
+                                    ],
+                                  ),
+                                ),
+                                _OriginalMessageBuilder(message: message),
```
gnunicorn, line 182: I am wondering though, this can be arbitrary large messages ... should we cap this somehow? Only show the top 100px or something... but maybe we just wait until we see that as a problem being reported...

PR #816, opened 2023-08-02
Upgrade sdk

bitfriend:
- This update doesn't cover the latest `matrix-sdk`, because the latest version needs `proc-macro2 v1.0.66`. If `proc-macro2` is not `v1.0.55`, `genco` causes error `Your compiler does not support spans which is required by genco, see: https://github.com/rust-lang/rust/issues/54725`.
- `proc-macro2` version pinning is removed from `a3` project, and it is kept in `ffi-gen`.
- Changed some code in login, because new `matrix-sdk` introduced `AuthSession`.


native/acter/api.rsh
```diff
@@ -128,13 +128,13 @@ object NewsEntryDraft {
     fn add_image_slide(body: string, url: string, mimetype: string, size: Option<u32>, width: Option<u32>, height: Option<u32>, blurhash: Option<string>) -> Future<Result<bool>>;
 
     /// create news slide for audio msg
-    fn add_audio_slide(body: string, url: string, secs: Option<u32>, mimetype: Option<string>, size: Option<u32>);
+    fn add_audio_slide(body: string, url: string, mimetype: string, secs: Option<u32>, size: Option<u32>);
```
gnunicorn, line 131: am I missing something or how are these changes here and in the `attachments.rs` related to this upgrade?

native/acter/src/api/news.rs
```diff
@@ -431,78 +431,159 @@ impl NewsEntryDraft {
         Ok(true)
     }
 
-    pub fn add_audio_slide(
+    pub async fn add_audio_slide(
```
gnunicorn, line 434: same here ... these do not seem to be related to the sdk-upgrade, are they?

native/core/src/support.rs
```diff
@@ -10,5 +10,15 @@ pub struct RestoreToken {
     /// Server homebase url
     pub homeurl: Url,
     /// Session to hand to client
-    pub session: Session,
+    pub session: CustomAuthSession,
```
gnunicorn, line 13: eeeer... does this mean the old stored tokens do still deserialize?

gnunicorn (changes requested):
There seems to be quite some unrelated content in here... please clean that up first ...

native/core/src/support.rs
bitfriend: Yes, I introduced `CustomAuthSession` to support token in older format.

Cargo.toml
```diff
@@ -8,29 +8,30 @@ default-members = ["native/acter"]
 
 [workspace.dependencies.matrix-sdk]
 git = "https://github.com/matrix-org/matrix-rust-sdk"
-rev = "645af31c59c9310f822dabae87166c0ea7d29bd5"
+rev = "2b72df7c0611020bdfbaf5e089888d9383774c6f"
```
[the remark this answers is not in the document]
bitfriend: If we use the latest version, we will meet `indexmap` issue.
As I wrote in this PR's description, it occurs version conflict between several crates, and it looks hard to fix them.

native/acter/api.rsh
bitfriend: Fixed

native/acter/src/api/news.rs
bitfriend: Fixed

gnunicorn:
@bitfriend the integration test didn't fail on `main`... needs to be investigated, can't risk that spaces don't load ...

native/acter/src/api/spaces.rs
```diff
@@ -602,22 +602,13 @@ impl Client {
...
-                if let Some(space_alias) = space.inner.room.canonical_alias() {
-                    if space_alias == alias_id {
-                        return Ok(space);
-                    }
-                }
-            }
-            bail!("Room with alias not found");
         } else {
-            bail!("Neither roomId nor alias provided");
+            bail!("roomId not provided");
```
gnunicorn: I see what you did here... but wonder ... this is async ... shouldn't we instead have that `resolve_alias`-call be a fallback if we can't find the room here? I'd very much prefer that

bitfriend: Fixed

gnunicorn (approved):

gnunicorn (changes requested):
just noticed the ffi-gen branch this pins to now  is not on the latest state (missing changeset) compared to the one we had before. please upgrade the branch or merge it into the other branch on the ffi-gen-project before merging.

Issue #795, opened 2023-07-28
Show space-parent for chats and in sidebar


gnunicorn:
In the chat listing as well as on the sidebar in the desktop version we only show the avatar itself, but not the parent-space they might belong to. This leads to some confusion. We should show the related space icon (inclusive it being clickable, have a the tooltop and be liked to the proper space) on the bottom-right here as well -- same way we show it in the top bar of a space itself.

Two side notes to add:
1. this should not delay the loading of the outer icon but should only happen _after_ we have loaded the space itself
2. we should make that part of `acter-avatar` - to allow for nested avatars on all four corners.

gnunicorn:
Similarly, we want the space-icon in the chat-bar for each item, too.

PR #819, opened 2023-08-02
Remote child-space data

### LINKED ISSUE #790: Loading spaces failure
for an unknown reason, under various circumstances loading of rooms fails and the screen or section is just showing the message `Loading spaces failed: room not found`. This has at least been reproduced for cases where we attempt to load a room in relevance to our current room, but don't have that loaded yet (because we lack the access or just aren't done loading yet), like:

I am part of a space where but am not part of some related space
  - like e.g. in the top-bar, if I am not part of the parent-space
  - related spaces or chats in the overview as well as the "spaces"-section (both in terms of "parents", children or related space)

We probably want to attempt to load the space from remote if we can't find it in these situations, but we also have to think about the problem that we might still not be able to load that data at all because we lack permissions or currently don't have access to the right server: what should we show then?

gnunicorn:
- [x] load space-hierarchy data from remote
- [x] show that in related-spaces sub-bar
    - [x] pretty-show for subspaces 
- [x] allow to join when possible
- [x] deal with rooms we have left
- [x] deal with unknown parents
- [x] implement related chats in this model, too
- [x] de-duplicate lists (with already joined list)
- [x] clean up parent loading

properly-tagged-subspaces-and-convos

Fixes #790, #795


bitfriend (approved):
Great work

PR #821, opened 2023-08-02
activate chat reactions selection and improvements

gtalha07:
- [x] Activate chat reactions and visual improvements. Message tap to show message actions
- [x] Reply fixes: earlier wasn't showing encrypted events as part of reply content.
- [x] Chat list fixes: minor duplication bug.

### Bug: Redacted Events cannot be rendered as part of reply, as it couldn't fetch event id.

Screenshot 2023-08-03 at 12 45 42 AM
Screenshot 2023-08-03 at 12 46 02 AM
Screenshot 2023-08-03 at 12 46 22 AM

### DIFF
--- a/.changes/0821-chat-fixes.md
+++ b/.changes/0821-chat-fixes.md
@@ -1,0 +1,4 @@
+- [fix] activate chat reactions.
+- [improvement] chat reactions ui.
+- [fix] reply content fixes.
+- [fix] chat list duplication bug.
--- a/app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
+++ b/app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
@@ -5,17 +5,13 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 class ChatInputNotifier extends StateNotifier<ChatInputState> {
   ChatInputNotifier() : super(const ChatInputState());
 
-  void setReplyView(bool value) {
-    state = state.copyWith(showReplyView: value);
-  }
+  void toggleReplyView(bool value) =>
+      state = state.copyWith(showReplyView: value);
 
-  void showSendBtn(bool value) {
-    state = state.copyWith(sendBtnVisible: value);
-  }
+  void showSendBtn(bool value) => state = state.copyWith(sendBtnVisible: value);
 
-  void toggleAttachment() {
-    state = state.copyWith(attachmentVisible: !state.attachmentVisible);
-  }
+  void toggleAttachment(bool value) =>
+      state = state.copyWith(attachmentVisible: value);
 
   void emojiRowVisible(bool value) =>
       state = state.copyWith(emojiRowVisible: value);
--- a/app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
+++ b/app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
@@ -5,8 +5,8 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 class ChatInputNotifier extends StateNotifier<ChatInputState> {
   ChatInputNotifier() : super(const ChatInputState());
 
-  void toggleReplyView() {
-    state = state.copyWith(showReplyView: !state.showReplyView);
+  void setReplyView(bool value) {
+    state = state.copyWith(showReplyView: value);
   }
 
   void showSendBtn(bool value) {
@@ -17,11 +17,11 @@ class ChatInputNotifier extends StateNotifier<ChatInputState> {
     state = state.copyWith(attachmentVisible: !state.attachmentVisible);
   }
 
-  void emojiRowVisible() =>
-      state = state.copyWith(emojiRowVisible: !state.emojiRowVisible);
+  void emojiRowVisible(bool value) =>
+      state = state.copyWith(emojiRowVisible: value);
 
-  void emojiPickerVisible() =>
-      state = state.copyWith(emojiPickerVisible: !state.emojiPickerVisible);
+  void emojiPickerVisible(bool value) =>
+      state = state.copyWith(emojiPickerVisible: value);
 
   void setReplyWidget(Widget? child) {
     state = state.copyWith(replyWidget: child);
--- a/app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
+++ b/app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
@@ -10,14 +10,10 @@ import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter/features/home/providers/client_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:file_picker/file_picker.dart';
-import 'package:filesystem_picker/filesystem_picker.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 import 'package:mime/mime.dart';
-import 'package:open_app_file/open_app_file.dart';
-import 'package:path_provider/path_provider.dart';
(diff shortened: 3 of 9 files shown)


gnunicorn:
> Bug: Redacted Events cannot be rendered as part of reply, as it couldn't fetch event id.

is that still being worked on (we should probably have a placeholder for redacted messages anyways) or should we put that into a separate issue once this is merged?

.changes/0821-chat-fixes.md
```diff
@@ -0,0 +1,4 @@
+- [fix] activate chat reactions.
+- [improvement] chat reactions ui.
+- [fix] reply content fixes.
+- [fix] chat list duplication bug.
```
gnunicorn, line 4: Let's try to give more insights to actual end-users with our changes recordings (this isn't just for developers). How about:

```suggestion
- **Chat message emoji-reactions**: we now render chat-message reactions again, as well as allow you to send them by clicking in the UI.
- Chat replies fixed: rendering replies as well as composing them has been redesigned and is a lot more clear now
- messages appearing twice in a chat room should be fixed now. If you see that again, please let us know.
```

app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
```diff
@@ -17,11 +17,11 @@ class ChatInputNotifier extends StateNotifier<ChatInputState> {
     state = state.copyWith(attachmentVisible: !state.attachmentVisible);
   }
 
-  void emojiRowVisible() =>
-      state = state.copyWith(emojiRowVisible: !state.emojiRowVisible);
+  void emojiRowVisible(bool value) =>
+      state = state.copyWith(emojiRowVisible: value);
```
gnunicorn, line 21: Good. I prefer these hard bool-setters rather than toggle-switches ... it is a lot more clear from the UI point of view, what is expected to happen (and other calls in between do not mess with that state).

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -926,20 +989,23 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
     }
   }
 
-  ProfileData? getUserProfile(String userId) {
+  ProfileData? getUserProfile(String? userId) {
```
gnunicorn, line 992: why do we want to support getting a null here? Why would this ever be called with `null`? I think we should keep it as `String userId` and handle that non existence elsewhere.. 

It's more clear if there is a `userId != null ?  getUserProfile(userId) : null` that we know and handle `userid` to be `null` rather than wondering why a call of `getUserProfiel(userId)` doesn't return for the case (because the user is `null`)...

```suggestion
  ProfileData? getUserProfile(String userId) {
```

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -926,20 +989,23 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
...
     final chatProfiles = ref.watch(chatProfilesProvider);
-    if (chatProfiles.containsKey(userId)) {
-      return chatProfiles[userId];
+    if (userId != null) {
+      if (chatProfiles.containsKey(userId)) {
+        return chatProfiles[userId];
+      }
```
gnunicorn, line 997: subsequentially:
```suggestion
   if (chatProfiles.containsKey(userId)) {
     return chatProfiles[userId];
```

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -926,20 +989,23 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
...
     }
-    return ProfileData('', null);
+
+    return null;
```
gnunicorn, line 1000: I'd keep this though

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -48,22 +49,47 @@ class BubbleBuilder extends ConsumerWidget {
             onLeftSwipe: !isAuthor()
                 ? null
                 : () {
-                    ref.read(chatInputProvider.notifier).setReplyWidget(child);
-                    ref.read(chatRoomProvider.notifier).repliedToMessage =
-                        message;
-                    if (!ref.read(chatInputProvider).showReplyView) {
-                      ref.read(chatInputProvider.notifier).toggleReplyView();
+                    if (ref.read(chatInputProvider).emojiRowVisible) {
```
gnunicorn, line 52: this cries for being taken out here into a local function above to be referenced here.... considering that it is being used twice and hard to read in this indentation-level

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -151,22 +180,20 @@ class _ChatBubble extends ConsumerWidget {
                                       ActerAvatar(
                                         uniqueId:
                                             message.repliedMessage!.author.id,
+                                        displayName: replyProfile != null
```
gnunicorn, line 183: and considering that `displayName` is used twice here and below, I'd also pull out this assignment above, below the `replyProfile` and reuse the calculated value here and below.

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -94,6 +120,9 @@ class _ChatBubble extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final replyProfile = ref.watch(chatRoomProvider.notifier).getUserProfile(
+          message.repliedMessage?.author.id ?? '',
+        );
```
gnunicorn, line 125: yeah, why even? If `getUserProfile` accepted `null`, why put a `''` in there? but generally, I'd rewrite that...

```suggestion
    final authorId = message.repliedMessage?.author.id;
    final replyProfile = authorId != null ? ref.watch(chatRoomProvider.notifier).getUserProfile(authorId) : null;
```

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -118,32 +120,112 @@ class CustomChatInput extends ConsumerWidget {
...
-                children: [
-                  const _BuildAttachmentBtn(),
-                  const Expanded(
-                    child: Padding(
-                      padding: EdgeInsets.symmetric(horizontal: 10),
-                      child: _TextInputWidget(),
+        Visibility(
+          visible:
+              !ref.watch(chatInputProvider.select((ci) => ci.emojiRowVisible)),
+          replacement: Container(
```
gnunicorn, line 126: Can you explain to my your preference for the `Visibility`-Class rather than just toggling between the two potential states. For me, the visibility-class and the additional indentations make it less legible.

```suggestion
        !ref.watch(chatInputProvider.select((ci) => ci.emojiRowVisible)) ?  Container(
```


Also not sure if we have more "complex" sub widgets like this with their own state if they shouldn't be either an in-lined `Consumer` or better even be an externalized widget-class ...

This doesn't necessarily be addressed/changed in this PR but we might want to come to some agreement when to do what and thus I'd like to understand the reasoning behind using `Visibility`...

gnunicorn (approved):
Great work!

Some changes in code-organisation and some discussion points I'd like to hear your opinion. But no need to review again once addressed. Feel free to merge when done so.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: I guess this is been reformed from legacy-code, where it was written to be ```Visibility```. There's no reasoning behind using ```Visibility``` other than toggling between states, which I would also argue that using ```Visibility``` seems to be more in-line with the UI widget pattern and understandable than using conditional.

app/lib/features/chat/widgets/bubble_builder.dart
gtalha07: Yeah I guess it makes sense. Will make changes.

PR #831, opened 2023-08-04
fix image message flicker and reply UI improvements

gtalha07:
- [x] fix: where image starts flickering upon rendering as image message.
- [x] fix: reply UI should now show ellipsis overflow in case of larger messages.
- [x] Image reply visual improvements.
Screenshot 2023-08-04 at 8 17 37 PM
Screenshot 2023-08-04 at 8 19 51 PM
Screenshot 2023-08-04 at 8 21 07 PM


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -442,30 +436,40 @@ class _ReplyContentWidget extends StatelessWidget {
...
-          constraints: const BoxConstraints(maxHeight: 100, maxWidth: 125),
-          child: ClipRRect(
-            borderRadius: BorderRadius.circular(6.33),
-            child: Image.memory(
-              base64Decode(msg?.metadata?['base64']),
-              fit: BoxFit.fill,
-              cacheWidth: 125,
-            ),
-          ),
+        child: ImageMessageBuilder(
```
gnunicorn, line 443: nitpick: but this isn't a builder, it's a widget, isn't it?

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -442,30 +436,40 @@ class _ReplyContentWidget extends StatelessWidget {
...
+      bool enlargeEmoji = false;
+      if (textMsg.metadata!.containsKey('enlargeEmoji')) {
+        enlargeEmoji = textMsg.metadata!['enlargeEmoji'];
+      }
+      return Container(
+        alignment: Alignment.center,
+        constraints: const BoxConstraints(maxHeight: 120),
+        padding: const EdgeInsets.all(8),
+        child: enlargeEmoji
+            ? Text(textMsg.text, maxLines: 3, overflow: TextOverflow.ellipsis)
```
gnunicorn, line 460: why does it depend on whether we show emoji to show the HTML version? Shouldn't we always show the HTML version if given?

(sidenote that we should make clear if there is an HTML version or only the blank text version, but that is a Rust-API fix to do at another time)

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -442,30 +436,40 @@ class _ReplyContentWidget extends StatelessWidget {
...
+            : Html(
+                // ignore: prefer_single_quotes, unnecessary_string_interpolations
+                data: """${textMsg.text}""",
```
gnunicorn, line 463: yeah, but why though?  why not
```suggestion
                data: textMsg.text,
```

app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -1,101 +1,138 @@
...
+    if (decodedImage != null && decodedImage!.isNotEmpty) {
+      if (widget.isReplyContent) {
         return ClipRRect(
           borderRadius: BorderRadius.circular(15),
-          child: const SizedBox(
-            height: 60,
-            width: 60,
-            child: CircularProgressIndicator(strokeWidth: 6),
+          child: Container(
+            constraints: const BoxConstraints(maxHeight: 200, maxWidth: 200),
```
gnunicorn, line 49: why 200? it seems suspicious to me that this isn't relative to the outer size of space we have at all...

app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -1,101 +1,138 @@
...
+          errorWidget: (
             BuildContext context,
-            Object error,
-            StackTrace? stackTrace,
+            Object url,
+            dynamic error,
           ) {
             return Text('Could not load image due to $error');
           },
         ),
```
gnunicorn, line 125: Are we sure this works in encrypted rooms? Because there the data at the given uri is scrambled and unreadable unless you know how to decrypt it ... I think we need to return back to either room or client and hand that the uri to download and decrypt it for  us ...

(not sure we want to tackle that in this PR, but probably want to take note of it, if not)

gnunicorn (approved):
Looks good. Just minor remarks, that once addressed can be merged without anyother review.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: Yes, its a widget. Named it for the simplicity as its custom image renderer for```Chat``` widget.
Screenshot 2023-08-07 at 2 28 57 PM

app/lib/features/chat/widgets/custom_input.dart
gtalha07: I think that's how supposed to be according to package examples. But its working fine without adding double quotes. I will change to it.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: Yes, we should always show HTML version in reply content UI.

app/lib/features/chat/widgets/image_message_builder.dart
gtalha07: Technically, it would never reach to this for now as we're fetching binaries for all images of room. The uri will always be empty. This is just added as support in case if we plan to use uri rather than binaries for remote images.

app/lib/features/chat/widgets/image_message_builder.dart
gtalha07: We should constrain the image size while replying as it could get large enough for overflow. I suppose this max allowance could do that. But if you have suggestions, let me know :)

gnunicorn: I meant to use max-values based on the size available. Like use MediaQuery and stuff. Like 200 is very small on a screen like mine with 2880px for no clear reason (I know that it isn't actual pixels we use in flutter, but you get the point).

gnunicorn: though arguably the size is just a preview here to inform the user they picked the right thing, so it doesn't really matter... consider this resolved.

Issue #833, opened 2023-08-07
remove unimplemented chat features


gnunicorn:
Disable buttons in chat message
- [ ] voice 
- [ ] location
- [ ] gallery
- [ ] camera

in chat list:
- [ ] multi-select

Issue #817, opened 2023-08-02
Removing ```RoomPage``` required param ```Convo``` and listen its state through provider.


gtalha07:
Potentially we want to have ```GoRouter``` for ```RoomPage``` be flexible and should navigate with out the need of ```extra``` parameter and listening current ```Convo``` state through provider.

Issue #487, opened 2023-02-14
```stream.paginateBackwards()``` not returning updated bool value.


gtalha07:
```stream.paginateBackwards()``` which is used to fetch messages doesn't update to new boolean when there are no messages to fetch and just shows warning logs with ignored requests. 
**Logs:**
![Screenshot 2023-02-14 201957](https://user-images.githubusercontent.com/68579938/218780483-76a13970-e046-4589-9c04-129b0e0ee350.png)

gtalha07:
Closing as fixed in #834.

PR #834, opened 2023-08-07
Chat Invites and minor fixes

gtalha07:
- [x] Integrate chat invites functionality (analogous to space invites) from Space Invites.
- [x] Work around #817.
- [x] Chat Bubble fix where it gets overflow when the message content contains matrix pills. 
- [x] ```room.member.event``` visual improvements.
- [x] Fix Issue #833.
- [x] Fix issue #487.


gtalha07:
Screenshot 2023-08-09 at 8 47 04 PM
Screenshot 2023-08-09 at 8 48 38 PM
Screenshot 2023-08-09 at 8 49 11 PM

.changes/0834-chat-invites-fixes
```diff
@@ -0,0 +1,6 @@
+- [chat] Invites are now enabled for chat. After room creation, you can invite other users by going to room profile and adding them. (if have permissions).
+- [fix] Earlier some messages with mentions were getting overflow in some scenarios. This should not be happening now.
+- [fix] Chat pagination. Messages timeline should now correctly identify the end of history.
+- [enhance] Server messages, redacted and encrypted events are now distinct.
+- [fix] remove unimplemented features.
+- [enhance] room member events will present more info.
```
gnunicorn, line 6: Minor improvement to keep this clean and easy to understand for end-users:

```suggestion
- [enhance] chat room member events will present more info.
```

bitfriend (approved):
Looks good to me

app/lib/features/chat/pages/room_page.dart
```diff
@@ -37,8 +33,11 @@ class _RoomPageConsumerState extends ConsumerState<RoomPage> {
   void initState() {
     super.initState();
     WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-      ref.watch(chatRoomProvider.notifier).init(widget.convo.getRoomIdStr());
-      ref.watch(chatRoomProvider.notifier).fetchUserProfiles();
+      final convo = ref.watch(currentConvoProvider);
+      if (convo != null) {
+        ref.watch(chatRoomProvider.notifier).init(convo.getRoomIdStr());
+        ref.watch(chatRoomProvider.notifier).fetchUserProfiles();
```
gnunicorn, line 39: this feels kinda wrong here... like ... it already felt wrong before, but not even more so.

Shouldn't that be happening inside the `chatRoomProvider` listening to the `currentConvoProvider`? Rather than being initiated by a widget? If this is too much work to do now, I am okay with keeping it as-is... but I think generally, we rather want the providers coupling to happen at the provider-side rather than at the widget-level, don't you think?

app/lib/features/chat/pages/room_page.dart
```diff
@@ -154,189 +153,182 @@ class _RoomPageConsumerState extends ConsumerState<RoomPage> {
...
-                builder: (context, ref, child) {
-                  final convoProfile = ref.watch(
-                    chatProfileDataProvider(widget.convo),
-                  );
-                  return convoProfile.when(
-                    data: (profile) {
-                      if (profile.displayName == null) {
-                        return Text(
-                          AppLocalizations.of(context)!.loadingName,
+    return convo != null
```
gnunicorn, line 163: this feels like bad error management. If, for whatever reason, there is no convo, we should show an error message - especially if that is mainly a developer error - rather than just showing nothing and covering up the broken assumption with an unclear blank screen...

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+import 'package:go_router/go_router.dart';
+
+class RoomProfilePage extends ConsumerWidget {
+  const RoomProfilePage({
+    Key? key,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final room = ref.watch(currentConvoProvider)!;
```
gnunicorn, line 21: eehh... so... I looked at the router as well and I don't really when this is being set and thus if I just randomly route to this page, it will fail, is that correct? I don't like that... 

I leave more comment about that at the router...

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+    final members =
+        ref.watch(chatMembersProvider(room.getRoomIdStr())).requireValue;
```
gnunicorn, line 23: we can't be sure this has loaded yet though ... a large room, the user might click on the chat and the profile long before loading of the members (from remote) has succeeded. That should not block the user from going here and using that page. Let's refactor that to the a `members.when` and show a loading animation please.

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+          ),
+        );
+      }
+    }
+
+    return Scaffold(
+      body: CustomScrollView(
+        slivers: [
+          SliverAppBar(
+            elevation: 0.0,
```
gnunicorn, line 56: why is specifying the elevation needed?

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+            actions: <Widget>[
+              PopupMenuButton<int>(
+                icon: const Icon(Atlas.dots_vertical, size: 22),
+                itemBuilder: (context) => [
+                  PopupMenuItem<int>(
+                    value: 0,
+                    child: Row(
```
gnunicorn, line 63: am I missing something or does this action not have nothing it does when clicked? Then let's either remove that section or hook it up to the `room.leave()`.

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+                  data: (profile) => Column(
+                    children: [
+                      Center(
+                        child: Padding(
+                          padding: const EdgeInsets.only(
+                            top: 38,
+                            bottom: 12,
+                          ),
+                          child: SpaceParentBadge(
+                            badgeSize: 20,
```
gnunicorn, line 94: if the avatar is larger (75), we probably want the badge to be larger, too, no? (like 40 sounds about right)?

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -0,0 +1,146 @@
...
+                              size: 75,
+                            ),
+                          ),
+                        ),
+                      ),
+                      Text(
+                        profile.displayName ?? roomId,
+                        overflow: TextOverflow.clip,
+                        style: Theme.of(context).textTheme.titleLarge,
+                      ),
```
gnunicorn, line 110: may I suggest we _always_ show the roomId somewhere (whether we have a display name or not) and analogues to the UserProfile have a quick-copy button next, too it?

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -38,7 +38,12 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
     subscription = timeline?.diffRx().listen((event) async {
       await _parseEvent(event);
     });
-    await timeline?.paginateBackwards(10);
+    bool hasMore = false;
+    do {
+      hasMore = await timeline!.paginateBackwards(10);
+      // wait for diffRx to be finished
+      sleep(const Duration(milliseconds: 100));
```
gnunicorn, line 45: I have two problems with that:
1. sleep isn't `async` and thus _blocks_ the thread ... just want to avoid janking in the UI
2. 100ms might not be enough, is it? We can't guarantee that on slower phones ...

I wonder if there isn't a way for the `paginateBackwards(10)` to give us both information directly, rather than us relying on other channels hopefully having processed thing properly...

app/lib/features/chat/widgets/custom_message_builder.dart
```diff
@@ -43,12 +43,18 @@ class CustomMessageBuilder extends StatelessWidget {
...
-                  text,
-                  style: Theme.of(context).textTheme.bodySmall,
+                padding: const EdgeInsets.only(left: 10, bottom: 5),
+                child: RichText(
+                  text: TextSpan(
+                    text: message.author.id,
+                    style: Theme.of(context).textTheme.bodySmall,
+                    children: [
+                      TextSpan(
+                        text: '\t$text',
```
gnunicorn, line 53: `\t`? Shouldn't we just give it an internal padding on the left?

app/lib/features/chat/widgets/member_list_entry.dart
```diff
@@ -0,0 +1,195 @@
+import 'package:acter/common/dialogs/pop_up_dialog.dart';
```
gnunicorn, line 1: this is mainly a copy of the same for spaces, right? Isn't there any way we can reuse the existing code rather than having a second copy where just space -> convo?

app/lib/router/router.dart
```diff
@@ -325,9 +325,18 @@ List<RouteBase> makeRoutes(Ref ref) {
...
       },
+      routes: <RouteBase>[
+        GoRoute(
+          parentNavigatorKey: rootNavKey,
+          path: 'profile',
+          name: Routes.chatProfile.name,
+          redirect: authGuardRedirect,
+          builder: (context, state) => const RoomProfilePage(),
+        ),
+      ],
```
gnunicorn, line 339: yeah, that isn't a sufficient fix. It still relies on the outer caller of the parameter to _have set_ the convo... Which isn't working in the specific case that this is supposed to fix: when we come from a notification. In that context, all we have is the roomId and the knowledge it is a chat room. Which _should_ be enough to route there (push notifications later won't have any other way to interact than push a stringy-route). Same will also be true if you linked to another room in a chat or had any other internal link to a room (like we allow in Pins or News-bulletins).

Generally all routes that take a parameter like spaceId or roomId should be self-sufficient in that they figure out how to load that thing - even just for the fact that if you routed there and the other didn't properly update, you'd be looking at the wrong data. Take a look at how spaces does it for a reference.

gnunicorn (approved):
This doesn't actually fix #817, it just moves the responsibly around. What is meant with 817 is that no loading of convo should be required on the calling part of the route - but after this, it still is. Hence me updating the PR description.

Other than that, I left a bunch of smaller remarks and issues I see. I understand it's many and they might be work-intensive so I leave it up to you to resolve them or merge and just put them up as issues to be fixed later...

app/lib/features/chat/widgets/member_list_entry.dart
gtalha07: There is if we provide optional parameter of ```Convo``` in already written widget. And apply additional checks to determine if its meant for ```Space``` or ```Conversation```.

app/lib/features/chat/widgets/custom_message_builder.dart
gtalha07: I suppose it meant here for applying space between different text spans.

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
gtalha07: We could add Future delay here. Current problem with pagination is that it gives 10 events which are not specifically message events (in all cases) and it doesn't add up enough messages in chat list to trigger pagination from UI. So this check is needed to ensure messages list are populated to determine if room pagination is possible or not. This is how the mechanism of ```flutter_chat_ui``` is set up.

app/lib/features/chat/widgets/member_list_entry.dart
gnunicorn: yeah, I think I prefer that one. Just have it accept both `space` and `convo` optionally and check that at exactly one is checked and then just everywhere do `space != null ? space.function() : convo.function()`. that should be fine enough :) . I can see that we might later have to extend that for use-case-specific things, but for now that is good enough.

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
gnunicorn: yeah, just keep it as is for now...

app/lib/router/router.dart
gtalha07: I think the most probable (which is also ideal) way is to load ```Convo``` object through the future provider by feeding ```roomId```, essentially handling all required cases for data.

gtalha07: Example like the ```chatProvider``` we have already set up to return convo via id.

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
gtalha07: Without this loop call, the UI wouldn't populate enough messages to reach the ```scrollExtent``` of whole list interface, which would allow scroll up gesture.

app/lib/features/chat/pages/room_page.dart
gtalha07: Hmm yes, these events should happen on business logic layer and shouldn't bloat up on the interface. I will try to put these where intended.

app/lib/features/chat/pages/room_profile_page.dart
gtalha07: Do you want to have it enabled like tooltip (just as for acter-avatars)?

gnunicorn: no, we don't have tooltips on touch-devices, important information can never be put behind that. Just another entry below, same as for user-profile. And either hide the display name if not given or use the same pattern as the user-profile of allowing to edit an empty one if it isn't there yet. But as said, we can also push that into the another PR later.

app/lib/features/chat/pages/room_profile_page.dart
gtalha07: Here it only shows member list count. Anyways I converted it into ```AsyncValue```.

PR #835, opened 2023-08-07
non-acter spaces support 

gnunicorn:
- [x] show indicator in the top-bar of the space
- [x] on non-acter spaces:
  - [x] show warning above the "About" in the overview
      - [x] if the permissions allow for it, offer to convert to an acter space  
  - [x] disappear top-bar features that are acter-only 
  - [x] hide acter-only cards in overview


app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -6,88 +7,93 @@ import 'package:flutter/material.dart';
...
 import 'package:flutter_svg/flutter_svg.dart';
 
+typedef MakeIconFn = Widget Function(BuildContext);
+
 class TabEntry {
   final Key key;
   final String label;
   final String target;
-  final Widget icon;
+  final MakeIconFn makeIcon;
```
gtalha07, line 16: Why just not have it with ```Icon```?

app/lib/features/space/widgets/non_acter_space_card.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:go_router/go_router.dart';
+
+class NonActerSpaceCard extends ConsumerWidget {
+  final String spaceId;
+  const NonActerSpaceCard({super.key, required this.spaceId});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final myMembership = ref.watch(spaceMembershipProvider(spaceId));
+    const fallback = Text(
```
gtalha07, line 14: I feel like all these plain text strings should be resided in separate class under ```constants``` directory. But right now we can left it as it is till we properly do localisation.

app/lib/features/space/widgets/non_acter_space_card.dart
```diff
@@ -0,0 +1,79 @@
...
+            ),
+            const SizedBox(height: 10),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void upgradeSpace(BuildContext context, WidgetRef ref) async {
+    customMsgSnackbar(context, 'Converting to acetr space');
```
gtalha07, line 54: Minor typo :)

gnunicorn:
```suggestion
    customMsgSnackbar(context, 'Converting to acter space');
```

app/lib/features/space/widgets/non_acter_space_card.dart
gnunicorn: +1 for translations ... but until then ...

PR #837, opened 2023-08-08
Refactor code

bitfriend:
- Remove unnecessary imports
- Simplify inline crate using import statement
- Replace outer `Ok()` with inner `Ok()` about `RUNTIME.spawn()`
- Rename `OptionText` so that it is not confused with `acter_core::templates::Input::Text`

### DIFF
--- a/native/acter/src/platform/native.rs
+++ b/native/acter/src/platform/native.rs
@@ -6,7 +6,6 @@ use matrix_sdk::{Client, ClientBuilder};
 use matrix_sdk_sqlite::make_store_config;
 use parse_env_filter::eager::{filters, Filter};
 use std::{
-    fs, io,
     path::{Path, PathBuf},
     sync::{Arc, Mutex},
 };
@@ -30,10 +29,10 @@ pub async fn new_client_config(
                 tracing::warn!(
                     "{data_path:?} already existing. Moving to backup at {backup_path:?}."
                 );
-                fs::rename(&data_path, backup_path)?;
+                std::fs::rename(&data_path, backup_path)?;
             }
 
-            fs::create_dir_all(&data_path)?;
+            std::fs::create_dir_all(&data_path)?;
 
             let config = make_store_config(&data_path, None).await?;
             let builder = Client::builder()
@@ -95,7 +94,7 @@ pub fn init_logging(
     if let Some(console_logger) = console_logger {
         builder = builder.chain(console_logger);
     } else {
-        builder = builder.chain(io::stdout());
+        builder = builder.chain(std::io::stdout());
     }
 
     let mut path = PathBuf::from(log_dir.as_str());
--- a/native/acter/src/platform/native.rs
+++ b/native/acter/src/platform/native.rs
@@ -6,7 +6,8 @@ use matrix_sdk::{Client, ClientBuilder};
 use matrix_sdk_sqlite::make_store_config;
 use parse_env_filter::eager::{filters, Filter};
 use std::{
-    path::PathBuf,
+    fs, io,
+    path::{Path, PathBuf},
     sync::{Arc, Mutex},
 };
 
@@ -21,7 +22,7 @@ pub async fn new_client_config(
         .spawn(async move {
             let data_path = sanitize(&base_path, &home_dir);
 
-            if reset_if_existing && std::path::Path::new(&data_path).try_exists()? {
+            if reset_if_existing && Path::new(&data_path).try_exists()? {
                 let backup_path = sanitize(
                     &base_path,
                     &format!("{home_dir}_backup_{}", Local::now().to_rfc3339()),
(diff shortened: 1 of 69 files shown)


native/acter/src/platform/native.rs
```diff
@@ -94,7 +95,7 @@ pub fn init_logging(
     if let Some(console_logger) = console_logger {
         builder = builder.chain(console_logger);
     } else {
-        builder = builder.chain(std::io::stdout());
+        builder = builder.chain(io::stdout());
```
gnunicorn, line 98: I can't quite work out the rules here ... `std::io::stdout()` is not nice, but `io::stdout()` is? Can you clarify the guide and its reasoning?

native/cli/src/action/mock.rs
```diff
@@ -382,9 +383,9 @@ impl<'a> Mock<'a> {
     }
 
     pub async fn export(&mut self) -> Result<()> {
-        std::fs::create_dir_all(".local")?;
+        fs::create_dir_all(".local")?;
 
-        futures::future::try_join_all(self.users.values().map(|cl| async move {
+        try_join_all(self.users.values().map(|cl| async move {
```
gnunicorn, line 388: like here ... why is one prefixed with the module `fs` and the other is not (`try_join_all`)? And why even care?

native/core/src/store.rs
```diff
@@ -13,16 +14,16 @@ use crate::{
...
+    indizes: Arc<DashMap<String, Vec<String>>>,
+    dirty: Arc<DashSet<String>>,
 }
 
 static ALL_MODELS_KEY: &str = "ACTER::ALL";
 static DB_VERSION_KEY: &str = "ACTER::DB_VERSION";
 static CURRENT_DB_VERSION: u32 = 1;
 
-async fn get_from_store<T: serde::de::DeserializeOwned>(client: Client, key: &str) -> Result<T> {
+async fn get_from_store<T: DeserializeOwned>(client: Client, key: &str) -> Result<T> {
```
gnunicorn, line 26: It is common practice to keep the prefix `serde::de` to make apparent what this is (it's `serde` and not some local thing). I'd really like us to keep those please.

gnunicorn, line 26: It is common practice to keep the prefix `serde::de` to make apparent what this is (it's `serde` and not some local thing). I'd really like us to keep those please.
```suggestion
async fn get_from_store<T: serde::de::DeserializeOwned>(client: Client, key: &str) -> Result<T> {
```

native/core/src/templates.rs
```diff
@@ -209,7 +209,7 @@ fn execute_value_template(
             Ok(TomlValue::Array(items))
         }
         TomlValue::Table(t) => {
-            let mut new_table = toml::map::Map::with_capacity(t.len());
+            let mut new_table = TomlMap::with_capacity(t.len());
```
gnunicorn, line 212: not sure how this is meant to be better...

gnunicorn (changes requested):
hmmm... I am no fan of style-without-rules PRs like this. It touches 66 files (and thus makes for a huge review) to apply a style preference that is neither clear nor agreed upon, nor required by anything nor enforced/enforce-able. The next day someone else could come along and create a similarly large PR doing almost the opposite -- as it is just a matter of preference and taste. I do not care about rust style beyond rust format and clippy: everything else is up to the author or must be clearly written somewhere.

If we want any of these change to generally be applied we should codify them, document that in the docs and configure either rust-format or clippy lints for it so we can automatize its enforcement. If you care about any of these beyond just personal taste, please provide the guidelines in the docs (with proper reasoning for why that is a style we prefer) and add enforcable clippy lints or rust formatting rules.

native/acter/src/platform/native.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/837/commits/f505a9b2f23a2a43289e2b8ea73fd72128829fdc

native/cli/src/action/mock.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/837/commits/e32b76b5e24c7a2bfb3d33bd5b8f068c97b85884

native/core/src/store.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/837/commits/bcc038be8646acce71d816ca9fc4eb8b742f3a3c

native/core/src/templates.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/837/commits/a7da96c2e967c2afd110ac001b0058bb9c82cbf1

gnunicorn (approved):

PR #841, opened 2023-08-09
convert markdown to html for text message render

### LINKED ISSUE #667: chat doesn't handle markdown
Check the nightly channel: Markdown chat messages do appear unrendered:

Screenshot_20230627-110512

gtalha07:
- Fixes #667.
Screenshot 2023-08-09 at 10 55 07 PM

### DIFF
--- a/.changes/0841-chat-fixes.md
+++ b/.changes/0841-chat-fixes.md
@@ -1,0 +1,1 @@
+- [fix] chat should properly handle markdown content of messages now.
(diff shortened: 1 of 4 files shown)


gnunicorn:
No, we should not do that. The rust spec pretty clearly states that messages are to be [sent in plain-text and optionally in other formats, like html](https://spec.matrix.org/v1.6/client-server-api/#mtext). The problem is that this part of the UI doesn't seem to use the HTML-field supplied by the sender (which we know is being supplied) but only shows the plain text version. That's the fix we need to do here (as we've already done in other spaces like for Pins and Update-bulletins). 

We **should not** include a markdown renderer and interpret the plain text again (unless of course there was ever a markdown-format supplied, but considering that there is a lack of proper markdown spec, I doubt that will ever happen).

gtalha07:
Okay make sense. We know this happens in the case of server messages where it should send proper HTML format or plain text rather than markdown to be able to correctly render.

PR #842, opened 2023-08-10
Put alias resolving into a Runtime.

gnunicorn:
Fixes the crash on the chats page if there is a chat with a roomAliasId.

And a run-by fix on the canonical parent provider to handle missing access nicer.


bitfriend (approved):
Great work

PR #885, opened 2023-08-25
Improving pins 

### LINKED ISSUE #825: Finalize pins section
With Bens comment below, we should:

- [ ] Align on what is a pin? (Currently users can only create pins that is text or links, what about files?)
- [ ] Finalise the designs for both mobile and desktop (How to: Edit, Read, Update, Delete)

Current Designs:
https://www.figma.com/file/ps6Wa0fuRYkKNjPPaZqVaO/Acter-3.0---Design?type=design&node-id=4178-71815&mode=design&t=EmRT9iiNlFKgdXGq-4

### LINKED ISSUE #870: Rethink the "no parent space selected" labelling
The UI for “no parent space selected” is not self-explaining for when creating space and other objects. Rethink that part of the design...

relates to: [Product Terminology - Change Proposal](https://github.com/acterglobal/a3/issues/867)

gnunicorn:
Fixes #825

And a bunch of smaller fixes along the way:
- Fix routing of items within spaces to actually go full-screen
- Move space-selection into a separate, better looking form-field-widget, fixes #870 
- Markdown editor previews added to Pin, Updates
- Minor UI fixes in quick jumps

### DIFF
--- a/app/lib/common/widgets/md_editor_with_preview.dart
+++ b/app/lib/common/widgets/md_editor_with_preview.dart
@@ -15,6 +15,7 @@ class MdEditorWithPreview extends ConsumerStatefulWidget {
   final FormFieldValidator<String>? validator;
   final String hintText;
   final String labelText;
+  final TextEditingController? controller;
 
   const MdEditorWithPreview({
     Key? key,
@@ -22,6 +23,7 @@ class MdEditorWithPreview extends ConsumerStatefulWidget {
     this.validator,
     this.hintText = 'Description',
     this.labelText = 'Description',
+    this.controller,
   }) : super(key: key);
 
   @override
@@ -35,6 +37,7 @@ class _MdEditorWithPreviewState extends ConsumerState<MdEditorWithPreview> {
 
   @override
   Widget build(BuildContext context) {
+    final controller = widget.controller ?? _textCtr;
     return Expanded(
       child: Padding(
         padding: const EdgeInsets.symmetric(vertical: 10),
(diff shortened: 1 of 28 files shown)


app/lib/common/widgets/md_editor_with_preview.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:acter/common/providers/sdk_provider.dart';
+import 'package:acter/common/widgets/render_html.dart';
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+final markdownProvider =
```
gtalha07, line 7: Do we want to keep it publicly accessible? Otherwise have it labelled as ```private```. The convention is good to avoid conflicting names which could reduce possibility of introducing bugs.

app/lib/common/widgets/md_editor_with_preview.dart
```diff
@@ -0,0 +1,111 @@
...
+                      widget.onChanged ?? (value);
+                    },
+                  ),
+            Positioned(
+              right: 10,
+              bottom: 10,
+              child: Tooltip(
+                message: 'Toggle preview',
+                child: IconButton(
+                  onPressed: () => setState(() => _showPreview = !_showPreview),
```
gtalha07, line 99: Using ```setState``` is all good for widgets with not much deep tree. Makes code simpler. But alternatively we would keep states in provider for easier access and testing.

app/lib/common/widgets/md_editor_with_preview.dart
gnunicorn: I think it can be publicly accessible, that is fine for anything that wants to render markdown->html in the backend manner.

app/lib/common/widgets/md_editor_with_preview.dart
gnunicorn: Yeah, for simple toggles I do not quite agree to be honest ... a provider for a simple true/false flag that _literally_ only switches a particular view from one to the others seems like a lot of overhead...

PR #886, opened 2023-08-26
Revert getting of verification emojis from sas state to remote mode

bitfriend:
When other device verifies this device, getting emojis from SAS state works well.
When this device verifies other device, getting emojis from SAS state doesn't work.
So I will reactivate getting of emojis from remote server.

### DIFF
--- a/native/acter/src/api/verification.rs
+++ b/native/acter/src/api/verification.rs
@@ -331,24 +331,6 @@ impl VerificationEvent {
             .await?
     }
 
-    pub fn was_triggered_from_this_device(&self) -> Result<bool> {
-        let device_id = self
-            .client
-            .device_id()
-            .context("guest user cannot get device id")?;
-        warn!("was_triggered_from_this_device device_id: {}", device_id.to_string());
-        if let Some(other_device_id) = self.content.get("other_device_id") {
-            warn!("was_triggered_from_this_device other_device_id: {}", *other_device_id);
-            Ok(*other_device_id == *device_id)
-        } else if let Some(from_device) = self.content.get("from_device") {
-            warn!("was_triggered_from_this_device from_device: {}", *from_device);
-            Ok(*from_device == *device_id)
-        } else {
-            warn!("neither other_device_id nor from_device");
-            Ok(false)
-        }
-    }
-
     pub async fn accept_sas_verification(&self) -> Result<bool> {
         let client = self.client.clone();
         let controller = self.controller.clone();
--- a/native/acter/src/api/verification.rs
+++ b/native/acter/src/api/verification.rs
@@ -114,8 +114,63 @@ impl VerificationEvent {
         self.emojis.clone_from(&emojis);
     }
 
-    pub fn get_emojis(&self) -> Vec<VerificationEmoji> {
+    // when this device triggered verification of other device, it can get emojis from SAS state
+    pub fn emojis(&self) -> Vec<VerificationEmoji> {
         self.emojis.clone()
+    }
+
+    // when other device triggered verification of this device, it can get emojis from remote server
+    pub async fn get_emojis(&self) -> Result<Vec<VerificationEmoji>> {
+        let client = self.client.clone();
+        let sender = self.sender.clone();
+        let event_id = self.event_id.clone();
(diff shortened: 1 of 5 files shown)


native/acter/src/api/verification.rs
```diff
@@ -114,10 +114,65 @@ impl VerificationEvent {
...
+                                .iter()
+                                .map(|e| VerificationEmoji {
+                                    symbol: e.symbol.chars().next().unwrap() as u32, // first char in string
+                                    description: e.description.to_string(),
+                                })
+                                .collect::<Vec<VerificationEmoji>>();
+                            return Ok(sequence);
+                        } else {
+                            return Ok(vec![]);
+                        }
```
gnunicorn, line 147: Is this really a proper case? Shouldn't this raise an error that we didn't receive any emojis?

```suggestion
                        let Some(items) = sas.emoji() else {
                            bail!('No emojis found. Aborted.');
                        };
                        let sequence = items
                            .iter()
                            .map(|e| VerificationEmoji {
                                symbol: e.symbol.chars().next().unwrap() as u32, // first char in string
                                description: e.description.to_string(),
                            })
                            .collect::<Vec<VerificationEmoji>>();
                        return Ok(sequence);
```

native/acter/src/api/verification.rs
```diff
@@ -114,10 +114,65 @@ impl VerificationEvent {
...
+                    }
+                } else if let Some(txn_id) = txn_id {
+                    if let Some(Verification::SasV1(sas)) = client
+                        .encryption()
+                        .get_verification(&sender, txn_id.as_str())
+                        .await
+                    {
+                        if let Some(items) = sas.emoji() {
```
gnunicorn, line 155: same as above...

native/acter/src/api/verification.rs
```diff
@@ -114,10 +114,65 @@ impl VerificationEvent {
...
+                                .collect::<Vec<VerificationEmoji>>();
+                            return Ok(sequence);
+                        } else {
+                            return Ok(vec![]);
+                        }
+                    }
+                }
+                // request may be timed out
+                info!("Could not get verification object");
+                Ok(vec![])
```
gnunicorn, line 171: And here. This sounds like an error case to me...
```suggestion
                bail!("Could not get verification object");
```

gnunicorn (approved):
Generally looks good. Just, because this is a security relevant area, I'd rather we throw more errors in case things go unexpected (as commented).

Generally, I am not super happy that we have so little testing around this entire section ... makes me uncomfortable that we might break it without realizing (like we seem to have done this time) ... Any suggestion how we can increase the long-term confidence on this code highly appreciated!

native/acter/src/api/verification.rs
bitfriend: Other device may not respond in time and timeout is possible case.
If this function calls `bail!()`, flutter may occur exception.
I wonder if flutter should occur exception in possible case.

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/886/commits/d3b14ed0add97f31c0c55f897925f668dcccce6b

native/acter/src/api/verification.rs
gnunicorn: Yes, flutter should _not_ consider any case where we don't see the emoji a regular case of operation. Raising/throwing here is the right thing to do.

bitfriend: Understood

bitfriend: Should I replace `Ok(false)` with `bail!()` here too?
https://github.com/bitfriend/a3/blob/d3b14ed0add97f31c0c55f897925f668dcccce6b/native/acter/src/api/verification.rs#L204-L206

gnunicorn: yes, probably.

bitfriend: Fixed by https://github.com/acterglobal/a3/pull/886/commits/1ae92a8818f9db6af242ee3820ac9a45bfa7102a

gnunicorn (approved):

PR #895, opened 2023-08-29
User deactivation

### LINKED ISSUE #892: user deactivation
As we allow for users to register, by the rules of Apples AppStore the user must also be able to delete the account (data):

> Guideline 5.1.1(v) - Data Collection and Storage

> We noticed that your app supports account creation but does not appear to include an option to initiate account deletion.

> Apps that support account creation must also offer account deletion to give App Store users more control of the data they've shared while using your app.

> Next Steps

> If your app already supports account deletion, reply to this message and let us know how to locate this feature. If your app does not support account deletion, revise your app to include an option to initiate account deletion and delete all user data you are not legally required to retain.

> If you are unable to offer account deletion or need to provide additional customer service flows to facilitate and confirm account deletion, either because your app operates in a highly-regulated industry or for some other reason, reply to this message in App Store Connect and provide additional information or documentation. If you have questions regarding your legal obligations, check with your legal counsel.

> Keep these requirements in mind when updating your app to support account deletion:

> - Only offering to temporarily deactivate or disable an account is insufficient.
> - If users need to visit a website to finish deleting their account, include a link directly to the page on your website where they can complete the process.
> - You may include confirmation steps to prevent users from accidentally deleting their account. However, only apps in highly-regulated industries may require users to use customer service resources, such as making a phone call or sending an email, to complete account deletion.

---

However, matrix itself doesn't really support user deletion. We might want to do what the element-web version does though: remove all personal information, drop out of all spaces and rooms, remove all devices/sessions and put an impossible password on the account. That could do well enough and would block the user-id for the time being. Might be sufficient.

gnunicorn:
This PR adds the ability to deactivate ones account, of course with the big red text and password-confirmation before the user actually does it, as that the most permanent, GDPR compliant implementation of right-of-erasure you can implement on Matrix.

This comes with a rust-backend end-to-end-test and code to also remove the locally stored data right at the same time.

See it in action:
deactivate-and-destroy
And then if you tried log in again:
and-gone

fixes #892 .

### DIFF
--- a/app/lib/common/dialogs/deactivation_confirmation.dart
+++ b/app/lib/common/dialogs/deactivation_confirmation.dart
@@ -1,0 +1,124 @@
+import 'package:acter/common/dialogs/pop_up_dialog.dart';
+import 'package:acter/common/providers/sdk_provider.dart';
+import 'package:acter/common/themes/app_theme.dart';
+import 'package:acter/common/utils/routes.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+
+// Can be extended to be reusable dialog as riverpod states get added.
+// Ref can be used to read any provider which are declared.
+void deactivationConfirmationDialog(BuildContext context, WidgetRef ref) {
+  TextEditingController passwordController = TextEditingController();
+  showDialog(
+    context: context,
+    builder: (BuildContext ctx) {
+      return AlertDialog(
+        title: Text('Deactivate Account',
+            style: TextStyle(
+                color: AppTheme.brandColorScheme.error, fontSize: 32,),),
+        content: SingleChildScrollView(
+          child: ConstrainedBox(
+            constraints: const BoxConstraints(maxWidth: 650),
+            child: Wrap(
+              children: [
+                Text(
+                  'Careful: You are about to permanently deactivate your account ',
+                  style: TextStyle(
+                    color: AppTheme.brandColorScheme.error,
+                    fontSize: 24,
+                  ),
+                ),
+                const Text('If you proceed: \n  \n'
+                    '- All your personal data will be removed from your homeserver, including display name and avatar \n'
+                    '- All your sessions will be closed immediately, no other device will be able to continue their sessions \n'
+                    '- You will leave all rooms, chats, spaces and DMs that you are in \n'
+                    '- You will not be able to reactivate your account \n'
+                    '- You will no longer be able to log in \n'
+                    '- No one will be able to reuse your username (MXID), including you: this username will remain unavailable indefinitely \n'
+                    '- You will be removed from the identity server, if you provided any information to be found through that (e.g. email or phone number) \n'
+                    '- All local data, including any encryption keys, will be permanently deleted from this device \n'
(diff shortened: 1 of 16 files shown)


bitfriend (approved):
Great work

app/lib/common/dialogs/deactivation_confirmation.dart
```diff
@@ -0,0 +1,124 @@
...
+                  child: Text(
+                    'Please provide your user password to confirm you want to deactivate your account.',
+                    style: TextStyle(
+                      color: AppTheme.brandColorScheme.error,
+                      fontSize: 24,
+                    ),
+                  ),
+                ),
+                ConstrainedBox(
+                  constraints: const BoxConstraints(maxWidth: 250),
```
gtalha07, line 54: I feel like ```MediaQuery``` width should be used here. The content should expand itself based on current context.

native/acter/src/api/auth.rs
```diff
@@ -332,3 +342,19 @@ pub async fn register_with_token_under_config(
...
+impl Client {
+    pub async fn deactivate(&self, password: String) -> Result<bool> {
+        // ToDo: make this a proper User-Interactive Flow rather than hardcoded for
+        //       password-only instance.
+        let account = self.account()?;
+        RUNTIME
+            .spawn(async move {
+                let auth_data =
+                    AuthData::Password(Password::new(account.user_id().into(), password.clone()));
+                account.deactivate(None, Some(auth_data)).await?;
```
bitfriend, line 355: I think it is better to deactivate after logout.
Then `deactivate()` would be outside of `Client` crate.

gtalha07 (approved):
One minor suggestion for feedback and else is all good to merge!

app/lib/common/dialogs/deactivation_confirmation.dart
gnunicorn: the reason I bounded it is that it would normally expand to full width (+ some padding on the sides) but that makes the text pretty unreadable (12-words rule). Using a max-width I can assure we never overexpand, but do contract accordingly on smaller devices. What would be the benefit to look at mediaquery and do it in relation to that?

app/packages/rust_sdk/lib/acter_flutter_sdk.dart
```diff
@@ -377,6 +377,36 @@ class ActerSdk {
...
+
+    final userId = client.userId().toString();
+
+    // take it out of the loop
+    if (_index >= 0 && _index < _clients.length) {
+      _clients.removeAt(_index);
+      _index = _index > 0 ? _index - 1 : 0;
+    }
+    try {
+      if (!await client.deactivate(password)) {
```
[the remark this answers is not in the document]
gnunicorn: no. if you logged out, you can't call deactivate. It is a function only a logged-in user can do.

native/acter/src/api/auth.rs
gnunicorn: We can't deactivate after we are logged out. It requires a logged-in client to do.

app/lib/common/dialogs/deactivation_confirmation.dart
gtalha07: Thanks for reasoning, I understand it now.

Issue #891, opened 2023-08-29
Issues after latest nightly (29.08.23)


emilvincentz:
- [x] Emoji reaction view is showing users who reacted, but not the emojiis -> #900
- [x] emojis given, are not showing in chat. Also emojii snippet is not showing emojiis -> #900
- [x] When entering chat on IOS, there is no text input field -> #902 
- [x] chat sidebar is also showing on ios (should not show)  -> #902 
- [x] Pins seems to require description when link already inserted (should not be required) -> #904
- [x] Pins. Error in the link field. Although i already inserted link, it says: "Text or URL must be given". -> #904 
- [x] Chat room route push bug .i.e. navigating to another room while current route is room id will throw exceptions and keeps the room view blocked. -> #915

emilvincentz:
![Image](https://user-images.githubusercontent.com/77483554/264016990-c8ec9bb2-4e6a-4f47-a49e-f4f4f4a95daa.png)
![Image](https://user-images.githubusercontent.com/77483554/264016988-a0b22a34-b980-4128-9272-5012995f1457.png)
![Image](https://user-images.githubusercontent.com/77483554/264016989-a225ec82-12ff-4038-8891-64cf820b11b5.PNG)
![Image](https://user-images.githubusercontent.com/77483554/264016991-0c9860fa-d7b9-4635-8e0f-050b83e2a0ca.PNG)
![Image](https://user-images.githubusercontent.com/77483554/264016992-bfd0dd8d-85c2-48ab-89fa-ac98ff9c6a0e.png)
![Image](https://user-images.githubusercontent.com/77483554/264016993-723c279a-5d1f-428d-83fd-db63acf96cb2.png)
![Image](https://user-images.githubusercontent.com/77483554/264016994-efd9ab2e-5b2e-44d1-b32a-ef2fe26df06f.png)

PR #900, opened 2023-08-30
Enforce Platform Emojis

gnunicorn:
Looks like the font-fallback in #761 did not properly work on Windows, Mac and iOS, as reported in #891 .

For reasons unknown to the author, the included NotoEmoji-Font doesn't properly render on these platforms - maybe because it is an COLRv1 variable font - it is unclear. However, for the sake of support over consistency, this change fixes that by enforcing the usage of the system default emoji-font for these platforms, rather than first using NotoEmoji and only falling back to the system font if it isn't found (which was the previous pattern).

Because that makes the font non-`const` now (as checking Platform is - for some reason - not `const` either), this requires some trickery around the usage of the font setting in other places, too. 

Here you can see it working in all case: react-emoji, react-emoji-drawer, giant-sent-emoji, emoji-in-text-message and the emoji picker for both:

**MacOS**:
emojis-on-macos

and **Windows**:
emojis-on-windows

### DIFF
--- a/app/lib/common/themes/app_theme.dart
+++ b/app/lib/common/themes/app_theme.dart
@@ -2,11 +2,10 @@ import 'package:flutter/material.dart';
 import 'package:flutter/foundation.dart' show kIsWeb;
 import 'dart:io';
 
+@pragma('vm:platform-const')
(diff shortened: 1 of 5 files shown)


gtalha07 (approved):
LGTM!

PR #902, opened 2023-08-30
Fix chat sidebar

gnunicorn:
refs #891 

Replaces the `InChat` with the already existing `WithSidebar` widget for proper responsive behavior (not showing up on mobile) - with a new `preferSidebar`-flag and an outsourced and key-pinned widget we can reuse the same widget over multiple views. Nice.

Additionally contains two run-by fixes:
 - Clicking on the room from a space doesn't put the view inside the space-view anymore
 - Closing the room search does also reset the search value (rather than leave a subset of rooms visible).

Here you can see it on Android:

chat-on-android

And on Linux:
chat-on-linux


gtalha07 (approved):
LGTM!

PR #903, opened 2023-08-31
Message editing for all media

bitfriend:
- Implement text message editing
- Implement image/audio/video/file message editing


gnunicorn:
@bitfriend needs a conflict resolving now

PR #907, opened 2023-09-01
chat shell (Desktop) and fixes

### LINKED ISSUE #905: Use ```ShellRoute``` or ```StatefulShellRoute``` for chat overview.
Since chat list is a consistent view so we want to keep it as ```ShellRoute``` for routing to be working without any issues.

gtalha07:
- [x] Fix #905.
- [x] Chat nested navigation (mobile).
- [x] Simplify profile data providers.
- [x] Fix Space member avatars .i.e. some were not loading.
- [x] (Activities) chat notifications routing.


gtalha07:
Preview:
Screen Recording 2023-09-03 at 5 16 12 PM

gtalha07:
**Preview Chat : (Mobile Android)**

Screen Recording 2023-09-03 at 5 32 10 PM

app/lib/features/chat/widgets/emoji_reaction_item.dart
```diff
@@ -15,44 +16,49 @@ class EmojiReactionItem extends ConsumerWidget {
...
-                emoji,
-                style: EmojiConfig.emojiTextStyle,
-              ),
-            )
-            .toList(),
-      ),
+    final member = ref.watch(chatMemberProvider(userId));
+    return member.maybeWhen(
+      data: (member) {
+        final profile = ref.watch(memberProfileProvider(member));
```
gnunicorn, line 22: do I read these correctly: by splitting the provider into two, we now have a bunch of `.maybeWhen` that don't really equate to anything as the widgets aren't drawn unless there is a valid _userId_ .

May I suggest adding a provider that just encapsulates these two providers as one again? Call it `memberProfileFromUserIdProvider` and then we can still just have a single provider and thus a single level of indentation as the other is not needed either way ...

app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -37,7 +41,7 @@ class ImageMessageBuilder extends ConsumerWidget {
                 maxWidth: MediaQuery.of(context).size.width * 0.2,
               ),
               child: Image.memory(
-                data,
+                data!,
```
gnunicorn, line 44: Wait, what? You are changing the Provider from a `Uint8List` to a `Uint8List?`, with the explicit case that _on error_ it won't error but come back with `null` and then you have the UI do a `!` on the return value... 

Nope. I'm sorry. But there is plenty of reasons why loading the imageBinary might fail (network, IO, decryption, decoding ...) and just ignoring the error (it isn't even logged or printed!) and then going on but _hard fail_ on a `null`, that was explicitly returned isn't an acceptable way to deal with the situation. In general we need to decrease the hard-null `!` in our code base, it is mostly based on assumption of limited information and it leads to catastrophic failures with the end-user clients in cases, where - if we handled that better - we could at least given them an explanation of with an error message to understand what went wrong.  

The code was better before this change, as at least the error code was run when the image loading failed.

app/lib/features/chat/widgets/member_list.dart
```diff
@@ -36,7 +43,7 @@ class MemberList extends ConsumerWidget {
               padding: const EdgeInsets.all(8.0),
               child: MemberListEntry(
                 member: member,
-                convo: convo,
+                convo: roomNotifier.asyncRoom.requireValue,
```
gnunicorn, line 46: I really feel that this roomNotifier.asyncRoom didn't change things to the better, considering the increased number of `requireValue` that it comes with...

app/lib/features/home/pages/dashboard.dart
```diff
@@ -33,7 +33,9 @@ class _DashboardState extends ConsumerState<Dashboard> {
   @override
   void initState() {
     super.initState();
-    _checkIfSpacesPresent();
+    WidgetsBinding.instance.addPostFrameCallback((timestamp) {
+      _checkIfSpacesPresent();
```
gnunicorn, line 37: can you explain why we would delay this operation?

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -94,7 +96,12 @@ class QuickJump extends ConsumerWidget {
                 ),
               ),
               onPressed: () {
-                navigateTo(route: Routes.chat);
+                final roomId = ref.watch(roomIdProvider);
+                if (isDesktop) {
+                  navigateTo(target: '/chat/$roomId');
+                } else {
+                  navigateTo(route: Routes.chat);
+                }
```
gnunicorn, line 104: yeah, no, the increasing number of cases where routing is changing depending on whether we are `isDesktop` or not concerns me. I think this is the wrong approach.

gnunicorn (changes requested):
While the changes to the core UI for using the chat sidebar are fine, the surrounding changes in the providers and the side-effects that then ripple through the system indicate clearly to me that this is the wrong approach.

For once I think strongly that _routing_ must figure out what it meant to show _on its own_ without any external providers needing to be set to certain variable for it to work: we will have embedding links going to specific items like `/chat/!asdf91` from anywhere in the UI and we do not want to add hacks in the link-code to deal with the state-setting changes necessary nor figure out whether we do that differently based on whether we are a desktop or not. All we should be doing it route to the URL with the provided information and have the router deal with the rest.

(Arguably, even whether we are a desktop or not, is already a misnomer, as we do show the sidebar and thus should also show the sidebar of chat rooms on a tablet in landscape mode, if we have enough space.)

I have pointed out in various places why this disconnect of two dependent, yet independently managed states (of router and riverpod provider) will lead to further issues down the line. Which is just exacerbated through the number of `requireValue` and `data!` cases added in this PR (to deal with the fallout of changes in providers) that lead to catastrophic failures. We should attempt to remove `requireValue` and `null!` cases from our App, as they bring the app into an unrecoverable situation for the end-user rather than a nicely displayed error message, not add to the pile.

A few places places effected, arguably, aren't introduced by this PR and are attempted to fixed. But I think we need to refactor that entire understanding and agree on the underlying architecture we want to have here. Notifiers calling functions on other notifiers upon a successful data loading are just the wrong-way around the provider pattern. And even though the code existed prior, we need to change the way it is done as a whole, to fix the issues we are seeing now.

app/lib/common/widgets/chat/convo_card.dart
```diff
@@ -50,14 +49,18 @@ class _ConvoCardState extends ConsumerState<ConvoCard> {
...
         onTap: () {
-          ref
-              .read(currentConvoProvider.notifier)
-              .update((state) => widget.room);
-
-          context.pushNamed(
-            Routes.chatroom.name,
-            pathParameters: {'roomId': roomId},
-          );
+          ref.read(roomIdProvider.notifier).update((state) => roomId);
```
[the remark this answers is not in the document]
gtalha07: Agree, I think routing should be fine enough to update the pages state. ```roomIdProvider``` is there for keeping the selection state on desktop which could be modified to have it as internal chat list state. The goal was here to have room opened initially but I think its not better approach.

app/lib/features/activities/widgets/notification_card.dart
```diff
@@ -125,12 +129,16 @@ class NotificationCard extends ConsumerWidget {
...
-                  'roomId': roomId,
-                }, // FIXME: fails at the moment
-              );
+              ref.read(roomIdProvider.notifier).update((state) => roomId);
+              isDesktop
+                  ? context.goNamed(
+                      Routes.chatroom.name,
+                      pathParameters: {'roomId': roomId},
+                    )
+                  : context.pushNamed(
```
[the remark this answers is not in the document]
gtalha07: Yeah there could be function which sorts of handle the navigation stack on what the platform is running. But I think in the long run, this would also not be beneficial. I am of the idea that we should keep two go routers with separate navigation logic and loading up one from the root during app launch, based on what platform is of the context. This would simplify the core navigation and easier to apply changes without making them (interdependent) and also not introducing more problems underway.

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -15,18 +14,19 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
   Widget build(BuildContext context, WidgetRef ref) {
-    final room = ref.watch(currentConvoProvider)!;
-    final convoProfile = ref.watch(chatProfileDataProvider(room));
-    final roomId = room.getRoomIdStr();
-    final members = ref.watch(chatMembersProvider(roomId));
-    final myMembership = ref.watch(spaceMembershipProvider(roomId));
-    final location = ref.watch(currentRoutingLocation);
+    final roomNotifier = ref.watch(chatRoomProvider.notifier);
+    final convoProfile =
+        ref.watch(chatProfileDataProvider(roomNotifier.asyncRoom.requireValue));
```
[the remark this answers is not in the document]
gtalha07: This page is a part of shell and nested. We can't reach here directly outside of chat shell atm (which doesn't mean it won't  as feature gets more refined). We have a understanding right now that room is already cached before we even reach to this state. But yeah as I did addressed before about using ```requireValue``` is least preferable. We might have to rethink the whole idea of loading rooms and states of them in chat notifiers.

app/lib/features/chat/widgets/emoji_reaction_item.dart
gtalha07: Correct, I think we should merge these into one. I would like to keep ```memberProfileProvider``` as it is because its re-usable, needs ```Member``` object which doesn't need additional reference of being part of ```Convo``` or ```Space```.

gtalha07: Correct, we should merge these into one. But that leads to another question. How would we able to reference ```userId``` to ```Convo``` or ```Space```? I would still like to keep ```memberProfileProvider``` as taking ```Member``` since it doesn't need prior information of what it's a part of and thus being reusable. ```userId``` is dependent on additional context of which hierarchy it belongs.

app/lib/features/chat/widgets/image_message_builder.dart
gtalha07: I will change it. Apologies!

app/lib/features/home/pages/dashboard.dart
gtalha07: This doesn't necessary delays the operation, it's more like a safety approach to not altering any state of UI while widget is already building. This call would ensure we are running state code after widget is properly initialised. As I see we're using ```ref.watch()``` which is reactive to put changes in UI. Removing this would lead to exception, as I was getting when navigating to dashboard.

gtalha07: ```final spaces = await ref.watch(spacesProvider.future);```

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -38,10 +40,16 @@ final chatStreamProvider = StreamProvider<List<Convo>>((ref) async* {
...
-        .map((item) => (item['conversation']) as Convo)
+        .map((item) => (item['conversation']) as ffi.Convo)
         .toList();
     //FIXME: how to check empty chats ?
     if (conversations.isNotEmpty) {
+      if (isDesktop) {
+        ref
+            .watch(roomIdProvider.notifier)
+            .update((state) => conversations[0].getRoomIdStr());
+      }
```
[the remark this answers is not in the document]
gtalha07: Agree, this is not better way of setting rooms. I will remove it.

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -15,44 +14,54 @@ import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
...
-        'Some error occured loading room ${e.toString()}',
-      );
-    }
+    client = ref.read(clientProvider)!;
+
+    asyncRoom.when(
+      data: (room) async {
+        // reset messages when room is changed.
+        ref.read(messagesProvider.notifier).reset();
+        convo = room;
```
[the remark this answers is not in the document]
gtalha07: Yeah, as being informed of the bug that leads to showing wrong timeline. I need to restructure this entire flow.

app/lib/common/providers/common_providers.dart
```diff
@@ -61,50 +60,43 @@ final chatProfileDataProvider =
...
+    FutureProvider.family<ProfileData, ffi.Member>((ref, member) async {
+  try {
+    ffi.UserProfile profile = member.getProfile();
+    ffi.OptionString displayName = await profile.getDisplayName();
+    final avatar = await profile.getThumbnail(62, 60);
+    return ProfileData(displayName.text(), avatar.data());
+  } catch (e) {
+    debugPrint('$e');
+    return ProfileData('', null);
+  }
```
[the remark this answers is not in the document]
gnunicorn: I think we should find some agreed upon patterns and write them up, because some of these constrains are really detrimental for the user experience and we need to make sure those don't fall through the cracks.

app/lib/features/activities/widgets/notification_card.dart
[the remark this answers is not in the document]
gnunicorn: as pointed out below, I think anything that goes for a `isDesktop`-Style check will ultimately fail us as it doesn't take into account the complexity that is responsive layouts ... like tables in landscape format ... not a desktop, not a mobile either ...

app/lib/features/chat/widgets/emoji_reaction_item.dart
gnunicorn: Not saying we should merge them. They can exist as they are, but we can _add_ a provider (on top) that combines both providers. Everything else can stay the same.

app/lib/features/home/pages/dashboard.dart
gnunicorn: but the called function (internally) is async, this delaying doesn't "ensure" that unfortunately.

Yeah... ref-read and ref-watch in initState are problematic ... we might want to find some guidelines for that, too ...

gnunicorn:
superseeded by #915

PR #912, opened 2023-09-05
Block user and content

bitfriend:
- Implement user blocking
- Implement content blocking

### DIFF
--- a/native/acter/src/api/room.rs
+++ b/native/acter/src/api/room.rs
@@ -17,7 +17,10 @@ use matrix_sdk::{
     media::{MediaFormat, MediaRequest},
     room::{Room as SdkRoom, RoomMember},
     ruma::{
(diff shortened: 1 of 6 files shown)


native/acter/src/api/room.rs
```diff
@@ -2367,6 +2390,34 @@ impl Room {
...
+    pub async fn report_content(
+        &self,
+        event_id: String,
+        score: Option<i32>,
+        reason: Option<String>,
+    ) -> Result<bool> {
+        let room = if let SdkRoom::Joined(r) = &self.room {
+            r.clone()
+        } else {
+            bail!("Can't update power level in a room we are not in")
```
gnunicorn, line 2403: this is probably not the right error message...

gnunicorn (approved):

native/acter/src/api/room.rs
bitfriend: Fixed

Issue #908, opened 2023-09-04
Catching up on encrypted history


gnunicorn:
For encrypted spaces and chats, we have a unreadable history problem. For security reasons, the encryption keys are rotated every so often. That means if the device hasn't synced in a while, or is added newly to the roster of devices, chat previous messages are delivered to the device but can't be decrypted immediately when the keys are missing.

We already have a mechanism to help with that: device verification. When device is cross-verified with another device, and it encounters messages from history that it can't decrypt, it can ask any other cross-verified device for the decryption keys. Afaik, this happens more or less transparently in the matrix-rust-sdk (or the timeline object) and it deals with all the complicated hoops and loops.

You can even observe that process in element by: logging in with a new session, navigate to an encrypted chat channel with previous messages. You will see them come up but grayed out as "can't be decrypted". Once you cross-verified that sessions with another session, it will start fetching the decryption keys and you can see the messages being replaced with the decrypted ones.

This latter part of the process doesn't seem to be happening with Acter at the moment. Even after cross-verifying the new session, old messages can't be decrypted and their failed attempt is not replaced in the UI. It does work for new messages coming in, but not for old ones.

We need to fix this. For chat/conversations but also for spaces later, as there we need to load the _entire_ history to recreate the state machine ...

PR #915, opened 2023-09-05
Chat refactor

gnunicorn:
- [x] refactor chat riverpods
- [x] make its state updates be triggered by router
- [x] load chat list at startup
- [x] cache latest messages
- [x] keep separate chat message listings around
- ~~keep input state linked to chat room~~ postponed into #954
- [x] clean up chat providers
- [x] fix chat overlaying issue, fixes #899  
- [x] merge latest message and timeline object
    - [x] update latest  
    - [x] update if the latest changes (encrypted -> unencrypted)
- [x] move spaces and chats into eyeball-im observables
   - [x] track the diff of lists rather than the full list 
   - [x] define order mechanism in the rust-core

**Newly introduced bugs to fix**:
 - [x] search doesn't properly match names / shows no chats
 - [x] sort-order of chat rooms appears to be random
 - [x] rooms we've left shouldn't be shown
 - [x] spaces order is random
 - [x] images aren't loading
 - [x] chat back-button jumps back to previously selected chat
 -  ~~chat room is emebedded in relative space (namely space-shell)~~ refs #955
 - [x] timeline doesn't show latest events/incoming events
 - [x] List diff not properly applied to chatroom
 - [x] massive amount of room member failures:
```
flutter: Failed to load profile due to User not found among room members
flutter: ERROR loading avatar due to User not found among room members
```

PR #916, opened 2023-09-06
Request room key when unable to decrypt

bitfriend:
Will fix [this issue](https://github.com/acterglobal/a3/issues/908).

### DIFF
--- a/native/acter/src/api/convo.rs
+++ b/native/acter/src/api/convo.rs
@@ -30,7 +30,8 @@ use matrix_sdk::{
     },
     Client as SdkClient, RoomMemberships,
 };
-use std::{ops::Deref, path::PathBuf};
+use matrix_sdk_ui::timeline::RoomExt;
+use std::{ops::Deref, path::PathBuf, sync::Arc};
 use tracing::info;
 
 use super::{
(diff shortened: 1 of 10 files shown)


bitfriend:
The feature `automatic-room-key-forwarding` requests room key in `OlmMachine::decrypt_room_event()`.
And this feature contains `e2e-encryption` feature.

native/acter/src/api/convo.rs
```diff
@@ -69,19 +69,16 @@ impl Convo {
...
         }
     }
 
-    fn set_latest_message(&mut self, mut msg: RoomMessage) {
-        if let Some(mut event_item) = msg.event_item() {
-            msg.set_event_item(Some(event_item));
-        }
-        self.latest_message = Some(msg);
+    pub fn set_latest_message(&mut self, msg: Box<RoomMessage>) {
+        self.latest_message = Some(*msg);
```
gnunicorn, line 81: I am not convinced that piping the message through the fronted to then set it back on the backend is a good idea ...

but I as I am on a refactor for the entire section, it's fine for now...

gnunicorn (approved):

native/acter/src/api/convo.rs
gnunicorn: It's fine for now. I am currently looking at merging timeline and latest_room message in #915 .

PR #918, opened 2023-09-06
Extend blocking feature

bitfriend:
We're currently missing sender info for Pins, News and Events. We actually need that for:
- Check `userId` to make reporting not appear if user is same.
- Optionally add the ignore user option which would require getting the member via `userId`.

Currently `report_content()` falls under only `Convo` hierarchy. We want to implement it in `Space` too.


gnunicorn (approved):

gtalha07 (approved):
LGTM for now. But later we want to move that API into ```Client``` for global access

PR #927, opened 2023-09-09
Block Users UI

gnunicorn:
- [x] expose `isIgnored` on `Member`, `ignoredUsers`(list), `ignoreUser` and `unignoreUser` on `Account`
- [x] Add feature to block user from any Member List entry 
- [x] Small refactor of the Settings UI for improved usability
- [x] Improved Design for Sessions Screen to provide more clarity
- [x] Improve wording and UI around user blocking
- [x] Settings Screen to add and remove users from the block list

block-from-ui


app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -15,45 +17,149 @@ class SettingsMenu extends ConsumerWidget {
...
-            title: const Text('Labs'),
-            subtitle: const Text('Experimental Acter features'),
-            leading: const Icon(Atlas.lab_appliance_thin),
-            selected: isSelected(Routes.settingsLabs),
-            onTap: () {
-              isDesktop || size.width > 770
-                  ? context.goNamed(Routes.settingsLabs.name)
-                  : context.pushNamed(Routes.settingsLabs.name);
-            },
+      body: SettingsList(
```
gtalha07, line 40: I wonder if we can use this component for room profile. But we should consider that when focusing on room/chat profiles.

gtalha07 (approved):
LGTM! Please close all issues related to it when merged ;)

PR #945, opened 2023-09-14
Push Notifications for mobile

gnunicorn:
This contains:

1. Strong encrypted on rest:
   From this point forward all newly registration or logins will have a strong encrypted-on-rest-paradigm installed: upon creation the database we create a random password and use it to encrypt all data stored on disk (within the matrix-sdk store). All this is then kept in the secure enclave ('keychain') of the operation system. Old sessions will be moved over into that enclave though already existing database won't be migrated or encrypted. This was necessary because we couldn't open the unsecured session storage before from the iOS Service Extension (see below), but we can savely open this one.
2. A new "Push Notifications"-flag is added to the Labs for iOS & Android (with Google Services):
   When activated and upon restart, the user will be asked if they want to receive push notifications. When accepted, we are sending over a "pusher" to the homeserver using your newly added `$PUSH_SERVER` as target (sygnal-servers already deployed).
3. When receiving a push notification from the server, we are:
   a) having a new service that enables us to read and decrypt the event, understanding its content and showing a decrypted message
   b) for iOS background services, this PR already contains the necessary Service-Extension and uniffi-swift-plumbing to load the rust service (that's what those makefile changes are about), and it actually works. Unfortunately it isn't able to actually decrypt the data until we have deployed the sliding-sync mechanism, so we are running into a runtime error (theoretically this should _just work_ once we have switched to sliding-sync). In those cases it will show a generic message for the time being (but c) still works)
   c) Upon click, regardless of foreground or background, the app will open and route to the corresponding chat-room or space.
   Note that if we are in foreground and looking at the chat, not notification is issued (we are usually _seconds_ faster in showing the message in chat compared to when the push-message makes us to the device).

_Supported Platforms_: iOS & Android.
_Instructions for testing_: You need to set the `PUSH_SERVER` or the code will be skipped altogether. For our test-setup use the usual matrix-domain and prefix it with `sygnal.` (DM me for help if you need it):
```diff
diff --git a/.vscode/launch.json b/.vscode/launch.json
index 1d70c46a11..97ea6181bd 100644
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -50,7 +50,9 @@
             "flutterMode": "debug",
             "toolArgs": [
                 "--dart-define",
-                "RAGESHAKE_URL=http://localhost/api/submit"
+                "RAGESHAKE_URL=http://localhost/api/submit",
+                "--dart-define",
+                "PUSH_SERVER=sygnal.SERVER"
             ]
         },
         {
```

Stil to do:
- [x] changelogs
- [x] End-user documentation.
- [x] confirm this still builds on all systems
   - [x] MacOS
   - [x] Windows
   - [x] Linux
   - [x] iOS
   - [x] Android 
- [ ] limit uniffi-features behind the flag to keep the overhead for other platforms low

For later
- Desktop Push:
   - Windows: https://learn.microsoft.com/en-us/azure/developer/mobile-apps/notification-hubs-backend-service-flutter
   - MacOS (similar as for iOS?!?)
   - Linux ??? maybe with [unifiedPush](https://unifiedpush.org/)?

Android:
[android push.webm](https://github.com/acterglobal/a3/assets/40496/971f7835-00b1-4652-8084-83c7fb555738)

iOS:

https://github.com/acterglobal/a3/assets/40496/9d8adbd1-19ab-42ad-8a03-3c998cf480a6


app/ios/Podfile.lock
```diff
@@ -166,4 +174,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ef19549a9bc3046e7bb7d2fab4d021637c0c58a3
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.13.0
```
gtalha07, line 177: As per from your information, the cocoapods version change was causing some issue in iOS builds. Is this resolved now?

app/lib/common/dialogs/nuke_confirmation.dart
```diff
@@ -0,0 +1,79 @@
...
+          textAlign: TextAlign.left,
+          text: const TextSpan(
+            text: 'Attention: ',
+            style: TextStyle(color: Colors.white, fontSize: 15),
+            children: <TextSpan>[
+              TextSpan(
+                text:
+                    'Nuking removes all local data, including encryption keys. If this is your last signed-in device you might no be able to decrypt any previous content.',
+              ),
+              TextSpan(text: 'Are you sure you want to nuke?'),
```
gtalha07, line 32: Nuke seems a strong wording. I would probably go for ```Erase``` or ```Wipe``` in terms of easy understanding and no assumed double meanings.

app/lib/common/dialogs/nuke_confirmation.dart
```diff
@@ -0,0 +1,79 @@
...
+// Ref can be used to read any provider which are declared.
+void nukeConfirmationDialog(BuildContext context, WidgetRef ref) {
+  showDialog(
+    context: context,
+    builder: (BuildContext ctx) {
+      return AlertDialog(
+        title: const Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Icon(Atlas.bomb_bold,color: Colors.red,),
```
gtalha07, line 17: I would also be not be in favor of using this icon, could be brush or some more friendly warning sign...

app/ios/Podfile.lock
gnunicorn: I think it is resolved, yes.

app/lib/common/dialogs/nuke_confirmation.dart
gnunicorn: It is intentionally strong wording as this is super dangerous... and should only taken if there is no other course of action left ... it won't even back up ... Hence also using the bomb_icon and big-red-letters ...

I mean ... _ideally_ this screen never shows up to a regular user. It only does when we got the underlying SDK into an unrecoverable state - most likely upon startup.

app/lib/common/dialogs/nuke_confirmation.dart
gnunicorn: brush?

It was intended to be something that no one clicks without understanding the risk ... as this is a super-wipe-out without backups ... and see comments as above ...

app/lib/features/activities/providers/notifiers/notifications_list_notifier.dart
```diff
@@ -76,17 +71,17 @@ class NotificationsListNotifier extends StateNotifier<NotificationListState>
...
+        // );
 
-        final provider = ref.watch(featuresProvider);
-        if (provider.isActive(LabsFeature.showNotifications)) {
-          if (!ev.read()) {
-            final brief = extractBrief(ev);
-            notify(brief);
-          }
-        }
+        // final provider = ref.watch(featuresProvider);
```
gtalha07, line 78: Why is this commented out? Or its been moved somewhere else? If it is, you can remove comment.

app/lib/router/router.dart
```diff
@@ -64,16 +65,24 @@ Future<String?> authGuardRedirect(
...
+    }
+
+    if (autoGuestLogin) {
+      // if compiled with auto-guest-login, create an account
+      await acterSdk.newGuestClient(setAsCurrent: true);
+      return null;
+    }
+  } catch (error, trace) {
+    // ignore: deprecated_member_use
+    return state.namedLocation(
```
gtalha07, line 82: Very nice, a better handling approach.

docs/content/docs/dev/troubleshooting.md
```diff
@@ -32,14 +32,100 @@ We are using [`path_provider`'s `getApplicationDirectory`](https://pub.dev/packa
 Just opening a terminal and doing `rm -rf $PATH` (where `$PATH` is the necessary path above), usually resets all locally stored data. **Careful as this also removes all locally held crypto tokens**.
 
 ## Execution failed for task ':app:processDebugMainManifest'.
+
 > Unable to make field private final java.lang.String java.io.File.path accessible: module java.base does not "opens java.io" to unnamed module @7fbe8ef
 
 If you happen to come across this exception while compiling Android build for application launch, the reason for that is with the flutter fetching the JDK path from Android Studio directory as a default option. This is reported to be occuring in latest **Android Studio Flamingo (2022.2.1)** in reference to the Flutter issue [#106416](https://github.com/flutter/flutter/issues/106416).
```
gtalha07, line 38: this issue had been resolved in newly revised version. We can remove this :)

gtalha07 (approved):
Very well put! Excited to try out new stuff.....

app/lib/common/dialogs/nuke_confirmation.dart
gtalha07: As in means of cleaning as we can see some cleaning data apps, anyways I would like to hear what's the general opinion on this via feedback.

app/lib/features/activities/providers/notifiers/notifications_list_notifier.dart
gnunicorn: It was some previous code that would show local notifications just coming from our own local supply mechanism and show a local notification regardless of actual push ability on the device ... I suppose we could think about activating something similar again for desktop but we'd have to think about how to phrase that and actually make sure we get the needed data ... I'll remove this for now.

app/lib/features/activities/providers/notifiers/notifications_list_notifier.dart
```diff
@@ -76,17 +71,17 @@ class NotificationsListNotifier extends StateNotifier<NotificationListState>
...
+        // if (provider.isActive(LabsFeature.showNotifications)) {
+        //   if (!ev.read()) {
+        //     final brief = extractBrief(ev);
+        //     // notify(brief);
+        //   }
+        // }
```
gnunicorn, line 84: remove this:
```suggestion
```

bitfriend:
Great work

PR #948, opened 2023-09-14
switch default homeserver to staging environment

gnunicorn:
We have a development/staging environment for testing. This PR switches the default development environment over to use them for registration and login. It does not effect already logged in accounts. However, when you login again next you have to use the full-user-id or register a new account with the staging server to use it.

Nightly builds are not effected - they will continue to use the production environment.


gtalha07 (approved):
Looking good!

PR #956, opened 2023-09-16
Reworking tasks

gnunicorn:
Minimal side-project I've worked on during my off-time: Tasks & TaskLists.

- [x] Refactor to use latest riverpods infrastructure
   - [x] remove old stuff
   - [x] remove get(x)
- [x] minimalist UI
- [x] create lists
- [x] default-off app-setting
- [x] inline add-task

Simplified it now works like this:
new-simplified-tasks


app/lib/features/search/widgets/quick_actions_builder.dart
```diff
@@ -55,16 +55,16 @@ class QuickActionsBuilder extends ConsumerWidget {
...
-              ? OutlinedButton.icon(
-                  onPressed: () {
-                    navigateTo(route: Routes.actionAddTask, push: true);
-                    debugPrint('Add Task');
-                  },
-                  icon: const Icon(Atlas.plus_circle_thin),
-                  label: const Text('Task'),
-                )
-              : null,
+          // isActive(LabsFeature.tasks)
```
gtalha07, line 58: why its commented out, we don't want to have tasks being accessed from quick actions ?

app/lib/features/tasks/dialogs/create_task_list_sheet.dart
```diff
@@ -0,0 +1,169 @@
...
+          onPressed: () async {
+            if (_formKey.currentState!.validate()) {
+              DefaultDialog(
+                title: Text(
+                  'Posting TaskList',
+                  style: Theme.of(context).textTheme.titleSmall,
+                ),
+                isLoader: true,
+              );
+              try {
```
gtalha07, line 120: Can we have this non UI logic be a part of function which can be called here? Makes widget tree a lot clean and easier to refactor.

app/lib/features/tasks/pages/tasks_page.dart
```diff
@@ -0,0 +1,85 @@
...
+    return Scaffold(
+      body: CustomScrollView(
+        slivers: <Widget>[
+          PageHeaderWidget(
+            title: 'Tasks',
+            sectionDecoration: const BoxDecoration(
+              gradient: AppTheme.primaryGradient,
+            ),
+            actions: [
+              // IconButton(
```
gtalha07, line 29: We can remove that rather than commenting :)

app/lib/features/tasks/widgets/task_entry.dart
```diff
@@ -0,0 +1,139 @@
...
+                      ),
+                    );
+                  },
+                  error: (e, s) => Text('loading comments failed: $e'),
+                  loading: () => const SizedBox.shrink(),
+                ),
+      ),
+    );
+    final assignees = task.assignees();
+    if (assignees.isNotEmpty) {}
```
gtalha07, line 93: The purpose of this being here isn't clear to me. We do want to show something if its not empty???

app/lib/features/tasks/widgets/task_list_card.dart
```diff
@@ -0,0 +1,239 @@
...
+import 'package:flutter/material.dart';
+
+@immutable
+class TasksOverview {
+  final List<Task> openTasks;
+  final List<Task> doneTasks;
+  const TasksOverview({required this.openTasks, required this.doneTasks});
+}
+
+class TasksNotifier extends FamilyAsyncNotifier<TasksOverview, TaskList> {
```
gtalha07, line 17: Would suggest moving this all notifier related to ```notifiers```.

gtalha07 (approved):
Yahoo! We finally got rid of another framework (GetX) inside framework. Overall looks much nicer. There are few suggestions which can be done after merge also but feel free to do changes if needed.

app/lib/features/tasks/widgets/task_entry.dart
gnunicorn: ah, that's old stuff. Gotta remove it.

app/lib/features/tasks/widgets/task_list_card.dart
```diff
@@ -0,0 +1,239 @@
...
+    return retState;
+  }
+}
+
+final tasksProvider =
+    AsyncNotifierProvider.family<TasksNotifier, TasksOverview, TaskList>(() {
+  return TasksNotifier();
+});
+
+class InlineTaskAdd extends StatefulWidget {
```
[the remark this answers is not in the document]
gnunicorn: What order should it have? And I assume by private you mean starting with an underscore?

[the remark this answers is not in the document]
gtalha07: By ordering meant the widget should be at the top on the basis of the file naming .i.e. task_list_card. Providers should be private if only used here since they don't need to be kept global if not shared state. Otherwise too much common naming might conflict with other providers, unintentionally..

app/lib/features/search/widgets/quick_actions_builder.dart
gtalha07: Okay then we can remove that and re-add when its done ;)

PR #957, opened 2023-09-16
preserve chat input state across rooms

### LINKED ISSUE #954: keep input state linked to chat room
When we switch chat rooms, the input state should stay in memory for each room. So if I started typing or replying and then went to a different room but came back, it should still be in that state.

gtalha07:
- fixes #954.

### DIFF
--- a/app/lib/features/chat/widgets/custom_input.dart
+++ b/app/lib/features/chat/widgets/custom_input.dart
@@ -26,7 +26,8 @@ import 'package:intl/intl.dart' show toBeginningOfSentenceCase;
 import 'package:mime/mime.dart';
 
 // keep track of text controller values across rooms.
-final _textValuesProvider = StateProvider<Map<String, String>>((ref) => {});
+final _textValuesProvider =
+    StateProvider.family<String, String>((ref, roomId) => '');
 
 class CustomChatInput extends ConsumerStatefulWidget {
   final Convo convo;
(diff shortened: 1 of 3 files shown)


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -25,7 +25,8 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:intl/intl.dart' show toBeginningOfSentenceCase;
 import 'package:mime/mime.dart';
 
-final mentionKey = GlobalKey<FlutterMentionsState>();
+// keep track of text controller values across rooms.
+final _textValuesProvider = StateProvider<Map<String, String>>((ref) => {});
```
gnunicorn, line 29: Why use a MapState rather than just a state family?

```suggestion
final _textValuesProvider = StateProvider.family<String, String>((ref) => '');
```

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -607,15 +609,31 @@ class _FileWidget extends ConsumerWidget {
...
+
+    if (textValues[roomId] != null) {
+      textValues[roomId] = '';
+      String value = textValues[roomId]!;
+      value += mentionKey.currentState!.controller!.text;
+      textValues[roomId] = value;
+    } else {
+      textValues[roomId] = mentionKey.currentState!.controller!.text;
+    }
+    ref.read(_textValuesProvider.notifier).update((state) => textValues);
```
gnunicorn, line 634: Using a family would also shorten this to:

```suggestion
    var textValues = ref.watch(_textValuesProvider);
    textValues += mentionKey.currentState!.controller!.text;
    ref.read(_textValuesProvider(roomId).notifier).update((state) => textValues);
```

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -624,6 +642,8 @@ class _TextInputWidget extends ConsumerWidget {
     final width = MediaQuery.of(context).size.width;
     return FlutterMentions(
       key: mentionKey,
+      // restore input if available
+      defaultText: ref.watch(_textValuesProvider)[roomId],
```
gnunicorn, line 646:
```suggestion
      defaultText: ref.watch(_textValuesProvider(roomId)),
```

gnunicorn (approved):
Looks good, but why not use a familyprovider?

PR #968, opened 2023-09-18
fix fallback user avatar size

### LINKED ISSUE #930: Fallback Avatar of User too small
Register a new account and you'll find the default selected fallback-avatar is too small:

Image

gtalha07:
- fixes #930.
- hide user avatar in dashboard (desktop) since we have in navigation rail already.
Screenshot 2023-09-18 at 8 10 22 PM
Screenshot 2023-09-18 at 8 10 41 PM

### DIFF
--- a/app/lib/common/widgets/user_avatar.dart
+++ b/app/lib/common/widgets/user_avatar.dart
@@ -23,7 +23,7 @@ class UserAvatarWidget extends ConsumerWidget {
             ActerAvatar(
               mode: DisplayMode.User,
               uniqueId: client.userId().toString(),
-              size: size,
+              size: data.profile.hasAvatar() ? size : size * 2,
               avatar: data.profile.getAvatarImage(),
               displayName: data.profile.displayName,
             ),
--- a/app/lib/features/home/pages/dashboard.dart
+++ b/app/lib/features/home/pages/dashboard.dart
@@ -138,14 +138,17 @@ class _DashboardState extends ConsumerState<Dashboard> {
                         mode: DisplayMode.User,
                       ),
                     ),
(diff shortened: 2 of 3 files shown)


app/lib/common/widgets/user_avatar.dart
```diff
@@ -23,7 +23,7 @@ class UserAvatarWidget extends ConsumerWidget {
             ActerAvatar(
               mode: DisplayMode.User,
               uniqueId: client.userId().toString(),
-              size: size,
+              size: data.profile.hasAvatar() ? size : size * 2,
```
gnunicorn, line 26: it really feels like we have a bunch of place where we do this... like ... we should probably take a look at our package and figure out why that is a problem, rather than putting a conditional `size * 2` everywhere it is being used...

app/lib/features/home/pages/dashboard.dart
```diff
@@ -138,14 +138,17 @@ class _DashboardState extends ConsumerState<Dashboard> {
...
                     ),
-                    child: Container(
-                      key: Keys.avatar,
-                      margin: const EdgeInsets.all(8),
-                      child: InkWell(
-                        onTap: () => context.pushNamed(Routes.myProfile.name),
-                        child: const UserAvatarWidget(size: 20),
-                      ),
-                    ),
+                    child: !isDesktop
```
gnunicorn, line 141: arguably this should depend on the width/whether the sidebar is available rather than it "being desktop", but sure, let's do this for now...

gnunicorn (approved):
neither of these fixes is the fully properly one, but will do for now.

Issue #800, opened 2023-07-28
Clean up calendar events


gnunicorn:
We have several small bugs in the current UI for calendar events, we need to address

The widgets on home and the space overview:
 - [ ] the calender shown is not connected to the events - when scrolling the calendar the event list doesn't change/update
 - [ ] there is the #742 on the calendar
 - [ ] we should filter the list of events to shown to only upcoming events
 - [ ] and on home only the ones you agreed to or said `Maybe` to 
 - [ ] and limit them to the next 5, with a bubble on the header with the counter and the "see all"-link underneath if there are more.
 - [ ] and show the rsvp-status per event

 On the events page and the space-events page:
 - [ ] divide into upcoming and past events
 - [ ] show calendar, link it to filter for events
 - [ ] show rsvp-status per event

gtalha07:
Closed as completed via #1040.

PR #969, opened 2023-09-18
Dashboard/Overview and Calendar Event fixes

### LINKED ISSUE #944: Add "+" icon to "my spaces" header on home screen
As new users now automatically join a space (Acter.global + Sandbox) they no longer get prompted with the "create space" pop up. Thats why i think its important that we either 

1. add the + icon to "my spaces" header on home screen
2. or, make it more clear that the "my spaces" is clickable.

Personally i would go for nr.1 as it doesnt take up a lot of space and its the shortest click path

Image

If its not a big job, adding the + icon in the space menu on desktop (like element) would also make it easier for users to get started with creating their own space.

Image

gtalha07:
- fixes #944.
- See #800 tracking.

### DIFF
--- a/app/lib/features/events/pages/events_page.dart
+++ b/app/lib/features/events/pages/events_page.dart
@@ -2,7 +2,6 @@ import 'dart:core';
 import 'dart:math';
 
 import 'package:acter/common/providers/common_providers.dart';
-import 'package:acter/common/snackbars/custom_msg.dart';
 import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/utils/utils.dart';
 import 'package:acter/common/widgets/default_page_header.dart';
@@ -32,15 +31,6 @@ class EventsPage extends ConsumerWidget {
               gradient: AppTheme.primaryGradient,
             ),
             actions: [
-              IconButton(
-                icon: const Icon(Atlas.funnel_sort_thin),
-                onPressed: () {
-                  customMsgSnackbar(
-                    context,
-                    'Events filtering not yet implemented',
-                  );
-                },
-              ),
               IconButton(
                 icon: Icon(
                   Atlas.plus_circle_thin,
--- a/app/lib/features/events/pages/events_page.dart
+++ b/app/lib/features/events/pages/events_page.dart
@@ -3,31 +3,31 @@ import 'dart:math';
 
 import 'package:acter/common/providers/common_providers.dart';
 import 'package:acter/common/snackbars/custom_msg.dart';
+import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/widgets/default_page_header.dart';
+import 'package:acter/features/events/widgets/events_item.dart';
+import 'package:acter/features/home/providers/events.dart';
 import 'package:atlas_icons/atlas_icons.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:acter/features/pins/widgets/pin_list_item.dart';
-import 'package:acter/features/pins/providers/pins_provider.dart';
-import 'package:go_router/go_router.dart';
-import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/themes/app_theme.dart';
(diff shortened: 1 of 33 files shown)


app/lib/features/events/pages/events_page.dart
```diff
@@ -0,0 +1,100 @@
...
+            actions: [
+              IconButton(
+                icon: const Icon(Atlas.funnel_sort_thin),
+                onPressed: () {
+                  customMsgSnackbar(
+                    context,
+                    'Events filtering not yet implemented',
+                  );
+                },
+              ),
```
gnunicorn, line 43: Let's remove /disable buttons without function ...
```suggestion
```

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -76,8 +90,6 @@ class QuickJump extends ConsumerWidget {
                     icon: SvgPicture.asset(
                       'assets/images/tasks.svg',
                       semanticsLabel: 'tasks',
-                      width: 32,
-                      height: 32,
```
gnunicorn, line 80: that is too small now. Please revert his and the `iconSize: 48` above.
grafik

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -56,12 +55,27 @@ class QuickJump extends ConsumerWidget {
                     onPressed: () {
                       navigateTo(route: Routes.pins);
                     },
-                    icon: const Icon(Atlas.pin_thin, size: 32),
+                    icon: const Icon(Atlas.pin_thin),
```
gnunicorn, line 58:
```suggestion
                    icon: const Icon(Atlas.pin_thin, size: 32),
```

Pretty sure the size was set on purpose because otherwise this icon is crammed into the circle, appearing to pierce it:

grafik

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -98,11 +109,9 @@ class QuickJump extends ConsumerWidget {
               },
               icon: const Icon(
                 Atlas.chats_thin,
-                size: 32,
```
gnunicorn, line 101: If I am not mistaken that was smaller on purpose, too. As seen in the image above, it is just too large for the circle it is in...

app/lib/features/space/widgets/chats_card.dart
```diff
@@ -15,59 +15,62 @@ class ChatsCard extends ConsumerWidget {
...
+                  'There are no chats in this space',
+                  style: Theme.of(context).textTheme.bodySmall,
+                );
+              }
+              return Column(
+                children: [
+                  ListView.builder(
+                    shrinkWrap: true,
+                    physics: const NeverScrollableScrollPhysics(),
+                    itemCount: chats.length > 3 ? 3 : chats.length,
```
gnunicorn, line 43: how about?
```suggestion
                    itemCount: min(chats.length, 3), 
```

gnunicorn (approved):
thanks, mate. I have some UI fixes, but we can postpone them if necessary...

app/lib/features/pins/pages/pins_page.dart
```diff
@@ -67,9 +71,10 @@ class PinsPage extends ConsumerWidget {
               }
               return SliverGrid.builder(
                 itemCount: pins.length,
-                gridDelegate: SliverGridDelegateWithFixedCrossAxisCount(
+                gridDelegate:
+                    SliverGridDelegateWithFixedCrossAxisCountAndFixedHeight(
                   crossAxisCount: max(1, min(widthCount, minCount)),
-                  childAspectRatio: 6,
+                  height: MediaQuery.of(context).size.height * 0.1,
```
gnunicorn, line 77: I think this causes some ugly design, now:

grafik

gtalha07: Let me take a look at it and do quick fix!

gtalha07: Increased the height of card to contain chip: ``` height: MediaQuery.of(context).size.height * 0.15,```

PR #971, opened 2023-09-19
Fix parsing error from rust analyzer

bitfriend:
Will fix [E0023](https://doc.rust-lang.org/stable/error_codes/E0023.html).
image


gnunicorn:
Is that a change in enum_dispatch? Because it used to work before, did it not?

gnunicorn (approved):

bitfriend:
Yeah, `enum_dispatch` seems to occur [E0023](https://doc.rust-lang.org/stable/error_codes/E0023.html) now.

PR #972, opened 2023-09-19
Fix crash in RSVP sending

bitfriend:
When sending RSVP, app crashes.
Will fix it.
image

### DIFF
--- a/native/acter/api.rsh
+++ b/native/acter/api.rsh
@@ -315,20 +315,24 @@ object CalendarEvent {
 object CalendarEventUpdateBuilder {
     /// set title of the event>
     fn title(title: string);
+
     /// set description text
     fn description_text(body: string);
+
     /// set utc start in rfc3339 string
-    fn utc_start_from_rfc3339(utc_start: string);
+    fn utc_start_from_rfc3339(utc_start: string) -> Result<()>;
     /// set utc start in rfc2822 string
-    fn utc_start_from_rfc2822(utc_start: string);
+    fn utc_start_from_rfc2822(utc_start: string) -> Result<()>;
     /// set utc start in custom format
-    fn utc_start_from_format(utc_start: string, format: string);
+    fn utc_start_from_format(utc_start: string, format: string) -> Result<()>;
+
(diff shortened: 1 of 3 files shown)


bitfriend:
@gnunicorn explained me about it.

>I recommend looking closely at all types handed out in the API calls around the crashing calls, see if any of them might be references or expecting referenced types ...
>I had the a similar problem in tasks using `description()` not returning a clone and so rust would discard the object after and any further call of that function or object would return garbled memory ...

bitfriend:
`RsvpDraft::status()` returns `&mut Self` on rust side.
But it should return `void` on flutter side.
Otherwise app crashes when calling of `RsvpDraft::status()` on flutter side.

native/acter/api.rsh
```diff
@@ -387,7 +387,7 @@ object RsvpManager {
 
 object RsvpDraft {
     /// set status of this RSVP
-    fn status(status: string) -> RsvpDraft;
+    fn status(status: string);
```
gnunicorn, line 390: doesn't that still have an `unreachable` internally?

gnunicorn (approved):
I'm fine with merging this as is, but there is still an unreachable we need to remove, too.

native/acter/api.rsh
gnunicorn: as said, let's merge as-is

native/acter/src/api/rsvp.rs
```diff
@@ -75,19 +78,32 @@ pub struct RsvpDraft {
...
 impl RsvpDraft {
     pub fn status(&mut self, status: String) -> &mut Self {
-        let Ok(s) = RsvpStatus::from_str(&status) else {
-            unreachable!("Wrong status about RSVP")
-        };
-        self.inner.status(s);
+        if let Ok(s) = RsvpStatus::from_str(&status) {
+            self.inner.status(s);
+        } else {
+            error!("Wrong status about RSVP");
```
gnunicorn, line 84: yeah, that's fine.

PR #978, opened 2023-09-20
Refactor media news slide & media message

bitfriend:
- Improve media news slides
- Organize the order of arguments in media messages
- Parse mimetype when sending media msg
- Parse filesize when sending media msg


native/core/src/events/news.rs
```diff
@@ -101,50 +98,6 @@ pub struct NewsSlide {
     pub references: Vec<ObjRef>,
 }
 
-impl NewsSlide {
```
gnunicorn, line 104: I think I already mentioned in a previous review that I am not fond of moving APIs that are generic and can be used rather broadly outside the ffi-layer (so, without a `RUNTIME`) out of `core` and into the higher level. I'd rather want the higher level as thin as possible: everything that needs RUNTIME must be there, like things that interact with client, but things that do not need it should be located in the underlying crate. That is especially true for the instantiation of types from this crates that the upper crate ideally doesn't do directly but always uses the helper functions for, as existed here before.

native/acter/src/api/news.rs
```diff
@@ -386,54 +387,55 @@ pub struct NewsEntryDraft {
...
     pub fn add_text_slide(&mut self, body: String) -> &mut Self {
+        let text_content = TextMessageEventContent::markdown(body);
         self.slides.push(NewsSlide {
             client: self.client.clone(),
             room: self.room.clone().into(),
-            inner: news::NewsSlide::new_text(body),
+            inner: news::NewsSlide {
+                content: NewsContent::Text(text_content),
+                references: Default::default(),
+            },
```
gnunicorn, line 397: I prefer it the way it was before. I explain more in the comment around the section change.

gnunicorn (changes requested):
Looks fine, but the entanglement of types in the higher layer should be reverted IMHO. Left a comment on the section to explain the thoughts behind that idea.

native/core/src/events/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/978/commits/ef091de0739d8d31a253cec981d154fb4d4895ec

native/acter/src/api/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/978/commits/ef091de0739d8d31a253cec981d154fb4d4895ec

native/acter/src/api/common.rs
```diff
@@ -442,19 +442,19 @@ impl LocationDesc {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ReactionRecord {
     sender_id: OwnedUserId,
-    timestamp: MilliSecondsSinceUnixEpoch,
+    ts: MilliSecondsSinceUnixEpoch,
```
gnunicorn, line 445: any particular reason this expressive name has been replaced by a short less clear name? I see this also extended to the API and I am afraid this causes more confusion than it is helpful...

gnunicorn (changes requested):
The core feature set seems fine but there are seemingly unrelated API changes of u64 to u32 conversion that a) I understand and b) now can induce a crash of the App. Please revert those.

native/acter/src/api/common.rs
bitfriend: `ReceiptRecord` has `ts` field too.
So I wanted to unify those fields in name.

native/acter/src/api/search.rs
```diff
@@ -43,8 +43,12 @@ impl PublicSearchResultItem {
...
     }
 
-    pub fn num_joined_members(&self) -> u64 {
-        self.chunk.num_joined_members.into()
+    pub fn num_joined_members(&self) -> u32 {
+        let count = u64::from(self.chunk.num_joined_members);
+        if count > u32::MAX as u64 {
+            panic!("count of joined members overflowed");
+        }
+        count as u32
```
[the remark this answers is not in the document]
bitfriend: Using of `u32` was intended for javascript.
If you don't agree with `u32`, I will revert all occurrences of `u32` to `u64`.

native/acter/src/api/search.rs
```diff
@@ -101,8 +105,14 @@ impl PublicSearchResult {
         self.resp.prev_batch.clone()
     }
 
-    pub fn total_room_count_estimate(&self) -> Option<u64> {
-        self.resp.total_room_count_estimate.map(Into::into)
+    pub fn total_room_count_estimate(&self) -> Option<u32> {
```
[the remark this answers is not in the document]
bitfriend: If we don't take account of javascript, I will revert all occurrences of `u32` to `u64`.

native/acter/src/api/room.rs
```diff
@@ -171,6 +172,14 @@ impl Member {
         self.membership_status().to_string()
     }
 
+    pub fn power_level(&self) -> u32 {
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/978/commits/cde0cd8b038c74f0d70eed95eccb9e8940b9203f

native/acter/src/api/search.rs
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/978/commits/cde0cd8b038c74f0d70eed95eccb9e8940b9203f

native/acter/src/api/search.rs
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/978/commits/cde0cd8b038c74f0d70eed95eccb9e8940b9203f

gnunicorn (approved):

PR #992, opened 2023-09-25
Profile page UI update

kokodavid:


gnunicorn:
@kokodavid got some screenshots for us here? It's hard imagine from the code, what this actually looks like ...

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -185,7 +167,8 @@ class MyProfile extends ConsumerWidget {
                   ),
                 ),
                 Container(
-                  width: MediaQuery.of(context).size.width - 100,
+                  width: isDesktop ? 400 : MediaQuery.of(context).size.width -15,
```
gnunicorn, line 170: Okay, I think we need to clarify _when_ `isDesktop` is okay to be used... because here it seems out of place again. A local width should not be informed by the platform but by the available space. The latest iPad in Landscape-Mode has a higher resolution than a MacBook Air 2017 - yet the later would be understood as a `isDesktop` while the former is not. And in neither case can you be sure you are on full-screen view (I almost never have Acter on full-screen). We need to think in a responsive pattern rather than this black-and-white pattern.

If you have a default size it should be that is meant to be responsive (like the MediaQuery used here - very good!) but it must tap-out for certain sizes, please use [`BoxConstraints`](https://api.flutter.dev/flutter/rendering/BoxConstraints-class.html):

```suggestion
                  constraints: const BoxConstraints(maxWidth: 400, width: MediaQuery.of(context).size.width -15),
```

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -235,42 +231,66 @@ class MyProfile extends ConsumerWidget {
...
+                        ),
+                        child:    Column(
+                          children: [
+                             ProfileItemTile(icon: Atlas.check_shield,title: 'Linked Devices',onPressed: () =>
+                            context.pushNamed(Routes.settingSessions.name),color: Colors.white,),
+                            const Divider(indent: 40,endIndent: 10,),
+                             ProfileItemTile(icon: Atlas.gear,title: 'Settings',onPressed: () =>
+                            context.pushNamed(Routes.settings.name),color: Colors.white,),
+                          ],
+                        ),
```
gnunicorn, line 249: this looks like incorrect formatting to me ...

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -235,42 +231,66 @@ class MyProfile extends ConsumerWidget {
...
-                      const SizedBox(height: 30),
-                      OutlinedButton.icon(
-                        icon: const Icon(Atlas.exit_thin),
-                        onPressed: () => logoutConfirmationDialog(context, ref),
-                        label: const Text('Logout'),
+                      const SizedBox(height: 5),
+                        Container(
+                        margin: const EdgeInsets.symmetric(vertical: 10),
+                        decoration: BoxDecoration(
+                          color: const Color(0xFF122D46),
```
gnunicorn, line 238: eeehh... let's not hard-code colors. Please use the AppThemes and add a color if necessary.

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -235,42 +231,66 @@ class MyProfile extends ConsumerWidget {
...
+                            color: const Color(0xFF122D46),
+                            borderRadius: BorderRadius.circular(10),
+                          ),
+                          child:    Column(
+                            children: [
+                               ProfileItemTile(icon: Atlas.bell_reminder,title: 'Notifications',onPressed: () =>
+                              context.pushNamed(Routes.settingSessions.name),color: Colors.white,),
+                              const Divider(indent: 40,endIndent: 10,),
+                               ProfileItemTile(icon: Icons.star_border_outlined,title: 'Rate us',onPressed: () =>
+                              context.pushNamed(Routes.settings.name),color: Colors.white,),
```
gnunicorn, line 268: same as above: style seems odd, please do not hard-code colors

gnunicorn, line 268: Let's not add more dead-buttons. Either we put something useful behind this button (not just moving to `settings`) or it shouldn't appear in the UI. We are trying to remove the "buttons that don't do anything and thus appear broken to the user" from the UI...

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -235,42 +231,66 @@ class MyProfile extends ConsumerWidget {
...
+                    Container(
+                        margin: const EdgeInsets.symmetric(vertical: 10),
+                        decoration: BoxDecoration(
+                          borderRadius: BorderRadius.circular(10),
+                          border: Border.all(color: AppTheme.brandColorScheme.onError),
+                        ),
+                        child:    Column(
+                          children: [
+                             ProfileItemTile(icon: Atlas.exit,title: 'Logout',onPressed: () =>
+                            logoutConfirmationDialog(context, ref),color: Colors.red,),
```
gnunicorn, line 287: colors and style again.

app/lib/features/profile/widgets/profile_item_tile.dart
```diff
@@ -0,0 +1,37 @@
...
+          mainAxisAlignment: MainAxisAlignment.spaceBetween,
+          children: [
+            Row(
+              children: [
+                 Icon(icon,color: color,),
+                 const SizedBox(width: 10,),
+            Text(title,style: TextStyle(color: color),),
+              ],
+            ),
+            Icon(Icons.arrow_forward_ios,size: 19,color: color,),
```
gnunicorn, line 31: are we mixing spaces and tabs or what is going on here?

gnunicorn (approved):
I've a few remarks regarding style, broken buttons and potential down-the-line issue with colors and responsiveness. Please improve and provide screenshots. Other than that it is ready to go.

app/lib/features/profile/pages/my_profile_page.dart
kokodavid: Is there a format used in naming the custom color schemes ?

app/lib/features/profile/pages/my_profile_page.dart
kokodavid: Notifications and Rate us buttons are not visible to the users 
image

gnunicorn: what happened to "deactivate account"? We need that (Apple AppStore requirement)

gnunicorn:
Good to see the screenshot, otherwise I wouldn't have noticed: We need to keep the "deactivate Account" in there, it's an Apple AppStore Requirement that we have it. Maybe we can make the "Logout" an orange/yellow and the "Deactivate" in the red?

app/lib/features/profile/pages/my_profile_page.dart
kokodavid: Regarding that , I noticed that we have logout and deactivate account in the Settings Menu , Or should I just include it in MyProfile page ?
image

gnunicorn: @kokodavid ah, right. no, that's fine. my bad. keep it as is.

app/lib/features/profile/pages/my_profile_page.dart
gnunicorn: @kokodavid well, being specific I'd say. Like `DangerBoxDecorationColor`, is what I'd propose here.

PR #1005, opened 2023-09-26
update room profile section, colors

gtalha07:
- Update Acter component theme to be more in sync with overall system.
- Correct acter gradient patterns.
- Updated look on profile section of rooms with the support of split view on desktop.
- fix for #1006.
- overflow fixes for:
  - Activities, Pins and Events screen bars.
  - Pins, Event and Space cards.


gtalha07:
Screenshot 2023-09-27 at 7 29 14 PM
Screenshot 2023-09-27 at 7 29 57 PM
Screenshot 2023-09-27 at 7 30 23 PM

app/lib/features/chat/pages/chats_shell.dart
```diff
@@ -12,22 +14,33 @@ class ChatShell extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final currentId = ref.watch(selectedChatIdProvider);
+    final isExpanded = ref.watch(hasExpandedPanel);
     return LayoutBuilder(
       builder: (context, constraints) {
         if (constraints.maxWidth >= sidebarMinWidth) {
+          WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+            ref.read(inSideBarProvider.notifier).update((state) => true);
+          });
```
gnunicorn, line 23: doing these kinda responsive things with providers always feels very weird too many. And using `addPostFrameCallback` even more ...  I feel like often times this call  has caused us quite some harm as it isn't actually reactive...

but let's do this for now ...

app/lib/features/chat/pages/room_page.dart
```diff
@@ -73,193 +75,220 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
...
+                  ),
+                ),
+              ],
+            ),
+            SliverFillRemaining(
+              child: Container(
+                decoration: const BoxDecoration(
+                  gradient: AppTheme.primaryGradient,
+                ),
+                child: Chat(
```
gnunicorn, line 185: I assume all that is only because of indentation changes...

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -23,216 +25,248 @@ class RoomProfilePage extends ConsumerWidget {
...
+                    padding: const EdgeInsets.only(
+                      top: 38,
+                      bottom: 12,
+                    ),
+                    child: SpaceParentBadge(
+                      badgeSize: 20,
+                      roomId: roomId,
+                      child: convoProfile.when(
+                        data: (profile) => ActerAvatar(
+                          mode: profile.isDm
```
gnunicorn, line 88: same as above. add a FIXME please.

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -23,216 +25,248 @@ class RoomProfilePage extends ConsumerWidget {
...
+                      data: (membership) => SettingsTile.navigation(
+                        onPressed: (ctx) {
+                          membership!.canString('CanInvite')
+                              ? ctx.pushNamed(
+                                  Routes.spaceInvite.name,
+                                  pathParameters: {'spaceId': roomId},
+                                )
+                              : customMsgSnackbar(
+                                  ctx,
+                                  'Not enough power level for invites, ask room administrator to change it',
```
gnunicorn, line 167: eh... the invite button should rather not be shown at all if the user doesn't have the power level to do so ...

but I am okay if you put that into a ticket post-merge..

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -23,216 +25,248 @@ class RoomProfilePage extends ConsumerWidget {
...
                           ),
                         );
                         customMsgSnackbar(
                           context,
-                          'Room ID copied to clipboard',
+                          'Room ID: $roomId copied to clipboard',
                         );
                       },
+                      title: Text(
+                        'Copy Room Link',
```
gnunicorn, line 200:
```suggestion
                        'Copy Room Id',
```

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -23,216 +25,248 @@ class RoomProfilePage extends ConsumerWidget {
...
+                        'Leave Room',
+                        style: tileTextTheme!.copyWith(
+                          color: Theme.of(context).colorScheme.badgeUrgent,
+                        ),
+                      ),
+                      leading: Icon(
+                        Atlas.trash_thin,
+                        size: 18,
+                        color: Theme.of(context).colorScheme.badgeUrgent,
+                      ),
```
gnunicorn, line 262: no `onPressed`? should probably have one that
1. confirms they want to leave
2. removes them from the room
3. routes somewhere else (`/chat`?)

if you don't want to do that before merge, then remove the section for now and put that into a ticket for later. But a button that again doesn't work isn't really acceptable to end users...

app/lib/features/pins/pages/pins_page.dart
```diff
@@ -43,9 +43,13 @@ class PinsPage extends ConsumerWidget {
                 ),
               ),
             ],
-            expandedContent: const Text(
-              'Pinned items from all the Spaces you are part of',
-            ),
+            expandedContent: size.width <= 600
+                ? null
+                : Text(
```
gnunicorn, line 48: I wonder if we can't just outsource that into a specific widget rather than copying that `<= 600` everywhere (also... why that number and why is it based on the `width`?)?

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -161,30 +161,34 @@ class QuickJump extends ConsumerWidget {
...
-                color: Colors.white,
+    return Scaffold(
+      body: Padding(
+        padding: const EdgeInsets.all(16.0),
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          mainAxisAlignment: MainAxisAlignment.start,
+          children: <Widget>[
+            Expanded(
+              child: TextField(
```
gnunicorn, line 172: nope. this is wrong. for pointer-devices, we do want the content close together. Only on the phone where the button is on the bottom do we want the quick-actions to be at the bottom.

This on desktop, is wrong:
grafik

Please revert.

app/lib/features/spaces/widgets/public_spaces_selector.dart
```diff
@@ -206,71 +207,80 @@ class PublicSpaceItem extends ConsumerWidget {
...
               ),
             ),
           ),
-          Padding(
-            padding: const EdgeInsets.only(
-              left: 20,
-              top: 2,
-              bottom: 2,
+          Flexible(
+            child: Html(
```
gnunicorn, line 277: do we allow html in the topic? That sounds dangerous... especially to use here ... 5 lines ... that is a lot of text for a rather small widget...

yeah. no... I looked it up, we aren't allowing `space.topic()` to use HTML in the about-card either. Please revert this.

gnunicorn (approved):
Thanks, mate!

However, there are a few things that can't be merged as-is and other that need at least a FIXME around them...

app/lib/features/chat/pages/chats_shell.dart
gtalha07: The code is okay but not a pattern. It's very much go against the riverpod practices as it frequently changes the state based on ```constraints``` but ideally we should have some separate notifier which listens to it and do under the hood state change rather than doing inside ```build()```. This can be looked later but I'll add comment for improvement.

app/lib/features/chat/pages/room_page.dart
gtalha07: Precisely.

app/lib/features/chat/pages/room_profile_page.dart
gtalha07: I remember this wasn't already working though?? OR got removed somewhere in other PR's commit history. I'll add the functionality back in any case.

app/lib/features/pins/pages/pins_page.dart
gtalha07: We probably need to re-think about responsive architecture here. Maybe having some 3rd party package dealing with responsiveness or making a whole upper layer class wrapping all of it.

app/lib/features/search/widgets/quick_jump.dart
gtalha07: Actually this was misplaced here. I was looking this view on Android and it doesn't appear well .i.e. search field is overflowing the system status bars and obscuring the view. This should be refactored, but on the other hand, will remove ```Expanded```.

gnunicorn (approved):

PR #1012, opened 2023-09-27
directly invite if the full username is given

### LINKED ISSUE #1000: Invite doesn't detect username anymore
if we searched for a fully qualified username, a button should appear below to invite them directly. Doesn't happen anymore :( ...

grafik

gnunicorn:
fixes #1000 .

Matches fully qualified user-ids, but also if you forgot to add the leading `@` ...

and remove some old files we don't need anymore.


app/lib/common/dialogs/invite_to_room_dialog.dart
```diff
@@ -9,6 +9,18 @@ import 'package:atlas_icons/atlas_icons.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
+final userNameRegExp = RegExp(
```
gtalha07, line 12: Would be handy if we put this in ```utils``` if we have more use-cases than here ;)

gtalha07 (approved):
LGTM!

app/lib/common/dialogs/invite_to_room_dialog.dart
gnunicorn: fair... but do we have more use cases for this?!?

PR #1017, opened 2023-09-28
make sidesheet footer stick to bottom and scroll on overflow

### LINKED ISSUE #999: Submit of sidesheets hidden behind keyboard on mobile
You can access the submit buttons. I think we want to remodel the dialog to have a sticky bottom-bar with the buttons and if there isn't enough space for all the things, they need to be scrolled...

Screenshot_1695720199

gtalha07:
- Also include sidesheet improvements .i.e. visual and validation handling


gnunicorn:
Why did you move all these sidesheets from the (more specific) `dialog`-folder into the (more generic) `widgets`-folder? And as far as I can see you didn't track that move with `git mv` so it is pretty hard to review (as they appears as completely new file and it is unclear what actually changed) ...

gtalha07:
> Why did you move all these sidesheets from the (more specific) `dialog`-folder into the (more generic) `widgets`-folder? And as far as I can see you didn't track that move with `git mv` so it is pretty hard to review (as they appears as completely new file and it is unclear what actually changed) ...

Imo they don't really comprehend to be dialogs and having a navigation route, they are more like pages or widgets. CRUD sheet widgets didn't really got changed much aside from reducing the code lines and post CRUD flow correction. 

(My mistake to not properly address the change in a better way).

gnunicorn:
> they are more like pages or widgets.

I feel they are distinct enough from regular widgets to not lump them together (to make them easier to find). And `pages` sounds like full-page things to me, which these aren't. As these are overlays, I thought `dialog` is good term, but what about `sheets` then? That is a pretty distinct category of UI elements in our app.

/me had just removed a bunch of unused files in `widgets` that we didn't notice were unused now. @gnunicorn thinks that having a huge folder with plenty of files that aren't clear where they are used leads to more of that and wants to prevent from us keeping useless files around.

gtalha07:
> > they are more like pages or widgets.
> 
> I feel they are distinct enough from regular widgets to not lump them together (to make them easier to find). And `pages` sounds like full-page things to me, which these aren't. As these are overlays, I thought `dialog` is good term, but what about `sheets` then? That is a pretty distinct category of UI elements in our app.
> 
> /me had just removed a bunch of unused files in `widgets` that we didn't notice were unused now. @gnunicorn thinks that having a huge folder with plenty of files that aren't clear where they are used leads to more of that and wants to prevent from us keeping useless files around.

Yeah I'd like to keep them as ```overlays``` or ```sheets``` in ```common``` folder  instead. What do you think?

gnunicorn:
if so, I'd prefer `sheets`. Not sure they all belong into `common` though ... shouldn't most of them be in their respective feature-folder?

app/lib/features/events/widgets/create_event_sheet.dart
```diff
@@ -0,0 +1,358 @@
...
+          children: <Widget>[
+            Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: <Widget>[
+                const Padding(
+                  padding: EdgeInsets.only(bottom: 5),
+                  child: Text('Name'),
+                ),
+                InputTextField(
+                  hintText: 'Type Name',
```
gnunicorn, line 71:
```suggestion
                  hintText: 'Name of the event.',
```

app/lib/features/space/widgets/edit_space_sheet.dart
```diff
@@ -190,96 +165,51 @@ class _EditSpacePageConsumerState extends ConsumerState<EditSpacePage> {
...
+          ? null
+          : () async {
+              // check permissions before updating space
+              bool havePermission = await permissionCheck();
+              if (!havePermission) {
+                // ignore: use_build_context_synchronously
+                customMsgSnackbar(
+                  context,
+                  'Cannot edit space with no permissions',
+                );
```
gnunicorn, line 180: shouldn't we return here?
```suggestion
                );
                return;
```

app/lib/features/space/widgets/edit_space_sheet.dart
```diff
@@ -190,96 +165,51 @@ class _EditSpacePageConsumerState extends ConsumerState<EditSpacePage> {
...
+                final roomId = await _handleUpdateSpace(context);
+                debugPrint('Space Updated: $roomId');
+                // We are doing as expected, but the lints triggers.
+                // ignore: use_build_context_synchronously
+                if (!context.mounted) {
+                  return;
+                }
+                context.pop();
+                context.pop();
```
gnunicorn, line 202: we need a better handling of this popping... it is hard to follow...

```suggestion
                context.pop(); // pop the loading screen
                context.pop(); // pop the edit sheet
```

app/lib/features/spaces/widgets/create_space_sheet.dart
```diff
@@ -279,6 +218,7 @@ class _CreateSpacePageConsumerState extends ConsumerState<CreateSpacePage> {
         return;
       }
 
+      Navigator.of(context, rootNavigator: true).pop();
       Navigator.of(context, rootNavigator: true).pop();
```
gnunicorn, line 222:
```suggestion
      Navigator.of(context, rootNavigator: true).pop(); // pop the loading screen
      Navigator.of(context, rootNavigator: true).pop(); // pop the side sheet
```

app/macos/Podfile.lock
```diff
@@ -88,4 +88,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce7dbe26c78bfc7ba46736094c1e2d25982870fa
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.1
```
gnunicorn, line 91: we need this to stay with 1.11.3 for release reasons ...
```suggestion
COCOAPODS: 1.11.3
```

gnunicorn (approved):
just minor updates.

however, my confidence in reviews of the sidesheets has fallen over the last weeks as they grew ever more complex and are hard to fully comprehend now ...  we need more tools to ensure we don't accidentally break those.

PR #1019, opened 2023-10-02
Use GoRouter rathern than SubShell

### LINKED ISSUE #955: Refactor routing with StatefulShellRoute
We want to switch from ShellRoutes to StatefulShellRout in it for the space-view and the chat view to keep the sub-route state. We want: a chat-branch, an activites-branch, a profile-branch (with settings and stuff), the search-branch and home/dashboard which contains all other item and listing pages. Creation-Sheets, Popups, etc need to stick on the top level.

Source: https://codewithandrea.com/articles/flutter-bottom-navigation-bar-nested-routes-gorouter/

gnunicorn:
fixes #955 .

This fixes the long-standing context-embedding bug:

[routing-fixes-le-deux.webm](https://github.com/acterglobal/a3/assets/40496/57ec180c-5d4c-476a-8187-e9acf06bfe97)
[showing-the-routing-fixes.webm](https://github.com/acterglobal/a3/assets/40496/ea5ffded-967d-4e25-9b0f-3fe854398af8)

### DIFF
--- a/app/lib/router/router.dart
+++ b/app/lib/router/router.dart
@@ -674,14 +674,20 @@ List<RouteBase> makeRoutes(Ref ref) {
         ),
 
         /// Space subshell
-        ShellRoute(
-          navigatorKey: spaceNavKey,
-          pageBuilder: (context, state, child) {
+
+        GoRoute(
+          name: Routes.space.name,
+          path: Routes.space.route,
+          redirect: authGuardRedirect,
+          pageBuilder: (context, state) {
+            tabKeyNotifier.switchTo(const Key('overview'));
             return NoTransitionPage(
               key: state.pageKey,
               child: SpaceShell(
                 spaceIdOrAlias: state.pathParameters['spaceId']!,
-                child: child,
+                child: SpaceOverview(
+                  spaceIdOrAlias: state.pathParameters['spaceId']!,
+                ),
               ),
             );
           },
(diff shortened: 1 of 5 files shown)


app/lib/router/router.dart
```diff
@@ -777,22 +779,11 @@ List<RouteBase> makeRoutes(Ref ref) {
...
-            GoRoute(
-              name: Routes.space.name,
-              path: Routes.space.route,
-              redirect: authGuardRedirect,
-              pageBuilder: (context, state) {
-                tabKeyNotifier.switchTo(const Key('overview'));
-                return NoTransitionPage(
-                  key: state.pageKey,
-                  child: SpaceOverview(
+                  child: SpaceShell(
```
gtalha07, line 782: I am not really digesting it of switching to normal nested navigation view with a common wrapper over all the routes, its solving one thing but raises other many. We are effectively making it re-render the whole page instead of  nested one and because tab view is build on demand, there's no persistence of state between tabs with this change. ```ShellRoute``` keeps that state and I would like to keep as it is there but maybe refactoring it solely would be something I would vouch for.

gnunicorn: The problem with keeping the old style is that it was impossible to get the proper navigation flow to work. And it makes sense once you read up on how ShellRoutes are _meant_ to be used: they are understood as two separate routes (even on the same screen) - e.g. have a split screen and each its own router with independent routing. The way we use it isn't really how that is meant to be used ... it isn't meant for hierarchy nested routes ... that's what the `GoRoute(routes: [...])` is meant to do...

gtalha07: Alright, I think we can merge as it is and can work on in a separate PR.

gtalha07 (approved):

gnunicorn:
the bug is actually https://github.com/flutter/flutter/issues/125752 . Let's do a workaround instead

Issue #967, opened 2023-09-18
we should filter the list of events to shown to only upcoming events


gnunicorn:
and on home only the ones you agreed to or said yes or Maybe to ...

We should have an API for that on the rust side:
 - [ ] all_upcoming_events
 - [ ] my_upcoming_events
 - [ ] my_previous_events
 - [ ] an easy way to learn about the current rsvp-status per event

PR #1028, opened 2023-10-03
calendar event fixes

gtalha07:
- See tracking #800.
- [x] Implement #967.
- [x] Implement (show calendar, link to filter the events).
- [x] Implement (divide into upcoming and past events) for all events
- [x] show RSVP status per event.

Issue opened  after merge: #1033 
**Currently missing API: Upcoming and Past event for spaces.**


gtalha07:
Screenshot 2023-10-07 at 12 59 06 AM
Screenshot 2023-10-07 at 12 59 32 AM
Screenshot 2023-10-07 at 12 59 57 AM

gnunicorn (changes requested):
What is this new `features/events/application/events` folder about? Why move things into further subfolders? And what does "application", "data/repository" and "domain/repositories" and "presentation" do now? Why is there an "application" within a "feature" folder? that doesn't make any sense..

It feels like every other PR the folder structure/layout is changed without explanation...

Either apply a folder structure across the entire application (so that you get the "pattern by familiarity"-effect) in a PR and add some useful explanation of the pattern (so others also understand how to use it and why it is done the way it is) in the docs so that we can discuss the benefits and drawbacks of that proposal on its own, or stick with the existing pattern, please (even if you feel that it is messy). But mixing a new (not-agreed) pattern in with a new feature PR isn't the way to go.

gtalha07:
> What is this new `features/events/application/events` folder about? Why move things into further subfolders? And what does "application", "data/repository" and "domain/repositories" and "presentation" do now? Why is there an "application" within a "feature" folder? that doesn't make any sense..
> 
> It feels like every other PR the folder structure/layout is changed without explanation...
> 
> Either apply a folder structure across the entire application (so that you get the "pattern by familiarity"-effect) in a PR and add some useful explanation of the pattern (so others also understand how to use it and why it is done the way it is) in the docs so that we can discuss the benefits and drawbacks of that proposal on its own, or stick with the existing pattern, please (even if you feel that it is messy). But mixing a new (not-agreed) pattern in with a new feature PR isn't the way to go.

This folder structuring/layout represents Domain driven design where there are each independent layers relying on ```domain``` layer which contains abstract interface of business logic implementation. To summarise:
- API/Repository layer: This essentially contains API calls and any error handling using ```Failure/Data``` unions.
- Domain Layer: Abstract implementation of Repository  layer and every other layer is dependent on it.
- Application: Feature use cases which are essentially communication between UI and business logic layer. This contains UI methods. 
- Presentation: Pure UI widgets which are independent but relying on domain. 

This design promotes easy understanding and testing of components and thus being independent, allow us to modify or refactor code with much flexibility. A dev can understand where the changes needed to be made and even doing large refactors, it doesn't need modifications over large scale of files, since each layer is independent (except domain). 

Since the addition of riverpod, the goal was to adapt a scalable DDD architecture with Riverpod+Freezed, which would allow us to maintain source code effectively as the codebase gets increasing. And this is more like the first steps to that approach.

Of course the familiarity with overall current structuring isn't there yet (as even sorting these to DDD principles) took up a huge amount of time, but I can introduce that on short PR basis or maybe keeping this opened until full revamp and review. 

Currently right now, any bug fixation or new addition requires large amount of code changes and represents filenames which doesn't conform to the actual change terminology. With these principles, I think we can reduce the amount of complexity in understanding components logic and methods and improve the reviews.

gtalha07:
More reference about DDD principles and understanding:
https://resocoder.com/2020/03/09/flutter-firebase-ddd-course-1-domain-driven-design-principles/

gnunicorn:
> This design promotes easy understanding and testing of components and thus being independent, allow us to modify or refactor code with much flexibility. A dev can understand where the changes needed to be made and even doing large refactors, it doesn't need modifications over large scale of files, since each layer is independent .

Almost every structure pattern is meant to do that, yet I have still to experience one that actually does that "intuitively" (as they also often claim) and actually being able to sustain that in the long run (rather than being in the way ... like most Java abstractions are).

I am not very fond of cramming things into large patterns with many folder and thousands of files just because some guide by someone (from not even the same domain) said so. I already dislike the somewhat existing weird distinction between "models" and "providers" (and "notifiers") we have in that parts of the app, where very often, it feels like just having the these things that clearly belong together in one file is just easier to see, comprehend and import. It all feels very enterprisy and less practical oriented - it is always more work to navigate many files and directories to get the entire field of view for what is going. And this seems to just exacerbate this case.

But my point was to explicitly _not discuss this here_ but instead:
> add some useful explanation of the pattern (so others also understand how to use it and why it is done the way it is) in the docs so that we can discuss the benefits and drawbacks of that proposal on its own

Let's do that.

But before we do that, I'd like to raise the question: What actual problem is that meant to solve? And do we have a ticket about that with a higher urgency than the other tickets we have open? I don't think so. In that case, I'd argue, we shouldn't actually be spending time much time on this at all...

gtalha07:
> > This design promotes easy understanding and testing of components and thus being independent, allow us to modify or refactor code with much flexibility. A dev can understand where the changes needed to be made and even doing large refactors, it doesn't need modifications over large scale of files, since each layer is independent .
> 
> Almost every structure pattern is meant to do that, yet I still have to come across one that actually does that "intuitively" (as they also often also claim) and actually being able to sustain that in the long run (rather than being in the way ... like most Java abstractions and most of these patterns than often are).
> 
> I am not very fond of cramming things into large patterns with many folder and thousands of files just because some guide by someone (from not even the same domain) said so. Like we already have a quaint infrastructure by having the entire Rust, then the FFI and its objects and then the providers - I don't think I have seen anyone else trying to use these and thus have an informed understanding of navigating that in their proposed pattern.
> 
> I already dislike the somewhat existing weird distinction between "models" and "providers" (and "notifiers") we have in that parts of the app, where very often, it feels like just having the these things that clearly belong together in one file is just easier to see, comprehend and import. It all feels very enterpris-y and less practical oriented - it is always more work to navigate many files and directories to get the entire field of view for what is going. And this seems to just exacerbate this case.
> 
> But my point was to explicitly _not discuss this here_ but instead:
> 
> > add some useful explanation of the pattern (so others also understand how to use it and why it is done the way it is) in the docs so that we can discuss the benefits and drawbacks of that proposal on its own
> 
> Let's do that.
> 
> But before we do that, I'd like to raise the question: What actual problem is that meant to solve? And do we have a ticket about that with a higher urgency than the other tickets we have open? I don't think so. In that case, I'd argue, we shouldn't actually be spending time much time on this at all...

I see. I guess we aren't much focused around this aspect for now neither we have concrete examples that would validate this flow on current project architecture, so I'll discard the structure changes and re-write according to what current we're familiar with.

PR #1050, opened 2023-10-16
restore right swipe on chat messages.

### LINKED ISSUE #1044: Swipe on message to reply stopped working
with the latest release, its no longer possible to swipe on a message to reply.
Only on the messages that i as a user has sent.

gtalha07:
- Will fix #1044.


gnunicorn (approved):
This is just a revert of the part that was unintentionally removed, right!?! Go for it.

PR #1070, opened 2023-10-23
Fix textual updates, new integration test suite for flutter

### LINKED ISSUE #1053: Text updates only creates black/emptty update
Image

gnunicorn:
Fixes #1053 .

Also includes a revamp of the integration test suite from gherkin (fully removed) and replaced with `convenient_test`. Next to login and registration smoketest, there is an additional tests creating a space and posting a textual news item - proofing that it works now (and will continue to do so). In order to make that happen, more `Keys` are introduced throughout the UI, and various smaller fixes are made to allow for items to be clickable with the UI runners. Additionally, the PR contains documentation on how to use the test runner yourself.

Open Tasks:
 - [ ] github workflows run on requests and on `main` (already in but defunct). 
 - [x] can we create api docs for the tester?
 - [x] the rescaling issue in the onboarding screen is fixed in a wrong way that now makes the font super small even if the space is given.


app/lib/common/providers/common_providers.dart
```diff
@@ -32,6 +32,15 @@ Future<ProfileData> getProfileData(Account account) async {
   return ProfileData(displayName.text(), avatar.data());
 }
 
+final myUserIdStrProvider = StateProvider((ref) {
+  final client = ref.watch(clientProvider);
```
gtalha07, line 36: Why aren't we using ```alwaysClientProvider``` here ??

app/lib/features/events/sheets/create_event_sheet.dart
```diff
@@ -209,9 +209,9 @@ class _CreateEventSheetConsumerState extends ConsumerState<CreateEventSheet> {
       draft.utcEndFromRfc3339(utcEndDateTime);
       draft.descriptionText(description);
       final eventId = await draft.send();
-      final client = ref.read(clientProvider);
+      final client = ref.read(alwaysClientProvider);
```
gtalha07, line 212: Good to have improved it, but why not using it everywhere else?

gnunicorn: I only tried to improve the problem that the integration tests failed with `null!` issues and replaced it where I found that pattern. We can extend that to other places, but it is only a softer version of it, so places where this throwing isn't acceptable, isn't good...

app/lib/common/providers/common_providers.dart
gnunicorn: came up with it after, but we could ;)

app/lib/features/onboarding/pages/login_page.dart
```diff
@@ -54,7 +54,9 @@ class _LoginPageState extends ConsumerState<LoginPage> {
           // no message means, login was successful.
           context.goNamed(Routes.main.name);
         } else {
-          customMsgSnackbar(context, loginSuccess);
+          EasyLoading.showError(
```
gtalha07, line 57: could use ```showSuccess``` here.

gtalha07 (approved):
Looks sleek and good to merge!! Except few commented parts which can be addressed on your behalf ;)

app/lib/features/onboarding/pages/login_page.dart
gnunicorn: no... if we have a message it is an "error message" ... this is a bit confusing in the API but that's how it is ...

PR #1073, opened 2023-10-24
Revert "Merge pull request #978" - media changes.

gnunicorn:
This reverts commit 9257891a27307ee58309b7f235de07ddcadd8537, reversing changes made to a9aed6a3fdacc3a486ce6bf9709f59dea0efe3ae.

It appears those changes broke media upload support for both chat and news updates. Thus we are reverting these changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,6 @@ dependencies = [
  "fern",
  "ffi-gen",
  "ffi-gen-macro",
- "filesize",
  "futures",
  "futures-signals",
  "lazy_static",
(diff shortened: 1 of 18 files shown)


bitfriend:
`content_type` is determined by `mime_guess`.
`mime_guess` detects content type using file extension.
If anyone wants to use temporary file supported by system, it will have no file extension and `mime_guess` would fail to parse content type.

gnunicorn:
> `content_type` is determined by `mime_guess`.
> `mime_guess` detects content type using file extension.

Then picking `mime_guess` was a bad choice. Mime guessing based on the file extension is generally a bad choice.

> If anyone wants to use temporary file supported by system, it will have no file extension and `mime_guess` would fail to parse content type.

But we aren't only seeing this fail on the test cases. We can _confirm_ that uploading images via chat isn't working on `main`, while it works fine with this revert. Even if that `mime_guess` was the intended choice, the feature in the end-product doesn't work anymore. So, it needs to be fixed or reverted.

bitfriend:
Then I will do 2 things:
1. use `filesize` crate` to remove `size` param from media upload
2. convert `u32` to `u64`

@gnunicorn What do you think?

PR #1083, opened 2023-10-26
Minor fixes 

### LINKED ISSUE #941: Logging in first time, the update page is shown by default, but home button is highlighted.
Image

### LINKED ISSUE #1024: Links in chat can't be clicked
Found on Android in the nightly channel: clicking links doesn't do anything.
Screenshot_20231002-203336

### LINKED ISSUE #1025: Route breaks on back from chat profile in android
See:

https://github.com/acterglobal/a3/assets/40496/863d3500-804a-4000-a235-119afb419cec

### LINKED ISSUE #1056: Support swipe down to hide keyboard
### Feature Request
On iOS, at least, the ability to hide the keyboard by swiping downwards is common and nice to have.

### Steps to reproduce
1. Open Acter
2. Navigate to the **Chat** section
3. Open a chat, for example **Acter Foyer**
4. Click on the text input to open the keyboard
5. Swipe downwards on the screen

#### Expected
It hides the keyboard and goes out of text input mode (i.e. the cursor in the text input field disappears).

#### Actual
The keyboard stays open and you are still in text input mode. The chat history is scrolled.

### Versions
- Acter App: 1.23.1012/iOS
- Device: iPhone 11 Pro

### LINKED ISSUE #1079: Disable sending empty messages
Currently a user can send empty messages, this should be disabled.

Image

kumarpalsinh25:
This PR fixes general issues as below.

- Resolve route breaks issue on back from chat profile in android, fixes #1025 
- Resolve link click issue on chat message, fixes #1024 
- Chat Screen: Support swipe down to hide keyboard for iOS, fixes #1056 
- Disable sending empty messages, fixes #1079

### DIFF
--- a/app/android/app/src/main/AndroidManifest.xml
+++ b/app/android/app/src/main/AndroidManifest.xml
@@ -40,4 +40,21 @@
             android:name="flutterEmbedding"
             android:value="2" />
     </application>
+    <!-- Provide required visibility configuration for API level 30 and above -->
+    <queries>
+        <!-- If your app checks for SMS support -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="sms" />
+        </intent>
+        <!-- If your app checks for call support -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="tel" />
+        </intent>
+        <!-- If your application checks for inAppBrowserView launch mode support -->
+        <intent>
+            <action android:name="android.support.customtabs.action.CustomTabsService" />
+        </intent>
+    </queries>
 </manifest>
--- a/app/android/app/src/main/AndroidManifest.xml
+++ b/app/android/app/src/main/AndroidManifest.xml
@@ -40,21 +40,4 @@
             android:name="flutterEmbedding"
             android:value="2" />
     </application>
-    <!-- Provide required visibility configuration for API level 30 and above -->
-    <queries>
-        <!-- If your app checks for SMS support -->
-        <intent>
-            <action android:name="android.intent.action.VIEW" />
-            <data android:scheme="sms" />
-        </intent>
-        <!-- If your app checks for call support -->
-        <intent>
-            <action android:name="android.intent.action.VIEW" />
-            <data android:scheme="tel" />
-        </intent>
-        <!-- If your application checks for inAppBrowserView launch mode support -->
-        <intent>
-            <action android:name="android.support.customtabs.action.CustomTabsService" />
-        </intent>
-    </queries>
 </manifest>
--- a/app/lib/features/chat/widgets/custom_input.dart
+++ b/app/lib/features/chat/widgets/custom_input.dart
@@ -517,6 +517,7 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
   }
 
   Future<void> onSendButtonPressed() async {
+    if(mentionKey.currentState!.controller!.text.isEmpty) return;
     final roomId = widget.convo.getRoomIdStr();
     final inputNotifier = ref.read(chatInputProvider(roomId).notifier);
     final mentionReplacements =
--- a/app/lib/features/chat/widgets/text_message_builder.dart
+++ b/app/lib/features/chat/widgets/text_message_builder.dart
@@ -10,7 +10,6 @@ import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 import 'package:flutter_link_previewer/flutter_link_previewer.dart';
 import 'package:flutter_matrix_html/flutter_html.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:url_launcher/url_launcher.dart';
 
 class TextMessageBuilder extends ConsumerStatefulWidget {
   final Convo convo;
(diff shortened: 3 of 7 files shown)


app/android/app/src/main/AndroidManifest.xml
```diff
@@ -40,4 +40,21 @@
             android:name="flutterEmbedding"
             android:value="2" />
     </application>
+    <!-- Provide required visibility configuration for API level 30 and above -->
```
gnunicorn, line 43: Can you elaborate what these are for? I'd rather we keep the number of permissions we ask from your users to an absolute minimum...

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -143,6 +144,11 @@ class _TextWidget extends ConsumerWidget {
               maxLines: isReply ? 3 : null,
             )
           : Html(
+              onLinkTap: (url) async {
```
gnunicorn, line 147: We have a `openLink` in `app/lib/common/utils/utils.dart` that tries to figure out whether something is an app internal link (e.g. to a Task-item) or something we try to open with the external launcher. Please use that helper here.

app/lib/features/home/providers/navigation.dart
```diff
@@ -12,7 +12,7 @@ import 'package:acter/common/themes/app_theme.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
 const fallbackSidebarIdx = 1;
-const fallbackBottomBarIdx = 0;
+const fallbackBottomBarIdx = 1;
```
gnunicorn, line 15: I am not sure this is correct. The fallback is used _every time_ we are not matching correctly - so all `/tasks` `/pins`-pages as well. I think if you used the quick-jump and selected an item (search -> one of the icons) you'll find the wrong icon highlighted on the bottom now: not the home but the news is highlighted ....

That's not the behavior we want.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -517,6 +517,7 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
   }
 
   Future<void> onSendButtonPressed() async {
+    if(mentionKey.currentState!.controller!.text.isEmpty) return;
```
gnunicorn, line 520: arguably, if there is nothing to send, we do not only want to avoid that we want to disable the entire button, don't we?

gnunicorn (changes requested):
A few improvements, but at least one attempt also introduces another bug (I believe) and thus needs to be dealt with before merging.

app/android/app/src/main/AndroidManifest.xml
kumarpalsinh25: Above android 11 OS, this are the queries tabs we need to on the android manifest file to support default link actions.
Ref: https://pub.dev/packages/url_launcher

app/lib/features/chat/widgets/text_message_builder.dart
kumarpalsinh25: Sure, let me do changes accordingly

app/lib/features/home/providers/navigation.dart
kumarpalsinh25: There is conditional routes for Mobile and Desktop.
 
Screenshot 2023-10-26 at 5 03 34 PM

On Desktop: Home Tab is considered as default tab and route
On Mobile: Update Tab is considered as default tab but when we do login at that time index was returning as -1 and that's why home bottom was selected even though Update page is opened

Please share your thoughts here.

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Actually, "onSendButtonPressed" is called from many places to send the message so that's why I have putted empty message condition here to avoid any black message from anywhere.

app/android/app/src/main/AndroidManifest.xml
gnunicorn: I see... but then we should add the [same for `iOS`](https://pub.dev/packages/url_launcher#ios), too, no?

gnunicorn: On another note: we aren't actually checking whether the host can open it. In the utils we are checking whether we consider it an internal link and if not we send it forward ... We might want to reconsider that approach then...

app/lib/features/home/providers/navigation.dart
gnunicorn: Yeah, but the code is more complex. As described in the comment: the fallback is used for more than just this one case. If you follow the path described above: does it show the wrong icon? Then the patch isn't sufficient.

app/android/app/src/main/AndroidManifest.xml
kumarpalsinh25: I have made the required changes. Thanks

app/lib/features/home/providers/navigation.dart
kumarpalsinh25: I have removed this point from this PR. I will handle it on subsequent PR.

gtalha07 (approved):
LGTM other than suggested changes...

gnunicorn (approved):

PR #1088, opened 2023-10-26
space header bug fixes

### LINKED ISSUE #1031: Space header is missing when clicking on space avatar in update.
Space header is missing when clicking on space avatar in update.
Also wierd space widget is shown when clicking on chat avatar.

https://github.com/acterglobal/a3/assets/77483554/4e0b5b54-1194-4c19-b682-59075fec5788

### LINKED ISSUE #1047: When clicking on a space’s parent avatar, the space changes, but the header does not change.

gtalha07:
Includes:
- fix #1031. 
- fix #1047.

**Issue**: ```context.push()``` doesn't work with ```ShellRoute``` or it appears buggy. Hence its been changed to ```context.go()```.

### DIFF
--- a/app/lib/common/widgets/spaces/space_parent_badge.dart
+++ b/app/lib/common/widgets/spaces/space_parent_badge.dart
@@ -1,4 +1,5 @@
 import 'package:acter/common/providers/room_providers.dart';
+import 'package:acter/common/utils/routes.dart';
 import 'package:acter_avatar/acter_avatar.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -51,9 +52,10 @@ class SpaceParentBadge extends ConsumerWidget {
                     message: parent.profile.displayName,
                     child: InkWell(
                       onTap: onTap ??
-                          () {
-                            context.push('/$parentId');
-                          },
+                          () => context.goNamed(
+                                Routes.space.name,
+                                pathParameters: {'spaceId': parentId},
+                              ),
                       child: ActerAvatar(
                         mode: DisplayMode.Space,
                         displayName: parent.profile.displayName,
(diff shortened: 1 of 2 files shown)


app/lib/common/widgets/spaces/space_parent_badge.dart
```diff
@@ -51,9 +52,10 @@ class SpaceParentBadge extends ConsumerWidget {
                     message: parent.profile.displayName,
                     child: InkWell(
                       onTap: onTap ??
-                          () {
-                            context.push('/$parentId');
-                          },
+                          () => context.goNamed(
+                                Routes.space.name,
+                                pathParameters: {'spaceId': parentId},
+                              ),
```
gnunicorn, line 58: but that won't work for the other cases where this widget is used. Namely, if you clicked this from a chat now, it wouldn't push the route and you wouldn't have any `back` on top ...

I suggest we create another param `this.push = true` that the surrounding part can set to `false` (like in the space header) that switches from `pushNamed` to `goNamed` behaviour.

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -83,7 +83,7 @@ class NewsSideBar extends ConsumerWidget {
         const SizedBox(height: 10),
         InkWell(
           onTap: () {
-            context.pushNamed(
+            context.goNamed(
```
gnunicorn, line 86: We need to fix the router overall, this should really be a push as the user probably wants to navigate back easily, but we can't offer that here. This fix is fine, this is just a remark how bad our state of router is (:( .

gnunicorn (changes requested):
Fixes it, yes. But will break another behaviour. We need a more comprehensive fix. Suggested one in the comment.

app/lib/common/widgets/spaces/space_parent_badge.dart
gtalha07: How we would determine if this is a push ?? Even if we do, it would still make shell disappear as ```push``` is executed in either case. The problem lies how the ShellRoute works with pushed route unfortunately because of the [issue](https://github.com/flutter/flutter/issues/120665)...

gnunicorn:
superseeded by #1141

PR #1103, opened 2023-10-30
chat creation fixes

### LINKED ISSUE #1091: app chrashes when trying to type user in DM creation on IOS
Ive managed to create dm group chat before. Howeve now when i try it keeps chrashing .

https://github.com/acterglobal/a3/assets/77483554/61346a0e-dc06-4d06-9fac-8fdd5f7db33f

### LINKED ISSUE #1094: DM avatar needs restart to show properly
When DM is created, it doesnt show the right icon to begin with. First when you restart the app it shows properly

https://github.com/acterglobal/a3/assets/77483554/739ec0d1-926a-4818-a06e-25cb755610e7

### LINKED ISSUE #1095: DM creation: not able to scroll down to select user user (IOS)
When searching for users, its not possible to scroll on the list.

https://github.com/acterglobal/a3/assets/77483554/2166ac30-874f-4a70-8f6c-48baaee19794

### LINKED ISSUE #1099: Wording of chat creation
When creating the chat, its currently not clear that "Start Chat" is the "group/space chat" or that you create a DM (single DM og group DM) by starting to type in the search field.
Just calling the "group chat" = "Start Chat" is confusing, as both "DM's" and "group chats" are "Chats"

Image

gtalha07:
- [x] fix #1099.
- [x] fix #1094.
- [x] fix #1095.

### DIFF
--- a/.changes/1103-chat-creation-fixes.md
+++ b/.changes/1103-chat-creation-fixes.md
@@ -1,5 +1,4 @@
 - Chats (Create)
-  - Improve wordings
   - fix: bug where DM/Group DM avatars do not load correctly after creation in chat room.
   - fix: Hide keyboard when scrolling over the found users list after search.
   - fix: Selecting users in found users list will now automatically make page scroll over to top.
--- a/.changes/1103-chat-creation-fixes.md
+++ b/.changes/1103-chat-creation-fixes.md
@@ -1,0 +1,6 @@
+- Chats (Create)
+  - Improve wordings
+  - fix: bug where DM/Group DM avatars do not load correctly after creation in chat room.
+  - fix: Hide keyboard when scrolling over the found users list after search.
+  - fix: Selecting users in found users list will now automatically make page scroll over to top.
+  - `Not Found Users` state will show over the page if there's no user existing comparative to search term.
--- a/app/lib/features/chat/widgets/room_avatar.dart
+++ b/app/lib/features/chat/widgets/room_avatar.dart
@@ -56,7 +56,7 @@ class RoomAvatar extends ConsumerWidget {
           );
         }
 
-        //Type == DM : Handle avatar according to the members counts
+        //Type == DM and no avatar: Handle avatar according to the members counts
         else {
           return dmAvatar(ref);
         }
--- a/app/lib/features/chat/widgets/room_avatar.dart
+++ b/app/lib/features/chat/widgets/room_avatar.dart
@@ -46,6 +46,14 @@ class RoomAvatar extends ConsumerWidget {
             avatar: profile.getAvatarImage(),
             size: 18,
           );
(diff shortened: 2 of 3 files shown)


.changes/1103-chat-creation-fixes.md
```diff
@@ -0,0 +1,6 @@
+- Chats (Create)
+  - Improve wordings
```
gnunicorn, line 2: too generic, not relevant to actual end-users.
```suggestion
```

app/lib/features/chat/widgets/room_avatar.dart
```diff
@@ -38,17 +38,17 @@ class RoomAvatar extends ConsumerWidget {
...
+        if (!convo.isDm()) {
           return ActerAvatar(
             uniqueId: roomId,
-            mode: profile.isDm ? DisplayMode.User : DisplayMode.Space,
+            mode: DisplayMode.Space,
             displayName: profile.displayName ?? roomId,
             avatar: profile.getAvatarImage(),
             size: 18,
           );
         }
```
gnunicorn, line 49: but a DM that has an Avatar _also_ should be showing that avatar ... that's why the previous check was right and the current one wouldn't be, no?

```suggestion
        } else if (profile.hasAvatar()) {
                  return ActerAvatar(
            uniqueId: roomId,
            mode: DisplayMode.User,
            displayName: profile.displayName ?? roomId,
            avatar: profile.getAvatarImage(),
            size: 18,
          );
          }
        
```

gnunicorn (approved):
wording thing but also some important fix around the avatar image...

app/lib/features/chat/widgets/room_avatar.dart
gnunicorn: > We are handling that case here.  

[where](https://github.com/acterglobal/a3/pull/1103/files#diff-4015547fb05cb94d3fa13a84efb72ab7c8fc980c44c4e6f8ff2b65cc5d34c249R70-R97)? `dmAvatar` does not check whether the chat has an Avatar set, it uses the _members_ Avatars regardless ...

>  Also profile.hasAvatar() don't seem useful here as ActerAvatar would return fallback if avatar is not present.

I don't get that comment. `profile` is the `ConvoProfile` here. If that chat has an avatar, regardless of whether it is a DM or not, we want to show _that_ avatar. That is all this is about. Not the fallback or anything else.

The fallback is different depending on `isDm == true` or not, but if we have an avatar for the room, we _always_ want to show the configured avatar. isDm only decides whether we use a circle or the square ...

gtalha07: Okay seems reasonable. I will address it.

gnunicorn (approved):

PR #1111, opened 2023-10-31
Enable push by default

### LINKED ISSUE #996: User is bugged about push notifications on each start
on Android ... even if they already said no previously. We said we wanted to not do that, if they said no and store the response locally, but it appears that we aren't doing that correctly...

### LINKED ISSUE #1090: Notifications lead to update feed (IOS)
Okey, i think i have figured out the pattern. 
Its when the app is closed, that the notifications directs to the update feed.

https://github.com/acterglobal/a3/assets/77483554/14ccc2d6-d098-4a17-bbe6-1aa2d5eb6203

gnunicorn:
- [x] do not bother the User by asking for push rights as start of the app/register/login, fixes #996 
- [x] When user leaves the activities-screen, we check for push permissions
- [x] when the user reactivates push settings, we directly request permissions - if that fails, we send them to the settings to activate the push feature from there.
- [x] open the right chat when clicking a background notification that opens the cold-app on iOS, fixes #1090 .
- [x] put push notifications into default-on for iOS & Android
- [x] Document changes

Implementors note: especially the latter feature requires quite some hoops, that why I've outsourced the widget into a separate one that is being used for both occasions.


app/lib/common/notifications/notifications.dart
```diff
@@ -380,44 +424,52 @@ Future<bool> setupPushNotifications(
...
-  return await onNewToken(client, token);
+  return await onToken(client, token);
+}
+
+Future<bool> onNewToken(String token) async {
+  debugPrint(
+    'Received the update information for the token. Updating all clients.',
+  );
+  // ignore: use_build_context_synchronously
+  final sdk = rootNavKey.currentContext!.read(sdkProvider).requireValue;
```
gtalha07, line 459: This I presume is valid in all cases, especially since this is triggering on startup ? Or if not I think we would need ```valueOrNulll``` case and handle error call in next line.

app/lib/features/home/providers/notifiers/client_notifier.dart
```diff
@@ -18,9 +17,6 @@ class ClientNotifier extends StateNotifier<Client?> {
       asyncSdk.writeLog(stackTrace.toString(), 'error');
       return true; // make this error handled
     };
-    final client = state = asyncSdk.currentClient;
-    if (client != null) {
-      await setupPushNotifications(client);
-    }
+    state = asyncSdk.currentClient;
```
gtalha07, line 20: So the sdk is guaranteed returning client now?

gtalha07 (approved):
LGTM!

app/lib/common/notifications/notifications.dart
gnunicorn: I was thinking about that, too as both currentContext and the provider could type-wise be `null`, but I don't see any case that can actually be true in practice. And even if it was this is in a callback for a stream, so we'd see a traceback in the logs but nothing else...

PR #1112, opened 2023-11-01
Switch msg sending from room to timeline

bitfriend:
Will fix https://github.com/acterglobal/a3/issues/1102


bitfriend:
@gnunicorn Some packages in `pubspec.lock` were downgraded.
Please check if they are critical to `flutter_match_fork`.

gtalha07 (approved):
LGTM!

native/acter/src/api/room.rs
```diff
@@ -961,196 +948,149 @@ impl Room {
             .await?
     }
 
-    pub async fn read_receipt(&self, event_id: String) -> Result<bool> {
+    pub async fn image_binary(&self, event_id: String) -> Result<FfiBuffer<u8>> {
```
gnunicorn, line 951: these three functions basically do the same just match for different message-types, no? Can't we just have a single `message_binary` that simply fetches internally?

native/acter/src/api/stream.rs
```diff
@@ -80,19 +94,195 @@ impl TimelineStream {
         matches!(self.room.state(), RoomState::Joined)
     }
 
-    pub async fn edit(&self, new_msg: String, original_event_id: String) -> Result<bool> {
+    // for integration test, not api
```
gnunicorn, line 97: maybe we should add a feature-flag for these non-regular-apis that we only enable for the integration tests?!?
```suggestion
    // for integration test, not api
    #[cfg(feature="testing")]
    #[doc(hidden)]
```

native/acter/src/api/stream.rs
```diff
@@ -80,19 +94,195 @@ impl TimelineStream {
...
+                let item = timeline
+                    .item_by_event_id(&event_id)
+                    .await
+                    .context("The event not found")?;
+                Ok(item)
+            })
+            .await?
+    }
+
+    pub async fn send_plain_message(&self, message: String) -> Result<bool> {
```
gnunicorn, line 112: I wonder if we want to rethink this entire API. In particular I wonder if might want to have a `MessageBuilder` that the user can just use directly to construct their message and then have a singular API can consume via a `timeline.send(message.into()).await`. Would make it easier to add features than having to add a new function every time ...

gnunicorn (commented):
just some parts to think about in the future.

native/acter/src/api/room.rs
bitfriend: I have no ideas about how can I abstract these functions into single.
image

gnunicorn: well, the input and output are the same types, so we can just collapse everything around and make an internal `match` rather than a `let`. something along the lines of:

```rust
  match &m.content.msgtype {
     MessageType::Image(content) => {
       // do what we do with the image case
       //..
       return Ok(FfiBuffer::new(data));
     },
     MessageType::Audio(content => {
       // do what we do with the image case
       //..
       return Ok(FfiBuffer::new(data));
     }
     // ... and other cases
     // ..
     _ => { bail!("Invalid file format"); }
     }
  }
```

native/acter/src/api/stream.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1175/commits/33a25e307e0ab1cf1c9a03b262a3d61615effe15

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1175/commits/02b4aea043cc6cfe010f2f0789059630f5fdd220

native/acter/src/api/stream.rs
gnunicorn: @bitfriend when I click the commit it tells me this is outside any branch or PR... when/how are you intending on proposing these to `main`?

PR #1126, opened 2023-11-06
Implement edited flag in room message

bitfriend:
- We need the flag that shows whether a message is the edited version of other.
- Edited message should show the latest body, not original body.
- Timeline-based integration test should be implemented.

### DIFF
--- a/native/acter/src/api/room.rs
+++ b/native/acter/src/api/room.rs
@@ -1065,7 +1065,7 @@ impl Room {
                     event_id.to_owned(),
                     MessageType::text_plain(new_msg.to_string()).into(),
                 );
-                let mut edited_content = RoomMessageEventContent::text_markdown(new_msg);
+                let mut edited_content = RoomMessageEventContent::text_plain(new_msg);
                 edited_content.relates_to = Some(Relation::Replacement(replacement));
 
                 let txn_id = TransactionId::new();
--- a/native/test/src/tests/msg_edit.rs
+++ b/native/test/src/tests/msg_edit.rs
@@ -1,4 +1,5 @@
-use anyhow::{bail, Result};
+use acter::RoomMessage;
+use anyhow::Result;
 use core::time::Duration;
 use futures::{pin_mut, stream::StreamExt, FutureExt};
 use tokio::time::sleep;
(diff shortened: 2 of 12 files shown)


native/acter/src/api/room.rs
```diff
@@ -1065,7 +1065,7 @@ impl Room {
                     event_id.to_owned(),
                     MessageType::text_plain(new_msg.to_string()).into(),
                 );
-                let mut edited_content = RoomMessageEventContent::text_markdown(new_msg);
+                let mut edited_content = RoomMessageEventContent::text_plain(new_msg);
```
gnunicorn, line 1068: why though?

bitfriend: Because here is `edit_plain_message()`.

native/test/src/tests/msg_edit.rs
```diff
@@ -1,41 +1,85 @@
...
+    pin_mut!(stream);
+
+    let event_id = convo.send_plain_message("Hi, everyone".to_string()).await?;
     println!("event id: {event_id:?}");
 
-    let edited_id = space
+    let edited_id = convo
         .edit_plain_message(event_id.to_string(), "This is message edition".to_string())
         .await?;
+    println!("edited id: {edited_id:?}");
```
gnunicorn, line 26:
```suggestion
    info!("edited id: {edited_id:?}");
```

native/test/src/tests/msg_edit.rs
```diff
@@ -1,41 +1,85 @@
...
-
-    let Ok(AnyTimelineEvent::MessageLike(AnyMessageLikeEvent::RoomMessage(MessageLikeEvent::Original(r)))) = ev.event.deserialize() else {
-        bail!("not the proper room event")
-    };
-
-    let msg = RoomMessage::room_message_from_event(r, space.deref().deref().clone(), false);
-    let Some(item) = msg.event_item() else {
-        bail!("This item should be event item not virtual item")
-    };
+    while let Some(diff) = stream.next().await {
```
gnunicorn, line 28: Assuming the check doesn't apply (and thus break) this tests will just idle forever asking for a next one that never comes, won't it?

How about changing this into a retry-loop with a definite end-time and failure after a few attempts instead?

native/test/src/tests/msg_edit.rs
```diff
@@ -1,41 +1,85 @@
...
+        match diff.action().as_str() {
+            "Append" => {}
+            "Insert" => {}
+            "Set" => {
+                info!("diff index: {:?}", diff.index());
+                info!("diff value: {:?}", diff.value());
+                info!("diff values: {:?}", diff.values());
+                let Some(value) = diff.value() else {
```
gnunicorn, line 36: within tests you can just do `.expect()` ... (which is the same as `else { bail!(...` effectively and shorter to read and write) ...

native/test/src/utils.rs
```diff
@@ -1,13 +1,36 @@
...
-    CreateSpaceSettingsBuilder, SyncState,
+    CreateConvoSettingsBuilder, CreateSpaceSettingsBuilder, SyncState,
 };
 use acter_core::templates::Engine;
 use anyhow::Result;
 use futures::{pin_mut, stream::StreamExt};
 use tracing::trace;
 use uuid::Uuid;
 
+pub async fn random_user_with_random_convo(prefix: &str) -> Result<(Client, OwnedRoomId)> {
```
gnunicorn, line 11: 👍

native/test/src/tests/msg_edit.rs
```diff
@@ -1,41 +1,85 @@
...
+                assert_eq!(text_desc.body(), "This is message edition");
+                break;
+            }
+            "Remove" => {}
+            "PushBack" => {}
+            "PushFront" => {}
+            "PopBack" => {}
+            "PopFront" => {}
+            "Clear" => {}
+            "Reset" => {
```
gnunicorn, line 57: I am not sure I understand ... we expect either Reset or set? Don't we surely know which one?

gnunicorn (approved):
yay, tests!

Just some minor remarks about the details.

native/test/src/tests/msg_edit.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1126/commits/455a87e9276216c462259b1e1137f58970e98207

native/test/src/tests/msg_edit.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1126/commits/ee1a9f2225cad25fe1bdc8c465be3121067f40c1

native/test/src/tests/msg_edit.rs
bitfriend: We get `Hi, everyone` msg under `PushBack` action in usual case.
We may get that msg under `Reset` action, when calling of `now_or_never()` loop was too late and timeline already arranged `Hi, everyone` into stream.
We get `This is message edition` msg under `Set` action.
As we know, `Set` action means by record modification.

PR #1137, opened 2023-11-09
Support encrypted for edit/reply of media msg

bitfriend:
Now edit/reply of media msg is supported for only unencrypted room.
We should support the encrypted room too.


gnunicorn (approved):

bitfriend:
Should I implement editing of audio or video msg too?

gnunicorn:
> Should I implement editing of audio or video msg too?

yes. But it could happen in a second PR if that would mean delaying this for very long..

bitfriend:
No, I think it will not take a long time.

Issue #1139, opened 2023-11-10
Mac desktop app:  Fatal Error - Reference model not found.


kumarpalsinh25:
Every-time when I run app on mac desktop, it is giving below error. 

<img width="792" alt="Screenshot 2023-11-10 at 4 10 44 PM" src="https://github.com/acterglobal/a3/assets/51526480/f19bd595-173d-455d-a345-29f7611f68c2">

gnunicorn:
It appears this is a [general data-not-found error in the core db fetching code](https://github.com/acterglobal/a3/blob/fd3dacaf25f449bc44780da3002948f83d26ada5/native/core/src/store.rs#L29). I'll add some more information to the error message so we have some starting point for debugging.

PR #1140, opened 2023-11-10
Attempting at debugging "Reference Model not found"-Error

gnunicorn:
Found the error being thrown from our internal code when trying to create the storage and land on a key being stored that we can't find the model for anymore. Not sure where this inconsistency originates but it is save to assume that we therefore don't have that model and should go on without trying to decipher it more.

refs #1139 , maybe even fixes that, but needs confirmation

### DIFF
--- a/native/core/src/error.rs
+++ b/native/core/src/error.rs
@@ -20,8 +20,8 @@ pub enum Error {
     #[error("Error with the matrix sdk Store")]
     Store(#[from] matrix_sdk::StoreError),
 
-    #[error("Reference Model not found.")]
-    ModelNotFound,
+    #[error("Model not found at {0}.")]
+    ModelNotFound(String),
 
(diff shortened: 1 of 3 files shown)


kumarpalsinh25:
@gnunicorn,
This solution seems to be working. I had "Reference Model Error" issue on mac desktop but but after taking this PR changes, it works fine.

bitfriend:
> Found the error being thrown from our internal code when trying to create the storage and land on a key being stored that we can't find the model for anymore. Not sure where this inconsistency originates but it is save to assume that we therefore don't have that model and should go on without trying to decipher it more.
> 
> refs #1139 , maybe even fixes that, but needs confirmation

Yeah, some models were defined before but they may be deleted from source code now.

bitfriend (approved):
Looks good to me

PR #1149, opened 2023-11-14
fix breaking changes from ActerAvatar

gtalha07:
- Updated ```ActerAvatar``` version introduces breaking changes which deemed necessary to refactor avatars UI. 
- Updated ```flutter_chat_ui``` package to newer version and fix inconsistencies.


app/lib/features/chat/widgets/room_avatar.dart
```diff
@@ -32,26 +33,54 @@ class RoomAvatar extends ConsumerWidget {
   Widget chatAvatarUI(Convo convo, WidgetRef ref) {
     //Data Providers
     final convoProfile = ref.watch(chatProfileDataProvider(convo));
+    final canonicalParent =
+        ref.watch(canonicalParentProvider(convo.getRoomIdStr()));
```
gnunicorn, line 37: I feel like that is probably something we want the caller to be able to configure. As at least in the space-chats we do not want to show parent-badges .... and if I am understanding this correctly, we would do this now.

app/lib/features/chat/widgets/room_avatar.dart
```diff
@@ -130,37 +165,63 @@ class RoomAvatar extends ConsumerWidget {
...
               ),
-              child: Text(
-                '+${members.length - 2}',
-                style: const TextStyle(fontSize: 8),
+              for (int i = 2; i < members.length; i++)
+                AvatarInfo(
+                  uniqueId: members[i].userId().toString(),
+                ),
+            ],
+            orElse: () => [],
```
gnunicorn, line 194: shouldn't we fallback to just using all userIds?

```suggestion
            orElse: () => [
              for (int i = 1; i < members.length; i++)
                AvatarInfo(
                  uniqueId: members[i].userId().toString(),
                ),
              ]
```

app/lib/features/chat/widgets/room_avatar.dart
```diff
@@ -130,37 +165,63 @@ class RoomAvatar extends ConsumerWidget {
...
+                displayName: secData.displayName,
+                avatar: secData.getAvatarImage(),
               ),
-            ),
+              for (int i = 2; i < members.length; i++)
+                AvatarInfo(
+                  uniqueId: members[i].userId().toString(),
+                ),
+            ],
+            orElse: () => [],
```
gnunicorn, line 220: same as above.

app/lib/features/home/providers/navigation.dart
```diff
@@ -41,23 +42,51 @@ final spaceItemsProvider = FutureProvider.autoDispose
...
-          softWrap: false,
+        orElse: () => SidebarNavigationItem(
+          icon: ActerAvatar(
+            mode: DisplayMode.Space,
+            avatarInfo: AvatarInfo(
+              uniqueId: roomId,
+              displayName: info.displayName,
+              avatar: info.getAvatarImage(),
+            ),
+            avatarsInfo: const [AvatarInfo(uniqueId: '!')],
```
gnunicorn, line 80: what is this?

app/lib/features/home/providers/navigation.dart
```diff
@@ -41,23 +42,51 @@ final spaceItemsProvider = FutureProvider.autoDispose
...
+            ),
+            avatarsInfo: parent != null
+                ? [
+                    AvatarInfo(
+                      uniqueId: parent.space.getRoomIdStr(),
+                      displayName: parent.profile.displayName,
+                      avatar: parent.profile.getAvatarImage(),
+                    ),
+                  ]
+                : [],
```
gnunicorn, line 62: other than this, the code is the same, no? And we don't care about any difference between loading and error... so why not go for `canonicalParent.valueOrNull != null ...` allowing us to collapse that entire code and remove the `orElse` ...

app/lib/features/space/settings/widgets/space_settings_menu.dart
```diff
@@ -31,16 +32,35 @@ class SpaceSettingsMenu extends ConsumerWidget {
           children: [
             ...spaceProfile.when(
               data: (spaceProfile) => [
-                SpaceParentBadge(
-                  roomId: spaceId,
-                  badgeSize: 18,
-                  child: ActerAvatar(
+                canonicalParent.maybeWhen(
```
gnunicorn, line 35: same as before. if all we do is show a different default in `null` why bother with a `maybeWhen` and not just use `canoncialParent.valueOrNull`?!?

app/lib/features/space/widgets/shell_header.dart
```diff
@@ -16,20 +16,42 @@ class ShellHeader extends ConsumerWidget {
...
+    final canonicalParent = ref.watch(canonicalParentProvider(spaceId));
     return Padding(
       padding: const EdgeInsets.only(left: 10),
       child: Row(
         children: <Widget>[
-          SpaceParentBadge(
-            roomId: spaceId,
-            badgeSize: 40,
-            child: ActerAvatar(
+          canonicalParent.maybeWhen(
```
gnunicorn, line 24: same same

gnunicorn (approved):
Minor code reduction suggestions.

app/lib/common/themes/chat_theme.dart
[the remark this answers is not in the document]
gtalha07: These changes and others unrelated are coming from the newer version of ```flutter_chat_ui```.

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -77,16 +76,15 @@ class RoomProfilePage extends ConsumerWidget {
...
-                Center(
-                  child: Padding(
-                    padding: const EdgeInsets.only(
-                      top: 38,
-                      bottom: 12,
-                    ),
-                    child: SpaceParentBadge(
-                      badgeSize: 20,
-                      roomId: roomId,
+                Flexible(
```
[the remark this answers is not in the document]
gtalha07: Oh its not needed here , must've forgot removing. I'll address it. 👍🏻

app/lib/features/chat/widgets/room_avatar.dart
gtalha07: Yeah you're correct. Thanks for noticing it out.

app/lib/features/home/providers/navigation.dart
gtalha07: An error fallback case, but its not needed now we will favor ```canonicalParent.valueOrNull```.

PR #1150, opened 2023-11-15
Bug fixing

### LINKED ISSUE #1127: scrolling issue on iOS (linking existing chat)
this only happens when linking chat, not linking space

https://github.com/acterglobal/a3/assets/77483554/7b246a53-549e-4d81-b2a9-c1894865cfc4

### LINKED ISSUE #1143: should not show full sub-space description on spaces pagw
Currently we show the full description. this gets very messy.
Instead we should just show the first 1 or 2 lines

Image

kumarpalsinh25:
- Resolve scroll issue when link to a sub-chat, fixes #1127
- Manage space description with 1 line and read more option, fixes #1143

### DIFF
--- a/app/lib/common/providers/room_providers.dart
+++ b/app/lib/common/providers/room_providers.dart
@@ -4,8 +4,10 @@ import 'dart:core';
 import 'package:acter/common/models/profile_data.dart';
 import 'package:acter/common/providers/notifiers/room_notifiers.dart';
 import 'package:acter/common/providers/space_providers.dart';
+import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:acter/common/providers/chat_providers.dart';
 
 class RoomItem {
   final Member? membership;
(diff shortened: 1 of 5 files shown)


app/lib/common/providers/room_providers.dart
```diff
@@ -76,6 +78,57 @@ final briefRoomItemWithMembershipProvider =
...
+  for (final element in allRooms) {
+    final room = await ref.watch(maybeRoomProvider(element.getRoomIdStr()).future);
+    if (room == null) {
+      throw 'Room ${element.getRoomIdStr()} not found';
+    }
+
+    final profileData = await ref.watch(roomProfileDataProvider(element.getRoomIdStr()).future);
+    final name = profileData.displayName ?? element.getRoomIdStr();
+
+    if (name.toLowerCase().contains(searchValue!.toLowerCase())) {
```
gtalha07, line 118: ```searchValue``` has potential for throwing null, how about handling it prior to reaching this line...

app/lib/common/widgets/spaces/space_hierarchy_card.dart
```diff
@@ -160,8 +165,17 @@ class SpaceHierarchyCard extends ConsumerWidget {
     final roomId = space.roomIdStr();
     final profile = ref.watch(spaceHierarchyProfileProvider(space));
     final topic = space.topic();
-    final Widget? subtitle =
-        topic != null && topic.isNotEmpty ? Text(topic) : null;
+    final Widget? subtitle = topic != null && topic.isNotEmpty
+        ? ReadMoreText(
```
gtalha07, line 169: I'd suggest using [expandable_text](https://pub.dev/packages/expandable_text), that's currently existing as utility package for handling cases like these.

app/pubspec.yaml
```diff
@@ -120,6 +120,7 @@ dependencies:
   flutter_local_notifications: ^16.0.0
   email_validator: ^2.1.17
   app_settings: ^5.1.1
+  readmore: ^2.2.0
```
gtalha07, line 123: We can remove this and instead use ```expandable_text``` already existing ;)

gtalha07 (changes requested):
Overall good but couple of bug potential and redundant cases I am seeing which needs addressed before merge!

app/lib/common/providers/room_providers.dart
kumarpalsinh25: Good notice. 👍 
I have will do the required changes here.

app/lib/common/widgets/spaces/space_hierarchy_card.dart
kumarpalsinh25: Oh I see. I will reuse the existing library here.  Thanks for letting me know about the same. :)

app/pubspec.yaml
kumarpalsinh25: Yes sure :)

app/lib/common/providers/room_providers.dart
```diff
@@ -76,6 +78,62 @@ final briefRoomItemWithMembershipProvider =
...
+
+  for (final element in allRooms) {
+    final room =
+        await ref.watch(maybeRoomProvider(element.getRoomIdStr()).future);
+    if (room != null) {
+      final profileData = await ref
+          .watch(roomProfileDataProvider(element.getRoomIdStr()).future);
+      final name = profileData.displayName ?? element.getRoomIdStr();
+
+      if (name.toLowerCase().contains(searchValue!.toLowerCase())) {
```
gtalha07, line 122: How ```searchValue``` null value is handled right now??? This comparison might break the provider itself. How about doing this?
```suggestion
      if(searchValue != null){
      
      if (name.toLowerCase().contains(searchValue!.toLowerCase())) {
         .....
      }
    }
    else {
        throw error
    }
```

gtalha07:
Also append the doc changes ;)

app/lib/common/providers/room_providers.dart
kumarpalsinh25: My Bad!
Updated code which will cover above use-case. :)

gtalha07 (approved):

PR #1152, opened 2023-11-16
Fix bug in user search by full name

bitfriend:
- [User search failed](https://github.com/matrix-org/synapse/issues/7588) when namespace contains `.`, for example `miles:acter` works but `miles:acter.global` doesn't work.
- Add integration test for user search

### DIFF
--- a/native/test/src/tests.rs
+++ b/native/test/src/tests.rs
@@ -11,6 +11,7 @@ mod receipt;
 mod redact;
 mod reply;
 mod rsvp;
+mod search_users;
 mod spaces;
 mod tasks;
 mod templates;
--- a/native/test/src/tests/search_users.rs
+++ b/native/test/src/tests/search_users.rs
@@ -1,0 +1,52 @@
+use acter::{
+    api::login_new_client,
+    ruma_common::{OwnedUserId, UserId},
+};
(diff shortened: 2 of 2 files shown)


bitfriend:
Actually rust side is too simple and clear so it can't have bug.
https://github.com/matrix-org/matrix-rust-sdk/blob/e1feefcbb0bed8badec62fe46f59e7a5dd01de60/crates/matrix-sdk/src/client/mod.rs#L387-L408

bitfriend:
SQL query seems to have issue when wildcard contains dot.
https://github.com/matrix-org/synapse/blob/3d87847ecc943c689c4587c5327d744e4a8f92c2/synapse/storage/database.py#L2409

bitfriend:
[This document](https://stackoverflow.com/questions/18636697/sql-like-command-doesnt-work-with-strings-containing-dots) looks helpful.

bitfriend:
> [This document](https://stackoverflow.com/questions/18636697/sql-like-command-doesnt-work-with-strings-containing-dots) looks helpful.

@gnunicorn It proposes regexp instead of wildcard.
Should I change source code of synapse?

gnunicorn:
closing for now.

PR #1154, opened 2023-11-16
Fix MacOS build & publishing flow

gnunicorn:
- [x] fix path of uploaders for publish-action
- [x] set build-target for macos rust lib to 11
- [x] fix debug running of macos app by quiting signing script early.


gtalha07 (approved):
LGTM!

PR #1159, opened 2023-11-17
Upgrade sdk

bitfriend:
Will test if new sdk fixed bug that removing of pushed rule doesn't work.


gnunicorn (approved):
No obvious problems.

PR #1160, opened 2023-11-17
Integrating Synapse Super Invites

gnunicorn:
Implement [Super Invites](https://pypi.org/project/synapse-super-invites/) in Acter: On supporter servers (this is checked), the user can create invitation-tokens and associate any number of rooms to it that upon redeeming the redeeming user is added to. Allowing for a quicker and easier space-wide-onboarding, especially when the server is configured to allow these as registration tokens and the registration-tokens are sent to the API after registration -- which is what we do.

Next to the rust API, and a Flutter UI to allow users to create, manage and redeem invitations from their profile settings, this comes with a full end-to-end UI flutter test, not only covering the regular add, but also the registration brings up the correct spaces and chats and that editing works as expected.

Here you can see the full-on multi-space-multi-chat-test in action: 

https://github.com/acterglobal/a3/assets/40496/a5a24ff6-92aa-4ad7-9e17-49764f3837b9

Since [0.8.1](https://pypi.org/project/synapse-super-invites/) this now also supports proper deletion and a test ensures that creating the same token again fails as expected.


kumarpalsinh25 (approved):
Very nice job! Looks good. 
Excited about this feature.

PR #1153, opened 2023-11-16
Fix math fork build problem by pinning to PR commit

gnunicorn:
Fixes:
```
Changed 8 dependencies!
24 packages have newer versions incompatible with dependency constraints.
Try `flutter pub outdated` for more information.

Running Gradle task 'bundleRelease'...                          
/opt/hostedtoolcache/flutter/stable-3.16.0-x64/.pub-cache/hosted/pub.dev/flutter_math_fork-0.7.1/lib/src/widgets/selection/gesture_detector_builder_selectable.dart:38:22: Error: Type 'TapDragUpDetails' not found.
  void onSingleTapUp(TapDragUpDetails details) {
                     ^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/flutter/stable-3.16.0-x64/.pub-cache/hosted/pub.dev/flutter_math_fork-0.7.1/lib/src/widgets/selection/gesture_detector_builder_selectable.dart:38:22: Error: 'TapDragUpDetails' isn't a type.
  void onSingleTapUp(TapDragUpDetails details) {
                     ^^^^^^^^^^^^^^^^
Target kernel_snapshot failed: Exception
````

By pinning the `flutter_math_fork` dependency to [the fix from this PR](https://github.com/simpleclub-extended/flutter_math_fork/pull/87)

PR #1161, opened 2023-11-20
switch to math_fork 0.7.2

gnunicorn:
clears #1153 .


bitfriend (approved):
Thank you for this patch.
`flutter pub get` works on windows without any change in `pubspec.lock`.

PR #1164, opened 2023-11-21
Edit message

### LINKED ISSUE #888: Edit chat message
allow user to edit their own chat messages.

### LINKED ISSUE #964: cache chat images and optimisation.
Chat Image messages aren't cached and gets a re-render on scrolling. Moreover, images decoded size should be optimised for chat feature as performance aspect.

kumarpalsinh25:
This is included the implementation of the Edit Message feature.

- User can edit the text the text message, fixes #888 
- Resolve image loading issue on the chat, fixes #964 

Reference:

https://github.com/acterglobal/a3/assets/51526480/b771ac90-d39a-46b9-8a9f-566ffbfd39fb


app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
```diff
@@ -6,11 +6,16 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 class ChatInputNotifier extends StateNotifier<ChatInputState> {
   ChatInputNotifier() : super(const ChatInputState());
 
+  void toggleEditView(bool value) =>
```
gnunicorn, line 9: this is not a toggle, but a set. A toggle would always `value = !value`
(the one below is misnamed, too for that matter..)

app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -38,6 +38,7 @@ class _ImageMessageBuilderState extends State<ImageMessageBuilder> {
             .imageBinary(widget.message.id)
             .then((value) => value.asTypedList()),
       );
+      setState(() {});
```
gnunicorn, line 41: what is this for? maybe add a doc comment because it looks like something someone might not understand what it is for and thus just deletes it in the future..

app/macos/Podfile.lock
```diff
@@ -100,4 +100,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce7dbe26c78bfc7ba46736094c1e2d25982870fa
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.12.1
```
gnunicorn, line 103: we should be doing something about this coming up again and again ... 
```suggestion
COCOAPODS: 1.13.0
```

gnunicorn (approved):
Looking good to me. Would appreciate if @gtalha07 could also comment on this.

gnunicorn:
@kumarpalsinh25 very nice! Does it work with other message types, too? Like images? If so, how?

app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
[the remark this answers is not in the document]
kumarpalsinh25: Yes sure, I will rename it.

**Sidenote:**
I have copied this naming from the previous implementation **(toggleReplyView)** similar this issue as below. ;)

```
 void toggleEditView(bool value) =>
      state = state.copyWith(showEditView: value);

  void toggleReplyView(bool value) =>
      state = state.copyWith(showReplyView: value);
``` 

Do I have to renamed `toggleReplyView` also?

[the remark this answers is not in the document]
gnunicorn: well, I'd say, yes... better code hygiene ... a run-by-fix for it.. thanks, mate!

app/lib/features/chat/widgets/image_message_builder.dart
kumarpalsinh25: Yes sure, I will add doc comment over there.

This is related to `image message` loading issue on the chat messages.
Actually, there was `async` method called from` initState()` to load image data but prior to that `build()` call and initialise with no image and going for loading forever.

So i have added  `setState(() {});` to update the image view when image data are loaded in the variable.

```

class _ImageMessageBuilderState extends State<ImageMessageBuilder> {
  AsyncValue<Uint8List> imageData = const AsyncValue.loading();

  @override
  void initState() {
    super.initState();
    loadImage();
  }

  Future<void> loadImage() async {
    try {
      imageData = AsyncValue.data(
        await widget.convo
            .imageBinary(widget.message.id)
            .then((value) => value.asTypedList()),
      );
      setState(() {});
    } catch (e, s) {
      imageData = AsyncValue.error(e, s);
    }
  }

  @override
    final size = MediaQuery.of(context).size;
    return imageData.when(
      data: (data) {
        return ClipRRect(
          borderRadius: widget.isReplyContent
              ? BorderRadius.circular(6)
              : BorderRadius.circular(15),
          child: ConstrainedBox(
            constraints: BoxConstraints(
              maxHeight:
                  widget.isReplyContent ? size.height * 0.2 : size.height * 0.3,
              maxWidth:
                  widget.isReplyContent ? size.width * 0.2 : size.width * 0.3,
            ),
            child: Image.memory(
              data,
              frameBuilder: (
                BuildContext context,
                Widget child,
                int? frame,
                bool wasSynchronouslyLoaded,
              ) {
                if (wasSynchronouslyLoaded) {
                  return child;
                }
                return AnimatedOpacity(
                  opacity: frame == null ? 0 : 1,
                  duration: const Duration(seconds: 1),
                  curve: Curves.easeOut,
                  child: child,
                );
              },
              errorBuilder: (
                BuildContext context,
                Object url,
                StackTrace? error,
              ) {
                return Text('Could not load image due to $error');
              },
              fit: BoxFit.cover,
            ),
          ),
        );
      },
      error: (e, st) => Text('Error loading image $e'),
      loading: () => const CircularProgressIndicator(),
    );
  }
}
```

app/macos/Podfile.lock
kumarpalsinh25: Yes sure, I will revert it.

And yes need to do something about this type of configuration files changes. May we can add it git-ignore to discard the local changes while commit.

kumarpalsinh25:
@gnunicorn. Thank you :)
Currently, Edit message is implemented for Text Messages only.

For image and other files type, we need to think about some UI-UX first before implementing it. Also many popular chat apps like WhatsApp, Skype etc don't have edit option for images and files.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -234,6 +279,47 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+
+                        final chatInputFocusState =
+                        ref.read(chatInputFocusProvider.notifier);
+                        WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+                          FocusScope.of(context)
+                              .requestFocus(chatInputFocusState.state);
+                        });
+                      },
+                      child: const Text(
+                        'Edit',
```
gtalha07, line 319: Minor typo . Should be `Editing Message`.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -505,8 +597,52 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+    WidgetRef ref,
+    Widget? child,
+  ) {
+    final roomId = widget.convo.getRoomIdStr();
+    final inputNotifier = ref.watch(chatInputProvider(roomId).notifier);
+    return Row(
+      children: [
+        const SizedBox(width: 5),
+        const Text(
+          'Edit:',
```
gtalha07, line 627: Minor typo, should be `Editing Message`.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -505,8 +597,52 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+          style: TextStyle(
```
gtalha07, line 628: We should avoid hardcoding theme variables. You can use ```Theme.of(context).textTheme``` appropriate for this.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -234,6 +279,47 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+                        final chatInputFocusState =
+                        ref.read(chatInputFocusProvider.notifier);
+                        WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+                          FocusScope.of(context)
+                              .requestFocus(chatInputFocusState.state);
+                        });
+                      },
+                      child: const Text(
+                        'Edit',
+                        style: TextStyle(color: Colors.white),
```
gtalha07, line 320: Avoid hardcoding theme variables.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -505,8 +597,52 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+            inputNotifier.toggleEditView(false);
+            inputNotifier.setReplyWidget(null);
+            inputNotifier.setEditWidget(null);
+            inputNotifier.setRepliedToMessage(null);
+            inputNotifier.setEditMessage(null);
+            FocusScope.of(context).unfocus();
+          },
+          child: const Icon(
+            Atlas.xmark_circle,
+            color: Colors.white,
```
gtalha07, line 609: Avoid hardcoding theme variables.

gtalha07 (approved):
Well put!! 👏🏻  I have pointed out some minor corrections and possible UI improvement for edited message indicator which can be addressed and merged without further explicit approval.

app/macos/Podfile.lock
gtalha07: @kumarpalsinh25  Agreed. This is somewhat annoying! :D

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: I think `Editing Message` will not be good typo as this is not a label for some process indication. This is the instruction label and that's why I have used just `Edit`.

I have checked other app also for this. Telegram is using just `Edit Message` label.

Telegram Reference Image:
Telegram

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -131,35 +139,49 @@ class _TextWidget extends ConsumerWidget {
...
+                                    .withOpacity(0.5)
+                                : null,
+                          ),
+                  maxLines: isReply ? 3 : null,
+                ),
+        ),
+        Visibility(
+          visible: wasEdited,
+          child: Text(
+            'Edited',
```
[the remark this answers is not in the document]
kumarpalsinh25: We ask design team to have thoughts this but I think current implementation is good.

I just checked WhatsApp and Telegram app. They are also showing `Edited` label within Bubble same as I did it currrently.

Reference Images:

WhatsApp Image 2023-11-21 at 7 29 44 PM
WhatsApp Image 2023-11-21 at 7 30 20 PM

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Actually, I have followed similar code that was already there for other menu options label as below.

 ```
InkWell(
                    onTap: () => customMsgSnackbar(
                      context,
                      'More options not implemented yet',
                    ),
                    child: const Text(
                      'More',
                      style: TextStyle(color: Colors.white),
                    ),
                  ),
```

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Actually, I have copies this same from the current code base which was already implemented for replyView.

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Ah, I just removed `style: TextStyle(color: Colors.white),` as default textTheme have white color.

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Ah, I just see that default Icon Theme have white color so just removed  `color: Colors.white,`.

PR #1167, opened 2023-11-22
chat invite stalling bug fix

### LINKED ISSUE #1133: Accepting chat invite is stalling indefinitely
When i click on accept invites its stalling.
However looking on element i see i alreaedy accepted the invite.
I can fix it by restarting the app

https://github.com/acterglobal/a3/assets/77483554/c324abbe-3a20-4000-8ce3-6281e4c2cb2e

gtalha07:
- [x]  fixes #1133.


app/lib/features/activities/widgets/invitation_card.dart
```diff
@@ -99,4 +81,19 @@ class InvitationCard extends ConsumerWidget {
...
+    EasyLoading.show(status: 'Joining', dismissOnTap: false);
+    bool res = await invitation.accept();
+    if (!res) {
+      EasyLoading.dismiss();
+      EasyLoading.showError('Failed to join');
+      return;
+    }
+    EasyLoading.dismiss();
+    if (ctx.mounted) {
+      ctx.goNamed(Routes.chatroom.name, pathParameters: {'roomId': roomId});
```
kumarpalsinh25, line 96: What if the invitation is of space type?

app/lib/features/activities/widgets/invitation_card.dart
```diff
@@ -62,26 +63,7 @@ class InvitationCard extends ConsumerWidget {
...
-                      if (await invitation.accept() != true) {
-                        if (context.mounted) {
-                          customMsgSnackbar(context, 'Failed to join.');
-                        }
-                      } // FIXME: route there? But we don't know if room or space ... :(
-                      if (context.mounted) {
-                        Navigator.of(context, rootNavigator: true).pop();
-                      }
-                    },
+                    onPressed: () => _onInviteAccept(context, data.roomId),
```
kumarpalsinh25, line 66:
```suggestion
                    onPressed: () => _onInviteAccept(context, data.roomId),
```

```suggestion
                    onPressed: () => _onTapAcceptInvite(context, data.roomId),
```

kumarpalsinh25 (commented):
Looks good. One minor query to be addressed before merging. :)

app/lib/features/activities/widgets/invitation_card.dart
gtalha07: Yeah I just tested, it must have a different route. However, it seems difficult knowing it'll be space or room before redirecting...

gtalha07: @gnunicorn Any thoughts on this?

gtalha07: I suppose we could check with provider via roomId whether its space or room before routing.

gnunicorn: we indeed do not currently know whether an invite is for a space or room and that is pretty annoying... element seems to be able to distinguish that somehow though ... we might want to open a ticket to investigate that and leave it as is for now?

kumarpalsinh25: If we can have `roomType` param in [invitation_profile.dart](https://github.com/acterglobal/a3/blob/main/app/lib/common/models/invitation_profile.dart) then it will solve this issue.

Temporary we can use relevant provider to get `roomType` from `roomId` before routing so that it will not misbehave with specific roomType.

gtalha07: > If we can have `roomType` param in [invitation_profile.dart](https://github.com/acterglobal/a3/blob/main/app/lib/common/models/invitation_profile.dart) then it will solve this issue.
> 
> Temporary we can use relevant provider to get `roomType` from `roomId` before routing so that it will not misbehave with specific roomType.

I tried using ```maybeSpaceProvider``` and ```maybeRoomProvider``` to return either `Space` or `Convo` to determine roomType but its not working as supposedly should. It's returning in both cases. We can leave redirection for now and open separate ticket as @gnunicorn said.

PR #1168, opened 2023-11-22
Rust proxy support

gnunicorn:
Allow us to set a http proxy at compile time for testing. Shows it in profile -> settings -> app info for confirmation.

Example usage by setting `--dart-define=HTTP_PROXY=http://localhost:8090/`, or via vscode config:
```diff
diff --git a/.vscode/launch.json b/.vscode/launch.json
index 0f8604193b..f6ffb9355f 100644
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -99,7 +99,9 @@
             "flutterMode": "debug",
             "toolArgs": [
                 "--dart-define",
-                "RAGESHAKE_URL=http://localhost/api/submit"
+                "RAGESHAKE_URL=http://localhost/api/submit",
+                "--dart-define",
+                "HTTP_PROXY=http://localhost:8118/",
             ]
         },
         {
```


kumarpalsinh25 (approved):
LGTM!

gnunicorn:
I wonder if we want to have either support for changing the proxy locally from within the app (similar to the debug-logging-setting with storing it locally in the user-settings) and/or we want to add support for reading the system-wide-proxy setting and set those (using [platform_proxy](https://gitlab.com/yamsergey/platform_proxy)) but I don't have time for that now and I don't think this has a high priority ...

kumarpalsinh25:
@gnunicorn 
As matrix is not providing HTTPS logs, we can make this point as low priority for making this as dynamic from app settings.

PR #1174, opened 2023-11-23
implement local echo and retry state for timeline event.

gtalha07:
- [x]  implement  send state and expose API from rust side.
- [x] implement local echo states and retry state on flutter side.
- [x] implement retry functionality for ```types.Message```. 
- [x] fix bug where user receipts were only enabled for ```m.room.message``` and has been generalised to support all room event cases.


bitfriend:
Why your changed file is only `message.rs`?
You will have changes in `api.rs` and `api.rsh` too.
After compilation, you will get change in `acter_flutter_sdk_ffi.dart`.

gtalha07:
> Why your changed file is only `message.rs`? You will have changes in `api.rs` and `api.rsh` too. After compilation, you will get change in `acter_flutter_sdk_ffi.dart`.

I haven't pushed it yet. please wait....

gtalha07:
**Preview for local echo sending:**
Screen Recording 2023-11-28 at 8 27 03 PM

Retry state
Screenshot 2023-11-29 at 5 57 17 PM

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -429,6 +434,17 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
     String eventId = eventItem.uniqueId();
 
     String? inReplyTo = eventItem.inReplyTo();
+
+    // user read receipts for timeline event item
```
gnunicorn, line 438: I am unclear about these receipt changes in here ... are they meant to be in this PR? What do they relate to in the description?

app/lib/features/chat/widgets/receipts_builder.dart
```diff
@@ -1,18 +1,114 @@
...
 import 'package:acter_avatar/acter_avatar.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show Convo, EventSendState;
+import 'package:atlas_icons/atlas_icons.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:quds_popup_menu/quds_popup_menu.dart';
+import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 
 class ReceiptsBuilder extends ConsumerWidget {
```
gnunicorn, line 12: So the ReceiptsBuilder is more than just about who has seen the message now? Then how about renaming it into something more appropriate? Like ... `MessageMetadataBuilder` or something?

native/acter/src/api/message.rs
```diff
@@ -82,11 +82,55 @@ use super::common::{
...
+            SdkEventSendState::Sent { event_id } => {
+                evt_id = Some(event_id.clone());
+                "Sent".to_string()
+            }
+        };
+        EventSendState {
+            state,
+            error: err,
+            event_id: evt_id,
+        }
```
gnunicorn, line 112: minor style thing. I prefer using direct assignments rather than mutable variables. I'd propose:

```suggestion
        let (state, error, event_id) = match inner {
            SdkEventSendState::NotSentYet => ("NotSentYet".to_string(), None, None),
            SdkEventSendState::Cancelled => ("Cancelled".to_string(), None, None),
            SdkEventSendState::SendingFailed { error } =>
                ("SendingFailed".to_string(),  Some(error.to_owned().to_string()), None),
            SdkEventSendState::Sent { event_id } =>
                ("Sent".to_string(), None, Some(event_id.clone())),
        };
        EventSendState {
            state,
            error,
            event_id,
        }
```

gnunicorn (approved):
Congrats on your first rust PR, @gtalha07 : Great work. Looks all good, just some minor changes you can but are not obliged to use.

app/pubspec.lock
```diff
@@ -2027,5 +2027,5 @@ packages:
     source: hosted
     version: "3.1.2"
 sdks:
-  dart: ">=3.1.0 <4.0.0"
+  dart: ">=3.2.0-194.0.dev <4.0.0"
```
kumarpalsinh25, line 2030: Revert all the changes of `app/pubspec.lock` if those not relevant to this PR.

app/lib/features/chat/widgets/receipts_builder.dart
```diff
@@ -1,18 +1,114 @@
...
+              width: 8,
+              child: CircularProgressIndicator(),
+            );
+          case 'SendingFailed':
+            return Row(
+              children: <Widget>[
+                GestureDetector(
+                  onTap: () => _handleCancelRetrySend(),
+                  child: Text(
+                    'Cancel Send',
```
kumarpalsinh25, line 44: Suggestion:
Does `cancel send` button require?

Because most of the messaging app, if message gets failed to send the it keeps showing with retry only action however if user don't want to retry and cancel it then user generally have option for delete message.

app/lib/features/chat/widgets/receipts_builder.dart
```diff
@@ -1,18 +1,114 @@
...
+                  ),
+                ),
+                const SizedBox(
+                  width: 10,
+                ),
+                GestureDetector(
+                  onTap: () => _handleRetry(),
+                  child: RichText(
+                    text: TextSpan(
+                      text: 'Failed to sent: ${sendState.error()}. ',
```
kumarpalsinh25, line 58: [B](text: 'Failed to sent: ${sendState.error()}.)
```suggestion
                      text: 'Failed to sent.',
```

Better to now show technical errors to end users.

gnunicorn: Disagreed. As it makes a huge difference to the user whether the message wasn't sent because there is no internet vs. they are not authorized to sent it. As we don't have proper types for the errors and thus can't show a more nuanced message, at least showing the actual error (even just so we can see it in a screenshot they report), is better than nothing.

app/lib/features/chat/widgets/receipts_builder.dart
gtalha07: I just inferred it based on the local echo states, as they include a `Cancelled` state and fn for ```TimelineItem```.i.e. ```cancel_retry()``` which is supposed to be functional?

kumarpalsinh25: No issue with the current implementation. I am just thinking for `Cancel Send` option needs as user will have delete message option on long-press event.

Issue #1115, opened 2023-11-03
Open and view files and/or images/videos


gnunicorn:
when clicking on an image within the chat, nothing happens. Instead a large zoom in of the image should be made available so the user can look at it more closely. Same for videos. For other files, the user should be prompted if they want to download and open it and if agreed do so ;) .

gnunicorn:
for files: on click, ask the user to "open" or "download".

PR #1176, opened 2023-11-24
General bug fixes and some user experience improvements

### LINKED ISSUE #1121: Fix/improve user path
**Chats**
Currently if a user go to "Space A" and then to the spaces page and click on a Sub space (Space B), when the user then click on the back button, the user is sent back to the home page. 

Instead, the user should be directed back to the previous page. In this example it means, that when the user click on the back button from Space B, the user should be directed back to Space A.

**Chats**
The same should be for chats, meaning if a user click and enter a chat from a space and the user click on the back button, the user should be directed back to the space, not the chat overview. The user only be directed back to the chat overview, if the user entered the chat from the chat overview.

### LINKED ISSUE #1146: remove the option to create pins and events for users who dont have permission.
Its confusing and frustrating that you as a user can go all the way to the point that its loading in creation process, but then shows error.

Image

kumarpalsinh25:
- Changes in back navigation for Sub-spaces, fixes #1121 
- Managed post ping/event option based on the permission, fixes #1146 
- Minor correction on routing: After redacting of pin, it should redirect back to pin listing instead of going to event list
- Added support for full screen image view in chat, ref #1115

### DIFF
--- a/.changes/1176-general-fixes.md
+++ b/.changes/1176-general-fixes.md
@@ -1,0 +1,3 @@
+- [bugfix] : Improved back navigation between sub-spaces and space
+- [bugfix] : Ping/Event creation is now based on the required permission
+- [enhancement] : User can now see chat image in full screen view
--- a/.changes/1176-general-fixes.md
+++ b/.changes/1176-general-fixes.md
@@ -1,3 +1,3 @@
 - [bugfix] : Improved back navigation between sub-spaces and space
-- [bugfix] : Ping/Event creation is now based on the required permission
+- [bugfix] : Pin/Event creation is now based on the required permission
 - [enhancement] : User can now see chat image in full screen view
--- a/app/lib/features/chat/widgets/image_message_builder.dart
+++ b/app/lib/features/chat/widgets/image_message_builder.dart
@@ -53,7 +53,7 @@ class _ImageMessageBuilderState extends State<ImageMessageBuilder> {
       barrierDismissible: false,
       useRootNavigator: false,
       builder: (ctx) => ImageDialog(
-        title: 'Image',
+        title: widget.message.name,
         imageData: data,
       ),
     );
--- a/app/lib/features/chat/widgets/image_message_builder.dart
+++ b/app/lib/features/chat/widgets/image_message_builder.dart
@@ -1,3 +1,4 @@
+import 'package:acter/common/widgets/image_dialog.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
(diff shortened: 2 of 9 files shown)


.changes/1176-general-fixes.md
```diff
@@ -0,0 +1,3 @@
+- [bugfix] : Improved back navigation between sub-spaces and space
+- [bugfix] : Ping/Event creation is now based on the required permission
```
gnunicorn, line 2:
```suggestion
- [bugfix] : Pin/Event creation is now based on the required permission
```

app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -46,48 +47,64 @@ class _ImageMessageBuilderState extends State<ImageMessageBuilder> {
     }
   }
 
+  void showFullScreenImage(Uint8List data) {
+    showAdaptiveDialog(
+      context: context,
+      barrierDismissible: false,
+      useRootNavigator: false,
+      builder: (ctx) => ImageDialog(
+        title: 'Image',
```
gnunicorn, line 56: shouldn't the image-message have a name as well?
```suggestion
        title: widget.message.name(),
```

app/lib/features/space/pages/pins_page.dart
```diff
@@ -19,81 +19,106 @@ class SpacePinsPage extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final spaceRelationsOverview =
```
gnunicorn, line 22: that's not the right provider for this. there is a provider that directly yields the space profile with the membership. We don't need that entire relationship thing here.

app/lib/features/space/pages/pins_page.dart
```diff
@@ -19,81 +19,106 @@ class SpacePinsPage extends ConsumerWidget {
...
+                            color: Theme.of(context).colorScheme.neutral5,
+                          ),
+                          iconSize: 28,
+                          color: Theme.of(context).colorScheme.surface,
+                          onPressed: () => context.pushNamed(
+                            Routes.actionAddPin.name,
+                            queryParameters: {'spaceId': spaceIdOrAlias},
+                          ),
+                        ),
+                      ),
```
gnunicorn, line 73: If we do that entire spaceRelationsOverview only to get the canPostPin figured out here, we probably do not want to wait for that to show the pins. Instead I'd suggest to replace the Visibility with a Consumer or a widget specific that then does the provider call ..

looking at that and how often we have that pattern we might want to create a common reusable Widget `UserCan(roomId, canString, child: IconButton( ... ) )` that does a `roomProviderMembership`-lookup, checks for that `canString` and if that comes with false, renders a `SizedBox.shrink()`.

gnunicorn (approved):
Some improvement suggestions.

app/lib/features/space/pages/pins_page.dart
kumarpalsinh25: Good notice.
I will create reusable components for this.

gtalha07 (approved):
LGTM!

Issue #1109, opened 2023-10-31
Bug in redacting pins/updates/event


emilvincentz:
Although you redact/delete a pin/update it still shows. 
Event when restartin the app it still shows. However usually an undefinite time, they go away.

kumarpalsinh25:
@bitfriend 
Can you check for below method regarding this issue? It is giving success response for pin and/or event reduction but it is not actually deleted.

```
 /// redact an event from this room
  /// reason - The reason for the event being reported (optional).
  Future<bool> redactContent(
    String eventId,
    String? reason,
  )
```

gnunicorn:
the internal state machine for our objects doesn't yet support redacting ... same as it doesn't remove items that are part of a specific space if you left it ... we should probably make some tracking issue for these somewhere...

gnunicorn:
fixed in https://github.com/acterglobal/a3/pull/1448 .

PR #1178, opened 2023-11-25
Bugfix in redacting pins/updates

bitfriend:
Will fix https://github.com/acterglobal/a3/issues/1109

### DIFF
--- a/app/lib/common/widgets/redact_content.dart
+++ b/app/lib/common/widgets/redact_content.dart
@@ -19,6 +19,7 @@ class RedactContentWidget extends ConsumerWidget {
   final bool isSpace;
   final void Function()? onRemove;
   final Function()? onSuccess;
+
   const RedactContentWidget({
     super.key,
     this.title,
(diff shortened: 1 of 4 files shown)


gtalha07:
Is it possible to include working case for this fix in form of short video? That'd be helpful for review!

app/lib/common/widgets/redact_content.dart
```diff
@@ -94,44 +94,25 @@ class RedactContentWidget extends ConsumerWidget {
...
-                  title: 'Close',
-                  isOutlined: true,
-                ),
-              ],
-            ),
-          );
+      if (ctx.mounted) {
+        Navigator.of(ctx, rootNavigator: true).pop();
+        Navigator.of(ctx, rootNavigator: true).pop(true);
+        customMsgSnackbar(ctx, 'Content successfully removed');
```
gtalha07, line 113: I'd propose to change all this structure using ```EasyLoading``` indicators, its quite handy in code reduction and easy understanding of what's happening ;)

gtalha07 (approved):

gnunicorn (approved):
The code looks fine and should probably be applied, but I am fairly certain that won't fix the actual bug in question - as the problem is within the inner state machine not properly handling the received redaction on the existing object.

I've updated the description to reflect that.

PR #1183, opened 2023-11-27
Ben fix 938

### LINKED ISSUE #938: Unverified session bug for new users
new users have 2 unverified sessions, although they never signed in on other devices with that account

Image

### LINKED ISSUE #1180: Remove notifications messages from activities screen

gnunicorn:
Potentially best reviewed changeset by changeset.

- fixes #938 : turns out that registrating-then-login-pattern creates two sessions for us upon startup. Unfortunately, that is caused by a [matrix-sdk bug](https://github.com/matrix-org/matrix-rust-sdk/pull/2888) but by switching to our own fork, we can avoid that for now. Dropping the number of sessions to 1. However, that still means an "unverified" devices case shows up in the activities stream, so in a second part, we also need to remove our own device from the list of unknown devices.
  This comes with an integration regression tests ensuring that no unverified devices are found upon simple registration (actually, it was even built in test-driven fashion: first made the test, then figured out the fix).
- This, too, fixes a bug reported just earlier today: for reasons unknown we were disallowing spaces in the display name upon registration. This, too, comes with an integration regression test ensuring the display name is taken over properly.
- Lastly, this removes the notification listing from the activities screen, as it served no purpose nor value. Fixes #1180 .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2903,7 +2903,7 @@ dependencies = [
 [[package]]
 name = "matrix-sdk"
 version = "0.6.2"
-source = "git+https://github.com/matrix-org/matrix-rust-sdk?rev=7e53c6821be8ec7ab5adcb84e43e478773e79c30#7e53c6821be8ec7ab5adcb84e43e478773e79c30"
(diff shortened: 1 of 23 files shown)


gtalha07 (approved):
LGTM!

PR #1187, opened 2023-11-29
Subspace view permissions fix

gnunicorn:
Turns out, we forgot to update the join_rule when linking a space/chat after we've created it, making it only show up for users already part of that chat/space. This PR:

- Adds two UX e2e-integration regression tests for being able to see subspaces when created right with the parent as well as linked later
- Fixes the problem by checking the `join_rules` of the child-room and prompting the user about whether we should update them if they can't be seen.


app/integration_test/support/super_invites.dart
gnunicorn, line 1: This is just a move from the test file into a reusable support file

emilvincentz:
@gnunicorn but does this fix the sub-spaces not showing in WeRise for kumar?
All except one or two I created directly within the WeRise space, but apparently still doesnt show

app/lib/features/space/sheets/link_room_sheet.dart
```diff
@@ -349,6 +357,62 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
...
       if (room != null) {
         room.addParentRoom(selectedParentSpaceId, true);
+        final joinRule = room.joinRuleStr();
+        List<String> currentRooms = [];
+        bool parentCanSee = joinRule == 'public';
+        String newRule = 'restricted';
+        if (joinRule == 'restricted' || joinRule == 'knock_restricted') {
+          currentRooms =
+              room.restrictedRoomIdsStr().map((t) => t.toString()).toList();
+          parentCanSee = currentRooms.contains(selectedParentSpaceId);
```
kumarpalsinh25, line 367: Just to confirmed,

 when we have `restricted` or `knock_restricted` still it doesn't sufficient for space member can see it within space. 
 
 specific `room` should have `parentId` in `restrictedRoomIdsStr` to apply this join rule.
 
 `restricted` or `knock_restricted` without setting `parentId` in `restrictedRoomIdsStr` can be consider same as `invite` join_rule?

app/lib/features/space/sheets/link_room_sheet.dart
```diff
@@ -349,6 +357,62 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
       final room = await ref.watch(maybeRoomProvider(roomId).future);
       if (room != null) {
         room.addParentRoom(selectedParentSpaceId, true);
+        final joinRule = room.joinRuleStr();
```
kumarpalsinh25, line 360: Better to have:
- Can we make separate function for `join logic` for code separation?

app/lib/features/space/sheets/link_room_sheet.dart
```diff
@@ -349,6 +357,62 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
...
+          // ignore: use_build_context_synchronously
+          bool shouldChange = await showDialog(
+            context: context,
+            builder: (context) {
+              return AlertDialog(
+                title: const Text('Not visible'),
+                content: const Wrap(
+                  children: [
+                    Text(
+                      "The current join rules of the child mean it won't be visible in the parent space to non-members. Should we update the join rules to allow for any parent space member to see and join the sub space?",
```
kumarpalsinh25, line 381: Suggestion:
Although, this feature is mainly used by admin or pro users, we can make this make less technical and more user friendly.

kumarpalsinh25 (approved):
Looks good. Some queries where need some information. 

And one general question.
- What happen if we unlink any sub-space/sub-chat? Do we need to change any join_rules at that time?

app/lib/features/space/sheets/link_room_sheet.dart
gnunicorn: That is correct. The "restricted" fields have a list roomIds (parents), who's member the access it restricted to. See https://spec.matrix.org/v1.8/client-server-api/#restricted-rooms

app/lib/features/space/sheets/link_room_sheet.dart
gnunicorn: Only if you have a better suggestion...

app/lib/features/space/sheets/link_room_sheet.dart
gnunicorn: yeah, I can do that.

gnunicorn:
@kumarpalsinh25 I only focused on linking for now. For the unlink it doesn't really matter as the server is never selecting the child room to be shown, if they aren't no child no more. Thus I left that out for now. Instead, I'd very much like to think about a UI for managing room visibility on both chat and space at some point and allow to manage things from there more specifically.

app/lib/features/space/sheets/link_room_sheet.dart
kumarpalsinh25: Just a suggestion. You can modify or use it based on what best fit for the requirement.

"The current join rules of the {roomName} mean it won't be visible in the {parentSpaceName}'s space members. Should we update the join rules to allow for {parentSpaceName}'s space member to see and join the {roomName}?"

app/lib/features/space/sheets/link_room_sheet.dart
```diff
@@ -349,6 +357,62 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
...
+              );
+            },
+          );
+          if (shouldChange) {
+            currentRooms.add(selectedParentSpaceId);
+
+            final sdk = await ref.read(sdkProvider.future);
+            final update = sdk.newJoinRuleBuilder();
+            update.joinRule(newRule);
+            for (final roomId in currentRooms) {
```
kumarpalsinh25, line 410: Do we need to update `newRule` for other parent's also?

gnunicorn:
> @gnunicorn but does this fix the sub-spaces not showing in WeRise for kumar? All except one or two I created directly within the WeRise space, but apparently still doesnt show

I sure hope so, but from my investigation I can't tell what setup the subspaces have and thus am unclear if this is the problem. Either way you'd have to unlink and link the spaces again (in an app with this change) before the change would happen to any existing space setup.

Issue #1185, opened 2023-11-28
Problems with verifying sessions


emilvincentz:
Problems on IOS (latest nightly 28.11.23): 

- [ ] camera verification doesnt work
- [ ] yes and no icons only shows "t"
- [ ] When succesfully paired, it shows "verification was NOT done" although it actually was.

![Image](https://github.com/acterglobal/a3/assets/77483554/1fff88ea-f09c-4922-8c99-316ceed2687d)
![Image](https://github.com/acterglobal/a3/assets/77483554/47717adc-4762-4650-a19c-15da59660f0e)

gnunicorn:
on the last step, for me trying to verify latest main with element, element tells me that verification was cancelled from the other side ... I think there is something wrong in the verification flow ...

PR #1202, opened 2023-12-10
Fix bug in early calling of SasVerification::is_done()

bitfriend:
refs https://github.com/acterglobal/a3/issues/1185


bitfriend:
As Kevin said, we are supposed to use to-device events to verify our other devices, and sync events to verify other users.

gtalha07 (approved):
LGTM!

gnunicorn (approved):
but I see at least two more bugs in the initial report this doesn't fix. I'll change the description.

bitfriend:
> but I see at least two more bugs in the initial report this doesn't fix. I'll change the description.

Yeah, it was my bad.

PR #1204, opened 2023-12-11
ReadPrivate type & receipt thread

bitfriend:
- Append ReadPrivate type to receipt handling
- Expose receipt thread to flutter


gnunicorn (approved):
code looks fine, though I don't understand what this does or how that is meant to work for the user ...

bitfriend:
I just wanted to implement the missed part that may be used by flutter someday.

PR #1208, opened 2023-12-11
[Updates]: Reactions implementation

gtalha07:
Done:

https://github.com/acterglobal/a3/assets/40496/321d58e5-8552-46f0-97d8-9e4e26f25209


native/test/src/tests/news.rs
```diff
@@ -304,3 +304,157 @@ async fn news_png_image_with_text_test() -> Result<()> {
...
+        }
+    })
+    .await?;
+
+    let slides = space.latest_news_entries(1).await?;
+    let final_entry = slides.first().expect("Item is there");
+    let reaction_manager = final_entry.reactions().await?;
+    let entry_evt_id = final_entry.event_id().to_string();
+    reaction_manager
+        .redact_reaction(entry_evt_id, "".to_string(), "".to_string())
```
bitfriend, line 436: We have to send reaction prior to redact it.

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -47,13 +49,40 @@ class NewsSideBar extends ConsumerWidget {
...
+                ? reactions.valueOrNull!.length.toString()
+                : '0',
+            style: style,
+            color: fgColor,
+            index: index,
+            onTap: () async {
+              final client = ref.read(alwaysClientProvider);
+              final status = await news.myLikeStatus();
+              final manager = await news.reactions();
+              debugPrint('STATUS: $status');
```
gnunicorn, line 69:
```suggestion
```

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -47,13 +49,40 @@ class NewsSideBar extends ConsumerWidget {
...
-          likeCount: news.likesCount().toString(),
-          style: style,
-          color: fgColor,
-          index: index,
+        SizedBox(
+          height: 75,
+          width: 75,
+          child: LikeButton(
+            isLiked: isLikedByMe.valueOrNull ?? false,
+            likeCount: reactions.valueOrNull != null
```
gnunicorn, line 59: `likesCount` this should probably not be a `String` but an `int`, or ... it is a `count` afterall...

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -47,13 +49,40 @@ class NewsSideBar extends ConsumerWidget {
...
+                );
+
+                await client.waitForReaction(eventId.toString(), null);
+              } else {
+                final reactions = await manager
+                    .reactionEntries()
+                    .then((value) => value.toList());
+                final eventId = reactions[0].eventIdStr();
+                await manager.redactReaction(eventId, '', '');
+              }
```
gnunicorn, line 83: the manager is specific to this news item ... why does the API user have to figure out if there was `like` by me and additionally look up the specific entry and redact that... This is all code that should be _in_ the rust SDK API not here...

native/acter/api.rsh
```diff
@@ -481,6 +491,8 @@ object CalendarEvent {
     fn update_builder() -> Result<CalendarEventUpdateBuilder>;
     /// get RSVP manager
     fn rsvp_manager() -> Future<Result<RsvpManager>>;
+    /// get the reaction manager
+    fn reaction_manager() -> Future<Result<ReactionManager>>;
```
gnunicorn, line 495: we should either go for `reactions` (like on `News`) or `reactions_manager` (like here) and use that consistently. I'd propose to drop `_manager` and just go for `reactions`.

native/acter/api.rsh
```diff
@@ -592,6 +604,52 @@ object Rsvp {
...
+//  ## ##    ## ##    ##  ##  ##         ##        ##    ##   ##   ## ##    #####   
+//  ## ##    ##       ## ###  ##         ##        ##    ##   ##   ##  ##      ###  
+//  ##  ##   ##  ##   ##  ##  ##   ##    ##        ##    ##   ##   ##  ##  ##   ##  
+// #### ##  ### ###  ###  ##   ## ##    ####      ####    ## ##   ###  ##   ## ## 
+
+
+object ReactionManager {
+
+    /// get my reaction status
+    fn my_status() -> Future<Result<bool>>;
```
gnunicorn, line 619: this is a bit vague. What does `my_status = true` mean? I assume that is meant to imply the user has reacted? then let's call it that `have_i_reacted`...

native/acter/api.rsh
```diff
@@ -592,6 +604,52 @@ object Rsvp {
...
+    fn total_reaction_count() -> u32;
+
+    /// get reaction entries
+    fn reaction_entries() -> Future<Result<Vec<Reaction>>>;
+
+    /// send the reaction
+    fn send_reaction(event_id: string, key: string) -> Future<Result<EventId>>;
+
+    /// redact the reaction
+    fn redact_reaction(event_id: string, reason: Option<string>, txn_id: Option<string>) -> Future<Result<EventId>>;
```
gnunicorn, line 634: why do we have `event_id`s here? These should be know to the manager, no?

native/acter/src/api/news.rs
```diff
@@ -342,6 +347,21 @@ impl NewsEntry {
...
+
+        RUNTIME
+            .spawn(async move {
+                let inner =
+                    models::ReactionManager::from_store_and_event_id(client.store(), &event_id)
+                        .await;
+                Ok(crate::ReactionManager::new(client, room, inner))
+            })
+            .await?
+    }
```
gnunicorn, line 363: this code appears to be the same here and for the `calendar_events`... maybe we can put that onto a helper onto `create::ReactionManager` that given `client, room, event_id` does the entire thing?

native/acter/src/api/news.rs
```diff
@@ -381,12 +401,37 @@ impl NewsEntry {
...
+    pub async fn comments_count(&self) -> Result<u32> {
+        let me = self.clone();
+
+        RUNTIME
+            .spawn(async move {
+                let manager = me.comments().await?;
+                Ok(manager.stats().total_comments_count().to_owned())
+            })
+            .await?
+    }
```
gnunicorn, line 413: if this triggers the build of `me.comments()` (which is then used on the UI afterwards anyways), the this shouldn't be here but instead only be a helper on the comments_manager ... this avoids us creating that internal thing and loading from the database twice ...

native/acter/src/api/news.rs
```diff
@@ -381,12 +401,37 @@ impl NewsEntry {
...
-        19
+    pub async fn my_like_status(&self) -> Result<bool> {
+        let me = self.clone();
+
+        RUNTIME
+            .spawn(async move {
+                let manager = me.reactions().await?;
+                manager.my_status().await
+            })
+            .await?
```
gnunicorn, line 434: The comment about `comments_count` applies here equally. Move all of that onto the ReactionsManager to avoid loading it twice.

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+        let my_id = client.user_id().context("User not found")?.to_owned();
+        let event_id = EventId::parse(&event_id)?;
+
+        RUNTIME
+            .spawn(async move {
+                let member = room
+                    .get_member(&my_id)
+                    .await?
+                    .context("Couldn't find me among room members")?;
+                if !member.can_send_message(MessageLikeEventType::RoomMessage) {
```
gnunicorn, line 113: You sure this is the right check? Shouldn't we check for Annotations? I think they have a different permission-level ...

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub fn relates_to(&self) -> String {
+        self.inner.relates_to.event_id.to_string()
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ReactionManager {
+    client: Client,
+    room: Room,
+    inner: models::ReactionManager,
```
gnunicorn, line 78: this should store the `event_id`, too.

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+        inner: models::ReactionManager,
+    ) -> ReactionManager {
+        ReactionManager {
+            client,
+            room,
+            inner,
+        }
+    }
+
+    pub async fn send_reaction(&self, event_id: String, key: String) -> Result<OwnedEventId> {
```
gnunicorn, line 101: let's us avoid posting the `event_id` here.

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+            })
+            .await?
+    }
+
+    pub async fn redact_reaction(
+        &self,
+        event_id: String,
+        reason: Option<String>,
+        txn_id: Option<String>,
+    ) -> Result<OwnedEventId> {
```
gnunicorn, line 132: this should figure out our own redaction without the user providing the `event_id`. The code that currently happens inline in the flutter side, should happen here.

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+            .await?
+    }
+
+    pub fn stats(&self) -> models::ReactionStats {
+        self.inner.stats().clone()
+    }
+
+    pub async fn my_status(&self) -> Result<bool> {
+        let manager = self.inner.clone();
+        let my_id = self.client.user_id().context("User not found")?;
```
gnunicorn, line 168: considering how often the user_id is needed, we might want to consider loading that once at creation time and just store it in the struct, too...

native/acter/src/api/reactions.rs
```diff
@@ -0,0 +1,218 @@
...
+                Ok(response.event_id)
+            })
+            .await?
+    }
+
+    pub fn stats(&self) -> models::ReactionStats {
+        self.inner.stats().clone()
+    }
+
+    pub async fn my_status(&self) -> Result<bool> {
```
gnunicorn, line 166: this entire helper sounds like something that is more efficient happening on the core-manager directly...

native/acter/src/testing.rs
```diff
@@ -112,7 +112,7 @@ pub async fn ensure_user(
     };
 
     if let Some(token) = reg_token {
-        info!("Login for {username} failed: {e}. Trying to register with token instead.");
+        error!("Login for {username} failed: {e}. Trying to register with token instead.");
```
gnunicorn, line 115: this isn't an error-level but supported behaviour.
```suggestion
        info!("Login for {username} failed: {e}. Trying to register with token instead.");
```

native/acter/src/testing.rs
```diff
@@ -123,7 +123,7 @@ pub async fn ensure_user(
         )
         .await
     } else {
-        info!("Login for {username} failed: {e}. Trying to register instead.");
+        error!("Login for {username} failed: {e}. Trying to register instead.");
```
gnunicorn, line 126: same as above.
```suggestion
        info!("Login for {username} failed: {e}. Trying to register instead.");
```

native/core/src/models.rs
```diff
@@ -59,10 +65,12 @@ use crate::{
...
 pub enum Capability {
-    // someone can comment on this
+    // someone can add reaction on this
+    Reactable,
+    // someone can add comment on this
     Commentable,
-    // someone can add attchments on this
-    HasAttachments,
+    // someone can add attachment on this
+    Attachable,
```
gnunicorn, line 73: attachable means something else though ... like you can _attach this to_ something and that this can have attachments... that's why we had the different name ...

how about ...
```suggestion
    Attachmentable,
```

native/core/src/models.rs
```diff
@@ -501,6 +512,37 @@ impl TryFrom<AnyActerEvent> for AnyActerModel {
...
+                        origin_server_ts: r.origin_server_ts,
+                    },
+                    reason: r.unsigned.redacted_because,
+                }),
+            },
+
+            AnyActerEvent::Redaction(e) => {
+                info!("AnyActerEvent::Redaction - {:?}", e);
+                match e {
+                    RoomRedactionEvent::Original(m) => Ok(AnyActerModel::Redaction(m.into())),
```
gnunicorn, line 533: Okay, now I am confused. what is that RedactionEvent if we already have a `MessageLikeEvent::Redacted`? Why is that even needed?

native/core/src/models.rs
```diff
@@ -270,6 +278,9 @@ pub enum AnyActerModel {
     AttachmentUpdate(AttachmentUpdate),
 
     Rsvp(Rsvp),
+    Reaction(Reaction),
+
+    Redaction(Redaction),
```
gnunicorn, line 283: this doesn't make sense to me. See below. a "Redaction" isn't meant to be a special type of event ...

native/core/src/models/redaction.rs
gnunicorn, line 1: This doesn't make any sense to me. `redactions` already exist in the matrix spec and our models and should be handled as a parameter/type change _on_ the actual object, applied directly by the `executor.redact`-function (which also applies that redaction to the underlying model) ... I don't understand why we are wrapping that redaction into a model now ... what is going on?

gnunicorn (changes requested):
Great work, @gtalha07 & @bitfriend :tada: !!!

I have a few questions regarding the API design and where what code should live and would ask that we change that to be more coherent and intuitive.

On a second part I do not understand what the need of the new `RedactionModel` is (we already have one) and why that entire part of the infrastructure is needed now rather than updating the existing path so that it works for that new use case ... please explain that or refactor it accordingly.

native/core/src/models/redaction.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/d50b2841f52227a6a85d51b8dc2d0fe16464a7f4

native/acter/api.rsh
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/cdd64defa739398635677d93e9a0efffa3df78ba

app/lib/features/news/widgets/news_side_bar.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/f186f3f9f7921396db40aaa16d87394e68384528

app/lib/common/widgets/like_button.dart
```diff
@@ -125,6 +125,7 @@ class _LikeButtonState extends State<LikeButton> with TickerProviderStateMixin {
                 children: [
                   InkWell(
                     onTap: () {
+                      debugPrint('like click --------------------------------');
                       if (widget.onTap != null) {
                         widget.onTap!();
                       }
```
gnunicorn, line 131:
```suggestion
                    onTap: widget.onTap
```

Is there ever a realistic chance we have no onTap at all? Maybe we should make that field required, too...

app/lib/features/news/widgets/news_side_bar.dart
gnunicorn: this should probably be changed to:
```suggestion
            likeCount: reactions.valueOrNull?.length ?? 0
```

gnunicorn (commented):
this is still in the works, right?!?

app/lib/features/news/widgets/news_side_bar.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/4286329716aadd590bbbb1306d9331d4e3934b6c

app/lib/features/news/widgets/news_side_bar.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/38a5cbc2d7976cf2464c65823f99898495a46390

native/acter/api.rsh
bitfriend: Yeah, it is possible to remove `event_id` from api param in `send_reaction()`.
We can use event id of news without passing of event id as fn param.
But for `redact_reaction()`, we have to use reacted event id, not news event id.
So we can't remove `event_id` from api param in `redact_reaction()`.

native/acter/src/api/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/6fcfe4b08f1441b3cc86194fade3d1d79cc2c7e9

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -47,13 +50,44 @@ class NewsSideBar extends ConsumerWidget {
...
+                index: index,
+                onTap: () async {
+                  final client = ref.read(alwaysClientProvider);
+                  final status = await manager.myStatus();
+                  debugPrint('STATUS: $status');
+                  if (!status) {
+                    final eventId = await manager.sendReaction(
+                      news.eventId().toString(),
+                      '\u{dd25}',
+                    );
```
gnunicorn: I'd suggest a rust-method "sendLike" and "sendUnlike" which deal with the details internally in rust. Also, again, this API should probably not require `news.eventId()` at all, considering that it is derived from `news` and thus should implicitly know that context ...

bitfriend: When we want `sendLike` and `sendUnlike`, we should use `U+1F44D` and `U+1F44E` as symbol?

bitfriend: And we should keep both `reaction_count()` and `like_count()`?

gnunicorn: we are internally just falling back [the ruma `Annotation`](https://docs.ruma.io/ruma/events/relation/struct.Annotation.html), right? Which just uses some `String`. So, yeah, let's agree that we interpret the unicode code points for the [Red Hearts emoji](https://emojipedia.org/red-heart) (`U+2764`&`U+FE0F` together) as our Acter-Like. Makes it somewhat compatible with existing Matrix and we can expand the counter of what else we consider a like later as well.

I recommend to put that as a static str somewhere then reference in the rust-code-base @bitfriend .

native/acter/src/api/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/e5e7c6a86780009aeeba43cbbfa5d677891a4fd9

native/acter/src/api/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/e5e7c6a86780009aeeba43cbbfa5d677891a4fd9

native/acter/src/api/reactions.rs
bitfriend: `event_id` can be known by `inner.event_id()`.

native/acter/src/api/reactions.rs
bitfriend: I thought it is correct to check `MessageLikeEventType::RoomMessage` because `room.send()` is called below that.
Anyway fixed [here](https://github.com/acterglobal/a3/pull/1208/commits/ec6cbeba3c242b3aabfdb3475e3b8b9d0279a988).

native/acter/src/api/reactions.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/26b6ec8159c04106454ba0884f67a02c6ae713fc

native/acter/src/api/reactions.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/87f23b9832180a2405fea636bf58679fdcfe70f2

native/acter/src/api/reactions.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/ad5a88076fe8b1bf80bb23dccedea00ee70575fb

native/acter/src/testing.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/b7b8a9b1a3f7b414661e413d687da13809d4bd9b

native/acter/src/testing.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/b7b8a9b1a3f7b414661e413d687da13809d4bd9b

native/core/src/models.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/e2b3ca95232d7fedd48a158f1c42657de5d9e794

native/core/src/models.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/d50b2841f52227a6a85d51b8dc2d0fe16464a7f4

native/core/src/models.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/d50b2841f52227a6a85d51b8dc2d0fe16464a7f4

app/lib/common/widgets/like_button.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/2e8359d97d1a136895ad8ed60567bf368815bc10

app/lib/features/news/widgets/news_side_bar.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1208/commits/4a9a1127db3f96f86a98f464c2fe55b4337c3b4f

gnunicorn (approved):
there's quite a few unrelated changes in here, most notably around the rsvp manager, the comments and activating the reactions on objects that we don't support them on yet ...

I'll clean up a bit more, but overall it's fine now.

bitfriend:
I implemented unlike feature according to [your suggestion](https://github.com/acterglobal/a3/pull/1208#discussion_r1503966421).
But I found you deleted unlike feature [here](https://github.com/acterglobal/a3/pull/1208/commits/9ca4b70dbf36d0d0582de239a15d4c5852e5b423).
Unlike feature is not needed?
Otherwise I had a mistake in understanding of your suggestion?

bitfriend:
And `redact_reaction()` is not needed too?

bitfriend:
> I implemented unlike feature according to [your suggestion](https://github.com/acterglobal/a3/pull/1208#discussion_r1503966421). But I found you deleted unlike feature [here](https://github.com/acterglobal/a3/pull/1208/commits/9ca4b70dbf36d0d0582de239a15d4c5852e5b423). Unlike feature is not needed? Otherwise I had a mistake in understanding of your suggestion?

I guess I was confused between `unlike` and `dislike`.

gnunicorn:
@bitfriend yes, in this PR there was a lot of stuff that was _anticipated_ for some later version, but the baseline for what we need _now_ wasn't quite there. It seemed like a case of too early generalization. I removed the parts that were in the way for me to get the specific case to work now. Thus, this PR contains a few unrelated and somewhat half-working untested code.

Re like & unlike: I might have badly phrased that. It really meant that you should be able to _like_ and remove that same like you made, not a thumbs-down-button (as Facebook had it in its early days). So I removed that entire thing.

PR #1236, opened 2023-12-18
Remove unwrap

bitfriend:
Will replace `unwrap()` with `expect()` or `context()`.

### DIFF
--- a/app/packages/rust_sdk/lib/acter_flutter_sdk_ffi.dart
+++ b/app/packages/rust_sdk/lib/acter_flutter_sdk_ffi.dart
@@ -12407,6 +12407,37 @@ class Api {
           int Function(
             int,
           )>();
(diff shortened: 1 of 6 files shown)


gnunicorn (changes requested):
None of these changes should be made! None of them is an `unwrap`. `?` is not an `unwrap`, it doesn't `panic` but returns. all `?` does is unwrap a `Ok` or immediately return if the value is an `Err` - and it automatically `.into()` as well, which for most cases is the best we want as we use `anyhow`. `.contexr` should only ever be used if you have a `Option` but want to immediately return upon `None` with a custom message. Using `.context` on a `Result` is almost always wrong, as it _hides_ the actual traceback as it starts a new one at that location - which makes it complicated if that is an inner problem, e.g. a http-timeout as it won't surface the actual error.

PR #1238, opened 2023-12-19
New vagrant based test-server setup incl docs

gnunicorn:
This moves the `docker-compose` and its local info into the `util/test_server`-subfolder, adds a vagrant setup for non-linux-systems with nice `cargo make`-commands for easy usage and docs explaining how to use it. All so that creating and running end-2-end UI tests a lot easier for all desktop development systems.

Additionally contains fixes to make the tests work again:
 - upgrade dependencies for flutter 3.16
 - small refactor in the logout procedures to remove a nasty exception on `alwaysClientProvider` upon `logout`
 - added missing `await` in test cases
 - fix shake to only attempt to register on actual devices - throws exception on simulators


kumarpalsinh25:
@gnunicorn 
I don't have much idea about `docker` and `test-server`. So I would suggest that someone else who knows this can review this PR.

gnunicorn:
@kumarpalsinh25 no one here does ;) . Best way to review is:
1. check out the branch locally
2. follow the new docs added to try it 
3. confirm it works for you, too.

kumarpalsinh25:
> @kumarpalsinh25 no one here does ;) . Best way to review is:
> 
> check out the branch locally
> follow the new docs added to try it
> confirm it works for you, too.

Ahh, I see.
No worries. I will take checkout of the [ben-vagrant-test-setup](https://github.com/acterglobal/a3/tree/ben-vagrant-test-setup) branch and check the details.

kumarpalsinh25:
@gnunicorn 
I'm getting below error when I run `cargo make test-server`.

```
No usable default provider could be found for your system.

Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.
```

Below steps I have performed.
1) Installed Vagrant Virtual Machine on my mac system. (Ref : https://developer.hashicorp.com/vagrant/install)
2) Taken a pull from [ben-vagrant-test-setup](https://github.com/acterglobal/a3/tree/ben-vagrant-test-setup) branch
3) Run `cargo make test-server` at the root directory of the project

gnunicorn:
@kumarpalsinh25 you didn't finish the vagrant setup. From the docs:
> and a corresponding provider (probably [VirtualBox](https://www.virtualbox.org/wiki/Downloads)).

gnunicorn:
@kumarpalsinh25 you could check if another vagrant provider is more useful. or use a [virtualbox test build](https://www.virtualbox.org/wiki/Testbuilds) which has support for the macos ARM chips.

kumarpalsinh25:
@gnunicorn 
Any idea about below errors?

Screenshot 2024-01-01 at 12 57 24 PM

gnunicorn:
@kumarpalsinh25 no idea. looks like a problem with the virtual box setup. can you start the virtual box instance manually (virtual box comes with a UI)? Does the VM actually start up and run?

Makefile.toml
```diff
@@ -628,3 +628,82 @@ script = [
...
+    if eq ${os} "linux"
+        exec docker-compose up -d
+    else
+        exec vagrant up --provision
+    endif
+    """
+]
+
+[tasks.ui-test-manager]
+private = true
```
gnunicorn, line 658: make this public

```suggestion
```

gtalha07 (approved):
The code part looks good but still based on my research, the VirtualBox test builds aren't ready to boot up VM on Apple silicon based chips, I constantly see this error while try to run ```cargo test-server```. 

```
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Connection reset. Retrying...
The guest machine entered an invalid state while waiting for it
to boot. Valid states are 'starting, running'. The machine is in the
'aborted' state. Please verify everything is configured
properly and try again.

If the provider you're using has a GUI that comes with it,
it is often helpful to open that and watch the machine, since the
GUI often has more helpful error messages than Vagrant can retrieve.
For example, if you're using VirtualBox, run `vagrant up` while the
VirtualBox GUI is open.

The primary issue for this error is that the provider you're using
is not properly configured. This is very rarely a Vagrant issue.
```

Makefile.toml
gtalha07: Unfortunately this still doesn't resolve the error I am facing....

gnunicorn:
Is there any other vagrant supported vm provider that has apple silicone support?

gtalha07:
> Is there any other vagrant supported vm provider that has apple silicone support?

[VMware](https://www.vmware.com/products/fusion/fusion-evaluation.html)seems to be able to run and even perform better on silicon based chips but comes with free trial. One another provider I have installed is UTM which has support for Apple silicon chips but discussion is still ongoing here when vagrant will adopt it: https://github.com/hashicorp/vagrant/issues/12518

gnunicorn:
Well, if the vagrant setup doesn't work for either of our Mac people, we might want to discard this approach.

PR #1245, opened 2023-12-21
Remove unwrap

bitfriend:
Will replace `unwrap()` with `expect()` or `context()`.

### DIFF
--- a/native/acter/src/api/stream.rs
+++ b/native/acter/src/api/stream.rs
@@ -412,7 +412,7 @@ impl TimelineStream {
 
     pub async fn retry_send(&self, txn_id: String) -> Result<bool> {
         let timeline = self.timeline.clone();
-        let transaction_id = OwnedTransactionId::from(txn_id);
+        let txn_id = OwnedTransactionId::try_from(txn_id)?;
 
         let room = self.room.clone();
         let my_id = room
@@ -431,7 +431,7 @@ impl TimelineStream {
                     bail!("No permission to send message in this room");
                 }
 
-                timeline.retry_send(&transaction_id).await?;
+                timeline.retry_send(&txn_id).await?;
                 Ok(true)
             })
             .await?
@@ -439,7 +439,7 @@ impl TimelineStream {
 
     pub async fn cancel_send(&self, txn_id: String) -> Result<bool> {
         let timeline = self.timeline.clone();
-        let transaction_id = OwnedTransactionId::from(txn_id);
+        let txn_id = OwnedTransactionId::try_from(txn_id)?;
 
         let room = self.room.clone();
         let my_id = room
@@ -458,7 +458,7 @@ impl TimelineStream {
                     bail!("No permission to send message in this room");
                 }
 
-                timeline.cancel_send(&transaction_id).await;
+                timeline.cancel_send(&txn_id).await;
                 Ok(true)
             })
             .await?
--- a/native/acter/src/api/stream.rs
+++ b/native/acter/src/api/stream.rs
@@ -286,9 +286,10 @@ impl TimelineStream {
                     bail!("No permission to send message in this room");
                 }
 
-                let Some(reply_item) = timeline.item_by_event_id(&event_id).await else {
-                    bail!("Not found which item would be replied to")
-                };
+                let reply_item = timeline
+                    .item_by_event_id(&event_id)
+                    .await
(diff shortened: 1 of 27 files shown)


native/acter/src/api/stream.rs
```diff
@@ -411,7 +412,7 @@ impl TimelineStream {
 
     pub async fn retry_send(&self, txn_id: String) -> Result<bool> {
         let timeline = self.timeline.clone();
-        let transaction_id = OwnedTransactionId::from(txn_id);
```
gnunicorn, line 414: this one I don't understand. Why replace a never failing `from` with a sometimes failing `try_from(...)?`?

native/acter/src/api/stream.rs
```diff
@@ -430,15 +431,15 @@ impl TimelineStream {
...
-                timeline.retry_send(&transaction_id).await;
+                timeline.retry_send(&txn_id).await?;
                 Ok(true)
             })
             .await?
     }
 
     pub async fn cancel_send(&self, txn_id: String) -> Result<bool> {
         let timeline = self.timeline.clone();
-        let transaction_id = OwnedTransactionId::from(txn_id);
```
gnunicorn, line 441: same as above...

native/acter/src/api/stream.rs
```diff
@@ -457,7 +458,7 @@ impl TimelineStream {
                     bail!("No permission to send message in this room");
                 }
 
-                timeline.cancel_send(&transaction_id).await;
```
gnunicorn, line 460: why does this not get a `?`?

native/acter/src/api/verification.rs
```diff
@@ -1066,7 +1070,7 @@ impl VerificationController {
                                 msg.set_content("short_authentication_string".to_string(), short_authentication_string.join(","));
                             }
                             StartMethod::ReciprocateV1(content) => {
-                                let secret = serde_json::to_string(&content.secret).unwrap();
+                                let secret = serde_json::to_string(&content.secret).expect("couldn't convert secret to string");
```
gnunicorn, line 1073: considering that this is in a callback, we probably don't want to panic here but instead do log if this doesn't work...
```suggestion
                            
let secret = match serde_json::to_string(&content.secret) {
    Ok(e) => e;
    Err(e) => {
       error!("Parsing RecipV1 secret failed", error);
       return;
    }
};
```

native/acter/src/api/verification.rs
```diff
@@ -1151,7 +1155,7 @@ impl VerificationController {
                             evt.sender,
                         );
                         msg.set_content("keys".to_string(), evt.content.keys.to_string());
-                        let mac = serde_json::to_string(&evt.content.mac).unwrap();
+                        let mac = serde_json::to_string(&evt.content.mac).expect("couldn't convert mac to string");
```
gnunicorn, line 1158: same logging as above instead of panic...

gnunicorn (approved):
Nice. I like replacing those `else bail` with `.context()` makes for smaller less expanded code blocks. Just one remark on an `expect` in a callback that we probally don't want to crash on if that failed.

native/acter/src/api/stream.rs
bitfriend: I just wanted to capture incorrect usage in development stage.

native/acter/src/api/stream.rs
bitfriend: `Timeline::cancel_send()` doesn't return `Result<>`.

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1245/commits/a183faf7571b2a83fa1b0d3e33d6a15d2fb4f03b

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1245/commits/a183faf7571b2a83fa1b0d3e33d6a15d2fb4f03b

native/acter/src/api/stream.rs
bitfriend: And I don't know why both `from()` and `try_from()` exist in `OwnedTransactionId`.
If failure is possible like `OwnedEventId` or `OwnedUserId`, `from()` should not exist.

PR #1248, opened 2023-12-22
context & expect

bitfriend:
- Will remove incorrect usage of `.context()?` for `Result<>`
- Will replace `.expect()` with `.context()?` for `Option<>`

### DIFF
--- a/native/acter/src/api/client.rs
+++ b/native/acter/src/api/client.rs
@@ -957,7 +957,7 @@ impl Client {
                     .encryption()
                     .get_device(&user_id, device_id!(dev_id.as_str()))
                     .await?
-                    .context("client should get device")
+                    .context("Unable to find device")
                     .map(|x| x.is_verified())
             })
             .await?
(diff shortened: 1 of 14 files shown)


native/acter/src/api/client.rs
```diff
@@ -951,16 +949,16 @@ impl Client {
...
         RUNTIME
             .spawn(async move {
-                let dev = c
+                client
                     .encryption()
                     .get_device(&user_id, device_id!(dev_id.as_str()))
                     .await?
-                    .context("client should get device")?;
-                Ok(dev.is_verified())
+                    .context("client should get device")
```
gnunicorn, line 960: We should make the error more understandable, I'd say.
```suggestion
                    .context("Unable to find device")
```

gnunicorn (approved):
Just a minor fix on an error message to be more comprehensible

native/acter/src/api/client.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1248/commits/dcabd9c2c0230aeb14a5336afa87f21c48051d79

PR #1252, opened 2023-12-28
Improve description for error

bitfriend:
Will give more clearer description for error case.


native/acter/src/api/room.rs
```diff
@@ -753,24 +753,24 @@ impl Room {
...
-            bail!("Can't set name to a room we are not in");
+            bail!("Unable to set name to a room we are not in");
         }
         let room = self.room.clone();
 
         let my_id = room
             .client()
             .user_id()
-            .context("User not found")?
+            .context("UserId not found")?
```
gnunicorn, line 763: This isn't more clear. `client.user_id()` should probably always have a context saying `context("You must be logged in to do that")`...

native/acter/src/api/room.rs
```diff
@@ -1002,7 +1002,7 @@ impl Room {
                 let event_content = evt.event.deserialize_as::<RoomMessageEvent>()?;
                 let original = event_content
                     .as_original()
-                    .expect("Couldn't get original msg");
+                    .expect("Original msg should be got from event content");
```
gnunicorn, line 1005: this is not more clear, leave the earlier. But also move it from `expect` to `context` probably...

native/acter/src/api/room.rs
```diff
@@ -1353,7 +1353,7 @@ impl Room {
                     },
                 };
                 let Some(request) = request else {
-                    // content info or thumbnail source doesn't exist
+                    warn!("Content info or thumbnail source not found");
```
gnunicorn, line 1356: :+1:

native/acter/src/api/room.rs
```diff
@@ -1412,7 +1412,7 @@ impl Room {
                 let event_content = evt.event.deserialize_as::<RoomMessageEvent>()?;
                 let original = event_content
                     .as_original()
-                    .expect("Couldn't get original msg");
+                    .expect("Original msg should be got from event content");
```
gnunicorn, line 1415: same as above ... probably used a few more times...

native/acter/src/api/verification.rs
```diff
@@ -120,7 +120,11 @@ impl VerificationEvent {
                         let sequence = items
                             .iter()
                             .map(|e| {
-                                let chr = e.symbol.chars().next().expect("first char not found");
+                                let chr = e
+                                    .symbol
+                                    .chars()
+                                    .next()
+                                    .expect("At least one character should be present");
```
gnunicorn, line 127: we are confident this crashing the app is a good choice?!? Shouldn't we instead have a `filter_map(` here and instead return `None` in cases we can't read the char? Probably return a broken emoji list and the user can then cancel the process..

native/acter/src/api/verification.rs
```diff
@@ -139,7 +143,11 @@ impl VerificationEvent {
...
```
gnunicorn, line 150: same.

native/acter/src/api/verification.rs
```diff
@@ -517,20 +525,20 @@ async fn request_verification_handler(
         request
             .accept_with_methods(methods)
             .await
-            .expect("Can't accept verification request");
+            .expect("Unable to accept verification request with methods");
```
gnunicorn, line 528: there is any type of reasons why this might fail - think network issues - so this should _not_ use `expect` but probably instead return an `Err` in those cases.

native/acter/src/api/verification.rs
```diff
@@ -517,20 +525,20 @@ async fn request_verification_handler(
...
     } else {
         request
             .accept()
             .await
-            .expect("Can't accept verification request");
+            .expect("Unable to accept verification request without methods");
```
gnunicorn, line 533: same as above.

native/acter/src/api/verification.rs
```diff
@@ -720,7 +728,11 @@ async fn sas_verification_handler(
                         .emojis
                         .iter()
                         .map(|e| {
-                            let chr = e.symbol.chars().next().expect("first char not found");
+                            let chr = e
```
gnunicorn, line 731: same as above. I wonder if we don't want to put that into a singular location and only call off it... like have a `VerificatonEmoji::from(emoji)?` or something...

native/core/src/spaces.rs
```diff
@@ -210,16 +210,16 @@ impl CoreClient {
                 // local uri
                 let path = PathBuf::from(avatar_uri);
                 let guess = mime_guess::from_path(path.clone());
-                let content_type = guess.first().expect("MIME type should be given");
-                let buf = std::fs::read(path).expect("File should be read");
-                let upload_resp = client.media().upload(&content_type, buf).await?;
+                let content_type = guess.first().expect("don't know mime type");
+                let buf = std::fs::read(path).expect("file reading failed");
```
gnunicorn, line 214: let's switch those to `.context()?` pls.

gnunicorn (changes requested):
By doing that mass-replacing you pointed me to various places that we use `expect` but should better use `context` and some other code improvements, that I think we should do now, too ...

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/b8628e84f7cc4497b318ff0e28569c4ff937d00f

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/8d49b591e076f69a6611b094d2dbe73e0ca1db8e

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/8d49b591e076f69a6611b094d2dbe73e0ca1db8e

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/577ad8847d232d969c8478ff9b59311427486f3b

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/577ad8847d232d969c8478ff9b59311427486f3b

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/7d8760ae4ea5b2873de5d72ae7eed146743f318a

native/core/src/spaces.rs
bitfriend: `CoreClient::create_acter_space()` returns `std::result::Result` not `anyhow::Result`.
So we can't use `anyhow::Context` here.

native/acter/src/api/verification.rs
bitfriend: `request_verification_handler()` is callback fn that is called using `tokio::spawn()`.
That is, `tokio::spawn()` is called in `RUNTIME.spawn()` and we can't wait for `tokio::spawn()` to be finished.
It means we needn't to receive return value from `request_verification_handler()`.
Then `request_verification_handler()` needn't to return `Result<>` and we needn't return `Err` for error case of `VerificationRequest::accept_with_methods()`.

bitfriend: Appended error handling to verification callback.
https://github.com/acterglobal/a3/pull/1252/commits/0c03a76369d3c19272ec69587bd2b0663fe3e6f8
But I wonder how `tokio::spawn()` will handle the return value of verification callback.

native/core/src/spaces.rs
gnunicorn: but `std::fs::read(path)` returns a `std::result::Result<..>` ... why are we not just doing a `?` here?

native/acter/src/api/verification.rs
```diff
@@ -119,17 +120,7 @@ impl VerificationEvent {
...
-                                    .symbol
-                                    .chars()
-                                    .next()
-                                    .expect("At least one character should be present");
-                                VerificationEmoji {
-                                    symbol: chr as u32,
-                                    description: e.description.to_string(),
-                                }
-                            })
+                            .filter_map(VerificationEmoji::new)
```
gnunicorn, line 123: love it!

native/acter/src/api/verification.rs
```diff
@@ -516,21 +505,23 @@ async fn request_verification_handler(
...
     if let Some(methods) = methods {
-        request
-            .accept_with_methods(methods)
-            .await
-            .expect("Unable to accept verification request with methods");
+        let result = request.accept_with_methods(methods).await;
+        if let Err(e) = result {
+            error!("Unable to accept verification request with methods");
+            return Err(e.into());
+        }
```
gnunicorn, line 518: why can't we just do ...
```suggestion
        let result = request.accept_with_methods(methods).await?;
```

native/acter/src/api/verification.rs
```diff
@@ -727,25 +721,15 @@ async fn sas_verification_handler(
...
-                            VerificationEmoji {
-                                symbol: chr as u32,
-                                description: e.description.to_string(),
-                            }
-                        })
+                        .filter_map(VerificationEmoji::new)
                         .collect::<Vec<VerificationEmoji>>();
                     msg.set_emojis(sequence);
                 }
                 let value = match serde_json::to_string(&decimals) {
```
gnunicorn, line 728: same question for just using `?` here..?

native/acter/src/api/verification.rs
```diff
@@ -699,13 +691,15 @@ async fn sas_verification_handler(
...
+) -> Result<()> {
     info!(
         "Starting verification with {} {}",
         &sas.other_device().user_id(),
         &sas.other_device().device_id()
     );
-    sas.accept().await.unwrap();
+    if let Err(e) = sas.accept().await {
+        return Err(e.into());
+    }
```
gnunicorn, line 702: Why not just ..

```suggestion
    sas.accept().await?;
```

gnunicorn (approved):

bitfriend:
`std::fs::read()` returns `std::io::Result<>`.
Now `acter_core::error::Error` doesn't support `std::io::Error`, so I appended `IoError` there.
As we know, `CoreClient::create_acter_space()` returns `acter_core::error::Result` not `anyhow::Result`.
`acter_core::error::Result` handles error case with `acter_core::error::Error`.

native/core/src/spaces.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/8d12c22301b0fc4e33875dd74b4305d8ccbcaed4

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1252/commits/f159e9e35eec03c6d91ecaff88d5b45562b67edc

PR #1256, opened 2024-01-03
Fix bug in reaction test

bitfriend:
Will fix bug that `Timeline::toggle_reaction()` failed.
image


bitfriend:
The reason why `rfind_event_by_id()` failed is that I monitored sisko's events.
I wanted to toggle reaction of kyra or worf, but I monitored only sisko's event id.
That was my bug.

Issue #1244, opened 2023-12-21
Remove stateful `currentConvoProvider`


gnunicorn:
We have that stateful `currentConvoProvider` in our `convo_providers` that provides a stateful, yet often unobserved global static item. Especially considering many providers being async, there is the chance that we attempt to do things while the `currentConvoProvider` has changed under us and thus fail to properly load the information. Rather than reading some global state and keeping the fingers crossed providers should have all the required types and data as inputs and thus rely on `ref.watch`/the ref-watch-hierarchy tree to manage dependencies. Getting rid of the global `currentConvoProvider` is important step towards actually properly manage dependencies.

PR #1258, opened 2024-01-04
fix: #1244 remove ```currentConvoProvider```

gtalha07:
### PR Changes:
This essentially removes ```currentConvoProvider``` from our state system and instead goes with the approach of providing data to the riverpod convo dependencies through the use of ```typedef``` feature available in dart 3 which serves as alias for combining different type objects, thereby making it easier to pass as argument to family providers .i.e. they are limited to take only one type argument. To summarize the changes:

1. ```currentConvoProvider``` is removed whose state is ambiguous and instead pass ```typedef``` which is a ```Record``` in dart term and takes multiple type objects.
2.  Make convo providers take required arguments in hierarchal manner instead of watching global convo variable.


app/lib/common/utils/utils.dart
```diff
@@ -290,3 +291,8 @@ const List<(RSVP, String)> rsvpOptions = <(RSVP, String)>[
   (RSVP.Maybe, 'Maybe'),
   (RSVP.Yes, 'Yes'),
 ];
+
```
gnunicorn, line 294: uhh... I like that. Very nice!

app/macos/Podfile.lock
```diff
@@ -149,4 +149,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce7dbe26c78bfc7ba46736094c1e2d25982870fa
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.12.1
```
gnunicorn, line 152: we really like messing with this file ...

gnunicorn (approved):
Thanks heaps. This is great. Just some minor inquiries and questions from my side.

kumarpalsinh25 (approved):
Much needed code refactor. Great work!

app/lib/common/utils/utils.dart
gtalha07: could place them in providers folder but looks cleaner here. Should move it?

app/macos/Podfile.lock
gtalha07: I am getting frustrated with this, the thing is it gets updated when you run it on your local for macos/ios platform. Can we add it in ```.gitignore``` list.

bitfriend (approved):
Great work

PR #1262, opened 2024-01-08
Repair the broken image path in dev doc

bitfriend:
Will restore broken images in dev doc.


gnunicorn (approved):
I assume this is fine.

PR #1265, opened 2024-01-09
[Small fix] Improve integration test on rust side

bitfriend:
- Rename test fn
- Update descriptions


bitfriend:
The username in random user is composed as `it-{prefix}-{uuid}`.
It is why https://github.com/acterglobal/a3/pull/1265/commits/1bf2a7a388c48dfc5f7b46327c72fc27575ebc73 was made.

gnunicorn (approved):
Don't see the need for this, but it also doesn't hurt..

bitfriend:
This patch will be used in https://github.com/acterglobal/a3/pull/1208

PR #1267, opened 2024-01-10
[Small fix] Implement timeline stream of space

bitfriend:
This feature will be used for [news reaction test](https://github.com/acterglobal/a3/pull/1208).

### DIFF
--- a/native/acter/src/api/spaces.rs
+++ b/native/acter/src/api/spaces.rs
@@ -26,6 +26,7 @@ use matrix_sdk::{
     room::{Messages, MessagesOptions, Room as SdkRoom},
     ruma::api::client::state::send_state_event,
 };
+use matrix_sdk_ui::timeline::RoomExt;
 use ruma_common::{
     directory::RoomTypeFilter, serde::Raw, OwnedRoomAliasId, OwnedRoomId, RoomAliasId, RoomId,
     RoomOrAliasId, ServerName,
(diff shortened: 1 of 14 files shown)


native/acter/src/api/spaces.rs
```diff
@@ -34,19 +35,20 @@ use ruma_events::{
...
-use crate::{Client, PublicSearchResult, Room, RUNTIME};
+use crate::{Client, PublicSearchResult, Room, TimelineStream, RUNTIME};
 
 use super::utils::{remap_for_diff, ApiVectorDiff};
 
 #[derive(Debug, Clone)]
 pub struct Space {
     pub client: Client,
     pub(crate) inner: Room,
+    timeline: Arc<Timeline>,
```
gnunicorn, line 51: Nope. Similar as for chat-room, we currently _don't_ ensure that `Space` is globally unique for the same `roomId`. It is created and re-created and cloned multiple times in the system and created ad-hoc. Storing a stateful object on it is not a suitable way to solve whatever problem this is meant to solve.

Implying that it was and there the `Timeline` would a unique and always alive object is not reasonable. Other from the fact that keeping a Timeline object around all the time for each space is too costly. We do not usually care about the _timeline_-based order for events in a space.

gnunicorn (changes requested):
Sorry, but no. This is a resource intensive and potentially hazardous change that isn't even explained for why it would be needed. We shouldn't do this.

native/acter/src/api/spaces.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1267/commits/029bdeae47415c616680404bdcb3e9454238fd1f

gnunicorn (approved):

Issue #1097, opened 2023-10-28
Roomid doesnt work when shared (its not clickable and also does not redirect to the chat)


emilvincentz:

gnunicorn:
Can you be a bit more precise where you are when you try to click it?

emilvincentz:
https://github.com/acterglobal/a3/assets/77483554/b8425c2e-1287-4913-baf0-7aa4bbf1e850

@gnunicorn

gnunicorn:
we should share the matrix-id-link rather than the ID and add support for clicking matrix-ids and show room previews.

PR #1268, opened 2024-01-11
Clickable roomId on message

kumarpalsinh25:
This PR fixes issue no - #1097.

- Feature Reference Video:
https://github.com/acterglobal/a3/assets/51526480/009763cf-6034-43fe-8552-b33bc64656a8

- Unit-test Reference Video:
https://github.com/acterglobal/a3/assets/51526480/c791f224-dcbe-4d24-856c-d51c97f68f8d

### DIFF
--- a/app/lib/features/chat/chat_utils/chat_utils.dart
+++ b/app/lib/features/chat/chat_utils/chat_utils.dart
@@ -51,3 +51,35 @@ UserMentionMessageData parseUserMentionMessage(
     displayName: displayName,
   );
 }
+
+String? getRoomIdFromLink(Uri uri) {
+  // Match regex for matrix room link
+  final urlRegexp = RegExp(
+    r'https://matrix\.to/#/(?<roomId>.+):(?<server>.+)+',
+    caseSensitive: false,
+  );
+  final matches = urlRegexp.firstMatch(uri.toString());
+
+  //Link is type of matrix room link
+  if (matches != null) {
+    final roomId = matches.namedGroup('roomId');
+    var server = matches.namedGroup('server');
+
+    //Check & remove if string contains "?via=<server> pattern"
+    server = server!.split('?via=').first;
+
+    //Create complete roomId with home server information
+    var roomIdWithServer = '$roomId:$server';
+
+    //For public groups - Replace encoded '%23' string with #
+    if (roomIdWithServer.startsWith('%23')) {
+      roomIdWithServer = roomIdWithServer.replaceAll('%23', '#');
+    }
+
+    //Return roomId
+    return roomIdWithServer;
+  }
+
+  //Link is other than matrix room link
+  return null;
+}
--- a/app/lib/features/chat/chat_utils/chat_utils.dart
+++ b/app/lib/features/chat/chat_utils/chat_utils.dart
@@ -1,4 +1,10 @@
-import 'package:html/dom.dart';
+import 'package:acter/common/providers/room_providers.dart';
+import 'package:acter/common/utils/rooms.dart';
+import 'package:acter/common/utils/routes.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+import 'package:html/dom.dart' as html;
 
 class UserMentionMessageData {
   String parsedMessage;
@@ -14,7 +20,7 @@ class UserMentionMessageData {
 
 UserMentionMessageData parseUserMentionMessage(
   String message,
-  Element aTagElement,
+  html.Element aTagElement,
 ) {
   String msg = message;
   String userName = '';
@@ -53,12 +59,14 @@ UserMentionMessageData parseUserMentionMessage(
 }
 
 String? getRoomIdFromLink(Uri uri) {
+  final link = Uri.decodeFull(uri.toString());
+
   // Match regex for matrix room link
   final urlRegexp = RegExp(
     r'https://matrix\.to/#/(?<roomId>.+):(?<server>.+)+',
     caseSensitive: false,
   );
-  final matches = urlRegexp.firstMatch(uri.toString());
+  final matches = urlRegexp.firstMatch(link);
 
   //Link is type of matrix room link
   if (matches != null) {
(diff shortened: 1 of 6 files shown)


app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -184,4 +186,38 @@ class _TextWidget extends ConsumerWidget {
       ],
     );
   }
+
+  Future<void> onLinkTap(Uri uri, BuildContext context, WidgetRef ref) async {
+    final roomId = getRoomIdFromLink(uri);
+
+    //If link is type of matrix room link
+    if (roomId != null) {
+      joinAndOrNavigateToChatRoom(context, ref, roomId);
```
gtalha07, line 195: So, clicking on link directly attempts to join to the room automatically, this doesn't look good. Ideally I would want some consensus for user here .i.e. showing confirmation or alert dialog prior to join room action.

app/test/features/chat/chat_message_test.dart
```diff
@@ -10,151 +10,239 @@ void main() {
...
-        msg = resultData.parsedMessage;
-      }
-
-      final messageDocument = parse(msg);
-      final messageBodyText = messageDocument.body?.text ?? '';
-      expect(messageBodyText, 'hello @acter2 @acter2');
-    });
-
-    test(
+  group(
```
gtalha07, line 13: Nice!!

gtalha07 (approved):
Great and useful changes overall, there's one change I'd like to have prior to merging!!

app/lib/features/chat/widgets/text_message_builder.dart
kumarpalsinh25: Updated code and reference video in the description area of the PR. :)

app/lib/features/chat/chat_utils/chat_utils.dart
```diff
@@ -51,3 +51,35 @@ UserMentionMessageData parseUserMentionMessage(
...
+    //Check & remove if string contains "?via=<server> pattern"
+    server = server!.split('?via=').first;
+
+    //Create complete roomId with home server information
+    var roomIdWithServer = '$roomId:$server';
+
+    //For public groups - Replace encoded '%23' string with #
+    if (roomIdWithServer.startsWith('%23')) {
+      roomIdWithServer = roomIdWithServer.replaceAll('%23', '#');
+    }
```
gnunicorn, line 77: I think you just want to use the [URI class](https://api.dart.dev/stable/1.20.1/dart-core/Uri-class.html) and its decoder functions here ... there could be more than just the starting character being encoded...

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -139,9 +139,10 @@ class RoomProfilePage extends ConsumerWidget {
                   tiles: [
                     SettingsTile(
                       onPressed: (ctx) {
+                        final serverName = roomId.split(':').last;
```
gnunicorn, line 142: This isn't really how `via` is meant to work. See the [matrix URI spec](https://spec.matrix.org/v1.9/appendices/#uris) for details. But ideally `via` tells you about (additional) servers that we know to be in the room as alternative ways to connect to the room (in particular if the homeserver isn't available). Just repeating the homeserver is a bit of a short-cut that doesn't in practice actually help doing the right resolution.

We can keep this for now, but should probably add a helper function on the rust side that compiles the actual correct information including `via`.

app/lib/features/chat/chat_utils/chat_utils.dart
```diff
@@ -51,3 +51,35 @@ UserMentionMessageData parseUserMentionMessage(
     displayName: displayName,
   );
 }
+
+String? getRoomIdFromLink(Uri uri) {
+  // Match regex for matrix room link
+  final urlRegexp = RegExp(
+    r'https://matrix\.to/#/(?<roomId>.+):(?<server>.+)+',
+    caseSensitive: false,
+  );
```
gnunicorn, line 60: `matrix://` is also a valid link format: https://spec.matrix.org/v1.9/appendices/#uris

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -56,7 +61,7 @@ class _TextMessageBuilderConsumerState
     //remove mx-reply tags.
     String parsedString = simplifyBody(widget.message.text);
     final urlRegexp = RegExp(
-      r'https://matrix\.to/#/@[A-Za-z0-9\-]+:[A-Za-z0-9\-]+\.[A-Za-z0-9\-]+',
+      r'https://matrix\.to/#/[@!#][A-Za-z0-9\-]+:[A-Za-z0-9\-]+\.[A-Za-z0-9\-]+',
```
gnunicorn, line 64: matrix:// is also a valid link we need to check for https://spec.matrix.org/v1.9/appendices/#uris

gnunicorn (changes requested):
There is another format missing, and while I can ignore the questionable way to create the `via`-param, we need to have support for linking to `spaces`, too!

app/lib/features/chat/pages/room_profile_page.dart
kumarpalsinh25: Ok got it. I will check for matrix specification and do required changes accordingly.

app/lib/features/chat/chat_utils/chat_utils.dart
kumarpalsinh25: How can I get link like `matrix://` from Acter or Element general?

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -184,4 +187,83 @@ class _TextWidget extends ConsumerWidget {
...
+    }
+
+    //If link is other than matrix room link
+    //Then open it on browser
+    else {
+      await openLink(uri.toString(), context);
+    }
+  }
+
+  void askToJoinGroup(
```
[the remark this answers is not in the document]
kumarpalsinh25: Understood, let me move this function to `chat_utils`

app/lib/features/chat/chat_utils/chat_utils.dart
gnunicorn: I am not sure which clients are sending those at the moment... the element-x iOS maybe?

kumarpalsinh25: No, I have check element web and element iOS. Both are sending `matrix.to` url.

And `matrix://` is not a link url,I believe It is `url schema` which is generally used in deep linking.

kumarpalsinh25: @gnunicorn 
- Is there any platform or way using with I can get `matrix://` and test it?
- If `matrix://` type of url are consider as url schema then still I have to handle condition for it?

gnunicorn: @kumarpalsinh25 there's a bunch of examples listed in the [spec](https://spec.matrix.org/v1.9/appendices/#matrix-uri-scheme):
image

gnunicorn: matrix.to is the old, `matrix:` is the new way of doing it.

kumarpalsinh25: > matrix.to is the old, matrix: is the new way of doing it.

@gnunicorn 
Sorry but I'm not agree with that.

Not sure but based on the my knowledge `matrix://` and `matrix.to` are different things and have their own purposes.

- `matrix://` is Matrix URI scheme which ideally suitable in deep-linking for direct intent calls
- `matrix.to` is Matrix Navigation URL and widely used by all the matrix platforms but general navigations.

For better clarification, I'm looking for any platform which is using `matrix://` (Matrix URI Scheme definition). 
If you have any idea about such matrix platforms then can you please share it so that I can look into the same and get detail understanding.

And if you don't mind then can we proceed with this PR as this PR is mainly for copying RoomID URL from profile page and user should able to handle it with click on that message like element and other platforms doing because `matrix://` is different thing and don't have any use-case at least in Acter App.

kumarpalsinh25: Thanks @gnunicorn for details understanding.
If `matrix://` is the new way then for navigating and deep-link then I would like to integrate that way in Acter and have support for both. 

@gnunicorn Do you have idea of any of the matrix platform which is using this new way?  I just wanted to confirm certain use-cases with that.

gnunicorn: >  Do you have idea of any of the matrix platform which is using this new way? I just wanted to confirm certain use-cases with that.

as said, if Element-X isn't using it, I wouldn't know if any does. [According to the original spec proposal](https://github.com/matrix-org/matrix-spec-proposals/pull/2312) fluffychat and nheko have support for it (don't know if they post use it primarily though).

FYI, when I say the "new way" of doing it ... the Spec was merged 2021 .... and looks like even Element doesn't use it primarily yet, so ....

kumarpalsinh25: Aha, I see..
Let me create create new ticket for `"matrix:-protocol-support"` and then we can move on with current `matrix.to` support.

PR #1275, opened 2024-01-15
Workaround to fix Android forgotten session.

### LINKED ISSUE #1194: Android Emulator: Frequent logout issue.
- App gets logout frequently on debug mode
- When we clicks on Acter app icon then also its gets logout and start from onboarding screen

gnunicorn:
It appears because of a bug in FlutterSecureStorage the first read goes into null. Just adding a fake read at the startup supposedly prevents that problem from occuring. Adding a section to do that.

And indeed it works here for the emulator...

Fixes #1194


kumarpalsinh25 (approved):
Big relief on day-to-day logout issue hadek! 

I have tested with various scenarios on Android Emulator. I can confirmed that the frequent logout issue is now fixed.

PR #1277, opened 2024-01-15
Fix up proxyman setup

gnunicorn:
Now able to use proxyman on:
 - [x] MacOS

By adding: `--dart-define HTTP_PROXY=http://PROXYMAN_IP:9090` (or `http://localhost9090/`) to the flutter app run command.

Screenshot 2024-01-15 at 17 24 08

Namely this does two things:
 - add support for reading the proxy-param for `login_with_token` in rust, thus actually enabling using the proxy on restarted apps
 - add support for using system certs on reqwest by enabling said feature for the build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,7 @@ dependencies = [
  "oslog",
  "parse-env-filter",
  "pulldown-cmark",
+ "reqwest",
  "ruma",
  "ruma-common",
  "ruma-events",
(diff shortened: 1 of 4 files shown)


kumarpalsinh25 (approved):
For Macos:
- - - - -
I can confirm that by running app with `flutter run -d macos --dart-define HTTP_PROXY=http://localhost:9090` it is showing logs on Proxyman.

gnunicorn:
android I couldn't get to work. comes back with a "self-signed cert"-issue :( ...

Issue #1273, opened 2024-01-15
Rework Pins tracking


gnunicorn:
From https://github.com/acterglobal/UX/issues/138

- [x] show pin content in overview list if it just a small text
- [x] clean up spacing, padding and designs
- [x] include pins in quick-jump search
- [x] add attachments
- [x] show main attachment in overview
- [x] add comments
- [x] add likes
- [ ] #1621
- [x] define new creation and editing methods
- [x] #1274

gnunicorn:
closing with only one item open.

PR #1288, opened 2024-01-17
Revamped Pins Editing flow and Appflowy inline editor integration

gtalha07:
Refs: #1273.

In order to keep minimal PR changes for big updates, this PR is trimmed and planned to merge on short basis and later requirements to be dealt separately. Notably this includes:

- [x] Search feature for Pins in Quick Jump Actions Page.
- [x] Pins general UI bug fixes.
- [x] Appflowy in-line editor integration for editing Pins content.
- [x] Revamped Pins editing flow.
- E2E test support will come in later PRs, as Pins internal content needs a little refactor and some UI changes .i.e. having links as part of attachments rather considering it a pin type.

Pins UI and editing flow Updated (Desktop):

https://github.com/acterglobal/a3/assets/68579938/f221e657-70ab-4df9-b405-be6e8e1dcee6

(Mobile):

https://github.com/acterglobal/a3/assets/68579938/a694879a-88c0-4c86-af2b-5fd754787294


app/macos/Podfile.lock
```diff
@@ -155,4 +155,8 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce7dbe26c78bfc7ba46736094c1e2d25982870fa
 
+<<<<<<< HEAD
 COCOAPODS: 1.14.2
+=======
+COCOAPODS: 1.14.3
+>>>>>>> main
```
gnunicorn, line 162: pretty sure this deosn't belong here.

native/acter/src/api/pins.rs
```diff
@@ -227,8 +227,13 @@ impl Pin {
             .and_then(|t| t.formatted.clone().map(|f| f.body))
     }
 
+<<<<<<< HEAD
     pub fn content_text(&self) -> Option<MsgContent> {
         self.content.content.as_ref().map(MsgContent::from)
+=======
+    pub fn content_text(&self) -> Option<String> {
+        self.content.content.clone().map(|t| t.body)
+>>>>>>> main
```
gnunicorn, line 236: smae here... broken merge...

app/macos/Podfile.lock
gtalha07: Yeah it doesn't, thanks for pointing out.

native/acter/src/api/pins.rs
gtalha07: Will fix it, thanks!

app/lib/features/pins/widgets/pin_item.dart
```diff
@@ -82,22 +83,25 @@ class _PinItemState extends ConsumerState<PinItem> {
...
-          onSave: (plain, htmlBody) async {
-            if (_formkey.currentState!.validate()) {
-              pinEditNotifier.setLink(_linkController.text);
-              pinEditNotifier.setPlainText(plain);
-              pinEditNotifier.setHtml(htmlBody);
-              final res = await pinEditNotifier.onSave(context);
-              debugPrint('PIN EDITED: $res');
-            }
-          },
+      Flexible(
```
gnunicorn, line 86: Love that it is inline, but I'm a bit puzzled by this. Does that mean we are always in the editor when we are looking at an individual Pin? That doesn't sound like a regular usecase though ... most of the time the person wants to look it up, not edit it. Editing features should probably be put behind some option-button->edit or something ...

app/lib/features/pins/widgets/pin_list_item.dart
```diff
@@ -9,11 +10,11 @@ import 'package:go_router/go_router.dart';
 import 'package:acter/common/utils/routes.dart';
 
 class PinListItem extends ConsumerStatefulWidget {
-  final ActerPin pin;
+  final String pinId;
```
gnunicorn, line 13: I understand the rational here, but I don't really agree with the approach taken. It just makes this Widget a lot harder to comprehend and follow. I'd rather we'd have created a wrapper Widget around it do the same:

```dart
class PinListItemById extends ConsumerStatefulWidget {
  final String pinId;
  final bool showSpace;
  const PinListItemById({
    super.key,
    required this.pinId,
    this.showSpace = false,
  });


    return ref.watch(pinProvider(pinId)).when(
      data: (acterPin) => PinListItem(pin: acterPin, showSpace: showSpace),
      loading: () => Skeletonizer(child: Card(child: Column( children: [ ... ]))),
      error: (err, st) => Text('Error loading pin: ${err.toString()}')
    };
}
```

gnunicorn (changes requested):
Thanks heaps. But the Widget always being in edit-mode I find a bit too strange for real usage ... let's put that behind an menu-option, maybe? Also left a remark on making the widget more complex rather than wrapping that complexity into an outer widget to use instead...

kumarpalsinh25 (commented):
Looks good overall. Exited to see inline html editor.

General feedback:
- Agreed with @gnunicorn regarding having view and edit screens as separate for better user experience
- If possible then can you share screen-recording of desktop app related to new Pins updates 
- There were few UI improvements required but we can do it on later stage.
- Added few minor comments

app/lib/features/pins/widgets/pin_item.dart
gtalha07: Hmm, that sounds sensible. I will change it into enabling/disabling edit mode 👍🏻 !

app/lib/router/shell_routers/quick_jump_shell_router.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:acter/common/utils/routes.dart';
+import 'package:acter/common/widgets/dialog_page.dart';
+import 'package:acter/features/search/pages/quick_jump.dart';
+import 'package:acter/router/router.dart';
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+List<RouteBase> makeQuickJumpShellRoutes(ref) {
```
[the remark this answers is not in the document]
gtalha07: This one is especially made for desktop case, the ```search_shell_router``` is for mobile navigation. Without this, this'd lead to error .i.e. quick jump shell opens up dialog rather than page view.

app/lib/router/general_router.dart
```diff
@@ -118,28 +108,6 @@ List<RouteBase> makeGeneralRoutes() {
...
       },
     ),
-    GoRoute(
-      parentNavigatorKey: rootNavKey,
-      name: Routes.editPin.name,
-      path: Routes.editPin.route,
-      pageBuilder: (context, state) {
-        return SideSheetPage(
-          key: state.pageKey,
-          transitionsBuilder: (context, animation, secondaryAnimation, child) {
```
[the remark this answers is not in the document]
gtalha07: No, the editing flow doesn't need special route in this case. We can use button toggle for switching to edit mode of pin view.

app/lib/features/space/pages/pins_page.dart
```diff
@@ -63,28 +62,35 @@ class SpacePinsPage extends ConsumerWidget {
...
-                              onPressed: () => context.pushNamed(
-                                Routes.actionAddPin.name,
-                                queryParameters: {'spaceId': spaceIdOrAlias},
-                              ),
-                              child: const Text('Share Pin'),
-                            )
-                          : null,
-                    ),
-                  ),
+                return Consumer(
```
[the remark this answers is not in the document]
gtalha07: Yes, there's no need of additional ```Consumer``` here, will remove it. 👍🏻

app/ios/Podfile.lock
```diff
@@ -58,19 +58,35 @@ PODS:
...
+    - Flutter
   - path_provider_foundation (0.0.1):
     - Flutter
     - FlutterMacOS
   - push_ios (0.0.1):
     - Flutter
   - ReachabilitySwift (5.0.0)
+  - rich_clipboard_ios (0.0.1):
+    - Flutter
+  - screen_brightness_ios (0.1.0):
```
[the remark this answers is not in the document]
gtalha07: Nothing I can remember of, will revert this. 👍🏻

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -189,34 +193,44 @@ class QuickJump extends ConsumerWidget {
...
+              child: Padding(
+                padding:
+                    const EdgeInsets.symmetric(vertical: 15, horizontal: 10),
+                child: TextField(
+                  autofocus: true,
+                  decoration: const InputDecoration(
+                    border: InputBorder.none,
+                    focusedBorder: UnderlineInputBorder(
+                      borderSide: BorderSide(color: Colors.white),
+                    ),
```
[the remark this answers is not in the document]
gtalha07: This is special case, as textfield here according to UI has underline border.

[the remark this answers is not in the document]
kumarpalsinh25: Better to keep same inputBox style here also as underline looks weird in given screen recordings.

app/lib/router/shell_routers/quick_jump_shell_router.dart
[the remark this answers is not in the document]
kumarpalsinh25: Understood but better to keep this routes in same files with appropriate naming and comments so that it is easy to get idea of this things.

kumarpalsinh25 (commented):
@gtalha07 
On the desktop view screen-recording, It looks like the editing color preference doen't getting saved. "Headline-2" text was set to brown color and saved. But after re-visit, it is reset to white color.

Can you share video recoding in which different text styles and html styling applied and getting preserved after saving it.

gtalha07:
> @gtalha07 On the desktop view screen-recording, It looks like the editing color preference doen't getting saved. "Headline-2" text was set to brown color and saved. But after re-visit, it is reset to white color.
> 
> Can you share video recoding in which different text styles and html styling applied and getting preserved after saving it.

Is html styling even supported in matrix ? I doubt that...

kumarpalsinh25:
> Is html styling even supported in matrix ? I doubt that...
@gnunicorn may have better idea here.

kumarpalsinh25 (approved):

app/lib/features/pins/models/pin_edit_state/pin_edit_state.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
```
gnunicorn, line 1: More often than not, I've come to find the riverpod-based form-models for simple data (like here title, link, etc) to be easily broken and not working correctly (e.g. because of lacking proper updates when the focus isn't gone yet, or simply not being properly in sync at all times or ugly refreshing and/or lack of proper cursor management) and found the much simpler form+formfields-with-validators with local edit-controller to be much easier to use and less fragile.

Unless there is a secondary source for changing the provider and therefore the need to sync the data (which isn't the case here nor for many of our editors in general), what is the benefit of using this over plain old edit-controller+form-fields?

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -83,152 +87,110 @@ class PinPage extends ConsumerWidget {
...
-              senderId: pin.sender().toString(),
-              isSpace: true,
-            ),
-          ),
-          child: Row(
-            children: <Widget>[
-              Icon(
-                Atlas.warning_thin,
-                color: Theme.of(context).colorScheme.error,
+          PopupMenuItem(
```
gnunicorn, line 90: I am confused: why did you drop the `else`? Doesn't that mean that other people can't report this pin anymore?  But anyone should be able to report a pin...

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -0,0 +1,144 @@
...
+
+  Future<List<ActerPin>> _getPins() async {
+    return (await arg.pins()).toList(); // this might throw internally
+  }
+
+  @override
+  Future<List<ActerPin>> build(Space arg) async {
+    final client = ref.watch(alwaysClientProvider);
+    final spaceId = arg.getRoomId();
+    _listener = client.subscribeStream('$spaceId::PINS'); // stay up to date
```
gnunicorn, line 66: yeah, this isn't ever triggered. the string must be `pins` (all lowercase), see https://github.com/acterglobal/a3/blob/f2e223d1c4fda5cc3722c3685e7f0603ea0ebc4a/native/core/src/statics.rs#L18

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -0,0 +1,144 @@
...
+    try {
+      final updateBuilder = pin.updateBuilder();
+      updateBuilder.title(state.title);
+      if (pin.isLink()) {
+        updateBuilder.url(state.link);
+      }
+      updateBuilder.contentText(state.plain);
+      if (state.htmlBody != null) {
+        updateBuilder.contentMarkdown(state.htmlBody!);
+      }
```
gnunicorn, line 128: I think _here_ is the actual problem you were facing. sending the HTML through the markdown parser isn't what you want to do, as it alters some stuff. You want a secondary setting-html-body-method ...

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -0,0 +1,144 @@
...
+      await updateBuilder.send();
+      final updatedPin = await pin.refresh();
+      // make sure we refresh the pin over list too.
+      ref.invalidate(pinsProvider);
+      ref.invalidate(spacePinsProvider);
+      ref.invalidate(pinsFoundProvider);
+      EasyLoading.dismiss();
```
gnunicorn, line 135: if you replace the existing one, you don't have to dismiss I think ...
```suggestion
```

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -0,0 +1,144 @@
...
+      updateBuilder.contentText(state.plain);
+      if (state.htmlBody != null) {
+        updateBuilder.contentMarkdown(state.htmlBody!);
+      }
+      await updateBuilder.send();
+      final updatedPin = await pin.refresh();
+      // make sure we refresh the pin over list too.
+      ref.invalidate(pinsProvider);
+      ref.invalidate(spacePinsProvider);
+      ref.invalidate(pinsFoundProvider);
```
gnunicorn, line 134:
```suggestion
```
This shouldn't be necessary, but it currently is as the `pins`-listening key was wrong...

app/lib/features/pins/widgets/pin_item.dart
```diff
@@ -0,0 +1,113 @@
...
+            validator: (value) {
+              if (value != null) {
+                final uri = Uri.tryParse(value);
+                if (uri == null || !uri.isAbsolute) {
+                  return 'link is not valid';
+                }
+              }
+              return null;
+            },
+          ),
```
gnunicorn, line 69: I am not quite sure how this works now. the user has to click the tiny icon on the left to open the link? I am pretty sure in more cases than not they want to open the link when they click it. not switch into an edit-mode ... if anything the UX should be the other way around: show a suffix-edit icon if the user can edit and only upon clicking that switch into edit-mode and on all other clicks, open the link...

app/lib/features/pins/widgets/pin_list_item.dart
```diff
@@ -50,31 +88,95 @@ class _PinListItemState extends ConsumerState<PinListItem> {
...
-                ? const Icon(Icons.open_in_full_sharp)
-                : const Icon(Icons.chevron_right_sharp),
-            onPressed: openItem,
-          ),
+        child: LayoutBuilder(
+          builder: (context, constraints) {
+            return Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 12),
+              child: Column(
+                key: Key(pinId), // FIXME: causes crashes in ffigen
```
gnunicorn, line 99: pretty sure this fixme isn't useful here anymore
```suggestion
                key: Key(pinId),
```

app/lib/features/search/pages/search_page.dart
```diff
@@ -26,7 +26,7 @@ class SearchPage extends ConsumerWidget {
           }
         }
       },
-      expand: true,
+      expand: false,
```
gnunicorn, line 29: wanna explain this change?

app/lib/features/search/providers/pins.dart
```diff
@@ -0,0 +1,43 @@
...
+    final title = pin.title();
+    final pinId = pin.eventIdStr();
+    if (searchValue.isNotEmpty) {
+      if (!(title.toLowerCase()).contains(searchValue)) {
+        continue;
+      }
+    }
+    finalPins.add(
+      PinDetails(
+        '/pins/$pinId',
```
gnunicorn, line 30: no hardcoding of routes please.

gnunicorn: hmpf... I see you are only doing what was done here before ... so ... let's add a FIXME here and say that we want the base delegate to support proper route creation...
```suggestion
        // FIXME: it would be nicer to have proper route delegation rather than final strings here
        '/pins/$pinId',
```

app/lib/features/search/widgets/quick_actions_builder.dart
```diff
@@ -72,7 +72,10 @@ class QuickActionsBuilder extends ConsumerWidget {
           canPostPin
               ? OutlinedButton.icon(
                   key: QuickJumpKeys.createPinAction,
-                  onPressed: () => context.pushNamed(Routes.actionAddPin.name),
+                  onPressed: () => navigateTo(
+                    route: Routes.actionAddPin,
+                    push: true,
```
gnunicorn, line 77: care to elaborate? this entire actions routing things is kinda fragile and frickely and changes like this can have unexpected consequences ... did you run the UI test-suite after doing this change? Does it still pass?

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -189,34 +193,42 @@ class QuickJump extends ConsumerWidget {
...
-                ref.read(searchValueProvider.notifier).state = value;
-              },
             ),
-          ),
-          ...body,
-        ],
-        // ),
+            ...body,
+          ],
+          // ),
```
gnunicorn, line 230:
```suggestion
```

native/acter/src/api/pins.rs
```diff
@@ -224,8 +226,8 @@ impl Pin {
             .and_then(|t| t.formatted.clone().map(|f| f.body))
     }
 
-    pub fn content_text(&self) -> Option<String> {
-        self.content.content.clone().map(|t| t.body)
+    pub fn content_text(&self) -> Option<MsgContent> {
```
gnunicorn, line 229: this clearly has the wrong name now...
```suggestion
    pub fn content(&self) -> Option<MsgContent> {
```

gnunicorn (changes requested):
It would have been better to split this into two PRs: one for the appflowy-editor-part (which looks mostly alright) and one about the quick-jump. There are several smaller problems on each side of it and while I think addressing most of them on the appflowy-side could allow for that to be fine, quickly, I am not sure this applies to the quick-jump side ...

app/lib/common/widgets/html_editor.dart
```diff
@@ -52,7 +53,8 @@ extension ActerEditorStateHelpers on EditorState {
   String intoHtml({
     AppFlowyEditorHTMLCodec? codec,
   }) {
-    return (codec ?? defaultHtmlCodec).encode(document);
+    final docHTML = (codec ?? defaultHtmlCodec).encode(document);
+    return HtmlUnescape().convert(docHTML);
```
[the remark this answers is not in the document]
gtalha07: So there's a case where encoding this directly will result in HTML having escaped attributes. How it brings that kind of HTML? : if a user directly injects HTML code in editor instead of using editor features, so upon save it renders literal HTML strings. Having it converted to plain HTML prevents that insecurity. 

I didn't knew the rust method ```.contentMarkdown()``` was indeed specific for  handling only markdown content. As in previous version, this method was also handling HTML content...

app/lib/common/widgets/html_editor.dart
```diff
@@ -182,10 +189,7 @@ class _HtmlEditorState extends State<HtmlEditor> {
               bulletedListItem,
               numberedListItem,
               linkItem,
-              buildTextColorItem(),
               buildHighlightColorItem(),
-              ...textDirectionItems,
-              ...alignmentItems,
```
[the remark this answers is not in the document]
gtalha07: Currently for some unknown reason, the text color and alignment doesn't get saved after editing. I had to disable it. Although I see the wire HTML content brings that attributes but the editor isn't rendering it somehow. This needs to look separately.

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
gtalha07: secondary setting method as in means of rust fn for ```updateBuilder``` ?

gtalha07:
@gnunicorn @kumarpalsinh25 Since the PR is already boluted and brought up lot of noise, it's better to split PR work into their relevant changes and bring it over on fresh basis for quick review and merge. What do you guys think?

gtalha07:
I will still address the feedback here, no second thoughts on that but doing changes here isn't good enough...

app/lib/features/search/widgets/quick_actions_builder.dart
gtalha07: reason is to to keep consistency .i.e. as I see many widgets here are using mixed up navigation methods so its better to have actions routing be confined to ```navigateTo``` only. I haven't ran the test-suite but tested a lot during PR work by navigating in-between a lot and seems no issue.

app/lib/features/search/providers/pins.dart
gtalha07: Ahh I forgot to change that, happens when you put your hands all over the code rather than actual fixation, will bring that in fresh PR :D

app/lib/features/search/pages/search_page.dart
gtalha07: I think this was causing UI overflow when the text field gets focus. Maybe not, I will see that in fresh PR. 👍🏻

app/lib/features/pins/widgets/pin_item.dart
gtalha07: The user can click on icon to launch url link, yes. Clicking on field will not switch it in edit mode as user has to specifically go in pin item actions to enable editing (which makes every field editable, then).

app/lib/features/pins/pages/pin_page.dart
gtalha07: Ahh this should not be this way, must have overlooked on it. Thanks for pointing it out!

app/lib/features/pins/models/pin_edit_state/pin_edit_state.dart
gtalha07: So one benefit of using riverpod based models is that you're controlling and managing UI data with one separate class rather than having separate providers for each state variable and potentially having multiple text controllers being passed through widget tree if that variable is needed at final state stage. This would make code a bit complex and hard to refactor. And we have a lot of providers right now which are hard to track which goes where,  having riverpod state class with internal state variables mitigates a lot of noise.

gtalha07: I know there are some potential problems with these models in regards of UI updates and sync problems, but to counter it we started adding UI suite which brings these issues forefront and allow us to be more helpful in investigating.

PR #1309, opened 2024-01-25
Refactoring App Theme

kumarpalsinh25:


app/lib/common/themes/components/bottom_sheet_theme.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:acter/common/themes/colors/color_scheme.dart';
+import 'package:flutter/material.dart';
+
+var bottomSheetTheme = BottomSheetThemeData(
```
gtalha07, line 4: Just a suggestion to ponder. I would really like to have these singleton themes be merged as one component theme file, makes it easier to change design stuff rather than looking over different files. But I would like to have general agreement on decision outcome!

app/lib/common/widgets/base_body_widget.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:acter/common/themes/colors/color_scheme.dart';
+import 'package:flutter/material.dart';
+
+/// Reusable base body widget.
+class BaseBody extends StatelessWidget {
```
gtalha07, line 5: Very good! 👏🏻

app/lib/common/widgets/member_list_entry.dart
```diff
@@ -239,13 +238,13 @@ class MemberListEntry extends ConsumerWidget {
...
                       actions: <Widget>[
-                        DefaultButton(
+                        ElevatedButton(
                           onPressed: () {
                             // close both dialogs
                             context.pop();
                             context.pop();
                           },
-                          title: 'Okay',
+                          child: const Text('Okay'),
```
gtalha07, line 247: Was thinking if there's a way to adapt font size based on devices, it's crucial based on user design perspective .i.e. people with disabilities would like to have shown larger font size than usual. Something like set up clamp on text scale factor would work well.

gtalha07 (approved):
Nice work overall!

app/lib/common/themes/components/bottom_sheet_theme.dart
kumarpalsinh25: This can be subjective based on the different thought process but ya we can surely code this on either way. My main goal here is to have all the design standard should be implemented from Flutter `ThemeData` itself to make app UI consistent.

PR #1310, opened 2024-01-25
New Media Cache

gnunicorn:
This PR adds support for storing media data _encrypted_ in a separate cache-folder on disk rather than in the sqlite store. This has two main advantages: 1) it is faster; 2) by having them separated the cache can be cleared without touching the important state data.

This is made possible through a new store-wrapper crate added in this PR that wraps any matrix-sdk-store (but in particular the sqlite-one, for which we have tests) and overwrites the media-storage functions with a local filesystem variant.


native/acter/src/platform/native.rs
```diff
@@ -196,3 +225,60 @@ fn get_log_filter(level: &str) -> Option<LevelFilter> {
...
+    };
+
+    let wrapped_state_store = matrix_sdk_store_media_cache_wrapper::wrap_with_file_cache(
+        sql_state_store,
+        media_path,
+        passphrase,
+    )
+    .await?;
+    Ok(config.state_store(wrapped_state_store))
+}
```
bitfriend, line 284: Yeah, this snippet is helpful to me.

bitfriend (approved):
Looks good to me

Issue #1302, opened 2024-01-19
Upgrading rust dependency tree


gnunicorn:
refs #1301 

Quite a few upgrades and releases have happened since we last upgraded our entire build environment. It seems we can do a long needed upgrade of it and should probably do it. Most notably we want:

- [x] switch to latest `nightly` (in `rust-toolchain`) to unpin from a very old version
- [ ] switch back to latest `matrix-sdk` on `main` (the `0.7.0` release unfortunately missed my PR right before)
- [x] upgrade ffigen to use the release genco again (probably)
- [ ] upgrade a few outdated dependencies:
```
acter: /DEV/acter/a3/native/acter/Cargo.toml
        derive_builder matches 0.10.2;  latest is 0.12.0
        strum matches 0.24.1;   latest is 0.25.0
        cbindgen matches 0.24.5;        latest is 0.26.0
        android_logger matches 0.12.0;  latest is 0.13.3

acter-core: /DEV/acter/a3/native/core/Cargo.toml
        chrono-tz matches 0.5.3;        latest is 0.8.5
        csscolorparser matches 0.5.0;   latest is 0.6.2
        derive-getters matches 0.2.1;   latest is 0.3.0
        derive_builder matches 0.11.2;  latest is 0.12.0
        minijinja matches 0.30.7;       latest is 1.0.11
        strum matches 0.24.1;   latest is 0.25.0
        toml matches 0.7.8;     latest is 0.8.8
        env_logger matches 0.10.2;      latest is 0.11.0

acter-cli: /DEV/acter/a3/native/cli/Cargo.toml
        dialoguer matches 0.9.0;        latest is 0.11.0
        env_logger matches 0.10.2;      latest is 0.11.0

acter-test: /DEV/acter/a3/native/test/Cargo.toml
        env_logger matches 0.10.2;      latest is 0.11.0

acter-tui: /DEV/acter/a3/native/tui/Cargo.toml
        crossterm matches 0.25.0;       latest is 0.27.0
        env_logger matches 0.10.2;      latest is 0.11.0

mr_minutes: /DEV/acter/a3/util/mr_minutes/Cargo.toml
        env_logger matches 0.10.2;      latest is 0.11.0
        git2 matches 0.17.2;    latest is 0.18.1
```

PR #1311, opened 2024-01-25
Update sdk

bitfriend:
Will fix https://github.com/acterglobal/a3/issues/1302


native/acter/Cargo.toml
```diff
@@ -40,8 +39,8 @@ mime = "0.3.17"
 mime2ext = "0.1.52"
 mime_guess = "2.0.4"
 parse-env-filter = "0.1.0"
-ruma = { workspace = true, features = ["html", "unstable-unspecified", "unstable-msc3488", "compat-unset-avatar", "unstable-msc3245-v1-compat"] }
-ruma-common = { workspace = true, features = ["client"] }
+ruma = { workspace = true, features = ["rand", "unstable-msc2448", "unstable-msc2965", "unstable-msc3930", "unstable-msc3245-v1-compat"] }
```
gnunicorn, line 42: `html` disappeared. is that on purpose?

native/acter/src/api/room.rs
```diff
@@ -156,7 +156,7 @@ impl Member {
         let tester: PermissionTest = match permission {
             MemberPermission::CanBan => return self.member.can_ban(),
             MemberPermission::CanInvite => return self.member.can_invite(),
-            MemberPermission::CanRedact => return self.member.can_redact(),
+            MemberPermission::CanRedact => return self.member.can_redact_own(),
```
gnunicorn, line 159: I don't think this change is correct. This would not allow admins to redact items anymore more. `can_redact_other()` is also relevant here. we might have to refactor the entire way this works ... at least also put up these two different modes of redaction up: split the current one into `MemberPermission::CanRedactOwn` and `MemberPermission::CanRedactOther` ...

native/acter/src/api/room.rs
```diff
@@ -1757,7 +1757,7 @@ impl Room {
                     .get_member(&my_id)
                     .await?
                     .context("Unable to find me in room")?;
-                if !member.can_redact() {
+                if !member.can_redact_own() {
```
gnunicorn, line 1760: same here we first need to figure out whether this is our own and then check accordingly ... but that means knowing the original author_id I suppose ... so I guess we have to alter the API to require that is also provided.

Or change the api-docs to say that this isn't checking for the permission and permission thus must have been checked before the call. Which I'd consider fine, too.

gnunicorn (changes requested):
Unfortunately, the change in the redact permission requires more changes in the entire system...

native/acter/Cargo.toml
bitfriend: Ah, I cloned params from matrix-rust-sdk project.
Will revert `html` feature.

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1311/commits/1989a5e1c1f64d6b0d0d0761ed32291247856c59

native/acter/src/api/room.rs
bitfriend: Sorry for my bad.
Fixed by https://github.com/acterglobal/a3/pull/1311/commits/1989a5e1c1f64d6b0d0d0761ed32291247856c59

native/acter/Cargo.toml
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1311/commits/099616d7834d0a61c7c2d8e83dffd34a27b71d9b

gnunicorn (changes requested):
Great update. I request only one more change then we can merge it.

bitfriend:
The matrix dev team is splitting `ruma` into `ruma-common`, `ruma-events` and `ruma-client-api`, etc.
Also we will use `ruma` instead of `matrix_sdk::ruma`.

gnunicorn (approved):

PR #1340, opened 2024-01-30
[Updates]: fix VideoSlide ```typeStr```

gtalha07:
- adds serde tag support for ```NewsContent``` to strictly parse according to type and fallback type to support existing ones.
- add video slide support in rust integration tests.


native/acter/src/api/news.rs
```diff
@@ -165,24 +166,49 @@ impl NewsSlide {
...
+            NewsContent::Video(VideoMessageEventContent { body, .. })
+            | NewsContent::Fallback(FallbackNewsContent::Video(VideoMessageEventContent {
+                body,
+                ..
+            })) => body.clone(),
+            NewsContent::Audio(AudioMessageEventContent { body, .. })
+            | NewsContent::Fallback(FallbackNewsContent::Audio(AudioMessageEventContent {
+                body,
+                ..
+            })) => body.clone(),
```
gnunicorn, line 203: aren't all these `body` of the same type? Could be one match in that case... but also we might want to consider moving that into `core` as it is a bit ... verbose... What do you think?

native/acter/src/api/news.rs
```diff
@@ -194,12 +220,32 @@ impl NewsSlide {
...
+            }
+
+            NewsContent::Video(content)
+            | NewsContent::Fallback(FallbackNewsContent::Video(content)) => {
+                MsgContent::from(content)
+            }
+            NewsContent::Text(content)
+            | NewsContent::Fallback(FallbackNewsContent::Text(content)) => {
+                MsgContent::from(content)
+            }
```
gnunicorn, line 248: :+1:

native/core/src/events/news.rs
```diff
@@ -30,61 +30,142 @@ pub enum NewsContent {
...
+    /// A text message.
+    Text(TextMessageEventContent),
+    /// An audio message.
+    Audio(AudioMessageEventContent),
+    /// A file message
+    File(FileMessageEventContent),
+    /// A location message.
+    Location(LocationMessageEventContent),
+
+    #[serde(untagged)]
```
gnunicorn, line 104: Let's add some docs here:
```suggestion
    /// Backwards-compatible fallback support for previous untagged version
    /// only for reading existing events.
    #[serde(untagged)]
```

gnunicorn (approved):
lgtm. just some minor things inline we might want to consider

native/acter/src/api/news.rs
gtalha07: Yes all cases have body as ```&String```. Could have possibility of moving in core, let me do it.

native/core/src/events/news.rs
gtalha07: Forgot it since you mentioned. Let me add!

Issue #1338, opened 2024-01-30
Bug reporting opens multiple times


gnunicorn:
When opening the bug reported via shaking accidentially and clicking the `back` it appears the windows is there a few times and it needs to be closed multiple times. The screen should probably be considered a singleton and ever only open once.

Issue #839, opened 2023-08-09
include user-id in bug-reporting


gnunicorn:
We are receiving reports by users now via the rageshake and it would very useful if we could contact them about it to ask follow up questions. To allow for that we need to include the submitters user-id. Let's add a (default-on) checkbox to `Include my user-id to allow the team to contact me about this issue` to the bug-reporting form-field and include the user-id in the submission if that is selected.

gnunicorn:
fixed in #1341  .

PR #1341, opened 2024-01-30
Improved bug reporter

gnunicorn:
:tada: with UI tests

- [x] fixes title reset, https://github.com/acterglobal/a3-meta/issues/66
- [x] add user id #839 
- [x] fix screenshot taking
- [x] fix open multiple times, https://github.com/acterglobal/a3/issues/1338
- [x] add previous user log
- [x] clear old cached logs and screenshot files

Look at this nice set of features available and end-2-end-tested!

grafik


app/lib/features/search/pages/quick_jump.dart
```diff
@@ -18,19 +18,24 @@ class QuickjumpDialog extends ConsumerWidget {
...
           }) async {
-            context.pop();
-            if (push) {
-              if (route == null) {
-                await context.push(target!);
-              } else {
-                await context.pushNamed(route.name);
-              }
+            if (custom != null) {
+              await custom(context);
```
kumarpalsinh25, line 24: What is `await custom(context);` ?

kumarpalsinh25 (approved):
Looks good overall. Great work!

app/lib/features/search/pages/quick_jump.dart
gnunicorn: It's a custom callback to run rather than specific routing - it's the callback we do for the bug reporter to close the quick jumper and take the screenshot before opening the bug reporter

PR #1345, opened 2024-01-31
Split ruma into sub crates

bitfriend:
The matrix dev team is splitting `ruma` into `ruma-common`, `ruma-events` and `ruma-client-api`, etc.
Also we will use `ruma` instead of `matrix_sdk::ruma`.


bitfriend:
`ruma_client_api` doesn't seem to work in the latest `ruma`.
Will discard this PR.

PR #1351, opened 2024-02-05
[small-fix] Improve flutter code

bitfriend:
- Replace `ref.read(clientProvider)!` with `ref.read(alwaysClientProvider)` so that it can be more meaningful
- Assemble path using platform directory separator


bitfriend:
We will use only `\` as directory separator in windows.
We will use only `/` as directory separator in linux and macos.

kumarpalsinh25 (approved):
LGTM!

PR #1359, opened 2024-02-07
News slide Colors

kumarpalsinh25:
- Now news slide can show the background color which selected by user at the time of the news posting
- Fixed image scaling issue
- Removed unwanted caption text
- changed the internal color-code from a complex struct to an ARGB-encoding `u32` as it is much simpler, takes less bandwidth and it is directly compatible with flutter.

Reference Video:
https://github.com/acterglobal/a3/assets/51526480/df5b9f67-b837-4891-ad76-4969d3ccdf56

### DIFF
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -51,17 +51,13 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
           ? Colors.transparent
           : selectedNewsPost?.backgroundColor,
       actions: [
-        Visibility(
-          visible: selectedNewsPost?.type == NewsSlideType.text,
-          child: IconButton(
-            key: NewsUpdateKeys.slideBackgroundColor,
-            onPressed: () {
-              ref
-                  .read(newsStateProvider.notifier)
-                  .changeTextSlideBackgroundColor();
-            },
-            icon: const Icon(Atlas.color),
-          ),
+        IconButton(
+          onPressed: () {
+            ref
+                .read(newsStateProvider.notifier)
+                .changeTextSlideBackgroundColor();
+          },
+          icon: const Icon(Atlas.color),
         ),
       ],
     );
@@ -288,17 +284,25 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
 
   Widget slideImagePostUI(BuildContext context) {
     final imageFile = selectedNewsPost!.mediaFile;
-    return Image.file(
-      File(imageFile!.path),
-      fit: BoxFit.contain,
+    return Container(
+      alignment: Alignment.center,
+      color: selectedNewsPost!.backgroundColor,
+      child: Image.file(
+        File(imageFile!.path),
+        fit: BoxFit.contain,
+      ),
     );
   }
 
   Widget slideVideoPostUI(BuildContext context) {
     final videoFile = selectedNewsPost!.mediaFile!;
-    return ActerVideoPlayer(
-      key: Key(videoFile.name),
-      videoFile: File(videoFile.path),
+    return Container(
+      alignment: Alignment.center,
+      color: selectedNewsPost!.backgroundColor,
+      child: ActerVideoPlayer(
+        key: Key(videoFile.name),
+        videoFile: File(videoFile.path),
+      ),
     );
   }
 }
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -289,7 +289,7 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
       color: selectedNewsPost!.backgroundColor,
       child: Image.file(
         File(imageFile!.path),
-        fit: BoxFit.contain,
+        fit: BoxFit.fitWidth,
       ),
     );
   }
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -289,7 +289,7 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
       color: selectedNewsPost!.backgroundColor,
       child: Image.file(
         File(imageFile!.path),
-        fit: BoxFit.fitWidth,
+        fit: BoxFit.contain,
       ),
     );
   }
--- a/app/lib/features/news/widgets/news_item_slide/image_slide.dart
+++ b/app/lib/features/news/widgets/news_item_slide/image_slide.dart
@@ -6,10 +6,14 @@ import 'package:flutter/material.dart';
 
 class ImageSlide extends StatefulWidget {
   final NewsSlide slide;
+  final Color bgColor;
+  final Color fgColor;
 
   const ImageSlide({
     super.key,
     required this.slide,
+    required this.bgColor,
+    required this.fgColor,
   });
 
   @override
@@ -41,10 +45,12 @@ class _ImageSlideState extends State<ImageSlide> {
       builder: (BuildContext context, AsyncSnapshot<Uint8List> snapshot) {
         if (snapshot.hasData) {
           return Container(
+            color: widget.bgColor,
+            alignment: Alignment.center,
             key: NewsUpdateKeys.imageUpdateContent,
             foregroundDecoration: BoxDecoration(
               image: DecorationImage(
-                fit: BoxFit.cover,
+                fit: BoxFit.fitWidth,
                 image: MemoryImage(
                   Uint8List.fromList(snapshot.data!),
                 ),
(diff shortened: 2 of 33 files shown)


app/lib/features/news/pages/add_news_page.dart
```diff
@@ -285,17 +282,25 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
...
     final imageFile = selectedNewsPost!.mediaFile;
-    return Image.file(
-      File(imageFile!.path),
-      fit: BoxFit.contain,
+    return Container(
+      alignment: Alignment.center,
+      color: selectedNewsPost!.backgroundColor,
+      child: Image.file(
+        File(imageFile!.path),
+        fit: BoxFit.fitWidth,
```
gnunicorn, line 290: why not `contain`? What happens if the image is _larger_ now? It will run out of space on the top and bottom? I think `contain` was correct, no?

app/lib/features/news/widgets/news_item_slide/image_slide.dart
```diff
@@ -40,9 +44,11 @@ class _ImageSlideState extends State<ImageSlide> {
       builder: (BuildContext context, AsyncSnapshot<Uint8List> snapshot) {
         if (snapshot.hasData) {
           return Container(
+            color: widget.bgColor,
+            alignment: Alignment.center,
             foregroundDecoration: BoxDecoration(
               image: DecorationImage(
-                fit: BoxFit.cover,
+                fit: BoxFit.fitWidth,
```
gnunicorn, line 51: okay, now I need this to be explained, please ...

app/lib/features/news/widgets/news_side_bar.dart
```diff
@@ -53,7 +40,7 @@ class NewsSideBar extends ConsumerWidget {
         LikeButton(
           likeCount: news.likesCount().toString(),
           style: style,
-          color: fgColor,
+          color: Colors.white,
```
gnunicorn, line 43: I understand why that was removed, but the idea of the fg-color is to allow the user to set a higher contrast to whatever they have as the background (e.g. a very light image) to keep these accessible. We probably want to use any `fg-color` (if given) here in the future regardless from it not yet being implemented...

app/packages/rust_sdk/lib/acter_flutter_sdk.dart
```diff
@@ -68,17 +68,11 @@ String userAgent = '$appName/$versionName';
 RegExp logFileRegExp = RegExp('app_.*log');
 RegExp screenshotFileRegExp = RegExp('screenshot_.*png');
 
-Color convertColor(ffi.EfkColor? primary, Color fallback) {
+Color convertColor(int? primary, Color fallback) {
```
gnunicorn, line 71: Arguably this is a one-liner now we might want to drop ... but that can happen at a later time ...

app/lib/features/news/pages/add_news_page.dart
```diff
@@ -51,16 +51,13 @@ class _AddNewsState extends ConsumerState<AddNewsPage> {
...
-            },
-            icon: const Icon(Atlas.color),
-          ),
+        IconButton(
+          onPressed: () {
+            ref
+                .read(newsStateProvider.notifier)
+                .changeTextSlideBackgroundColor();
+          },
+          icon: const Icon(Atlas.color),
```
gnunicorn, line 60: inkwell and long-press to allow for a custom color selector? Well, future issue, I guess...

gnunicorn (approved):
skipped the rust-side review as I wrote it, asking @bitfriend or @gtalha07 to review that, too.

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: Yes, it should be `contain`. I was testing something but my bad, forget to revert it.

app/lib/features/news/widgets/news_item_slide/image_slide.dart
kumarpalsinh25: I will make to `BoxFit.contain` as `BoxFit.cover` is scaling entire horizontal image. As per our last discussion, we need to remove auto-scaling of the image slide.

Screenshot 2024-02-07 at 5 34 41 PM

app/lib/features/news/widgets/news_side_bar.dart
kumarpalsinh25: Previously, `fgColor` was there based on the entire post and now it is based on the per slide mode. And also this side menus are applicable to entire post not for individual slides.

Screenshot 2024-02-07 at 5 44 13 PM

app/packages/rust_sdk/lib/acter_flutter_sdk.dart
kumarpalsinh25: @gnunicorn 
You may better know about this rust side change.

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: Yes, In future we can provider option to select custom color. But currently implement with fixed set of bg color look good to me and it is making things simple.

app/lib/features/news/widgets/news_side_bar.dart
gnunicorn: sure. future thing.

app/packages/rust_sdk/lib/acter_flutter_sdk.dart
gnunicorn: this is the dart-side ;) ...

```
convertColor = (int? primary, color Fallback) => primary != null ? Color(primary) : fallback
```

or maybe even `Color(primary ?? fallback)` is enough?

app/lib/features/news/widgets/news_item_slide/video_slide.dart
```diff
@@ -37,17 +41,21 @@ class _VideoSlideState extends State<VideoSlide> {
...
+      alignment: Alignment.center,
+      child: FutureBuilder<Uint8List>(
+        future: newsVideo.then((value) => value.asTypedList()),
+        builder: (BuildContext context, AsyncSnapshot<Uint8List> snapshot) {
+          if (snapshot.hasData) {
+            return ActerVideoPlayer(
+              videoFile: File.fromRawPath(snapshot.data!),
+            );
+          } else {
+            return const Center(child: Text('Loading video'));
```
gtalha07, line 55: This doesn't need to happen necessarily in this PR, but a placeholder here would be nice or having ```Skeletonizer()``` here.

gnunicorn: yeah, loading ahead the next screen below as well ... to decrease loading time in the first place. but both for the future...

native/acter/src/api/pins.rs
```diff
@@ -232,8 +232,8 @@ impl Pin {
         self.content.url.clone()
     }
 
-    pub fn color(&self) -> Option<Color> {
-        self.content.display.as_ref().and_then(|t| t.color.clone())
+    pub fn color(&self) -> Option<u32> {
```
gtalha07, line 235: Yes it's been there since long but didn't happen to ask on where this would be applied in Pins ?

native/acter/src/api/tasks.rs
```diff
@@ -236,8 +239,8 @@ impl TaskListDraft {
         self
     }
 
-    pub fn color(&mut self, color: Box<Color>) -> &mut Self {
-        self.content.color(Some(Box::into_inner(color)));
+    pub fn color(&mut self, color: Color) -> &mut Self {
```
gtalha07, line 242: Is this meant for bg color of task content?

gtalha07 (approved):
Rust and flutter side changes looks fine to me! Great work!

native/acter/src/api/pins.rs
gnunicorn: It's a common trait of notes apps to over to change the (background) color of the "post it" - most implement it even before any type of sync feature. The previous Effektio code base and tasks data model already anticipated that to be something people potentially want to customize their objects ... though we never implemented UIs for it.

native/acter/src/api/tasks.rs
gnunicorn: no, more like a way to highlight or mark tasks or task lists ... same as the comment for pins, we never fully implemented any product flow for this just anticipated it is something easy and quickly to add..

app/packages/rust_sdk/lib/acter_flutter_sdk.dart
kumarpalsinh25: Yes, I think this should be good. Thanks for the suggestion :)
`Color convertColor(int? primary, Color fallback) => Color(primary ?? fallback.value);`

PR #1360, opened 2024-02-07
Use only rsvp status from rust side & make its string representation as lowercase

bitfriend:
- Remove flutter definition for rsvp status
- Make `RsvpStatus` as `kebab-case` like `TaskAction`


bitfriend:
`Pending` in `enum RSVP` was needed to show empty status.
There's no real case scenario where user would want to clear his existing selected status.
Take an example of Google Calendar Rsvp.
Initially nothing is selected (Pending) but once selected you can either change selection to one of `Yes/Maybe/No`.

So we will be based on `RsvpStatus` enum from rust side that doesn't contain `Pending`.

native/acter/src/api/common.rs
```diff
@@ -53,6 +53,20 @@ impl OptionBuffer {
...
+
+impl OptionRsvpStatus {
+    pub(crate) fn new(status: Option<RsvpStatus>) -> Self {
+        OptionRsvpStatus { status }
+    }
+
+    pub fn status(&self) -> Option<RsvpStatus> {
+        self.status.clone()
+    }
+}
```
gnunicorn, line 68: This is the third of the same sort of code now... Shouldn't we unify these?

```suggestion
pub struct FfiOption<T> (Option<T>)

impl<T> FfiOption<T> where T: Clone{
    pub(crate) fn new(status: Option<T>) -> Self {
        FfiOption(inner)
    }

    pub fn inner(&self) -> Option<T> {
        self.0.clone()
    }
}

type OptionRsvpStatus = FfiOption<RsvpStatus>;
type OptionString = FfiOption<String>;
type OptionBuffer = FfiOption<FfiBuffer<u8>>;
```

?

gnunicorn (approved):
Just one question on whether we can unify some repeating code .... doesn't have to be in this PR necessarily though...

native/acter/src/api/common.rs
bitfriend: Great idea.
As you noted, `OptionXXXX` code is duplication and I wanted to optimize them, but I didn't know how to implement it.
Now will try your solution.

PR #1063, opened 2023-10-19
Default Notification Settings

gnunicorn:
On top of #1062 and #945  this adds a section to configure the default notification levels:

Screenshot_1697741807

PR #1364, opened 2024-02-08
Default Notification Modes

gnunicorn:
Resurrected and superseeded #1063 - this time it is actually working!

Main problem this fixes compared to the previous attempt was that updates didn't show in the UI. This was due to two parts: 1) the polling wasn't actually invalidating the same item due to type-vs-record problems; 2) the invalidation was just happening to quickly and thus not the updated value was actually shown but still the old one.

Both are addressed and fixed in this PR :) .


gtalha07 (approved):
LGTM!

Issue #1361, opened 2024-02-08
Large space name takes over slides UI in updates editor


gnunicorn:
> Not able to see update items, as the space name takes up the whole view. 

![Image](https://github.com/acterglobal/a3/assets/77483554/d661fb79-c1f3-49f1-bfa1-0ea6ca34f87b)

_Originally posted by @emilvincentz in https://github.com/acterglobal/a3/issues/793#issuecomment-1933632660_

PR #1370, opened 2024-02-09
News Update : General Improvements

kumarpalsinh25:
**1) Fixed the issue related long space name selection.**
fixes, #1361

Large space name

**2) Video slide indicator with play icon overlay:**
Video slide

**3) Make bottom slider more clear and move action button at the top:**
Clear slide bar

### DIFF
--- a/app/lib/features/news/news_utils/news_utils.dart
+++ b/app/lib/features/news/news_utils/news_utils.dart
@@ -56,6 +56,7 @@ class NewsUtils {
 
   //Add image slide
   static Future<void> addImageSlide(WidgetRef ref) async {
+    final clr = Colors.primaries[Random().nextInt(Colors.primaries.length)];
     XFile? imageFile = await ImagePicker().pickImage(
       source: ImageSource.gallery,
     );
@@ -63,6 +64,7 @@ class NewsUtils {
       final slide = NewsSlideItem(
         type: NewsSlideType.image,
         mediaFile: imageFile,
+        backgroundColor: clr,
       );
       ref.read(newsStateProvider.notifier).addSlide(slide);
     }
@@ -70,6 +72,7 @@ class NewsUtils {
 
   //Add video slide
   static Future<void> addVideoSlide(WidgetRef ref) async {
+    final clr = Colors.primaries[Random().nextInt(Colors.primaries.length)];
     XFile? videoFile = await ImagePicker().pickVideo(
       source: ImageSource.gallery,
     );
@@ -77,6 +80,7 @@ class NewsUtils {
       final slide = NewsSlideItem(
         type: NewsSlideType.video,
         mediaFile: videoFile,
+        backgroundColor: clr,
       );
       ref.read(newsStateProvider.notifier).addSlide(slide);
     }
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -1,20 +1,29 @@
 import 'dart:io';
 
 import 'package:acter/common/providers/chat_providers.dart';
+import 'package:acter/common/providers/sdk_provider.dart';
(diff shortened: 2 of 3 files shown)


gnunicorn:
Hmm.. the location of the space feels odd to me, it looks like it is a on-slide-thing. And even there we can still have an overflow issue if the space name is very long, no?

I'd propose to keep the space selector where it was but only show the icon after it was selected. Additionally we could show a very thin name above edit-bar maybe? But I'd assume the update posting user (who are power-users) to recognize the icon.

Another approach could be to ask for the space in the screen before or after, as it is unlikely to change while editing... But I think just the Icon upon the selection is best...

kumarpalsinh25:
@gnunicorn 
When we will have multiple action buttons like "Invite to chat", "Invite to space" etc will come at that time may be below section may look congested so that's why I have put that in top side.

> And even there we can still have an overflow issue if the space name is very long, no?
There are chances of that big name of space which may not fit in entire screen width? If yes then we need think about it and put some validation on naming as too bug name creates many UI glitches in multiple UI parts of the app.

> I'd propose to keep the space selector where it was but only show the icon after it was selected
I really like this flow but only concern about better clarity to user. If there are some space which don't have profile pic set then it tough to judge from icon and increase chances of posting news in wrong space.

> Another approach could be to ask for the space in the screen before or after
This may increase additional step in new posting

gnunicorn:
@kumarpalsinh25 
> Is there any chances of that big name of space which may not fit in entire screen width? 

Yes. There is no limit for the name of a space.

> I really like this flow but only concern about better clarity to user. If there are some space which don't have profile pic set then it tough to judge from icon and increase chances of posting news in wrong space.

Theoretically this is true, but I'd argue that a) you are most likely to post to spaces that haven an icon, as these are the most set-up and most public group, as well as the reader also needs to be able to associate the content and b) you did select the group either before from the UI or with the name shown in the selector. They are still power-users so I doubt this would be a problem. 

Thought: maybe we also want to show a similar small icon on it to indicate you can change it? Similar to the delete-icon on the slides?

kumarpalsinh25:
@gnunicorn 

> Yes. There is no limit for the name of a space.

We can think on putting some restriction on space name at later stage.

> Thought: maybe we also want to show a similar small icon on it to indicate you can change it? Similar to the delete-icon on the slides

Ok perfect. Let's go with this solution. Thanks for sharing thoughts.

gnunicorn:
> We can think on putting some restriction on space name at later stage.

not really. This comes [from the matrix spec](https://spec.matrix.org/v1.9/client-server-api/#mroomname) and we can't enforce it single handedly anyways...

kumarpalsinh25:
@gnunicorn 
Updated changes according to the suggestion.

Side-note:
I'm not able to uploading any image to PR description so it still may showing old screenshots.

gtalha07:
As I see in the [screenshot ](https://github.com/acterglobal/a3/pull/1370#issue-2126579863) you pasted above, the two buttons alignment .i.e plus icon which I think is for slide actions and other for changing colors should be interchanged to have action button at corner right as a UX enhancement.

gtalha07 (approved):
Looks good overall. I would like you to address some thoughts on my UX suggestions!

kumarpalsinh25:
@gtalha07 

> As I see in the https://github.com/acterglobal/a3/pull/1370#issue-2126579863you pasted above, the two buttons alignment .i.e plus icon which I think is for slide actions and other for changing colors should be interchanged to have action button at corner right as a UX enhancement.

Both buttons i.e. `Plus (Add action buttons)` and `Color Changes (For background color)` are applicable to individual slides. :)

kumarpalsinh25:
> Looks good overall. I would like you to address some thoughts on my UX suggestions!

Ya sure, happy to help on that part :)

PR #1374, opened 2024-02-09
Fix error in RsvpStatus parsing

bitfriend:
Will fix error in `camel-case` of `RsvpStatus`.

### DIFF
--- a/native/core/src/events/rsvp.rs
+++ b/native/core/src/events/rsvp.rs
@@ -14,11 +14,11 @@ use super::BelongsTo;
 #[serde(rename_all = "kebab-case", tag = "type")]
 #[strum(serialize_all = "kebab-case")]
 pub enum RsvpStatus {
-    #[serde(alias="Yes")]
+    #[serde(alias = "Yes")]
     Yes,
-    #[serde(alias="Maybe")]
+    #[serde(alias = "Maybe")]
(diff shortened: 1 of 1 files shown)


native/core/src/events/rsvp.rs
```diff
@@ -7,12 +7,18 @@ use strum::Display;
 use super::BelongsTo;
 
 /// RSVP status
+// previolusly accepted only camel-case
+// now accept not only serialization in kebab-case but also alias in camel-case
+// because a lot of RsvpStatus in camel-case were uploaded and if parsing under only kebab-case failed at RsvpStatus in camel-case
```
gnunicorn, line 12:
```suggestion
// previously accepted only PascalCase
// now accept will serialize to kebab-case but also deserialize the previously
// posted PascalCase variants
```


(starting uppercase letter isn't [camelCase, but PascalCase](https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/) ...)

gnunicorn (approved):
just updated the comment.

native/core/src/events/rsvp.rs
bitfriend: Thank you for your correction.

PR #1392, opened 2024-02-14
[Pins]: improve code modularity and readability

gtalha07:
This include enhancements according to general guidelines of code writing, readability and modularity.

### DIFF
--- a/app/lib/features/pins/pages/pin_page.dart
+++ b/app/lib/features/pins/pages/pin_page.dart
@@ -19,8 +19,9 @@ class PinPage extends ConsumerWidget {
   static const titleFieldKey = Key('edit-pin-title-field');
 
   final String pinId;
+  // ignore: use_key_in_widget_constructors
   const PinPage({
-    super.key,
+    Key key = pinPageKey,
     required this.pinId,
   });
 
@@ -52,6 +53,7 @@ class PinPage extends ConsumerWidget {
           ),
         );
       }
+
       if (memb.canString('CanRedactOwn') &&
           memb.userId().toString() == pin.sender().toString()) {
         final roomId = pin.roomIdStr();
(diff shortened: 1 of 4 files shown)


app/lib/features/pins/pages/pin_page.dart
```diff
@@ -21,31 +20,31 @@ class PinPage extends ConsumerWidget {
 
   final String pinId;
   const PinPage({
-    Key key = pinPageKey,
+    super.key,
```
gnunicorn, line 23: isn't this ` = pinPageKey` needed for the integration tests?

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -115,14 +93,52 @@ class PinPage extends ConsumerWidget {
...
+    required ActerPin pin,
+    required String roomId,
+  }) {
+    showAdaptiveDialog(
+      context: context,
+      builder: (context) => RedactContentWidget(
+        title: 'Remove this pin',
+        eventId: pin.eventIdStr(),
+        onSuccess: () {
+          if (context.mounted) {
```
gnunicorn, line 118:
```suggestion
          if (context.mounted && context.canPop()) {
```

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -145,28 +161,28 @@ class PinPage extends ConsumerWidget {
...
-                        decoration: InputDecoration(
-                          enabledBorder:
-                              pinEdit.editMode ? null : InputBorder.none,
-                          focusedBorder:
-                              pinEdit.editMode ? null : InputBorder.none,
-                          filled: false,
-                        ),
-                        onChanged: (val) => pinEditNotifer.setTitle(val),
-                      ),
+                title: Visibility(
```
gnunicorn, line 164: this probably should also be a function for the widget now...

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -145,28 +161,28 @@ class PinPage extends ConsumerWidget {
...
+                  visible: !pinEdit.editMode,
+                  replacement: TextFormField(
+                    key: PinPage.titleFieldKey,
+                    initialValue: acterPin.title(),
+                    readOnly: !pinEdit.editMode,
```
gnunicorn, line 169: this field is already checked before and can ever only be one value here.

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -145,28 +161,28 @@ class PinPage extends ConsumerWidget {
...
+                    style: Theme.of(context).textTheme.titleLarge,
+                    decoration: InputDecoration(
+                      enabledBorder: pinEdit.editMode ? null : InputBorder.none,
+                      focusedBorder: pinEdit.editMode ? null : InputBorder.none,
```
gnunicorn, line 173: same as above

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -115,14 +93,52 @@ class PinPage extends ConsumerWidget {
...
-    }
+    return Visibility(
+      visible: actions.isNotEmpty,
+      child: PopupMenuButton<String>(
+        key: PinPage.actionMenuKey,
+        icon: const Icon(Atlas.dots_vertical_thin),
+        itemBuilder: (ctx) => actions,
+      ),
+    );
+  }
```
gnunicorn, line 104: I understand this code was there before in a similar matter _but_ this actions may never be empty: If you can't edit or redact the item, you can always report it. The `if` above might not be 100% correct about this, but overall the menu is always existing...

```suggestion
    return PopupMenuButton<String>(
        key: PinPage.actionMenuKey,
        icon: const Icon(Atlas.dots_vertical_thin),
        itemBuilder: (ctx) => actions,
    );
  }
```

app/lib/features/pins/pages/pins_page.dart
```diff
@@ -18,76 +17,86 @@ class PinsPage extends ConsumerWidget {
...
-              ),
-            ),
-            loading: () => const SliverToBoxAdapter(
-              child: Center(
-                child: Text('Loading'),
-              ),
+          iconSize: 28,
+          color: Theme.of(context).colorScheme.surface,
+          onPressed: () => context.pushNamed(
+            Routes.actionAddPin.name,
```
gnunicorn, line 49: I know this wasn't technically changed in this PR but we should probably have the same _hasAnySpacesWithPermission_-check rather than always showing the button. If you don't want to do it in this PR, please take this and make an issue out of it so we can track it.

gnunicorn (approved):
fine overall. but some logic concerns

app/lib/features/pins/pages/pins_page.dart
gtalha07: added a FIXME and opening up issue!

PR #1401, opened 2024-02-16
Fix bug in getting of rsvp status

bitfriend:
`data.status()?.tag` compiled by `ffi-gen` is occurring error.
To avoid it, I will implement the additional method `status_str()` in `OptionRsvpStatus`.


gnunicorn:
Considering how often that broke the last few days alone, maybe we should consider adding an end-to-end-ui-test for RSVPing to events?!?

bitfriend:
Now status change works fine, but refresh doesn't work.
https://github.com/bitfriend/a3/blob/2c0cc077104c5ba6535bfbc3f1e17cda249e7d95/app/lib/features/events/pages/event_page.dart#L328-L331
Here `ref.refresh()` doesn't work.

app/lib/features/events/pages/event_page.dart
```diff
@@ -133,19 +132,16 @@ class CalendarEventPage extends ConsumerWidget {
...
-              break;
-            case RsvpStatusTag.Maybe:
-              rsvp = <RsvpStatusTag?>{RsvpStatusTag.Maybe};
-              break;
-            case RsvpStatusTag.No:
-              rsvp = <RsvpStatusTag?>{RsvpStatusTag.No};
-              break;
-          }
+        switch (data.statusStr(true)) {
+          case 'Yes':
```
kumarpalsinh25, line 136: Any reason for making switch case with hardcoded string instead of using enum?

kumarpalsinh25 (commented):
Looks good overall.

Side-note:
event_page.dart is going to be removed on https://github.com/acterglobal/a3/pull/1412.
So do rebase and required changes accordingly.

bitfriend:
> `data.status()?.tag` compiled by `ffi-gen` is occurring error. To avoid it, I will implement the additional method `status_str()` in `OptionRsvpStatus`.

This is why making switch case with hardcoded string instead of using enum.

app/lib/features/events/pages/event_page.dart
bitfriend: [This](https://github.com/acterglobal/a3/pull/1401#issue-2138226873) is that reason.

bitfriend:
Now editing page shows the delayed update of RSVP status.
Please watch log and segment button carefully.

https://github.com/acterglobal/a3/assets/48786056/332ff0d9-144f-4242-bc00-82b1a13f9d07

bitfriend:
Made `CalendarEventPage` as stateful to fix bug that rsvp status not updated.

app/lib/features/events/pages/event_page.dart
```diff
@@ -15,28 +14,60 @@ import 'package:flutter_easyloading/flutter_easyloading.dart';
...
+}
+
+class _CalendarEventPageState extends ConsumerState<CalendarEventPage> {
+  RsvpStatusTag? rsvpStatus;
+
+  @override
+  void initState() {
+    super.initState();
+    AsyncValue<OptionRsvpStatus> myRsvpStatus =
+        ref.read(myRsvpStatusProvider(widget.calendarId));
```
gnunicorn, line 35: why replace the watch with a singular read? This doesn't sound like a good idea..

native/acter/src/api/common.rs
```diff
@@ -69,6 +69,29 @@ impl OptionRsvpStatus {
...
+                    }
+                    RsvpStatus::Maybe => {
+                        return Some("Maybe".to_owned());
+                    }
+                    RsvpStatus::No => {
+                        return Some("No".to_owned());
+                    }
+                }
+            }
+            None
```
gnunicorn, line 89: this could still be a much briefer map ...
```suggestion
            status.map(|status| => 
                match status {
                    RsvpStatus::Yes => "Yes",
                    RsvpStatus::Maybe => "Maybe",
                    RsvpStatus::No => "No",
                }.to_owned())
```

native/acter/api.rsh
```diff
@@ -217,6 +217,9 @@ object OptionBuffer {
 object OptionRsvpStatus {
     /// get status
     fn status() -> Option<RsvpStatus>;
+
+    /// get status in string
+    fn status_str(pascal_case: bool) -> Option<string>;
```
gnunicorn, line 222: why do we want this to have pascal_case support? It doesn't make the mapping easier either ... Why not just map on `'yes', 'no' and 'maybe'` to generate the dart enum?

gnunicorn (commented):
I am not quite getting the approach taking and why certain things are necessary. please elaborate...

app/lib/features/events/pages/event_page.dart
bitfriend: As you see in below code, `setState()` is called when `Yes`, `Maybe` or `No` is tapped.
As we know, `setState()` is meaningless about `ref.watch()`.

native/acter/src/api/common.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1401/commits/9b29ff37073e4a5b50062bd1c4b1792a9e0686f5

native/acter/api.rsh
bitfriend: `pascal_case` is used [here](https://github.com/bitfriend/a3/blob/25f8077dc1591013109cf1f9a2c671746be6edfa/app/lib/features/events/widgets/events_item.dart#L92).

gnunicorn: okay, the code is like this now, but we shouldn't really do that. Just think of translations and it becomes clear that we can't just use the string-representation of the enum-case and show that. So we should rather remap it to the proper enum case and then only use that enum on the dart-side ...

app/lib/features/events/pages/event_page.dart
```diff
@@ -324,29 +332,35 @@ class CalendarEventPage extends ConsumerWidget {
...
+                      unselectedColor: Colors.transparent,
+                      selectedTextStyle: const TextStyle(color: Colors.white),
+                      unselectedTextStyle: const TextStyle(color: Colors.grey),
+                      // borderWidth: 0.7,
+                      // borderRadius: 32,
+                      onSegmentTapped: (value) async {
+                        await onRsvp(context, value, ref);
+                        setState(() => rsvpStatus = value);
+                        // refresh outer page
+                        ref.invalidate(myRsvpStatusProvider(widget.calendarId));
```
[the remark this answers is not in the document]
bitfriend: image
image
When 2nd page changes, 1st page should be updated too.
So I called `ref.invalidate()` about `myRsvpStatusProvider`.

gnunicorn (approved):

Issue #910, opened 2023-09-05
Document flutter & rust best practices


gnunicorn:
We have learned a few lessons using riverpod and flutter already, let's start documenting these - especially those that can not be enforced by lints - including a decent explanation as to why that matters. Like:

- avoiding `requireValue` and `ref.watch(clientProvider)!`
- general try-catches (rather use AsyncValues)
- router for state management
- proper dependency management (ref.watch rather than ref.read(someProvider.notifier).someFunction())`)

gnunicorn:
Flutter responsive advices:
 - Use `MediaQuery` for sizes
 - for MaxValue use `BoxConstraints`
 - use `isDesktop` rarely! It is almost always misleading in terms showing something because of sizes available (always remember: the latest iPad touch has a higher resolution than the also supported MacBook Air 2017! - `isDesktop` is an inadequate approximation of what we actually want and that is responsiveness according to the available space)

gnunicorn:
- Colors may not be hardcoded in the code, but should be pulled from the AppTheme.

gnunicorn:
Rust as well:
 - no `unwrap` outside of test-code - use `expect` if you must
 - use `expect` to clarify _developer_ assumptions. e.g. unpacking an Option that the current code _ensures_ is filled or if the surrounding `if` is meant to ensure the assumption can't be broken.
 - use `?` and `bail!(...)` to verify any user input, don't `panic` / `expect` / `unwrap` on user based input. Ensure the
 - use `.context(...)?` on `Option`s only, and ensure you are giving a user understandale error message.

bitfriend:
`?` is handling `Result<>`.
every time you have a `?`, it will effectively expand to:
```rs
let value = match value {
     Ok(v) -> v,
    Err(e) -> return Err(e))
};
```

gnunicorn:
riverpod:
 - don't use setState (it almost always silently fails if we do) but pin to the provider
 - do as much calculation in the providers
 - be as specific as possible on the data from your provider, use `.select()` if you only need one parameter, to avoid unnecessary re-renders
 - provider chaining is good, use it!

gnunicorn:
don't `requireValue!` but `valueOrNull?` always.

gnunicorn:
flutter/dart general: `!` only if you checked for `!= null` pretty much right before.

gnunicorn:
started draft here: https://ghost.acter.global/dev/code-style-best-practices/

gnunicorn:
[...]

PR #1404, opened 2024-02-16
Flutter code style documentation.

kumarpalsinh25:
Reference:
https://github.com/acterglobal/a3/issues/910

### DIFF
--- a/docs/content/docs/contributing/flutter-code-style-guide.md
+++ b/docs/content/docs/contributing/flutter-code-style-guide.md
@@ -1,0 +1,78 @@
+## Flutter - code style guide
+
+👉 This style guide describes the preferred style for code written as part of the [`acter` repository](https://github.com/acterglobal/a3)
+
+### Overview
+This document describes our approach to designing and programming Flutter development of the Acter App, from high-level architectural principles all the way to indentation rules.
+
+The primary goal of these style guidelines is to improve code readability so that everyone, whether reading the code for the first time or maintaining it for years, can quickly determine what the code does. Secondary goals are to design systems that are simple, to increase the likelihood of catching bugs quickly, and avoiding arguments when there are disagreements over subjective matters.
+
+That document is focused more about Flutter conventions.
+
+### App Theme
+Use theme classes for the generalise theme of the different components and avoid styling at the component level until and unless it is required.
+
+**Key things to note:**
+- Colors may not be hardcoded in the code, but should be pulled from the AppTheme.
+- Avoid component level styling
+
+Reference: https://docs.flutter.dev/cookbook/design/themes
+
+**Buttons Guide:**
+
+Primary Button
+```
+ ElevatedButton(
+     onPressed: () {},
+    child: ...,
+),
+```
+
+Secondary Button
+```
+ OutlinedButton(
+     onPressed: () {},
+    child: ...,
+),
+```
+
+### Code structure
+There are few basis standard we would like to follow in futter code.
+
+
+**1. General Guide**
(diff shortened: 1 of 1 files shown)


docs/content/docs/contributing/flutter-code-style-guide.md
```diff
@@ -0,0 +1,78 @@
...
+),
+```
+
+### Code structure
+There are few basis standard we would like to follow in futter code.
+
+
+**1. General Guide**
+- General try-catches (rather use AsyncValues)
+- Router for state management
```
gnunicorn, line 45: can you elaborate on these? I am not sure I understand them...

docs/content/docs/contributing/flutter-code-style-guide.md
```diff
@@ -0,0 +1,78 @@
...
+
+**1. Scaffold usage**
+Every parent page should use Scaffold widget to structure entire screen layout.
```
gnunicorn, line 48: I think we agreed to do that differently because of routing. This only applies to _full screen_-pages (which we have a smaller number of).

docs/content/docs/contributing/flutter-code-style-guide.md
```diff
@@ -0,0 +1,78 @@
...
+
+**2. Scrolling usage**
+- Use `SingleChildScrollView` in all the general use cases where need widgets need to be scrollable
+- Use `CustomScrollView` with its `Slivers` when complex scrolling behavior is required
+
+**3. Avoid code nesting**
+- Use module approach of widget layout and avoid huge nesting in code
+- Create sperate `methods` or `class` for common widgets
```
gnunicorn, line 56: I think examples would be useful. And we said that the rule of thumb is to reduce large nesting and reduce complexity...

docs/content/docs/contributing/flutter-code-style-guide.md
```diff
@@ -0,0 +1,78 @@
...
+
+**4. State management with Riverpod**
+- Avoid unneccesary usage of provider where things can be handle without it
+- Use appropriate River proivder type based on the use-case
```
gnunicorn, line 60:
```suggestion
- Use appropriate Riverpod provider type based on the use-case
```

docs/content/docs/contributing/flutter-code-style-guide.md
```diff
@@ -0,0 +1,78 @@
...
+- Avoiding `requireValue` and `ref.watch(clientProvider)!`
+- Proper dependency management (ref.watch rather than ref.read(someProvider.notifier).someFunction())`)
+
+**5. Naming Convention**
+- Use Names That Are Detailed and Clear
+- Observe the Dart Package Naming Guidelines
+- File names should contain singular nouns.
+- File Names with Prefixes for Related Components
+- Keep consistency and reuse in mind
```
gnunicorn, line 69:
```suggestion
- Keep consistency and reuse in mind
  e.g. `ThingDetailsPage` and `ThingListItem` are common practice.
```

gnunicorn (approved):
This is very brief so far. Would be good to extend it to a) show use cases and b) give the reasoning behind that guideline to explain _why_ we chose this (and to ensure that when the world progresses and those reasons aren't valid anymore we don't keep sticking to a rule for the sake of the rule)

docs/content/docs/contributing/flutter-code-style-guide.md
kumarpalsinh25: Yes true, no worries. I will remove this point as this is not applicable in most of the pages in our app.

PR #1412, opened 2024-02-19
Refactor : Event details UI-UX

kumarpalsinh25:
Event Details


app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+              _buildEventAppBar(calendarEvent),
+              _buildEventBody(calendarEvent),
+            ],
+          );
+        },
+        error: (error, stackTrace) => SliverToBoxAdapter(
+          child: Text('Error loading event due to $error'),
+        ),
+        loading: () => const SliverToBoxAdapter(
+          child: Center(child: CircularProgressIndicator()),
```
gnunicorn, line 47: can I has a skeleton here please?!?

gnunicorn, line 47: why are these `SliverToBoxAdatapter`s if they are just in the `body` and not in `slivers` of `Scaffold`? Does that even work?

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+                const Text('Remove Event'),
+              ],
+            ),
+          ),
+        ]);
+      }
+    }
+
+    //Report Event Action
+    else {
```
gnunicorn, line 143: please, for clarity.
```suggestion
    } else {
    //Report Event Action
```

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+      actions.add(
```
gnunicorn, line 144: also this is incorrectly bound to there not being a membership .... which would also mean you can't even report it though ... I think this should live in the actions regardless from there being any `membership`.

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+              const Text('Report Event'),
+            ],
+          ),
+        ),
+      );
+    }
+
+    if (actions.isEmpty) {
+      return const SizedBox.shrink();
+    }
```
gnunicorn, line 174: because there's always the option to report, this is never happening.
```suggestion
```

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+
+    return PopupMenuButton(
+      itemBuilder: (ctx) => actions,
+    );
```
gnunicorn, line 178: as this only happens later, why not push all that menu generation off to when the builder is called?

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -0,0 +1,412 @@
...
+            subtitle: Text('$startDate - $endDate'),
+          ),
+          const ListTile(
+            leading: Icon(Icons.link),
+            title: Text('Not set'),
+          ),
+          const ListTile(
+            leading: Icon(Atlas.map_navigation),
+            title: Text('Not set'),
+          ),
```
gnunicorn, line 381: Probably want to omit that for the time being.
```suggestion
```

gnunicorn (approved):
Just some minor style aspects. Good to go otherwise. But needs a merge conflict resolution.

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: > why are these SliverToBoxAdatapters if they are just in the body and not in slivers of Scaffold?

My bad, `SliverToBoxAdapter` is by  mistake placed here. Let me remove it.

> Does that even work?   

I don't think so it will work and lead to potential error.

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: I haven't change anything here, directly copy this menu options from previous implementation.
https://github.com/acterglobal/a3/blob/b8085a3f36831798485ab1257406e29222a48620/app/lib/features/events/pages/event_page.dart#L90

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: Sorry but not getting you. Can you please more details here?

app/lib/features/events/pages/event_details_page.dart
gnunicorn: I understand. It is still wrong though. But if you prefer not to change this right now, fine with me.

app/lib/features/events/pages/event_details_page.dart
gnunicorn: not that important ...

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: > But if you prefer not to change this right now, fine with me.

Nope, better to do it now. I was just not sure where it was correct or not. Let me do that change and show `Report Event ` regardless of any user membership.

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: Done :)

Screenshot 2024-02-23 at 5 54 38 PM

Issue #1413, opened 2024-02-19
Update display bugs


emilvincentz:
Viewed on iphone 12

![Image](https://github.com/acterglobal/a3/assets/77483554/bd87ec7a-84b5-4c05-a9bf-4b232cdf9bdd)
![Image](https://github.com/acterglobal/a3/assets/77483554/312cf8d8-3364-44cb-83a9-14a068dd0fcf)
![Image](https://github.com/acterglobal/a3/assets/77483554/5ecbc058-ff0d-4508-93fd-051b787425dd)

kumarpalsinh25:
@emilvincentz 
Are you testing on latest nightly build?

Issue #1411, opened 2024-02-17
News Update: Resume Text Update gets broken


kumarpalsinh25:
https://github.com/acterglobal/a3/assets/51526480/1d63ce7b-57e1-4fda-98d4-ac5e1b7ffa57

Issue #1407, opened 2024-02-16
News Update : Video doesn't fit to the screen size


emilvincentz:
The video preview is very small. But when in the creation form but also when displayed.

![Image](https://github.com/acterglobal/a3/assets/77483554/36b3d50a-6ead-423f-9aff-4d8bd0c964ee)
![Image](https://github.com/acterglobal/a3/assets/77483554/5752c27c-f7c0-4d43-9238-d0fd742f5833)

kumarpalsinh25:
@emilvincentz 
Can you share the video file that you are using in this news post?

emilvincentz:
yep.

emilvincentz:
https://github.com/acterglobal/a3/assets/77483554/5614174c-df20-4230-a7a8-ce7039bfab81

@kumarpalsinh25

Issue #1405, opened 2024-02-16
news editor: Color selector for images defunct 


gnunicorn:
Doesn't update the background anymore - but it appears the inner data is actually changed, just that the UI doesn't update it properly anylonger:

https://github.com/acterglobal/a3/assets/40496/079fa847-3232-4d31-bb4b-034db2072439

PR #1414, opened 2024-02-19
News Update : General Fixes

kumarpalsinh25:
- Fixes, #1405 
- Fixes, #1411 
- Fixes, #1407 

Reference Video:
https://github.com/acterglobal/a3/assets/51526480/981131c6-5a0d-46b3-9e81-179f0de87a62

- Fixes, #1389
- Fixes, #1400
- Fixes, #1413

Reference Video:
https://github.com/acterglobal/a3/assets/51526480/6ee35c3b-352f-48a7-bea2-56b8b0102b51

### DIFF
--- a/app/lib/common/widgets/acter_video_player.dart
+++ b/app/lib/common/widgets/acter_video_player.dart
@@ -54,14 +54,28 @@ class _ActerVideoPlayerState extends State<ActerVideoPlayer> {
 
   @override
   Widget build(BuildContext context) {
+    if (!_controller.value.isInitialized) {
+      return Container();
+    }
+
+    final Size screenSize = MediaQuery.of(context).size;
+    final double aspectRatio = _controller.value.aspectRatio;
+
     return AspectRatio(
-      aspectRatio: _controller.value.aspectRatio,
-      child: Stack(
-        alignment: Alignment.center,
-        children: [
-          VideoPlayer(_controller),
-          controlsOverlay(),
-        ],
+      aspectRatio: aspectRatio,
+      child: FittedBox(
+        fit: BoxFit.cover,
+        child: SizedBox(
+          width: screenSize.width,
+          height: screenSize.height,
+          child: Stack(
+            alignment: Alignment.center,
+            children: [
+              VideoPlayer(_controller),
+              controlsOverlay(),
+            ],
+          ),
+        ),
       ),
     );
   }
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -467,7 +467,7 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
       ref.invalidate(newsStateProvider);
       // Navigate back to update screen.
       Navigator.of(context).pop();
-      context.pushNamed(Routes.main.name); // go to the home / main updates
+      context.goNamed(Routes.main.name); // go to the home / main updates
     } catch (err) {
       EasyLoading.showError('$displayMsg failed: \n $err');
     }
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -42,9 +42,9 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
   EditorState textEditorState = EditorState.blank();
   NewsSlideItem? selectedNewsPost;
 
-  //Build UI
   @override
-  Widget build(BuildContext context) {
+  void initState() {
+    super.initState();
     ref.listenManual(newsStateProvider, fireImmediately: true,
         (prevState, nextState) async {
       if (nextState.currentNewsSlide != null && // we have a new one
@@ -81,6 +81,11 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
         setState(() => selectedNewsPost = nextState.currentNewsSlide);
       }
     });
+  }
+
+  //Build UI
+  @override
+  Widget build(BuildContext context) {
     return Scaffold(
       appBar: appBarUI(context),
       body: bodyUI(context),
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -45,7 +45,8 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
   //Build UI
   @override
   Widget build(BuildContext context) {
-    ref.listen(newsStateProvider, (prevState, nextState) async {
+    ref.listenManual(newsStateProvider, fireImmediately: true,
+        (prevState, nextState) async {
       if (nextState.currentNewsSlide != null && // we have a new one
               nextState.currentNewsSlide?.type ==
                   NewsSlideType.text && // and it is a text type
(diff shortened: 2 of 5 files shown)


app/lib/features/news/pages/add_news_page.dart
```diff
@@ -302,20 +274,37 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
...
 
   Widget slideTextPostUI(BuildContext context) {
-    final backgroundColor = ref.watch(
-      newsStateProvider.select((i) => i.currentNewsSlide?.backgroundColor),
-    );
+    //Get document data
+    final document = selectedNewsPost!.html != null
+        ? ActerDocumentHelpers.fromHtml(selectedNewsPost!.html!)
+        : ActerDocumentHelpers.fromMarkdown(selectedNewsPost!.text ?? '');
+    textEditorState = EditorState(document: document);
```
gnunicorn, line 281: so here is a thing I have seen you do resonable bit that I thought was a no-no: changing widget state within the `build` function (or its sub function). I _thought_ that a state change like that triggers a rebuild and therefore it should only be done outside of `buildContext` areas as we change the underlying state _while_ trying to render it ... that's why this is usually offloaded in some side function or better even direct callbacks rather than implicit changes triggered through a ref.watch-update cycle ....

Now, you have replaced that offloaded code by doing it implicitly during render time and that feels ... odd at least. Your thoughts?

gnunicorn: alternative approach in #1418 .

app/lib/common/widgets/acter_video_player.dart
```diff
@@ -54,14 +54,28 @@ class _ActerVideoPlayerState extends State<ActerVideoPlayer> {
 
   @override
   Widget build(BuildContext context) {
+    if (!_controller.value.isInitialized) {
+      return Container();
+    }
+
```
gnunicorn, line 60: how are we making sure the build is called when initialization updated?

app/lib/features/news/pages/add_news_page.dart
```diff
@@ -45,7 +45,8 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
   //Build UI
   @override
   Widget build(BuildContext context) {
-    ref.listen(newsStateProvider, (prevState, nextState) async {
+    ref.listenManual(newsStateProvider, fireImmediately: true,
```
gnunicorn, line 48: why did you switch this to `listenManual`? `listen` has the same features and according to the [docs], listenManual isn't save to use here:
> As opposed to [listen](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/WidgetRef/listen.html), [listenManual](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/WidgetRef/listenManual.html) is not safe to use within the build method of a widget. Instead, [listenManual](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/WidgetRef/listenManual.html) is designed to be used inside [State.initState](https://api.flutter.dev/flutter/widgets/State/initState.html) or other [State](https://api.flutter.dev/flutter/widgets/State-class.html) lifecycles.

If the plan was to move it to `initState` (which I think is a good goal), that move should also happen ;) .

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: > alternative approach in https://github.com/acterglobal/a3/pull/1418 .

Thanks @gnunicorn for this.

app/lib/features/news/widgets/news_item_slide/text_slide.dart
```diff
@@ -4,42 +4,91 @@ import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
...
+        duration: const Duration(milliseconds: 300),
+        curve: Curves.ease,
+      );
+    } else if (!topPosition && outOfRange && offset > maxScrollExtent + 150) {
+      widget.pageController.nextPage(
+        duration: const Duration(milliseconds: 300),
+        curve: Curves.ease,
+      );
+    }
+  }
```
gnunicorn, line 51: yeah not able to review this for sure ... so ... I just trust this is set up as we want ...

app/lib/features/news/widgets/news_item_slide/text_slide.dart
```diff
@@ -4,42 +4,91 @@ import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
...
+
+  @override
+  Widget build(BuildContext context) {
+    final slideContent = widget.slide.msgContent();
     final formattedText = slideContent.formattedBody();
+
     return Container(
-      padding: const EdgeInsets.all(20),
-      color: bgColor,
+      padding: const EdgeInsets.symmetric(horizontal: 26,vertical: 70),
```
gnunicorn, line 65:
```suggestion
      padding: const EdgeInsets.symmetric(horizontal: 26, vertical: 70),
```

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: > why did you switch this to listenManual?

Because `listen` doesn't have `fireImmediately`
Screenshot 2024-02-20 at 7 21 56 PM

kumarpalsinh25: > If the plan was to move it to initState (which I think is a good goal), that move should also happen ;) .

Ahh, Good notice. Let me move that to the `initState()` :)

app/lib/features/news/widgets/news_item_slide/text_slide.dart
kumarpalsinh25: Yes it is all about scroll behaviour things.

app/lib/features/news/widgets/news_item_slide/text_slide.dart
kumarpalsinh25: Which value is changed in suggestion?

app/lib/features/news/pages/add_news_page.dart
gnunicorn: @kumarpalsinh25 [according to the docs listen does have that](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/Ref/listen.html) ... That's how I knew about it, I didn't even know about `listenManual` before I saw it here ...

maybe we need to/want to upgrade our riverpod dependency?

app/lib/features/news/widgets/news_item_slide/text_slide.dart
gnunicorn: no value, just a space after the comma for style.

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: > @kumarpalsinh25 [according to the docs listen does have that](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/Ref/listen.html) ... That's how I knew about it, I didn't even know about listenManual before I saw it here ...

👍️

> maybe we need to/want to upgrade our riverpod dependency?

Yes we can go with riverpod upgrade to get some benefits of latest changes in dependency.

PR #1418, opened 2024-02-20
Fix several state related News Editor bugs

### LINKED ISSUE #1405: news editor: Color selector for images defunct

gnunicorn:
There was a misunderstanding in how the state is managed (stateful vs. stateless) that lead to a premature optimization on behalf of the stateful texteditor by trying to have its state recalculated the least often. This is now less selective and the state recalculation checks are expanded instead. Fixing several bugs like color changing problems.

fixes #1405 :

https://github.com/acterglobal/a3/assets/40496/1f5eb18f-3fc4-4532-a476-7eaddac84dab

### DIFF
--- a/app/lib/features/news/pages/add_news_page.dart
+++ b/app/lib/features/news/pages/add_news_page.dart
@@ -45,9 +45,14 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
   //Build UI
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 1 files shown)


kumarpalsinh25 (approved):
Great work. Thanks for noting it and doing require changes.

PR #1420, opened 2024-02-20
Refactoring of ref.watch & ref.read

bitfriend:
- miscellaneous refactoring
- replace useless `ref.watch()` with `ref.read()`. We should use `ref.read()` on button click.
- replace useless `ref.read()` with `ref.watch()`. We should use `ref.watch()` on rendering fn.


kumarpalsinh25 (approved):
LGTM!

PR #1448, opened 2024-02-26
Live redaction support for spaces.

### LINKED ISSUE #947: Live-redactions are not applied
Following #943 , if I redact any content (Pin, NewsEntry or Event), this will not remove them from the local view because:
 - [ ] a redaction that comes down live (rather than a redacted event) is currently not applied to the executor 
 - [ ] the executor does not issue a refresh of the corresponding UI component.

This unfortunately means that the user redacting the content will ignore the existing redaction issued and because we only sync new events never learns about the redaction. As well as anyone currently online, as their caches are not refreshed and the update not applied until they fresh-sync...

gnunicorn:
Fixes #947 by listening for redaction events coming down the pipe live and sending them over to the internal executor. 

Does not yet fix UI issues related to the redaction (like button missing for Pins or the area not disappearing for news)

See it in action (ignore that the other update has the same image, it is previous testing data but notice that the background color switches):

https://github.com/acterglobal/a3/assets/40496/b65e1e3c-06e8-46cf-8e86-0a7450201c39

### DIFF
--- a/.changes/1448-live-redactions.md
+++ b/.changes/1448-live-redactions.md
@@ -1,0 +1,1 @@
+- [fix] Redactions are now applied live when they are received. No more dangeling old objects (e.g. Updates, Events, Pins) in your spaces until a proper relogin but dealing with them on the fly.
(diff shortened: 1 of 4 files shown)


bitfriend (approved):
Looks good to me

Issue #1294, opened 2024-01-18
Add update : Support for adding embedded action(s) (starting with Share event for now).


kumarpalsinh25:
![Screenshot 2024-01-18 at 12 22 55 PM](https://github.com/acterglobal/a3/assets/51526480/9088298d-8eb4-4258-bdce-ac8b1a2e57a5)

PR #1450, opened 2024-02-27
News action buttons : Share Event

kumarpalsinh25:
**Work Summary:**
- - -
- Added news action button for "Share event"
- Removed the code of "Invite to space" & "Invite to chat" action button as they are no longer in use as of now

**Reference Video:**
- - -
https://github.com/acterglobal/a3/assets/51526480/9c71151a-94f2-4431-9e72-5a3102857727

**Issue covered:**
- - -
- fixes, #1294 
- fixes, #1316

### DIFF
--- a/.changes/1450-news-action-buttons.md
+++ b/.changes/1450-news-action-buttons.md
@@ -1,1 +1,1 @@
-- [new] : Now users can add "Share Event" action button in news update slides
+- [new] : You can now include links to Events right into the Updates sent to your space members. Allowing them to easily get access and respond to any upcoming Events right from the Updates screen.
--- a/.changes/1450-news-action-buttons.md
+++ b/.changes/1450-news-action-buttons.md
@@ -1,0 +1,1 @@
+- [new] : Now users can add "Share Event" action button in news update slides
--- a/app/lib/common/widgets/event/event_selector_drawer.dart
+++ b/app/lib/common/widgets/event/event_selector_drawer.dart
@@ -1,0 +1,90 @@
+import 'package:acter/features/events/providers/event_providers.dart';
+import 'package:acter/features/events/widgets/events_item.dart';
+import 'package:acter/features/events/widgets/skeletons/event_list_skeleton_widget.dart';
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+const Key selectEventDrawerKey = Key('event-widgets-select-event-drawer');
+
+Future<String?> selectEventDrawer({
+  required BuildContext context,
+  required String spaceId,
+  Key? key = selectEventDrawerKey,
+  Widget? title,
+}) async {
+  final selected = await showModalBottomSheet(
+    showDragHandle: true,
+    enableDrag: true,
+    context: context,
+    isDismissible: true,
+    builder: (context) => Consumer(
+      builder: (context, ref, child) {
+        final events = ref.watch(spaceEventsProvider(spaceId));
+        return events.when(
+          data: (eventsList) {
+            return Column(
+              key: key,
+              mainAxisAlignment: MainAxisAlignment.start,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              mainAxisSize: MainAxisSize.min,
+              children: <Widget>[
+                Padding(
+                  padding: const EdgeInsets.symmetric(horizontal: 20.0),
+                  child: Row(
+                    children: [
+                      Expanded(
+                        child: title ?? const Text('Select event'),
+                      ),
+                      OutlinedButton.icon(
+                        icon: const Icon(Atlas.minus_circle_thin),
+                        onPressed: () {
+                          Navigator.pop(context, null);
+                        },
+                        label: const Text('Clear'),
+                      ),
+                    ],
+                  ),
+                ),
+                Flexible(
+                  child: eventsList.isEmpty
+                      ? Container(
+                          height: 200,
+                          alignment: Alignment.center,
+                          child: const Text('No events found'),
+                        )
+                      : ListView.builder(
+                          padding: const EdgeInsets.all(8),
+                          itemCount: eventsList.length,
+                          itemBuilder: (context, index) {
+                            final event = eventsList[index];
+                            return EventItem(
+                              event: event,
+                              isShowRsvp: false,
+                              onTapEventItem: (event) {
+                                Navigator.pop(context, event);
+                              },
+                            );
+                          },
+                        ),
+                ),
+              ],
+            );
+          },
+          error: (error, stack) {
+            return Center(
+              child: Text('Failed to load: $error'),
+            );
+          },
+          loading: () =>
+              const SizedBox(height: 500, child: EventListSkeleton()),
+        );
+      },
+    ),
+  );
+  if (selected == '') {
+    // in case of being cleared, we return null
+    return null;
+  }
+  return selected;
+}
--- a/app/lib/features/events/widgets/events_item.dart
+++ b/app/lib/features/events/widgets/events_item.dart
@@ -9,16 +9,29 @@ import 'package:go_router/go_router.dart';
 
 class EventItem extends ConsumerWidget {
   final CalendarEvent event;
+  final Function(String)? onTapEventItem;
+  final bool isShowRsvp;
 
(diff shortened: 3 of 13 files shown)


gnunicorn:
OH THIS IS SOO COOL!

.changes/1450-news-action-buttons.md
```diff
@@ -0,0 +1 @@
+- [new] : Now users can add "Share Event" action button in news update slides
```
gnunicorn, line 1: we need to bolster this a lot more!
```suggestion
- [new] : You can now include links to Events right into the Updates sent to your space members. Allowing them to easily get access and respond to any upcoming Events right from the Updates screen.
```

app/lib/common/widgets/event/event_selector_drawer.dart
```diff
@@ -0,0 +1,90 @@
...
+              ],
+            );
+          },
+          error: (error, stack) {
+            return Center(
+              child: Text('Failed to load: $error'),
+            );
+          },
+          loading: () =>
+              const SizedBox(height: 500, child: EventListSkeleton()),
```
gnunicorn, line 80: :clap:

app/lib/common/widgets/event/event_selector_drawer.dart
```diff
@@ -0,0 +1,90 @@
...
+        );
+      },
+    ),
+  );
+  if (selected == '') {
```
gnunicorn, line 85: shouldn't this rather be a 
```suggestion
  if (selected.isEmpty) {
```

app/lib/features/events/widgets/events_item.dart
```diff
@@ -35,7 +48,10 @@ class EventItem extends ConsumerWidget {
               ),
             ),
             const SizedBox(width: 10),
-            _buildRsvpStatus(context, ref),
+            Visibility(
+              visible: isShowRsvp,
+              child: _buildRsvpStatus(context, ref),
+            ),
```
gnunicorn, line 54: There is a strong want in this team to use `Visibility` a lot more often than needed. And I'd like to quote the flutter docs about it again:
> Using this widget is not necessary to hide children. The simplest way to hide a child is just to not include it, or, if a child must be given (e.g. because the parent is a [StatelessWidget](https://api.flutter.dev/flutter/widgets/StatelessWidget-class.html)) then to use [SizedBox.shrink](https://api.flutter.dev/flutter/widgets/SizedBox/SizedBox.shrink.html) instead of the child that would otherwise be included. 

In this case, we do not do any animation or need to transition that item ever. So why go through the extra render at all?

Instead this should really just be:
```suggestion
            if (isShowRsvp) _buildRsvpStatus(context, ref),
```

app/lib/features/events/widgets/skeletons/event_item_skeleton_widget.dart
```diff
@@ -0,0 +1,43 @@
...
+  State<EventItemSkeleton> createState() => _EventListSkeletonState();
+}
+
+class _EventListSkeletonState extends State<EventItemSkeleton> {
+  @override
+  Widget build(BuildContext context) {
+    return Skeletonizer(child: _buildEventItemSkeletonUI());
+  }
+
+  Widget _buildEventItemSkeletonUI() {
```
gnunicorn, line 17: :clap: again.

why is this a stateful widget though?

Why not just?

```suggestion
class EventItemSkeleton extends StatelessWidget {
  const EventItemSkeleton({super.key});
  
  @override
  Widget build(BuildContext context) {
```

app/lib/features/news/widgets/news_item.dart
```diff
@@ -144,4 +151,29 @@ class _NewsItemState extends ConsumerState<NewsItem> {
...
+            data: (calendarEvent) {
+              return EventItem(
+                event: calendarEvent,
+              );
+            },
+            loading: () => const EventItemSkeleton(),
+            error: (e, s) => Center(child: Text('Event failed: $e')),
+          );
+    } else {
+      return const SizedBox();
```
gnunicorn, line 176: Probably want to show some "unsupported - please upgrade" in this case for Future Action types, no?

app/lib/features/news/pages/add_news_page.dart
```diff
@@ -405,6 +374,17 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
...
           );
+
+          if (slidePost.newsReferencesModel != null) {
+            final linkBuilder = sdk.api.newLinkRefBuilder(
+              slidePost.newsReferencesModel!.type.name,
+              slidePost.newsReferencesModel!.id ?? '',
+            );
+            final linkRef = linkBuilder.build();
+            final objBuilder = sdk.api.newObjRefBuilder(null, linkRef);
+            final objRef = objBuilder.build();
```
gnunicorn, line 385: looks being done repetitively, maybe we want to put that into a separate function and just call that?

app/lib/features/news/pages/add_news_page.dart
```diff
@@ -405,6 +374,17 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
               slidePost.backgroundColor?.value,
             ),
           );
+
```
gnunicorn, line 377: While you check in the UI that one can't select an event before having selected a space, there is still the unlikely case that the user changed the space after selecting the event, thus leading to a faulty reference.

We could check for that here, like have a `mustBeInSpaceId = ` on the `newsReferenceModel` and ensure that they match up and refuse to submit, or [when selecting a space](https://github.com/acterglobal/a3/pull/1450/files#diff-965bf3e8329d6f06301669ae25a78c85bc589f2d7846e778d7c34003305a9f6fR31-R36) we could clear the event to ensure this never happens ...

We need to do at least one of the two to prevent this accidental food gun to go off.

gnunicorn (approved):
Looking good. So excited this is landing now! Just one edge-case we need to cover and a few remarks on code style

app/lib/features/events/widgets/events_item.dart
gtalha07: That might not be necessary as a approach to cover all conditional render cases but I see in a lot of places, we are checking a lot of cases wrapped up in one condition that certainly doesn't look good in widget tree. In that ```Visibility``` widget makes it easier to understand what purpose it's serving. Although I agree, that for this simplistic case, its more easier to read it in one line ;)

kumarpalsinh25: Thanks @gnunicorn for highlight this things. Will take care of it.

app/lib/features/events/widgets/skeletons/event_item_skeleton_widget.dart
kumarpalsinh25: Opps! My Bad.
Yes it should be `StatelessWidget`. Let me change it.

app/lib/features/news/widgets/news_item.dart
kumarpalsinh25: Ahh, good notice. Let me do needful here.

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: Yupp! let me manage it in better way to avoid duplication of the code.

app/lib/features/events/widgets/events_item.dart
```diff
@@ -35,7 +48,10 @@ class EventItem extends ConsumerWidget {
               ),
             ),
             const SizedBox(width: 10),
-            _buildRsvpStatus(context, ref),
+            Visibility(
+              visible: isShowRsvp,
+              child: _buildRsvpStatus(context, ref),
```
gtalha07, line 53: Why not use ```Consumer``` in this widget method instead of making ```EventItem``` widget a consumer state? I don't ```ref``` usage in ```EventItem``` itself.

gtalha07 (approved):
Some code style suggestions that can be addressed with other feedbacks, overall good to merge! Nicely done.

app/lib/features/news/pages/add_news_page.dart
kumarpalsinh25: Let's clear event select upon Space Selection changes so that this issue may not exists.

Reference:
https://github.com/acterglobal/a3/assets/51526480/95dabfa6-ed21-4c40-af79-eea1d9a86dbd

PR #1452, opened 2024-02-27
Fix bug in rsvp update

bitfriend:
- Fix error in rsvp sending
- Fix bug in rsvp UI update


gnunicorn (approved):
code looks fine. but we really need some integration test for this now, this seems to be pretty fragile...

PR #1462, opened 2024-02-29
Miscellaneous improvements

bitfriend:
- Replace `ref.watch()` with `ref.read()` in non-`build()` fns
- Simplify conditional expression

### DIFF
--- a/app/lib/features/news/widgets/news_side_bar.dart
+++ b/app/lib/features/news/widgets/news_side_bar.dart
@@ -47,7 +47,7 @@ class NewsSideBar extends ConsumerWidget {
           color: Colors.white,
           index: index,
           onTap: () async {
-            final manager = await ref.watch(newsReactionsProvider(news).future);
+            final manager = await ref.read(newsReactionsProvider(news).future);
             final status = manager.likedByMe();
             debugPrint('my like status: $status');
             if (!status) {
--- a/app/lib/features/news/widgets/news_widget.dart
+++ b/app/lib/features/news/widgets/news_widget.dart
@@ -59,7 +59,7 @@ class _NewsWidgetState extends ConsumerState<NewsWidget> {
               LikeAnimation.run(index);
               final news = data[index];
               final manager =
-                  await ref.watch(newsReactionsProvider(news).future);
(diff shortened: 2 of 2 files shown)


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -191,16 +191,17 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
+                  children: editMessage != null
+                      ? [
+                          Consumer(builder: editMessageBuilder),
+                          _EditMessageContentWidget(
+                            convo: widget.convo,
+                            msg: editMessage,
+                          ),
+                        ]
+                      : [
+                          const SizedBox.shrink(),
```
kumarpalsinh25, line 203: `const SizedBox.shrink(),` doesn't requires here. Just `[]` should be sufficient.

app/lib/features/spaces/widgets/public_spaces_selector.dart
```diff
@@ -464,11 +462,11 @@ class PublicSpaceSelector extends ConsumerWidget {
         ),
       ),
     ];
-    if (val != null && val.isNotEmpty) {
+    if (val?.isNotEmpty == true) {
```
kumarpalsinh25, line 465: Do we really required` == true` here? 
I think `val?.isNotEmpty` should be sufficient.

kumarpalsinh25 (approved):
Add couple of minor suggestions but overall looks good.

app/lib/features/spaces/widgets/public_spaces_selector.dart
bitfriend: If ` == true` doesn't exist, lint error occurred.

app/lib/features/chat/widgets/custom_input.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1462/commits/738cb57c29a4ed3125aa82dadf2fb26b7eae993c

gtalha07 (approved):

PR #1469, opened 2024-03-01
Keep on syncing on "DeserializationFailure"

gnunicorn:
Before if the server failed or gave an incomplete response in sync, we'd show that error (and only that) on the home shell, blocking all other access and rendering the App unusable. This fixes that, and thus fixes https://github.com/acterglobal/a3-meta/issues/119 & fixes https://github.com/acterglobal/a3-meta/issues/118 .

Additionally this now:
1. moves that error reporting into the activities section, including indicating that an initial sync is still on-going
2. adds a new double-backing-off (with a max of 5mins) retry system with a countdown:

https://github.com/acterglobal/a3/assets/40496/6139a2e4-66ae-4ada-b57b-a25900ccfbb2

and then when it works again this goes away:

https://github.com/acterglobal/a3/assets/40496/727a97a0-d5fa-4419-b9db-e5d0cc88e9fe

---

Testing this isn't trivial as you need a running server and sync and then block the response in a specific way. The way I was able to produce this is by using the `cargo make desktop-proxyman`-debug-build and configure the http-proxy in acter to use the proxyman and then after some succesful sync select the post and make all future sync fail through the proxyman "map local editor" with the following settings (notice the `v3/sync/*` in the matcher):
```
HTTP/1.1 404 Not Found
access-control-allow-headers: X-Requested-With, Content-Type, Authorization, Date
access-control-allow-methods: GET, HEAD, POST, PUT, DELETE, OPTIONS
access-control-allow-origin: *
access-control-expose-headers: Synapse-Trace-Id, Server
cache-control: no-cache, no-store, must-revalidate
content-type: text/plain
date: Fri, 01 Mar 2024 19:12:27 GMT
server: Synapse/1.101.0
connection: close

This failed.
```
->
grafik


kumarpalsinh25 (commented):
Doesn't get much idea in this PR due to lack of understanding of sync architecture of the application. I would request @gtalha07 and @bitfriend to look into this PR once.

gtalha07:
Can see how the indicator can be useful for end-user but not sure if I would want to go for that message indication from the first look in the video. Could be more understanding and less technical. Can put exception string in log and just ```Initial Sync Failed. Retrying...```?

app/lib/common/providers/notifiers/sync_notifier.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'dart:math';
+
+import 'package:acter/common/models/sync_state/sync_state.dart';
+import 'package:acter/common/providers/space_providers.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' as ffi;
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+// ignore_for_file: avoid_print
+class SyncNotifier extends StateNotifier<SyncState> {
+  final ffi.Client client;
```
gtalha07, line 11: I see most of us use this alias to point towards package path and sometimes don't. It could cause some confusion for new contributors when they see ```ffi.Client``` and ```Client```. Should we have some general agreement here?

app/lib/features/activities/pages/activities_page.dart
```diff
@@ -20,110 +21,152 @@ class ActivitiesPage extends ConsumerWidget {
...
-        SliverToBoxAdapter(
-          child: Padding(
-            padding: const EdgeInsetsDirectional.symmetric(
-              horizontal: 10,
-              vertical: 5,
-            ),
-            child: Text(
-              'Invitations',
-              style: Theme.of(context).textTheme.headlineSmall,
+  Widget? renderSyncingState(BuildContext context, WidgetRef ref) {
```
gtalha07, line 24: nice!

app/lib/features/activities/pages/activities_page.dart
```diff
@@ -20,110 +21,152 @@ class ActivitiesPage extends ConsumerWidget {
...
-            child: EmptyState(
-              title: 'No Activity for you yet',
-              subtitle:
-                  'Notifies you about important things such as messages , invitations or requests.',
-              image: 'assets/images/empty_activity.svg',
+
+    final sessions =
+        allSessions.value!.where((session) => !session.isVerified()).toList();
+    if (sessions.isEmpty) {
+      return null;
```
gtalha07, line 106: Could've some placeholder indicating that all sessions verified with green badge ;)

gtalha07 (approved):
Looks good to me other than some minor remarks upon your disposal. Nice to have this information for end-user!

app/lib/common/providers/notifiers/sync_notifier.dart
gnunicorn: SyncState is existing once from ffi and once from the internals of flutter, that's why I used an alias here...

app/lib/features/activities/pages/activities_page.dart
gnunicorn: Not in this case. This page is meant to show the things you need to take care of. If everything is fine, it shouldn't show (here) but instead be removed to decrease the noise.

PR #1472, opened 2024-03-04
E2E Integration test for event module

kumarpalsinh25:
This PR contains E2E Integration tests for event module with below mentioned features

- Create event
- Change RSVP actions
- Edit event
- Delete event

Reference Video:
https://github.com/acterglobal/a3/assets/51526480/f128f37f-26b8-41a7-aa60-e2d0a20dab2e


bitfriend (approved):
Great work in overall.
BTW what about `calendar event` rather than `event` so that other developer can avoid be confused via technical terms?

PR #1479, opened 2024-03-05
Refactor core::Store

### LINKED ISSUE #958: When updating the the content of pin and events, it doesnt always work.
See video below

### LINKED ISSUE #1086: Issues with loading content
Note sure what exactly is happening, but sometimes experience problems with loading. 
Loading both avatars, sub-spaces and members. 

It doesnt really happen persistent.

And yes, I have good internet connection:)

https://github.com/acterglobal/a3/assets/77483554/4e156c16-2269-47ff-9134-1706dafa73a4

### LINKED ISSUE #1264: Some pins and updates are not showing on IOS and MacOS

### LINKED ISSUE #1278: too many requests upon starting up
other than the `/device`-requests (see #1276 ) just starting up and clicking around into chat and here and there a bit, there are too many requests taking place upon starting up with an existing session:

Screenshot 2024-01-15 at 17 30 36

We see:
 - a request to `$roomId/messages?dir=b&limit=20` for every space
 - a request to `$roomId/messages?dir=f&limit=100` for every space
 - several calls to `$roomId/members`
 - several calls to `thumbnail/$server/$ID` that should already have been cached previously - effectively no image changed since the last startup, why are we fetching remote content at all?
 - a request to `$roomId/messages?from=t1-XXXX` for every space

with a total of over 200 requests in just the first 16 seconds of startup...  too many!

gnunicorn:
This PR attempts to fix several data loading related problems as they seem to stem from the same inner origin. See below for descriptions of each part.

## The Problem
We have been using [`DashMap`], an asynchronously locking HashMap implement to keep the objects within our internal Store. Asynchronous lock-free Maps are _incredibly hard_, with a lot edge cases and conditions under which they do not perform correctly. Though we lacked concrete evidence, other projects in the past (most notably matrix-rust-sdk) have removed DashMap because it seemed to have several such problems.

### Evidence
When investigating the content loading issues reported several times, I switched on `acter_core::executor=trace` and noticed that despite all data coming in and thus seeing the `[2024-03-05][11:48:56.583073][acter_core::executor][TRACE] handle event_id=$EVENT_ID` indicating the start of the `executor.handle(event)` (and internally the `model.execute()`) there wasn't always a corresponding `handling done` for each event id. Most notably, it seemed to be exactly the number missing in the logs of updates and pins that were not showing up.

A further investigation into the internals showed that while the `matrix-sdk-sqlite`-store is sometimes slow especially for larger data to store (another PR is tackling that right after), it was our in-between DashMap-Driven-Store that wasn't returning. Effectively leading to endlessly pending futures that never fired.

### Effect

These future are awaited for in two main areas: for processing the following events from that message call--this is why it always seems to be happening to all objects of a particular space-- delaying all processing of that space's data indefinitely. Which is also the reason the _next_-position marker isn't properly updated and we are reloading the same set of data on every startup.

Equally that stalling future leads to the `nofify` not being called and thus the UI not being informed even though the internal Datastore is updating. This is leading to the effect that even though data has been synced and is being processed, the UI never updates and only a restart - which reloads the data from disk - actually applies them for the user to see.

## What this PR does

Next to adding some more specific trace-logging to enhance our capability to track this behavior, this PR tackles this problem in two main ways: It replaces the `DashMap` with the ,seemingly, more robust [`scc::HashMap`](https://docs.rs/scc/latest/scc/hash_map/struct.HashMap.html#method.scan) implementation for async locks, but probably more importantly it switches to a single locking mechanism around `dirty` and refactors the code to have to queue up more rather than concurrently write to the underlying store.

One important part of this refactor is that we are now holding the `lock` around dirty for the time we process the changed models synchronously but _before_ going to the store. Before we were somewhat cheeky about this and had the processing happen in between the writes and if they were happening concurrently and slightly slowed for some reason, we might have run into race conditions here. This code still isn't perfect as we don't have transactional support ensuring we revert back if there is a write failure nor can ensure that the model isn't being overwriting while we try to write it, but it is already a lot better.

### Evidence

With this PR applied, even a freshly logged in session now syncs up properly and shows the right Updates & Pins, as well as immediately shows updates on the Pins when edited. Additionally, the logs under the same conditions now show a matching count of `handle` and `handling done` indicating that all processing finished and no futures stalled.

---

- fixes #1264, fixes #1086 : While it is still unclear what exactly cause the same objects to be stalling on several machines, it seems apparent that there was a bug triggered in `DashMap` or our surrounding Store code that lead to this behavior.
- fixes #1278 : As a result of the endless stalling of the processing futures, the messages-processing-markers weren't updated ever and as a result we were fetching the same messages again on every start, just to fail to process them again. This is fixed with this PR 
- fixes #958 :  This was sometimes stalled internally as well, leading to neither the database nor the UI properly updating because the future never fired. This is fixed with this PR.


bitfriend (approved):
Great work.

gnunicorn:
Now with unit test for 73% of function and 90% of lines of the store:
grafik

PR #1490, opened 2024-03-06
UI-UX Improvement : Chat Filter and Search

kumarpalsinh25:

### DIFF
--- a/app/lib/features/chat/widgets/rooms_list.dart
+++ b/app/lib/features/chat/widgets/rooms_list.dart
@@ -65,25 +65,16 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
     String title = AppLocalizations.of(context)!.chat;
 
     if (ref.watch(hasRoomFilters)) {
-      String subFiltered = 'chats';
       final selection =
           ref.watch(roomListFilterProvider.select((value) => value.selection));
       switch (selection) {
         case FilterSelection.dmsOnly:
-          subFiltered = 'DMs';
+          title = 'DMs';
           break;
         case FilterSelection.favorites:
-          subFiltered = 'Bookmarked chats';
+          title = 'Favorites';
         default:
           break;
-      }
-
-      final searchTerm =
-          ref.watch(roomListFilterProvider.select((value) => value.searchTerm));
-      if (searchTerm != null && searchTerm.isNotEmpty) {
-        title = "Show $subFiltered with '$searchTerm'";
-      } else {
-        title = subFiltered;
       }
     }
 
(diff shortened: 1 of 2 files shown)


app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -65,26 +65,17 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
...
       final selection =
           ref.watch(roomListFilterProvider.select((value) => value.selection));
       switch (selection) {
         case FilterSelection.dmsOnly:
-          subFiltered = 'DMs';
+          title = 'DMs';
           break;
         case FilterSelection.favorites:
-          subFiltered = 'Bookmarked chats';
+          title = 'Favorites';
```
gnunicorn, line 75: nope. We call that "bookmarks" in Acter.

gnunicorn, line 75: nope. We call that "bookmarked" in Acter. Favorites implies a value judgement of love that isn't necessarily true nor feels correct. Bookmark is a more neutral term.


```suggestion
          title = 'Bookmarked';
```

app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -128,35 +140,13 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
...
-                .setSelection(newSelection.first);
-          },
-          selected: {selected},
-        ),
-        const SizedBox(height: 10),
+        filterChipsButtons(),
         Row(
           mainAxisAlignment: MainAxisAlignment.end,
           children: [
+            Expanded(child: searchTerms(context)),
```
gnunicorn, line 147: I understand correctly you removed the search term from the title and only show it while the box is visible now? What if you scrolled down and the search box disappears? The information of it being a search also disappears then? 

If we have a search term that we filter by that information must be sticky on screen or we'll see a bunch of "my chat's disappeared"-bug-reports of people forgetting that they had searched for things (the search info disappearing due to it not being synced back from the provider to the UI if the user left the UI was already an annoying bug we had in the past).

This information must be shown regardless of whether the filter-box is visible or not.

app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -177,12 +167,59 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
...
+            onSelected: (value) async {
+              await ref
+                  .read(roomListFilterProvider.notifier)
+                  .setSelection(FilterSelection.all);
+            },
+          ),
+          const SizedBox(width: 10),
+          FilterChip(
+            selected: selected == FilterSelection.favorites,
+            label: const Text('Favorites'),
```
gnunicorn, line 200:
```suggestion
            label: const Text('Bookmarked'),
```

gnunicorn (approved):
Design looks good but have two product/usability concern that need to be addressed.

app/lib/features/chat/widgets/rooms_list.dart
kumarpalsinh25: > If we have a search term that we filter by that information must be sticky on screen or we'll see a bunch of "my chat's disappeared"-bug-reports of people forgetting that they had searched for things (the search info disappearing due to it not being synced back from the provider to the UI if the user left the UI was already an annoying bug we had in the past).

Search terms doesn't show and cutting off as shown in below screenshot. We can make it for two line but again I feel that title bar should not be like that.

Screenshot 2024-03-06 at 5 59 45 PM


> I understand correctly you removed the search term from the title and only show it while the box is visible now? What if you scrolled down and the search box disappears? The information of it being a search also disappears then?
This information must be shown regardless of whether the filter-box is visible or not.

Let me think on it and do the needful.

PR #1496, opened 2024-03-07
Bump DM creation timeout

### LINKED ISSUE #1363: Error when sending creating dm
Wit shows error, however when tripping the app,the chat was actually created.
https://github.com/acterglobal/a3/assets/77483554/720bde02-8fb5-4f77-b77b-9c3a37338b07

### LINKED ISSUE #1415: create dm shows error but shows up anyway
creating a dm with myself, shows error but then shows up afterwards (but first after reopening the app)

https://github.com/acterglobal/a3/assets/77483554/d79eb999-86db-4930-93a0-3feb614b6b75

gnunicorn:
From 3seconds (which is arguably too low) to 30seconds - which should really cover most if not all cases of sync issues. Additionally contains some slight refactoring to make things less hard to read and follow: separate the UI into several inner Widget `fn`s and make the `handleCreateConvo` more self-standing and pure (less ref-read that is unnecessary and more type-specific inputs).

fixes #1363 , fixes #1415 .

### DIFF
--- a/app/lib/features/chat/widgets/create_chat.dart
+++ b/app/lib/features/chat/widgets/create_chat.dart
@@ -152,23 +152,21 @@ class _CreateChatWidgetState extends ConsumerState<CreateChatPage> {
         config.setParent(parentId);
       }
       final client = ref.read(alwaysClientProvider);
-      final roomId = await client.createConvo(config.build());
+      final roomIdStr = (await client.createConvo(config.build())).toString();
       // add room to child of space (if given)
       if (parentId != null) {
         final space = await ref.read(spaceProvider(parentId).future);
-        await space.addChildRoom(roomId.toString());
+        await space.addChildRoom(roomIdStr);
       }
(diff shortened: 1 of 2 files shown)


app/lib/features/chat/widgets/create_chat.dart
```diff
@@ -448,6 +248,217 @@ class _CreateChatWidgetConsumerState extends ConsumerState<_CreateChatWidget> {
...
+          ),
+        ),
+      ),
+    );
+  }
+
+  Widget renderPrimaryAction(BuildContext context) {
+    final selectedUsers = ref.watch(_selectedUsersProvider).toList();
+    return ListTile(
+      onTap: selectedUsers.isEmpty
```
gtalha07, line 325: I think we can also improve this tap handler having separate ```handleCreateAction()``` for readability.

gnunicorn: next time ;) .

PR #1498, opened 2024-03-07
Refactor Settings UI-UX

kumarpalsinh25:
This PR contains serval UI Improvements in the settings screens.

Most of the changes are in settings page and profile page. Other files changes are just due to naming changes. So out 18 files changes, there are only 2 or 3 file which have major changes other have minor things to look on :)

- - -
Video Reference
- - -
https://github.com/acterglobal/a3/assets/51526480/c7f5904a-2b26-4d2a-85e8-41e0c98993b3

- - -
Mobile
- - -
Settings
Settings (1)
- - -
Desktop
- - -
Settings (2)
Settings (3)


kumarpalsinh25:
@gnunicorn 

> The first entry for the profile isn't really clear as a settings option IMHO.

The first entry will look like below.
Screenshot 2024-03-07 at 7 03 05 PM

> How about changing the title to "My Profile" and the subtitle to show the displayName (user-id as fallback)?

Sorry but not getting you, can you explain bit more about this?

> Edit screen :  the way to change the avatar is completely unclear 

How about if we put pencil icon on bottom left corner of the avatar?

> that the display name and username have the same design despite one being change-able (display name) while the other isn't ... and then have a save-button that effectively only is about the display name (avatar is changed whenever you select a new one regardless of whether you clicked 'save') isn't really making sense ...

We can do one thing here..
1) Differentiate username field in such way that it will look no-editable
2) Make display as view only and give pencil icon at trailing side to open edit display name pop-up as it was there previously and remove save button

app/lib/features/home/widgets/sidebar_widget.dart
```diff
@@ -61,7 +61,7 @@ class SidebarWidget extends ConsumerWidget {
               key: Keys.avatar,
               margin: const EdgeInsets.only(top: 8),
               child: InkWell(
-                onTap: () => context.pushNamed(Routes.myProfile.name),
+                onTap: () => context.goNamed(Routes.settings.name),
```
gtalha07, line 64: You sure this wouldn't break the routing part ? Need to be careful thinking about changing route transition type...

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -1,121 +1,42 @@
...
       if (!context.mounted) {
         return;
       }
-      Navigator.of(context, rootNavigator: true).pop();
-      customMsgSnackbar(context, 'Display Name update submitted');
+
+      // close loading
+      EasyLoading.dismiss();
+
+      context.pop();
```
gtalha07, line 37: Where does this pop to? Back to settings menu? If that is, we don't want to navigate back. Not sure if its needed since Loader is changed and dismissed already by here.

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -129,56 +50,66 @@ class MyProfilePage extends ConsumerWidget {
...
-                  child: Column(
-                    mainAxisAlignment: MainAxisAlignment.center,
-                    crossAxisAlignment: CrossAxisAlignment.center,
-                    children: [
-                      GestureDetector(
-                        onTap: () => updateAvatar(data, context, ref),
+  AppBar _buildAppbar(BuildContext context) {
+    return AppBar(
+      title: Text(
+        'Profile',
```
gtalha07, line 83: Should be `My Profile` as addressed by ben.

gnunicorn:
> How about if we put pencil icon on bottom left corner of the avatar?

Yeah, sounds good.

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -129,56 +50,66 @@ class MyProfilePage extends ConsumerWidget {
...
+        style: Theme.of(context).textTheme.titleLarge,
+      ),
+    );
+  }
+
+  Widget _buildBody(BuildContext context) {
+    return Consumer(
```
gtalha07, line 90: We don't need this ```Consumer``` wrapper do we? Could be simply like :
```suggestion
    Widget _buildBody(BuildContext context){
    final account = ref.watch(accountProfileProvider);
    return account.when(
    ...
    }
```

app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -24,216 +23,227 @@ class SettingsMenu extends ConsumerWidget {
...
+        ),
+        const SizedBox(height: 20),
+      ],
+    );
+  }
+
+  Widget _settingMenuSection({
+    required BuildContext context,
+    required String sectionTitle,
+    bool isDanderZone = false,
```
gtalha07, line 181: minor word correction! ;)

app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -24,216 +23,227 @@ class SettingsMenu extends ConsumerWidget {
...
+          iconData,
+          color: enable ? iconColor : Theme.of(context).disabledColor,
+        ),
+        title: Text(
+          title,
+          style: titleStyles?.copyWith(
+            color: enable ? null : Theme.of(context).disabledColor,
+          ),
+        ),
+        subtitle: subTitle == null
```
gtalha07, line 235: Could be simplified to 
```suggestion
        subtitle: subTitle ?? Text(
                subTitle,
                style: titleStyles?.copyWith(
                  color: enable ? null : Theme.of(context).disabledColor,
                ),
              ),
```

app/lib/features/profile/pages/my_profile_page.dart
kumarpalsinh25: Yes, it is for going back to setting menu. Open for suggestion here :)

app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -24,216 +23,227 @@ class SettingsMenu extends ConsumerWidget {
...
 
     TextStyle titleStylesSelected(Routes route) {
-      return TextStyle(
-        color: colorSelected(route),
-      );
+      return TextStyle(color: colorSelected(route));
     }
 
-    final shouldGoNotNamed = isDesktop || size.width > 770;
+    final shouldGoNotNamed = isDesktop && size.width > 770;
```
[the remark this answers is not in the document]
kumarpalsinh25: Yes understand this but this is require change based on the recent changes and doesn't effecting any other places :)

kumarpalsinh25:
@gnunicorn 

1) 
Added 'Edit Profile' label
Screenshot 2024-03-07 at 8 24 21 PM

2) 
- Added pencil icon over the profile avatar
- Changed the color of username filed to make it like readonly
Screenshot 2024-03-07 at 8 24 43 PM

app/lib/features/profile/pages/my_profile_page.dart
gnunicorn: According to the design questionaire results we should speak of "You" and "Your". So "Your profile". But I think "profile" is fine enough here.

app/lib/features/settings/pages/settings_page.dart
```diff
@@ -0,0 +1,93 @@
...
+          _userProfileUI(),
+          const SettingsMenu(),
+        ],
+      ),
+    );
+  }
+
+  Widget _userProfileUI() {
+    return Consumer(
+      builder: (context, ref, child) {
```
gnunicorn, line 43: not necessary here either, if we made this a consumer-widget ... which I think, I'd prefer...

app/lib/features/settings/pages/settings_page.dart
```diff
@@ -0,0 +1,93 @@
...
+                  const Padding(
+                    padding: EdgeInsets.only(bottom: 10),
+                    child: Text('Edit Profile'),
+                  )
+                ],
+              ),
+            );
+          },
+          error: (e, trace) => Text('error: $e'),
+          loading: () => const Text('loading'),
```
gnunicorn, line 88: no skeletons :( ... :really_sad_panda:

app/lib/features/settings/widgets/settings_menu.dart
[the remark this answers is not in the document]
gnunicorn: why is it both though? why not just `size_width`-dependent? I'd really prefer to remove any instance of `isDesktop` we can get rid of ... it is almost always too generic...

app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -24,216 +23,227 @@ class SettingsMenu extends ConsumerWidget {
...
+        _settingMenuSection(
+          context: context,
+          sectionTitle: 'Account',
+          children: [
+            _settingMenuItem(
+              context: context,
+              iconData: Atlas.key_monitor_thin,
+              iconColor: colorSelected(Routes.settingSessions),
+              title: 'Sessions',
+              subTitle: 'Your active device sessions',
```
gnunicorn, line 52:
```suggestion
              subTitle: 'Your active devices',
```

let's simplify this.

app/lib/features/profile/pages/my_profile_page.dart
```diff
@@ -129,236 +50,193 @@ class MyProfilePage extends ConsumerWidget {
...
-                          ],
+                    ),
+                    const SizedBox(height: 25),
+                  ],
+                ),
+              ),
+            );
+          },
+          error: (e, trace) => Text('error: $e'),
+          loading: () => const Text('loading'),
```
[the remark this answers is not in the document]
kumarpalsinh25: How can I forgot this. Bad thing :(
Let me do it :)

app/lib/features/settings/pages/blocked_users.dart
```diff
@@ -68,7 +68,7 @@ class BlockedUsersPage extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final users = ref.watch(ignoredUsersProvider);
     return WithSidebar(
-      sidebar: const SettingsMenu(),
+      sidebar: const SettingsPage(),
```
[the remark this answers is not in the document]
kumarpalsinh25: Ahh, understood your thought here. Let me change it to `SettingsMenu`. :)

app/lib/features/settings/pages/settings_page.dart
kumarpalsinh25: My Bad! let me do it :)

app/lib/features/settings/pages/blocked_users.dart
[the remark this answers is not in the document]
kumarpalsinh25: We already have `SettingsMenu()` class in widget directory which container all the menus expect profile option.
`SettingsPage()` represents entire settings page which container upper profile menu and below side uses `SettingsMenu` component.

gtalha07 (approved):

PR #1501, opened 2024-03-07
Refactor Chat Room Messages 

### LINKED ISSUE #1460: Edited messages does not show
to replicate:
1. Create message
2. Edit message

gnunicorn:
Removes a lot of code that was previously needed but since the matrix-rust-sdk caught up with the messages parsing from the timeline can now be left to that directly. Allowing us to remove a lot now unneeded and (as it appears) partially buggy code.  

Fixes #1460 : 

https://github.com/acterglobal/a3/assets/40496/4d4ed6b9-008d-46e9-8595-58d16415b796


bitfriend (approved):
Great job.
Noticed you removed the old codes.
Thanks for this refactoring.

PR #1503, opened 2024-03-08
Refactor Profile Page UI

kumarpalsinh25:
Room Details Page

- - -

https://github.com/acterglobal/a3/assets/51526480/265dceed-315e-4be7-9eb1-c795a145c8ac

### DIFF
--- a/app/lib/features/chat/pages/room_profile_page.dart
+++ b/app/lib/features/chat/pages/room_profile_page.dart
@@ -181,8 +181,11 @@ class _RoomProfilePageState extends ConsumerState<RoomProfilePage> {
               context: context,
               iconData: Atlas.user_plus_thin,
               actionName: 'Invite',
+              actionItemColor: membership!.canString('CanInvite')
+                  ? null
+                  : Theme.of(context).colorScheme.onSurface,
               onTap: () {
-                membership!.canString('CanInvite')
+                membership.canString('CanInvite')
                     ? context.pushNamed(
                         Routes.spaceInvite.name,
                         pathParameters: {'spaceId': widget.roomId},
@@ -212,7 +215,7 @@ class _RoomProfilePageState extends ConsumerState<RoomProfilePage> {
                     .permalink();
             Share.share(
               roomLink,
-              subject: 'Room ID',
+              subject: 'Link to Chat',
             );
           },
         ),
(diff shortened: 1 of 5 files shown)


gnunicorn:
@kumarpalsinh25 we opted for not showing the members right away because that is a potentially large list that takes quite a bit to load. If that is offloaded into its own consumer/widget that doesn't block the other parts of the UI we can consider this design, otherwise we have to stick with it being a menu item ...

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -70,282 +74,314 @@ class RoomProfilePage extends ConsumerWidget {
...
+                      );
+              },
+            );
+          },
+          error: (e, st) => Text('Error loading tile due to $e'),
+          loading: () => const ActionItemSkeleton(
+            iconData: Atlas.user_plus_thin,
+            actionName: 'Invite',
+          ),
+        ),
```
gnunicorn, line 202: arguably showing the invite if you can't invite is a bit confusing. maybe we could show a disabled button with a tooltip stating the user can't invite to this room instead?

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -70,282 +74,314 @@ class RoomProfilePage extends ConsumerWidget {
...
+          context: context,
+          iconData: Icons.ios_share,
+          actionName: 'Share',
+          onTap: () async {
+            final roomLink =
+                await (await ref.read(chatProvider(widget.roomId).future))
+                    .permalink();
+            Share.share(
+              roomLink,
+              subject: 'Room ID',
```
gnunicorn, line 215:
```suggestion
              subject: 'Link to Chat',
```

gnunicorn (approved):
LGTM. Just some minor stuff, totally up for you if you want to do them before merging.

app/lib/features/chat/pages/room_profile_page.dart
kumarpalsinh25: Let me change colour of invite action button if user don't have permission.

kumarpalsinh25: Done :)
Screenshot 2024-03-08 at 10 03 51 PM

PR #1504, opened 2024-03-08
Show participates data in event details

kumarpalsinh25:
Screenshot 2024-03-08 at 7 36 48 PM

### DIFF
--- a/app/lib/features/events/pages/event_details_page.dart
+++ b/app/lib/features/events/pages/event_details_page.dart
@@ -1,3 +1,5 @@
+import 'package:acter/common/dialogs/member_info_drawer.dart';
+import 'package:acter/common/providers/chat_providers.dart';
 import 'package:acter/common/providers/room_providers.dart';
 import 'package:acter/common/themes/colors/color_scheme.dart';
 import 'package:acter/common/utils/routes.dart';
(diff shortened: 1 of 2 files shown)


app/lib/features/events/pages/event_details_page.dart
```diff
@@ -387,11 +411,51 @@ class _EventDetailPageConsumerState extends ConsumerState<EventDetailPage> {
...
+          return const Text('No participants going');
+        }
+
+        List<Widget> avtarList = [];
+        for (final participantId in eventParticipantsList) {
+          final memberInfo = ref.watch(
+            roomMemberProvider((roomId: roomId, userId: participantId)),
+          );
+          memberInfo.whenData((profileData) {
+            avtarList.add(
```
gnunicorn, line 437: this isn't really a safe way to generate that list, I'd say. I am not sure what exactly happens with the `avatarList` if the `memberInfo` is delayed or comes in later... it probably means it isn't the same order at least ... but what if any of them fails, for e.g. for timeout reasons? It would show the wrong count, wouldn't it?

I think this should instead be something like:

```
List<Widget> avatarList = eventParticipantsList.map((participantId) =>  ref.watch(
            roomMemberProvider((roomId: roomId, userId: participantId)),
          ).when(data: (profileData) => Padding( ... ), error: (e,s) = FallbackAvatar, loading: () => AvatarSkeleton))).toList();
```

gnunicorn (approved):
Looks good, but the way the participant list is generated doesn't look sound to me.

gnunicorn:
also ... might want to add the new member details box as a on-top function?

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: Ah, good notice. Let me do the needful changes here.

kumarpalsinh25:
> also ... might want to add the new member details box as a on-top function?

Sorry not getting you. You are talking about below user profile pop-up?
Screenshot 2024-03-12 at 7 33 15 PM

gnunicorn:
@kumarpalsinh25 yes.

kumarpalsinh25:
@gnunicorn 
Getting below error `onAvatarTap` , any idea?

Screenshot 2024-03-13 at 5 21 36 PM

Issue #1096, opened 2023-10-28
Subspaces and chat does not show in the space overview


emilvincentz:
https://github.com/acterglobal/a3/assets/77483554/e941c04b-7119-44ac-bd59-71d85c0490fc

kumarpalsinh25:
@emilvincentz 
Is this issue is generated in specific use-case? 
I have checked couple of spaces where I can see chats and sub-spaces on overview page.

emilvincentz:
hey @kumarpalsinh25 yes, I have invited you to the "Beacon" space. thats where the above currently happens.

kumarpalsinh25:
I have accepted invitation of the "Beacon" space. However I'm not able to see any chats and sub-spaces anywhere not just overview.



![Image](https://github.com/acterglobal/a3/assets/51526480/2c0a9f85-ea9b-4697-b2c6-ed2e888fbb2d)



![Image](https://github.com/acterglobal/a3/assets/51526480/59b90d70-06c3-4e63-9901-753c63bb0eba)


I have also cross checked element for this but got able to see any  chats and sub-spaces over there also.


![Image](https://github.com/acterglobal/a3/assets/51526480/be49e2e7-da9b-4986-8c35-6f04390377b5)

kumarpalsinh25:
hmm.. Strange! 
Is there any other spaces where we can reproduce the issue?

emilvincentz:
Strange indeed. I havent come across others  yet.

gnunicorn:
_sigh_ ... the way that the space tree is flatten per user is pretty ... messy ... and the implementations are unclear ...

I _suspect_ the server isn't showing @kumarpalsinh25 these rooms in the [hierarchy response]( https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv1roomsroomidhierarchy) it calculates as he wouldn't be able to access them - could it be that the rooms are "private" in visibility and _invite_only_ in invite mode? Then the server, it appears, filters them out of the response.

There is some more ugliness in there, as to even when it is a room you _could_ join, if you had been part of it and left it, the query also doesn't show it anymore. It is pretty mad.

emilvincentz:
@gnunicorn Well I havent joined any of the rooms or subspaces, but can still see them, so I expect that they are not invite only 
[...]

PR #1506, opened 2024-03-08
Refactor related spaces

gnunicorn:
Probably best viewed one commit at a time. On top of #1505 .

https://github.com/acterglobal/a3/assets/40496/acccc41c-914c-49fc-8a48-9e7ec3f2aa58

This does two things:
1. refactor the complicated related-spaces page in two steps: first switch all these sections into their own FNs for easier reading (first commit) then move that into a separate widget (second commit).
2. Reuse the same widget for the related-spaces section of the space overview without showing the parents, fixing #1096 .

In order to allow for the latter, I had to remove the grid view from the overview because those are not compatible with one another and I couldn't find a good way of making that work, considering that the widget itself uses the rather complex sliver-mechanism to allow for autoloading on scroll. As a side effect the screen looks a bit worse on desktop after this PR was applied :( (see screenshot below), but I guess that is an okay price to pay for now...

grafik

### DIFF
--- a/app/integration_test/tests/sub_spaces.dart
+++ b/app/integration_test/tests/sub_spaces.dart
@@ -1,6 +1,6 @@
 import 'package:acter/common/widgets/sliver_scaffold.dart';
-import 'package:acter/features/space/pages/related_spaces_page.dart';
 import 'package:acter/features/space/sheets/link_room_sheet.dart';
+import 'package:acter/features/space/widgets/relatest_spaces.dart';
 import 'package:convenient_test_dev/convenient_test_dev.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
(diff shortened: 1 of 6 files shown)


gtalha07:
> In order to allow for the latter, I had to remove the grid view from the overview because those are not compatible with one another and I couldn't find a good way of making that work, considering that the widget itself uses the rather complex sliver-mechanism to allow for autoloading on scroll. As a side effect the screen looks a bit worse on desktop after this PR was applied :( (see screenshot below), but I guess that is an okay price to pay for now...

Have you tried using https://pub.dev/packages/flutter_staggered_grid_view ? This seems optimal than the standard flutter ```GridView``` and wraps up content nicely like pins overview.

gtalha07 (approved):
Looks good overall. Reminder for adding docs if end-user is affected! ;)

PR #1509, opened 2024-03-11
refactor attachments as standalone

gtalha07:


gnunicorn:
@gtalha07 I don't quite get the function of this dialog: why is there so much empty space if we could also expand the image? What purpose does that serve?

grafik

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -327,92 +323,10 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
...
-                              attachmentConfirmationDialog(
-                                context,
-                                selectedFiles,
-                                _FileWidget(mimeType, selectedFiles.first),
-                                handleFileUpload,
-                                AttachmentType.file,
-                              );
-                            },
-                          ),
+                        onTap: () => showAttachmentSelection(
```
gnunicorn, line 326: can you elaborate why you felt the need to combine the attachment management for objects with this for chat? The `null` here in there kinda indicates pretty clearly to me that these are distinct enough to not combine them ... or is there any other place we send this over _without_ a manager attached?

or, alternatively, if you feel strong about them being the same components, how about not making the `showModalBottomSheet` any `manager` at all and instead manage this in a different layer above?

app/lib/common/dialogs/attachment_selection.dart
```diff
@@ -0,0 +1,110 @@
+import 'dart:io';
```
gnunicorn, line 1: I _assume_ this is mostly copy-paste from before ...

app/lib/common/providers/notifiers/attachments_notifiers.dart
```diff
@@ -36,3 +42,77 @@ class AttachmentsManagerNotifier
...
+  final AttachmentsManager manager;
+  final Ref ref;
+
+  AttachmentDraftsNotifier({required this.manager, required this.ref})
+      : super([]);
+
+  /// converts user selected media to attachment draft and adds to state list.
+  /// only supports image/video/audio/file.
+  Future<void> addDraft(File file) async {
+    EasyLoading.show(status: 'Adding draft');
```
gnunicorn, line 56: this is an odd message to show to the user, as they never said anything about doing a "draft". and now that this is a direct interaction API this probably shouldn't really care about adding draft and sending them as two steps ...?!?

app/lib/common/providers/notifiers/attachments_notifiers.dart
```diff
@@ -36,3 +42,77 @@ class AttachmentsManagerNotifier
...
+    final client = ref.read(alwaysClientProvider);
+    try {
+      final mimeType = lookupMimeType(file.path)!;
+      if (mimeType.startsWith('image/')) {
+        Uint8List bytes = await file.readAsBytes();
+        final decodedImage = await decodeImageFromList(bytes);
```
gnunicorn, line 62: this code looks warily similar to the one for chat attachment sending below, why is that not generalized somewhere?!?

app/lib/common/widgets/attachments/attachment_item.dart
```diff
@@ -0,0 +1,136 @@
...
+
+// Attachment item UI
+class AttachmentItem extends ConsumerWidget {
+  final Attachment attachment;
+  const AttachmentItem({super.key, required this.attachment});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    var msgContent = attachment.msgContent();
+    String? mimeType = msgContent.mimetype();
```
gnunicorn, line 15: attachments are strongly typed (`attachment.typeStr()`) why are we doing a mime-type-guessing-game here? Even if it is an image, if it was uploaded as a `file` to the network, we should just consider it a file ... not try to guess it to be something else...

app/lib/common/widgets/attachments/post_attachment_selection.dart
```diff
@@ -0,0 +1,185 @@
...
+  void handleChatAttachmentSend(
+    List<File> selectedAttachments,
+  ) async {
+    final roomId = widget.convo!.getRoomIdStr();
+    final client = ref.read(alwaysClientProvider);
+    final inputState = ref.read(chatInputProvider(roomId));
+    final stream = widget.convo!.timelineStream();
+
+    try {
+      for (var file in selectedAttachments) {
```
gnunicorn, line 121: This was clearly different before at least for chat and should be done in a similar way for attachments: instead of generalizing everything into files and then guess the content via a mime-type as we do here, before when a user selected an image, we sent it as an image, regardless of the internal types. We should do that, in particular because if the user selected something to be sent as a file or video or whatnot, it isn't our place to just replace that semantic information with something else we _think_ we detected. If it was selected as a file, we should sent it as a file, same for video and image... 

I understand this isn't a degression from before, so I won't consider this a blocker. But if we merge it this way, let's at least put up an issue asking for a refactor of this into the proper types provided by the user...

app/lib/common/widgets/attachments/post_attachment_selection.dart
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    if (inputState.repliedToMessage != null) {
+      final notifier = ref.read(chatInputProvider(roomId).notifier);
+      notifier.setRepliedToMessage(null);
+      notifier.setEditMessage(null);
+      notifier.showReplyView(false);
+      notifier.showEditView(false);
+      notifier.setReplyWidget(null);
+      notifier.setEditWidget(null);
```
gnunicorn, line 182: reads like we should have a `.clear()` on the notifier, rather than have it switch the state 7 times, we could just do that once ...

gnunicorn (changes requested):
Nice job :clap: . It isn't really standalone yet though. I was expecting that the Pin would just contain a `AttachmentsSectionWidget(manager: manager)` widget that we can use everywhere right away as long as we have a local `AttachmentsManager` (regardless of the surrounding type) ...

Along the same lines a few things have been generalized beyond what I understand and by moving some code around, I noticed a mime-type-guessing-misbehaviour as well. additionally, this PR just adds random pin content where it really shouldn't.

gtalha07:
> @gtalha07 I don't quite get the function of this dialog: why is there so much empty space if we could also expand the image? What purpose does that serve?

This dialog is a confirmation for user to send attachments. I agree the spacing here is abnormal, the file previews will align up here with wrapping. I will correct the concern!

app/lib/features/pins/widgets/pin_item.dart
```diff
@@ -292,19 +165,38 @@ class _PinDescriptionWidgetConsumerState
...
-                  : ActerDocumentHelpers.fromMarkdown(body);
-              textEditorState = EditorState(document: document);
+              if (body.trim().isNotEmpty) {
+                final document = html != null
+                    ? ActerDocumentHelpers.fromHtml(html)
+                    : ActerDocumentHelpers.fromMarkdown(body);
+                textEditorState = EditorState(document: document);
+              } else {
+                textEditorState = EditorState(
+                  document: ActerDocumentHelpers.fromMarkdown('No Description'),
```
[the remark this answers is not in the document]
gtalha07: There's currently bug with editor not taking up input cursor when try to focus on it if initially there's no content. I did not find reason for this yet but I will try to look at it again tomorrow. This has been added as temporal fix! But ideally I would like myself to have it reverted. 👍🏻

app/lib/features/pins/widgets/pin_item.dart
```diff
@@ -292,19 +165,38 @@ class _PinDescriptionWidgetConsumerState
...
+          onCancel: () {
+            final content = widget.pin.content();
+            if (content != null) {
+              textEditorState = EditorState(
+                document: ActerDocumentHelpers.fromMsgContent(content),
+              );
+            } else {
+              textEditorState = EditorState(
+                document: ActerDocumentHelpers.fromMarkdown('No Description'),
+              );
```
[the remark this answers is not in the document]
gtalha07: Addressed the reason above :)

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -203,4 +216,96 @@ class PinPage extends ConsumerWidget {
...
+    final attachmentTitleTextStyle = Theme.of(context).textTheme.labelLarge;
+    final attachments = ref.watch(pinAttachmentsProvider(pin));
+    final membership = ref.watch(roomMembershipProvider(pin.roomIdStr()));
+    final canPostAttachments = [];
+    if (membership.valueOrNull != null) {
+      final member = membership.requireValue!;
+      if (member.canString('CanPostPin')) {
+        canPostAttachments.add(_buildAddAttachment(pin));
+      }
+    }
```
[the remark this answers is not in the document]
gtalha07: It's still bound to pin yes, I wasn't sure how the ```standalone widget``` would serve in UI terms since the attachment component design could turn out differently for different features. If this is not the case, I can make one :)

app/lib/features/pins/pages/pin_page.dart
```diff
@@ -11,7 +13,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
 import 'package:skeletonizer/skeletonizer.dart';
 
-class PinPage extends ConsumerWidget {
+class PinPage extends ConsumerStatefulWidget {
```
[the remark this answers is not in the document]
gtalha07: Must had happened during refactoring phase, will change it! 👍🏻

app/lib/features/chat/widgets/custom_input.dart
gtalha07: They aren't same in a way that chat attachments don't need manager in any case, while the generic attachment do require a manager. My probable opinion to have this improved is to think it as different from ```manager``` since these attachments flow end up on ```timelineStream```. Shall I revert this specific change and keep the old flow?

app/lib/common/dialogs/attachment_selection.dart
gtalha07: Logically same, with the migration of some code to subsequent modals of it.

app/lib/common/providers/notifiers/attachments_notifiers.dart
gtalha07: Will remove this loading indication!

app/lib/common/providers/notifiers/attachments_notifiers.dart
gtalha07: Yeah this keeps duplicating, will add utility fn for it.

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: I guess I wasn't aware of it. Or maybe I inferred that wrong, will use this determination.

app/lib/common/widgets/attachments/post_attachment_selection.dart
gtalha07: Good point, but this change isn't related to this PR, this has just been moved and was same as before!

app/lib/features/pins/widgets/pin_item.dart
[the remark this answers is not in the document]
gnunicorn: do you have a bug report in the editor repo about this?

app/lib/features/pins/pages/pin_page.dart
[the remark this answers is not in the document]
gnunicorn: >  since the attachment component design could turn out differently for different features. If this is not the case, I can make one :)

Already for easier user adoption we should keep it similar across different parts of the UI, so: yes, for now (at least) it will look the same everywhere.

app/lib/features/chat/widgets/custom_input.dart
gnunicorn: Yeah, I'd rather we do not touch the existing and working chat-attachments-code in this PR but keep it the way it is. We can still think about unifying that later...

app/lib/common/providers/notifiers/attachments_notifiers.dart
gnunicorn: as said above, let's revert the chat code and keep this separated for now in this PR. Let's add a tech-debt issue that we want to address these, but not bloat up this PR even more...

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: Fixed by c6ea65a.

app/lib/common/providers/notifiers/attachments_notifiers.dart
gtalha07: Fixed by c6ea65a.

app/lib/features/pins/pages/pin_page.dart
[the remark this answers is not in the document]
gtalha07: Fixed by [21758f1](https://github.com/acterglobal/a3/pull/1509/commits/21758f16e4922c82b44a9114bf7d3ebe3a39c3d4)

app/lib/features/pins/widgets/pin_item.dart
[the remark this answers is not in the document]
gtalha07: Fixed by 41da803.

app/lib/features/pins/widgets/pin_item.dart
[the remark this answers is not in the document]
gtalha07: Fixed  by 41da803.

app/lib/common/widgets/attachments/post_attachment_selection.dart
gtalha07: Fixed by c53e1ac

app/lib/features/pins/pages/pin_page.dart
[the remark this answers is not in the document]
gtalha07: Fixed by 2e19583

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: Fixed by ac0da96

app/lib/common/widgets/attachments/attachment_options.dart
```diff
@@ -26,31 +27,27 @@ class AttachmentOptions extends StatelessWidget {
         if (!isDesktop)
           ListTile(
             onTap: () {
-              Navigator.of(context).pop();
```
kumarpalsinh25, line 29: Navigator.of(context).pop();

It should be here as it was before and need to have standalone route control of `AttachmentOptions`  itself instead of putting that one different class (Ex `attachmentConfirmationDialog`)

app/lib/common/widgets/attachments/attachment_widget.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:acter/common/widgets/attachments/attachment_item.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show AttachmentsManager;
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:skeletonizer/skeletonizer.dart';
+
+/// Attachment Section Widget
+class AttachmentWidget extends ConsumerWidget {
```
kumarpalsinh25, line 12:
```suggestion
class AttachmentSectionWidget extends ConsumerWidget {
```

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -541,6 +525,41 @@ class _CustomChatInputState extends ConsumerState<CustomChatInput> {
     return null;
   }
 
+  void attachmentConfirmation(
```
kumarpalsinh25, line 528: So this is different from `attachmentConfirmationDialog`?

kumarpalsinh25 (commented):
Good Job 👏 . Looks much better then previous code.

Added few queries and suggestions. :)

.changes/1508-refactor-attachments.md
```diff
@@ -0,0 +1,2 @@
+- [fix]: Redefined attachments .i.e. due to architectural constraints, they exists just like comments/likes and can be interacted on created pins.
+- [fix]: The url field in pins create/edit is now optional.
```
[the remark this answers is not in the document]
gtalha07: We already discussed on top about the issue of having attachments prior to pin creation. This is a constraint we have because of ```AttachmentManager``` object that lives inside higher class Pins/Events etc. For now, we agreed to have attachments be meant for something that can be interacted after pins are created.

app/lib/common/dialogs/attachment_confirmation.dart
```diff
@@ -1,61 +1,43 @@
 import 'dart:io';
 
 import 'package:acter/common/utils/utils.dart';
-import 'package:acter/common/widgets/default_dialog.dart';
+import 'package:acter/common/widgets/attachments/post_attachment_selection.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show AttachmentsManager;
```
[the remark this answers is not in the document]
gtalha07: This is import pattern where we just want to expose certain ffi object. We use it to avoid naming conflicts on flutter side. It's not important but recommended.

app/lib/common/dialogs/attachment_confirmation.dart
```diff
@@ -1,59 +1,43 @@
...
-  if (context.mounted) {
-    if (selectedFiles != null && selectedFiles.isNotEmpty) {
-      String fileName = selectedFiles.first.path.split('/').last;
+  BuildContext ctx,
+  AttachmentsManager manager,
+  List<AttachmentInfo>? selectedAttachments,
+) {
+  final size = MediaQuery.of(ctx).size;
+  if (selectedAttachments != null && selectedAttachments.isNotEmpty) {
+    if (isLargeScreen(ctx)) {
```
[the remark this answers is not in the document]
gtalha07: Alert/Critical dialogs which needs important actions sounds good. In this case, it shows attachment item list which user have selected and its growable. It'd be somewhat weird to have it dialog-like.

app/lib/common/dialogs/attachment_selection.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:file_picker/file_picker.dart';
+import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
+
+// reusable dialog for attachment selection
+void showAttachmentSelection(
+  BuildContext context,
+  AttachmentsManager manager,
+) async {
+  isLargeScreen(context)
```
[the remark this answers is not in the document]
gtalha07: This can be regarded as something outside of this PR scope. I would have open it as separate ticket if there's a general take. Besides that, desktop attachments design are still unclear as Balsmiq references mobile view.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: This is meant only for chat attachment dialog, it has been revised on ben feedback that chat attachments should have its own functionality and not be mixed up with generic attachments flow. Later we can open ticket for suggestive refactor!

app/lib/common/dialogs/attachment_selection.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show AttachmentsManager;
+import 'package:file_picker/file_picker.dart';
+import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
+
+// reusable dialog for attachment selection
+void showAttachmentSelection(
+  BuildContext context,
+  AttachmentsManager manager,
```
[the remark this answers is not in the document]
gtalha07: I think its bad-case where user selects file and it uploads directly before confirming anything. Not really sure we need control here. I've seen confirmation cases in one form or another in every chat/attachment related app. Element has it too :)

app/lib/common/dialogs/attachment_selection.dart
```diff
@@ -0,0 +1,104 @@
...
+  if (result != null) {
+    List<AttachmentInfo> newAttachments =
+        result.paths.map<AttachmentInfo>((path) {
+      var file = File(path!);
+      var attachment = (type: AttachmentType.file, file: file);
+      return attachment;
+    }).toList();
+
+    if (context.mounted) {
+      attachmentConfirmationDialog(
```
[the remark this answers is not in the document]
gtalha07: Not sure I quite agree here. This would be bad for end-user's consent.

app/lib/common/widgets/attachments/attachment_options.dart
gtalha07: Fixed by 0a89376

app/lib/common/widgets/attachments/attachment_widget.dart
gtalha07: Fixed by 2dc1643

app/lib/common/dialogs/attachment_selection.dart
```diff
@@ -84,25 +79,24 @@ void showAttachmentSelection(
...
     allowMultiple: true,
   );
   if (result != null) {
-    List<File> newAttachments = result.paths.map<File>((path) {
+    List<AttachmentInfo> newAttachments =
+        result.paths.map<AttachmentInfo>((path) {
       var file = File(path!);
-
-      return file;
+      var attachment = (type: AttachmentType.file, file: file);
```
gnunicorn, line 92: :+1:

app/lib/common/widgets/attachments/attachment_draft_item.dart
```diff
@@ -1,51 +1,43 @@
...
 
   // attachment preview handler
-  Widget _attachmentPreview(BuildContext context, File file) {
-    final mimeType = lookupMimeType(file.path)!;
-    if (mimeType.startsWith('image/')) {
+  Widget _attachmentPreview(
+    BuildContext context,
+    AttachmentInfo attachmentDraft,
+  ) {
+    final type = attachmentDraft.type;
```
gnunicorn, line 25: :+1:

app/lib/common/widgets/attachments/attachment_item.dart
```diff
@@ -12,31 +12,28 @@ class AttachmentItem extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     var msgContent = attachment.msgContent();
-    String? mimeType = msgContent.mimetype();
-    if (mimeType == null) {
-      return ErrorWidget(Exception('Invalid Message Content'));
-    }
-    if (mimeType.startsWith('image/')) {
-      return _AttachmentContainer(
+    String type = attachment.typeStr();
```
gnunicorn, line 15: :+1:

app/lib/common/widgets/attachments/attachment_section.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:acter/common/widgets/attachments/attachment_item.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show AttachmentsManager;
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:skeletonizer/skeletonizer.dart';
+
+/// Attachment Section Widget
+class AttachmentSectionWidget extends ConsumerWidget {
```
gnunicorn, line 12: oh, so yum :yum:  ! :+1:

app/lib/common/widgets/attachments/post_attachment_selection.dart
```diff
@@ -99,86 +92,57 @@ class _PostAttachmentSelectionState
...
-      for (var file in selectedAttachments) {
+      for (var selected in widget.attachments) {
+        final type = selected.type;
+        final file = selected.file;
         final mimeType = lookupMimeType(file.path)!;
-        if (mimeType.startsWith('image/')) {
-          final bytes = file.readAsBytesSync();
-          final image = await decodeImageFromList(bytes);
+        final manager = widget.manager;
+        if (type == AttachmentType.camera || type == AttachmentType.image) {
```
gnunicorn, line 107: :clap:

app/lib/common/widgets/attachments/post_attachment_selection.dart
```diff
@@ -99,86 +92,57 @@ class _PostAttachmentSelectionState
...
-      notifier.showReplyView(false);
-      notifier.showEditView(false);
-      notifier.setReplyWidget(null);
-      notifier.setEditWidget(null);
+      for (var draft in drafts) {
+        final res = await draft.send();
+        _log.info('attachment sent: $res');
+      }
+    } catch (e) {
+      EasyLoading.dismiss();
```
gnunicorn, line 144: we should show that error to the user, too!
```suggestion
      EasyLoading.error('Error sending attachments', e);
```

gnunicorn (approved):
looks good to me now :)

PR #1512, opened 2024-03-13
Starting comments on Pins

gnunicorn:
A minimal (UI-wise) but working implementation of a standalone comments-widget attached to a Pin. See:

https://github.com/acterglobal/a3/assets/40496/31e38a76-0948-47dc-8ff6-b785e3b3dd81

Please note:
 - requires the comments features to be turned on, and will stay behind the flag until the #1511
 - is only attached to pins at the moment

PR #1513, opened 2024-03-13
Simplified Manager model processing

gnunicorn:
This refactors the manager-driven models by removing the unnecessary `belongs_to` and simplifying the `execution` for when there can only be one manager (e.g. rsvps, reactions and attachments).

I was hoping this was the main cause for unnecessary UI refresh when liking or adding attachments, but they still seem to be happening :( . At least we can rule out this potential reason ...

on top of #1512 . Easiest reviewed just that last commit: https://github.com/acterglobal/a3/commit/315ea24947319ed4f6d4040e9c5d7ef8bacf8574


app/lib/features/comments/widgets/comments_list.dart
```diff
@@ -0,0 +1,62 @@
...
+
+  Widget onError(BuildContext context, Object error) {
+    return Column(
+      children: [
+        Text('Comments list error: $error'),
+      ],
+    );
+  }
+
+  Widget loading(BuildContext context) {
```
gtalha07, line 55: Why not using ```Skeletonizer``` here???

app/lib/features/comments/models.dart
```diff
@@ -0,0 +1,5 @@
+enum NewCommentLocation {
```
gtalha07, line 1: Why not have it on actual widget where it's used? OR maybe in ```/utils```. Placing it under ```/models``` sounds a bit confusing...

app/lib/features/comments/widgets/comments_section.dart
```diff
@@ -0,0 +1,71 @@
...
+  const CommentsSection({
+    super.key,
+    required this.manager,
+    this.newCommentLocation = NewCommentLocation.before,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final provider = ref.watch(featuresProvider);
+    if (!provider.isActive(LabsFeature.comments)) {
```
gtalha07, line 23: I'm still unsure which new stuff gets added in labs. What's the purpose of adding it there? Does that mean that feature isn't meant to be usable on daily basis? What's the criteria that sets lab features apart from live ones?

app/lib/features/comments/widgets/create_comment.dart
```diff
@@ -0,0 +1,95 @@
+import 'package:acter/common/widgets/html_editor.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_easyloading/flutter_easyloading.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class CreateCommentWidget extends ConsumerStatefulWidget {
```
gtalha07, line 8: We can convert this to simple ```Stateful```, don't see reason otherwise ;)

native/core/src/models/attachments.rs
```diff
@@ -145,35 +141,34 @@ impl ActerModel for Attachment {
...
-        for manager in managers {
+        if let Some(manager) = manager {
             updates.push(manager.save().await?);
         }
         Ok(updates)
     }
 
     fn belongs_to(&self) -> Option<Vec<String>> {
-        Some(vec![self.inner.on.event_id.to_string()])
+        None
```
gtalha07, line 171: If it returns ```None```, why it exists as fn ?

native/core/src/models/comments.rs
```diff
@@ -147,7 +177,7 @@ impl ActerModel for Comment {
     }
 
     async fn execute(self, store: &Store) -> Result<Vec<String>> {
-        let belongs_to = self.belongs_to().expect("we always have some as comments");
+        let belongs_to = self.belongs_to_inner();
```
gtalha07, line 180: 👍🏻

native/core/src/models/comments.rs
```diff
@@ -165,27 +195,17 @@ impl ActerModel for Comment {
...
-                r.event_ids
-                    .iter()
-                    .map(ToString::to_string)
-                    .collect::<Vec<_>>()
-            })
-            .unwrap_or_default();
-        references.push(self.inner.on.event_id.to_string());
-        Some(references)
+        // the higher ups don't need to be bothered about this
+        None
```
gtalha07, line 208: Not sure why we need ```belongs_to``` when it returns nothing...

native/core/src/models/rsvp.rs
```diff
@@ -138,35 +134,38 @@ impl ActerModel for Rsvp {
...
+        if let Some(manager) = manager {
             updates.push(manager.save().await?);
         }
         Ok(updates)
     }
 
     fn belongs_to(&self) -> Option<Vec<String>> {
-        Some(vec![self.inner.to.event_id.to_string()])
+        // the higher ups don't need to be bothered by this
+        None
```
gtalha07, line 168: Same here...

gtalha07 (commented):
Looks good refactor, except some questions from rust side and minor code styling issues on flutter side which feels addressing!

app/lib/features/comments/widgets/comments_section.dart
gnunicorn: > I'm still unsure which new stuff gets added in labs. What's the purpose of adding it there? Does that mean that feature isn't meant to be usable on daily basis? What's the criteria that sets lab features apart from live ones?

Fair enough. Generally my thinking is: everything that is still largely in flux and we don't want to present yet as _is_ should be a labs feature. For everything new that we want to have in `main` but not yet activated by default we should have that feature turned off by default. Once we think it is mature enough that we won't massively change it anymore, we can switch that to default-on and once that has not required any moving back, we should remove the flag entirely.

In this particular case, what I have been doing here is far from what the v1 ticket describes as a default-on-feature so I've put that behind a default-off-feature-flag. Thus we can merge it already and continue to work on it, improve the design and all and everyone, who wants to play with it can, but is clearly opting in to do so.

native/core/src/models/attachments.rs
gnunicorn: it's a trait impl. It has a default for `None` but I thought it might be useful to be explicit in these cases as it otherwise seems like it was just forgotten...

app/lib/features/comments/widgets/comments_list.dart
gnunicorn: That's the plan. But as we don't have any proper UI yet, creating a skeletonizer that copies it isn't much of a point ...

app/lib/features/comments/widgets/create_comment.dart
gnunicorn: editorState and whether the editor is shown (`opened`) are stateful. This is a stateful widget.

native/core/src/models/rsvp.rs
gnunicorn: left a code comment to clarify the explicit `None`.

app/lib/features/comments/models.dart
gtalha07: ```/models``` in general refers to mapped data objects. Going through ```/models```, one would except some data class.

Issue #1128, opened 2023-11-06
fixes for events


emilvincentz:
More important
- [x] #1433 
- [x] Remove "space" between hour and minute scroll (currently when you scroll to set a time, you can only scroll  if you press and scroll e.g. on the number 5 or 6, but if you scroll on the "space" between 5 and 6, you cant scroll.
- [x] Auto select end-date to be 1 hour after the start time. 
(Currently When creating event, the date and time picker Is making troubles. When adding a day and start time, without also jumping to end date, it says error.)
- [ ] #1475
- [x] #1473
- [x] Show the spaces that an event is nested under

less important
- [x] Make minute selector to 5 min intervals, insted 1 minute invervals.
- [x] #1474
- [x] Can we make another time picker solution for desktop? its really hard with the scroll

emilvincentz:
there are two other things I see:
1. Events are not displayed in chronological order
2. Also we need somehow differ between past and upcoming events

gnunicorn:
only one left in this tracking issue and it has its own issue, closing as mostly done.

PR #1514, opened 2024-03-13
Ensure events come in soonest-first order from the rust api

gnunicorn:
refs #1128 .

See how a newly created item isn't added to the end anymore but sorted in to the list depending on start time:

https://github.com/acterglobal/a3/assets/40496/cedba1d3-bca9-4378-b137-d5034639bd65


gtalha07 (approved):
Very nice!

PR #1515, opened 2024-03-13
Refactor roomMember Providers, user info drawer

### LINKED ISSUE #1049: mention user in chat, doesnt always seem to find freshly changed/joined user
https://github.com/acterglobal/a3/assets/77483554/22f3ca93-ffec-402b-8025-f5c486805804

gnunicorn:
Massively refactors the space/convo membership concept to:
1. only provider userIds for active users
2. reuse one single room-based membership provider for all profile data
3. drive-along fix for the way mentions for the user-listing are generated and what is searched for (includes the userId regardless of displayname set), fixes #1049


app/lib/common/dialogs/member_info_drawer.dart
```diff
@@ -13,18 +14,61 @@ import 'package:flutter/material.dart';
...
-class _MemberInfoDrawer extends ConsumerWidget {
-  final ProfileData memberProfile;
-  final String memberId;
+class _MessageUser extends ConsumerWidget {
   final Member member;
-  final String roomId;
-  const _MemberInfoDrawer({
-    super.key,
-    required this.memberProfile,
+  const _MessageUser({required this.member});
```
gnunicorn, line 21: just moved that into a separate widget. all else is a copy-paste.

app/lib/common/models/types.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:acter/common/models/profile_data.dart';
```
gnunicorn, line 1: this is a pure move from the utils into a more meaningful file.

app/lib/common/dialogs/member_info_drawer.dart
```diff
@@ -174,12 +178,108 @@ class _MemberInfoDrawer extends ConsumerWidget {
   }
 }
 
+class _MemberInfoSkeleton extends StatelessWidget {
```
kumarpalsinh25, line 181: Good to have it on separate skeleton file and folder (If have multiple skeletons).

app/lib/common/dialogs/member_info_drawer.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:acter/common/dialogs/block_user_dialog.dart';
```
kumarpalsinh25, line 1: I see there there are multiple class related to memberInfo features in this single file (`app/lib/common/dialogs/member_info_drawer.dart)`. I would suggest to create `member_info` folder and put different class in different files for better code readability.

app/lib/common/widgets/member_list_entry.dart
```diff
@@ -176,17 +177,84 @@ class _ChangePowerLevelState extends State<ChangePowerLevel> {
...
-  final Convo? convo;
+  final String memberId;
+  final String roomId;
   final Member? myMembership;
 
   const MemberListEntry({
     super.key,
+    required this.memberId,
+    required this.roomId,
+    this.myMembership,
```
kumarpalsinh25, line 225: Why to have `myMembership` in constructer even thought we can use `final myMembership = ref.watch(roomMembershipProvider(roomId));` in this class?

kumarpalsinh25: Better to have minimus argument to standalone component to make it independent of data dependency.

kumarpalsinh25 (approved):
Very nice! Much simplified code for `roomMemebrProviders`.

Add minor suggestions but you can do that on separate PR if needed. :)

app/lib/common/dialogs/member_info_drawer.dart
gnunicorn: They are all _intentionally_ internal (prefixed with a `_`). If I put them into several files or folders I would have to make them public and but I don't want any API public other than the `showMemberInfoDrawer()` fn. That's why I will leave it as is.
`

PR #1516, opened 2024-03-14
i18n lang support : Setup + Auth Module

kumarpalsinh25:
- - -
### READ ENTIRE DESCRIPTION BEFORE STARING CODE REVIEW
 - - -

This is the starting point of the `i18n - Multiple languages` work with the below features.

 - - -
1) fixes, #1334
- Changes related to i18n language integration
- Default language selection logic

   - **Scenario-1: When user opens app first time**
      - Select `Device lang` as default to Acter App if it is part of our Acter Supported language list
      - Select `English` lang as default to Acter App if `Device lang` is not the part of our Acter Supported language list
   
   - **Scenario-2: When user opens app second time**
      - Select `Default language` to Acter App based on the settings preference criteria
 
- Change language option from settings screen
   - Reference Video
      https://github.com/acterglobal/a3/assets/51526480/aada3172-99bc-4a2f-9a74-a60d6785b966
 - - -
 
2) ref, #1517
- Currently with this PR, entire auth module is covered. Rest of the screens will be cover in follow-up PRs.

 - - -
###  NOTES:
 - This PR contains 27 files changes but 90% of changes are just naming and key related changes.
 - Add few comments on code for better understanding during review
 - Only English lang arb is added (As it is consider as default template)
 - Added dutch lang arb file with single key only for working demonstration. We need to add all the keys (based on the template eng arb file) with relevant dutch lang translation. 
Screenshot 2024-03-14 at 5 40 18 PM


app/lib/features/onboarding/pages/intro_page.dart
```diff
@@ -11,11 +12,17 @@ class IntroPage extends StatefulWidget {
...
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    introTexts = [
+      L10n.of(context).simpleToUse,
+      L10n.of(context).secureWithE2EEncryption,
+      L10n.of(context).powerfulForAnyOrganizer,
+    ];
+  }
```
gnunicorn, line 25: why is this considered a state not just something we generate once at `build`?

gnunicorn, line 25: why is this a state rather than just a let in `build`?

app/lib/features/onboarding/pages/start_page.dart
```diff
@@ -16,229 +17,233 @@ class StartPage extends StatefulWidget {
...
-              text: 'Spaces',
-              style: TextStyle(
-                fontWeight: FontWeight.bold,
-                decoration: TextDecoration.underline,
+  List onBoardingPages = [];
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    onBoardingPages = [
```
gnunicorn, line 30: same here, why not just in `build`?

app/lib/l10n/app_de.arb
```diff
@@ -1,101 +1,3 @@
 {
-    "language": "Deutsch",
```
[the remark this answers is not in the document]
gnunicorn: "de" stands for German ("Deutsch" in its own tongue), not Dutch. Dutch is the language spoken in the netherlands and has the short-code `nl`. I know this is confusing ;) .

app/lib/l10n/app_de.arb
```diff
@@ -1,101 +1,3 @@
...
-    "verificationCodeNotice":"Vergleiche den Code mit dem Code auf dem Bildschirm deines Gegenübers.",
-    "verificationConclusionOkNotice":"Nachrichten mit diesem Gegenüber sind Ende-zu-Ende-verschlüsselt und können nicht von Dritten gelesen werden.",
-    "verificationConclusionOkSelfNotice":"Deine neue Sitzung ist jetzt verifiziert. Sie hat Zugriff auf deine verschlüsselten Nachrichten, und andere Benutzer sehen sie als vertrauenswürdig an.",
-    "verificationConclusionOkDone":"Sie haben %s erfolgreich bestätigt!",
-    "verificationConclusionNotSecure":"Nicht sicher",
-    "verificationConclusionCompromised":"Eine der folgenden Möglichkeiten kann beeinträchtigt sein:\n\n   - Dein Homeserver\n   - Der Homeserver mit dem dein Gegenüber verbunden ist\n   - Deine oder die Internetverbindung des Gegenüber\n   - Dein Gerät oder das Gerät des Gegenüber",
-    "verifyCancelledNotice":"Verifizierung abgebrochen. Du kannst sie erneut starten.",
-    "verifyInvalidQrNotice":"Dieser QR-Code sieht fehlerhaft aus. Bitte versuchen Sie es mit einer anderen Methode.",
-    "verificationCancelled":"Verifizierung abgebrochen"
+  "selectLanguage": "Selecteer Taal"
```
gnunicorn, line 2:
```suggestion
  "selectLanguage": "Sprache wählen"
```

this is German. The other was "dutch".

app/lib/common/utils/language.dart
```diff
@@ -0,0 +1,32 @@
...
+  final bool isLanguageContain = LanguageModel.allLanguagesList
+      .contains(LanguageModel.fromCode(deviceLanguageCode));
+  if (prefLanguageCode == null && isLanguageContain) {
+    await setLanguage(deviceLanguageCode);
+    ref.read(languageProvider.notifier).update((state) => deviceLanguageCode);
+  } else if (prefLanguageCode != null) {
+    ref.read(languageProvider.notifier).update(
+          (state) => prefLanguageCode,
+        );
+  }
```
gnunicorn, line 21: yeah, I don't think that is the logic we should do. If the user changed the system language between restarts and we have support for that language, we should use that, unless the user has explicitly set the language.

IMHO the logic should be:
1. user has set a language manually, use that
2. pick system language if available
3. fallback to english

But do not implicitly _write_ a language selection upon first start. Why would that be useful?

app/lib/l10n/app_de.arb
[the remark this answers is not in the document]
kumarpalsinh25: Oh got. I was not aware about this. Thanks for the information.

app/lib/features/onboarding/pages/intro_page.dart
kumarpalsinh25: I need to initialise `introTexts` array data but I can't directly do it at `variable declaration` time or in `initState` as for localise string, I need `context`.

So there were two options I had where I can do this `initialisation` with `context`.
1) `didChangeDependencies`
2) `build` method 

I preferred to do that on `didChangeDependencies` as just called right after `initState` and before `build` in lifecycle events. And our purpose it also kind of initialisation of static string values.

Reference:
https://api.flutter.dev/flutter/widgets/State/didChangeDependencies.html
Screenshot 2024-03-15 at 1 05 10 PM

app/lib/features/onboarding/pages/start_page.dart
kumarpalsinh25: Same as answer mentioned in above comment.

app/lib/common/utils/language.dart
kumarpalsinh25: It is the same logic.


> 1) user has set a language manually, use that:

This use-case is satisfied with below code. If user has set language manually then `prefLanguageCode` will not be `null` and below case will executed.

```
else if (prefLanguageCode != null) {
    ref.read(languageProvider.notifier).update(
          (state) => prefLanguageCode,
        );
}
```
  

> 2) Pick system language if available

This use-case is satisfied with below code. If user has not set language manually then `prefLanguageCode` will be `null` and if system language is available then below case will executed.

```
if (prefLanguageCode == null && isLanguageContain) {
    await setLanguage(deviceLanguageCode);
    ref.read(languageProvider.notifier).update((state) => deviceLanguageCode);
} 
```

> 3) fallback to english

If above two use-cases are not satisfied then we are using default "English" language with below code where we have already set 'en' as default language.

`final languageProvider = StateProvider<String>((ref) => 'en');
`

gtalha07 (approved):
LGTM!

app/lib/features/onboarding/pages/intro_page.dart
gnunicorn: the reason I dislike doing that in `didChangeDependencies` rather than in `build` is that the function is there to initialize _state_. But this isn't state. This is _technically_ static, never changing content. Thus it should not be acted upon as if it was ever changing - it's just confusing to make that a _variable_ where it being `final` (by just using it in `build`) would do.

app/lib/common/utils/language.dart
gnunicorn: This is only sufficient on first-start. If you changed the language _after_ it will not fallback to the system language. I don't understand why we'd want to force _write_ a local config if the user never changed it, it also logically doesn't make any sense to me to do that. What's the benefit?

kumarpalsinh25: I have followed the WhatsApp mechanism. Let me explain with the example.

Current Status:
- - -
Mobile OS : English Lang
WhatsApp : English Lang

STEP-1 : I have opened the `WhatsApp` and changed the lang to "Hindi". 
So now,
Mobile OS : English Lang
WhatsApp : Hindi Lang

STEP-1 : I have opened the `Mobile App Lang` settings and changed the lang to "Gujarati". 
So now,
Mobile OS : Gujarati Lang
WhatsApp : Hindi Lang (This is preserved here based on the my prior selection)

gnunicorn: That's not what I have a problem with. [Here is a PR explaining what I want to have changed](https://github.com/acterglobal/a3/pull/1521).

PR #1518, opened 2024-03-14
Miscellaneous refactoring

bitfriend:
- Remove unnecessary macro for github lint
- Mark send_verification_key as testing fn
- Replace flutter_riverpod with riverpod


app/lib/features/chat/widgets/message_metadata_builder.dart
```diff
@@ -5,10 +5,10 @@ import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
     show Convo, EventSendState;
 import 'package:atlas_icons/atlas_icons.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:quds_popup_menu/quds_popup_menu.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
+import 'package:flutter_riverpod/flutter_riverpod.dart';
```
gnunicorn, line 9: you removed it everywhere else, but here you kept `flutter_riverpod` ... why?

gnunicorn:
> Replace flutter_riverpod with riverpod

If so, why doesn't this remove the `flutter_riverpod`-dependency from the `pubspec.yaml`? there is at least one occurrence in there, I marked in my PR, but other than that, why stop there?

bitfriend:
`flutter_riverpod` contains `riverpod`.
I just wanted to reduce outer dependency of each module.

app/lib/features/chat/widgets/message_metadata_builder.dart
bitfriend: Provider may use `riverpod`, but UI should use `flutter_riverpod`.
I wanted to reduce the occurrence of `flutter_riverpod` usage as many as possible.

PR #1520, opened 2024-03-15
attachments redaction

gtalha07:
## Adds the ability to delete attachments from pins. 

**Notably this also includes the following UI changes and fixes:**
1. Fix context not in mounted state after picker opens and that makes confirmation dialog unable to appear. Commit 5df8b28.
2. Transition dialog state on mobile for selected attachments preview. (Bottom sheet is removed because it's complex to manage context pop offs while determining modal state .i.e. dialog/bottomsheet switch. Commit 5df8b28

Desktop: https://github.com/acterglobal/a3/assets/68579938/0dd0c96e-2000-4397-ab12-a5ddd2abeeac

Mobile: https://github.com/acterglobal/a3/assets/68579938/9b2c42b3-9297-4707-afa8-9a899d73beac


app/lib/common/widgets/attachments/attachment_section.dart
```diff
@@ -64,6 +73,127 @@ class AttachmentSectionWidget extends ConsumerWidget {
     );
   }
 
+  Widget _buildAttachmentItem(
+    BuildContext context,
+    Attachment item,
+    bool? canRedact,
```
kumarpalsinh25, line 79: No need to pass `canRedact` as it can be directly accessible from class variable.

app/lib/common/widgets/attachments/attachment_section.dart
```diff
@@ -64,6 +73,127 @@ class AttachmentSectionWidget extends ConsumerWidget {
...
+    return Stack(
+      children: [
+        AttachmentItem(
+          attachment: item,
+        ),
+        Positioned(
+          top: -12,
+          right: -12,
+          child: Visibility(
+            visible: canRedact!,
```
kumarpalsinh25, line 93: `canRedact` condition can be directly use at `Positioned` widget level

kumarpalsinh25 (approved):
Nice addition to the to attachment feature. 

Looks good Flutter side code.

kumarpalsinh25:
@gtalha07 
Just noted this glitch on screen- recording. I think we need to show video thumb image over there.
Screenshot 2024-03-20 at 1 35 19 PM

app/lib/common/widgets/attachments/attachment_section.dart
gtalha07: You're right. Thanks for pointing it out!

gtalha07:
> > @gtalha07 Just noted this glitch on screen- recording. I think we need to show video thumb image over there. !
> 
> I believe it's because of the bug `attachment.typeStr()` returning `image` for video type and UI treats it as an image instead of video. I will look into it if its a short fix 👍🏻.

@kumarpalsinh25 @gnunicorn  As anticipated, a similar fix had been done regarding news video slide appearing as image, I believe its the same bug we had fixed before by applying ```serde[tagged]``` over actual ```AttachmentContent``` and using fallback enum with ```serde[untagged``` as it takes up ```image``` over any other type. Can you point me the PR where its fixed? I believe it was Kumar's one.

app/lib/common/widgets/attachments/attachment_section.dart
gtalha07: Fixed by e704014.

PR #1535, opened 2024-03-21
i10 Support to settings module

kumarpalsinh25:


app/macos/Podfile.lock
```diff
@@ -156,4 +162,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce7dbe26c78bfc7ba46736094c1e2d25982870fa
 
-COCOAPODS: 1.14.3
\ No newline at end of file
+COCOAPODS: 1.14.2
```
gtalha07, line 165: This keeps changing, we should keep it in ```.gitignore``` :)

app/pubspec.lock
```diff
@@ -1293,6 +1293,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "3.16.7"
+  multiavatar:
```
gtalha07, line 1296: Same here, this changes aren't related to actual PR.

gtalha07 (approved):
Looks good, just keep an eye out on linter complains ;)

app/pubspec.lock
gnunicorn: but the pubspec.lock needs to go into the repo or it won't properly resolve

Issue #1444, opened 2024-02-26
Update Registration Screen


JMEFKT:
provide (linked) information on how Acter is safe+secure, why to always remember username, email address only optional, add #1441 & #1440 etc.

design iteratons: https://www.figma.com/file/ps6Wa0fuRYkKNjPPaZqVaO/Acter-3.0---Design?type=design&node-id=10416-11883&mode=design&t=zQexl6nXBrNXQwe0-4

examples for next level registration: https://miro.com/app/board/uXjVMlxEtAU=/?moveToWidget=3458764580888038792&cot=14

JMEFKT:
other examples: https://miro.com/app/board/uXjVMlxEtAU=/?moveToWidget=3458764580888038792&cot=14

MirekKolaja:
@kumarpalsinh25., the sign up form is now ready for Dev including these changes:

- [ ] Adding eye icon to see/unsee the password
- [ ] Spacing between input fields so there is a space for error Indicators
- [ ] Labels and the input fields are separately + there is a Plaholder text in the input fields
- [ ] renaming "Code" to "Invite Code"
- [ ] Invite Code tooltip with explanation with what it is
- [ ] Adding screens for "Save Username" and "Linking Email to Profile"

The final designs can be found here: 

https://www.figma.com/file/ps6Wa0fuRYkKNjPPaZqVaO/Acter-3.0---Design?type=design&node-id=11055-15617&mode=design&t=Sa698FCXCaPcxgGF-4

emilvincentz:
Hey @kumarpalsinh25 
Just saw that there is a few bugs on the registrations screens.
1. On iphone SE you can not see the login button, only signup and also you can scroll
2. There is a lot of black space at the top


![Image](https://github.com/acterglobal/a3/assets/77483554/950a5340-de38-4d7c-8728-ecf9e70cef2e)
![Image](https://github.com/acterglobal/a3/assets/77483554/9320042a-f056-4a5f-821e-a8744c02bd90)

kumarpalsinh25:
Opps! I will do the needful for smaller screens support.

JMEFKT:
the copy of the invite code explanation/tool tip needs to be adjusted & might also include an invite code itself (update beginning of next week)

JMEFKT:
not sure about the "e.g. for password recovery" part, that's a new addition, don't think we have that flow, that sounds like an automated process

[...]

PR #1536, opened 2024-03-22
Relative height fixes for onboarding screens.

kumarpalsinh25:
Intro
Intro (1)
Intro (2)
Intro (3)

fixes, https://github.com/acterglobal/a3/issues/1444#issuecomment-2012217442

### DIFF
--- a/app/lib/features/onboarding/pages/login_page.dart
+++ b/app/lib/features/onboarding/pages/login_page.dart
@@ -56,6 +56,8 @@ class _LoginPageState extends ConsumerState<LoginPage> {
   }
 
   Widget _buildLoginPage(BuildContext context) {
+    var screenHeight = MediaQuery.of(context).size.height;
+    var imageSize = screenHeight / 5;
     return Expanded(
       child: SingleChildScrollView(
         child: Container(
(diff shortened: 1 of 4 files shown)


kumarpalsinh25:
@emilvincentz 
This PR will fixed the UI issues related iPhone SE3.
Uploaded screenshot of different devices in the PR description for the better visualisation. :)

app/lib/features/onboarding/pages/login_page.dart
```diff
@@ -65,9 +67,8 @@ class _LoginPageState extends ConsumerState<LoginPage> {
             mainAxisAlignment: MainAxisAlignment.start,
             crossAxisAlignment: CrossAxisAlignment.stretch,
             children: [
-              const SizedBox(height: 12),
-              const LogoWidget(width: 150, height: 150),
-              const SizedBox(height: 20),
+              if (screenHeight > 650)
+                LogoWidget(width: imageSize, height: imageSize),
```
gnunicorn, line 71: we show now logo at all on smaller screens?

gnunicorn (approved):
lgtm

app/lib/features/onboarding/pages/login_page.dart
kumarpalsinh25: The reason it that, user will not able to see `LOGIN` button on first visualisation on very small device screens. They need to scroll a bit so that I think better to hide upper side logo on very small devices so that necessary data are visible to user without scroll.

PR #1548, opened 2024-03-26
i18n Support : Space module

kumarpalsinh25:
There are plenty of file need to be review but changes are pretty simple which replace static string with L18n string. There are no changes related to logical part or UI part.


gtalha07 (approved):
Looks good to me, nicely done. This replaces static strings with generic l10 string but I see we didn't have put ***Deutsch*** version for it, could you add it in comments or put ticket for others so that they know and seem capable to add it?

kumarpalsinh25:
> Looks good to me, nicely done. This replaces static strings with generic l10 string but I see we didn't have put _**Deutsch**_ version for it, could you add it in comments or put ticket for others so that they know and seem capable to add it?

Thanks @gtalha07 

Regarding `Deutsch` support, first goal is to replace all the static strings with L10 strings with english arb template file.
Once we have entire template english arb file ready then it is just all about adding additional language file.

Issue #1482, opened 2024-03-06
Pins: Attachments Previewing


gtalha07:
Pins Attachments should get previewed on user interaction based on media type .i.e. image, video, audio, file for now. Can use existing render components of other modules?? News ?

PR #1558, opened 2024-03-28
Attachments view

gtalha07:
Refs -> #1482.

### Attachments can now be viewed in app (supported for images and videos). Other file attachments can be shared natively.
1. On the rust-side, this includes adding ```download_media``` and ```media_path``` methods over attachments api, making it much easier to preview and save it over cache directory. 
2. Flutter side changes include adding dialog for attachment selection (Desktop), allowing user to choose file type selection just like on mobile version.

**Desktop Preview:** 
https://github.com/acterglobal/a3/assets/68579938/4bae4ba7-09b9-4438-bd2e-9673a1791028

**Mobile Preview:**
https://github.com/acterglobal/a3/assets/68579938/8ece8634-7f4a-45ba-866d-3f3c2fa29fa3


app/lib/common/widgets/attachments/attachment_item.dart
```diff
@@ -20,123 +23,18 @@ class AttachmentItem extends ConsumerWidget {
     if (type == 'image') {
       return AttachmentContainer(
         name: msgContent.body(),
-        child: _ImageAttachment(attachment: attachment),
+        child: ImageView(attachment: attachment, openView: openView),
```
kumarpalsinh25, line 26: We can re-used full screen image view common dialog which is already used in chat media.
`import 'package:acter/common/widgets/image_dialog.dart';`

app/lib/common/widgets/attachments/attachment_item.dart
```diff
@@ -20,123 +23,18 @@ class AttachmentItem extends ConsumerWidget {
...
-          child: Icon(Atlas.file_video_thin),
-        ),
-      );
-    } else if (type == 'audio') {
-      return AttachmentContainer(
-        name: msgContent.body(),
-        child: const Center(
-          child: Icon(Atlas.file_audio_thin),
-        ),
+        child: VideoView(attachment: attachment, openView: openView),
```
kumarpalsinh25, line 31: Same as above.
import 'package:acter/common/widgets/video_dialog.dart';

app/lib/common/widgets/attachments/attachment_item.dart
```diff
@@ -1,16 +1,19 @@
...
+import 'package:acter/common/widgets/attachments/views/video_view.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' show Attachment;
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:flutter_gen/gen_l10n/l10n.dart';
 
 // Attachment item UI
 class AttachmentItem extends ConsumerWidget {
   final Attachment attachment;
+  final bool? openView;
```
kumarpalsinh25, line 12: What is the use-case of `openView`?

kumarpalsinh25 (approved):
Overall looks good.

Added some queries and suggestions which need to be answered before merged. :)

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: The ```imageView``` uses that dialog. This widget is part of ```imageView```. See line 53.  
```
Widget imagePlaceholder(
    BuildContext context,
    Attachment attachment,
    AttachmentMediaState mediaState,
    WidgetRef ref,
  ) {
    final msgContent = attachment.msgContent();
    return InkWell(
      onTap: () async {
        if (mediaState.mediaFile != null) {
          showAdaptiveDialog(
            context: context,
            barrierDismissible: false,
            useRootNavigator: false,-
            builder: (ctx) => ImageDialog(                 <--- Component
              title: msgContent.body(),
              imageFile: mediaState.mediaFile!,
            ),
          );
        } else {
          ref
              .read(attachmentMediaStateProvider(attachment).notifier)
              .downloadMedia();
        }
      },
      ```

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: Same as above, ```VideoView``` has internal placeholder widget which uses ```VideoDialog``` component.

app/lib/common/widgets/attachments/attachment_item.dart
gtalha07: Should've add docs here. ```openView``` determines whether the pin attachment is clickable to be viewed in full screen or not. It's useful for pin lists where main attachment is shown and we don't want to have this opened unintentionally while opening pin list item.

PR #1561, opened 2024-03-29
Remove media cache path from permanent RestoreToken

### LINKED ISSUE #1422: Fatal error: operation not permitted (os error 1)
happens both on Iphone and MacOS. 
Note sure what causes it. But it seems to come when im opening the app after some time or close the app to reopen, but not always.

Image

gnunicorn:
This fixes #1422: "Fatal error: operation not permitted (os error 1)]()"

## Evidence:

Some further debugging on an effected instances showed that despite it working fine before, when trying to "recreate" the media cache dir from the restore token, we've gotten a the "Operation not permitted" Error thrown at us. A further attempt to just `ls` the folder that was stored in the restore token, told us that wasn't available ... and it contained what we identified as a "Container"-ID. 

## Explanation:

We have had seen a similar problem before with the `base_path`, where in some "sandboxed" environments (most notably iOS) the `getApplicationPath` would give us a different path upon every run of the App. The background is that while the files and folders are untouched, the app runs in a "container" and has read-write-access to that container only. Upon startup a fresh container is created and all previousl local data is hooked into that container, returning those changing path (as each container run gives it a new UUID and mount point on the file system).

For that reason we had to move out the `base_path` from the restore token and instead reconstruct the _actual path_ to data ourselves once we knew the new container UUID we were running on. Up until recently the `cache`-folder wasn't effected by this and we had stored it in the restore token and read from it without problems. But it appears that this has changed now and app-cache is also moved into the containerized environment-system now.

That explains why when trying to recreate the folder, we didn't have the permission by the system and why it only happened upon restarts. 

## The fix:
This PR fixes that issue by removing the `media_cache_dir`-field from the `RestoreToken` and instead goes the same way as with the `base_path` of reconstructing the same every time upon startup.

The only reason it was stored in the first place was to allow for older logins, that still used the existing db-base-cached to remain using that cache. This was a nice feature for the users, but is ultimately not necessary as a) we don't want this to be the case for long anyway and b) it would only fetch the data to be cached once and while it would store it twice now (once in the cache, once in the db) only a super small minority of users is ever going to be effected - a fresh relogin resolves that issue entirely.

## The proof

On a system that consistently showed this problem, this fix allowed @kumarpalsinh25 to use the app as always without running into the error again.

### DIFF
--- a/app/packages/rust_sdk/lib/acter_flutter_sdk.dart
+++ b/app/packages/rust_sdk/lib/acter_flutter_sdk.dart
@@ -263,11 +263,11 @@ class ActerSdk {
     return await client.getNotificationItem(roomId, eventId);
   }
 
-  Future<void> _maybeMigrateFromPrefs(appDocPath) async {
+  Future<void> _maybeMigrateFromPrefs(String appDocPath, String appCachePath) async {
     SharedPreferences prefs = await sharedPrefs();
     List<String> sessions = (prefs.getStringList(_sessionKey) ?? []);
     for (final token in sessions) {
-      ffi.Client client = await _api.loginWithToken(appDocPath, token);
+      ffi.Client client = await _api.loginWithToken(appDocPath, appCachePath, token);
       _clients.add(client);
     }
     _index = prefs.getInt('$_sessionKey::currentClientIdx') ?? 0;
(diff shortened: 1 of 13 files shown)


kumarpalsinh25 (approved):
Amazing Work!
This one not easy fix and figure out the root cause but with your great technical skills it is now fixed.

Highly Appreciated!

Issue #1556, opened 2024-03-28
Fix browsing of spaces user isn't part of (yet/anymore)


gnunicorn:
Given Space "Venus" and users Aisha and Burgho. Once Aisha invited Burgho to "Venus", it shows up in the space listing for Burgho and on the invites screen. However there are few screens that fail to render when Burgho surfs them - most notably the space-chat-screen.

We need to harden our story in two main parts here:
1. make sure a Space can be shown without too much trouble, even if you aren't part of it
2. indicator showing the user they aren't part of it - and in case of being kicked or banned the messages that came alongside with it

Also see https://github.com/acterglobal/a3-meta/issues/152

gnunicorn:
temporarily fixed by #1576

PR #1576, opened 2024-04-02
Improved invitation handling, remove not-joined Spaces from UI

gnunicorn:
- [x] only show spaces the user is currently part of (do not include invited and left spaces), fixes https://github.com/acterglobal/a3-meta/issues/158 , https://github.com/acterglobal/a3-meta/issues/152 . This is also a short-term fix for the problem of https://github.com/acterglobal/a3/issues/1556 by not showing them in any regular UI anymore.
- [x] Improve Invitation screen: now shows Avatar and User Info (as known), see below
- [x] Overall improvements of the underlying code, removing old unneeded files, cleaning up some rust and flutter stuff within the invitation parts
- [x] Separate out DMs from non-DMs and Spaces with distinct messages and design for more user clarity

This is what that looks like now:

grafik


native/acter/api.rsh
```diff
@@ -239,6 +239,12 @@ object UserProfile {
 }
 
 object RoomProfile {
+    /// get room id
+    fn room_id() -> RoomId;
```
gtalha07, line 243: What's the use of duplicating this information methods when we only seems to use string cases ? I don't recall having ```RoomId``` object being of any benefit over front-end...

native/acter/api.rsh
```diff
@@ -2543,14 +2549,20 @@ object Invitation {
...
-    fn room_name() -> Future<Result<string>>;
+    /// the RoomId as a String
+    fn room_id_str() -> string;
+
+    /// get the room of this invitation
+    fn room() -> Room;
 
     /// get the user id of this invitation sender
-    fn sender() -> UserId;
+    fn sender_id() -> UserId;
```
gtalha07, line 2559: Same goes for this ```UserId```. We only seem to use its string version which is readable and processable to flutter...

native/acter/src/api/profile.rs
```diff
@@ -111,6 +112,14 @@ impl RoomProfile {
         RoomProfile { room }
     }
 
+    pub fn room_id(&self) -> OwnedRoomId {
```
gtalha07, line 115: Here it goes again...I'm not sure if ```RoomId``` provides anything useful other than being a identifier which ```room_id_str``` does it.

gtalha07 (approved):
Nothing major on flutter or rust side changes other than some api method declaration questions which can be addressed regardless of merge.

kumarpalsinh25 (approved):
Flutter code looks good to me.

Flow for inviter information look confusing to me.

1) John invite you to join John : Looks weird for DM
Screenshot 2024-04-03 at 4 33 27 PM

2) Beacon invited you to join Ben : Don't know who invites what
Screenshot 2024-04-03 at 4 34 53 PM

Suggestion:
Invitation

native/acter/api.rsh
gnunicorn: technically, it would be a better if we were to pass only `RoomId` types in and out of the API, but that isn't really feasible until we've moved to uniffi. I have removed these from the api.rsh but left the actual `fn`s in place for future usage.

gnunicorn:
@kumarpalsinh25 how about this:

grafik

gnunicorn:
@gtalha07 , @bitfriend I did some significant refactoring of the invitation controller on the rust side to allow for us to have proper `Room`-Support and figure out the DM status from the invitations. I'd appreciate if you could review that to ensure I didn't mess anything up as well..

gtalha07 (approved):
Generally looks good rust-wise. Although I'd like @bitfriend to review logically if you have some doubts! :)

gtalha07:
Also you'd might want to take care flutter lints as what CI is complaining ;)

PR #1579, opened 2024-04-03
Cleaning up the l10n template file

bitfriend:
Add the missed records to en.arb


gnunicorn (approved):
This is fine. But for future record, other than adding and removing specific entries because we changed the code, all other changes of this file should be done through weblate.org which is the source of truth for this file and all other language files ...

Issue #1343, opened 2024-01-31
'please select room'-screen occasionally appears under shaky connection conditions


gnunicorn:
![Screenshot_20240131-125644](https://github.com/acterglobal/a3/assets/40496/b4cea4d8-69e8-4ac8-9f83-5ac8c44e382c)

gnunicorn:
Also see that shortly pop up when I navigate back from a room. Indicating that there is something wrong with the way the size-calculation works.

PR #1581, opened 2024-04-03
Refactor chat routing to avoid riverpod states

gnunicorn:
We've had several reports that through some oddities in navigation, the user could end up in a situation where rather than the chat list for selection they'd see the [chat-selection-empty-filler](https://github.com/acterglobal/a3/issues/1343).

This is due to the fact that in order to maintain the sidebar and expansion-bar (for the room profile page), we used some riverpod providers that were manually updated within build methods. As any manual updating of riverpod, this, too, is frickely and sometimes went out of sync with the actual screen view (maybe due to orientation changes and thus corresponding updates that aren't done).

This PR refactors the ChatLayoutBuilder and the internal pages to remove the riverpod states and instead properly pass along the right information within the widget-hierarchy. Through some clever changes this also removed a bunch of code for explicit back-button behaviour by now replacing most of that with proper "pushNamed"-calls rather than weird `goNamed` linking behaviours. Further more, the way the code has been refactored the empty-filler can only ever happen when we _also_ display the room list now.

Notice how in this screencast the screens and their arrows appear and disappear (or in case of the profile are replaced with a close button) purely depending on available view size and despite having routed into them in the large view, the proper link stacking behaviour is maintained regardless of resizes. 

https://github.com/acterglobal/a3/assets/40496/b24e0e96-aba0-4ca7-ad80-f1281cd0c8ed

Fixes https://github.com/acterglobal/a3-meta/issues/150 
Fixes https://github.com/acterglobal/a3/issues/1343

### DIFF
--- a/app/lib/features/chat/config.dart
+++ b/app/lib/features/chat/config.dart
@@ -1,0 +1,1 @@
+const sidebarMinWidth = 750;
--- a/app/lib/features/chat/pages/room_page.dart
+++ b/app/lib/features/chat/pages/room_page.dart
@@ -31,16 +31,18 @@ import 'package:skeletonizer/skeletonizer.dart';
 class RoomPage extends ConsumerWidget {
   static const roomPageKey = Key('chat-room-page');
   final String roomId;
+  final bool inSidebar;
 
   const RoomPage({
     required this.roomId,
+    required this.inSidebar,
     super.key = roomPageKey,
   });
 
   @override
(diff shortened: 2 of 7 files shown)


kumarpalsinh25 (approved):
Things are simplified!
Changes looks good to me!

gnunicorn:
did a rebase to fix the merge problem, but from a quick look it seems to have worked fine ...

Issue #1074, opened 2023-10-24
Enable user highligh text (so that one can copy the text)


kumarpalsinh25:

kumarpalsinh25:
Hello @emilvincentz @gnunicorn @gtalha07 

![Image](https://github.com/acterglobal/a3/assets/51526480/3ad583c5-498b-464e-bb57-bf5b3fcf1a66)


For this point, instead of provide hightligt text, can provide provide copy option button like WhatsApp, Telegram and other apps are doing?

Because Highlight Text is ideal option for Web Apps. For mobile and desktop app, generally copy button is more relevant.

Please share your thoughts.

emilvincentz:
hi @kumarpalsinh25 
I think this is the updated design. 
Is that you mean?
https://www.figma.com/file/ps6Wa0fuRYkKNjPPaZqVaO/Acter-3.0---Design?type=design&node-id=7354-56908&mode=design&t=bsdGVTonhxasuCAV-4

<img width="475" alt="Screenshot 2023-10-24 at 16 39 31" src="https://github.com/acterglobal/a3/assets/77483554/f398368e-40ba-4d77-b62b-d5ee85f14f00">

kumarpalsinh25:
Yes @emilvincentz,
This is exactly what I mean.

gnunicorn:
has been fixed in #1591

PR #1591, opened 2024-04-05
Chat Input Refactors

gnunicorn:
Trying to figure out why the send button sometimes disappears (see https://github.com/acterglobal/a3-meta/issues/172 e.g.), I noticed that the ChatInputState is unnecessarily complex with several interdependent booleans and refactored it for more specific, cleaner business-flow oriented system. In the process I also cleaned up several others aspects and refactored the Widgets involved as well.

I tried to keep each submit consistent in itself, clearly stating whether that is trying to change code or just refactor a piece. If the entire diff is too complicated to follow, reviewing commit-by-commit might be easier.

As a run-by this also does the following fixes (as separate commits):
 - [x] allow toggle of reactions 
 - [x] remove `overlay_support`: it's outdated and was only used at a single instance
 - [x] Fix the emoji selector colors (see video below)
 - [x] Removing a lot of manual notifier setting in favor of automatic dependency resolving of riverpod providers.
 - [x] Fixes https://github.com/acterglobal/a3-meta/issues/172 : As it turns out, we were missing the proper context to do reporting of sending errors and would fail to execute that code path. As a result the send-button would not be reactivated.
 - [x] Inlined Actions (see below), as per https://github.com/acterglobal/a3/issues/1074#issuecomment-1777370666 (but now background blur yet)
 - [x] Implement copy-message (for text messages). Fixes https://github.com/acterglobal/a3/issues/1074

The new design for emoji selector:

https://github.com/acterglobal/a3/assets/40496/f8aea531-fb2a-47b7-9b84-4f284b949fa9

The new inlined actions:

https://github.com/acterglobal/a3/assets/40496/908d734e-65bc-4b0e-9257-c8130d055b79

The new copy action:

https://github.com/acterglobal/a3/assets/40496/b4d0e6ba-1b6a-4a11-b9cc-c6b5365255fa


app/lib/features/chat/models/chat_input_state/chat_input_state.dart
```diff
@@ -1,24 +1,28 @@
-import 'package:flutter/material.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 
 part 'chat_input_state.freezed.dart';
 
+enum SelectedMessageState {
```
kumarpalsinh25, line 6: 👏

app/lib/features/chat/pages/room_page.dart
```diff
@@ -80,19 +80,7 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
...
-      inputNotifier.showEditButton(false);
-    }
-    if (ref.read(chatInputProvider(roomId)).showReplyView) {
-      inputNotifier.showReplyView(false);
-      inputNotifier.setReplyWidget(null);
-      inputNotifier.setEditWidget(null);
-    }
-    inputNotifier.setCurrentMessageId(message.id);
-    inputNotifier.emojiRowVisible(true);
+    inputNotifier.setActionsMessage(message);
```
kumarpalsinh25, line 83: 👏

app/lib/features/chat/pages/room_page.dart
```diff
@@ -224,16 +212,7 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
...
-              chatInputProvider(roomId).select((ci) {
-                return ci.emojiRowVisible;
-              }),
-            );
-            final inputNotifier = ref.read(chatInputProvider(roomId).notifier);
-            if (emojiRowVisible) {
-              inputNotifier.setCurrentMessageId(null);
-              inputNotifier.emojiRowVisible(false);
-            }
+            ref.read(chatInputProvider(roomId).notifier).unsetActions();
```
kumarpalsinh25, line 215: 👏

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -18,17 +19,32 @@ final chatInputProvider =
...
+final textValuesProvider =
+    StateProvider.family<String, String>((ref, roomId) => '');
+
 final chatInputFocusProvider = StateProvider<FocusNode>((ref) => FocusNode());
 final chatStateProvider =
     StateNotifierProvider.family<ChatRoomNotifier, ChatRoomState, Convo>(
   (ref, convo) => ChatRoomNotifier(ref: ref, convo: convo),
 );
 
+final isAuthorOfSelectedMessage =
```
kumarpalsinh25, line 32: 👏

app/lib/features/chat/models/chat_input_state/chat_input_state.dart
```diff
@@ -1,24 +1,28 @@
...
+  none,
+  actions,
+  replyTo,
+  edit,
+}
+
+enum SendingState {
+  preparing,
+  sending,
```
kumarpalsinh25, line 15: Do we not require other states like `sendingFail`, `sent`?

app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
```diff
@@ -1,72 +1,68 @@
...
-  void prepareSending() {
-    state = state.copyWith(sendBtnVisible: false, allowEdit: false);
+  void startSending() {
+    state = state.copyWith(sendingState: SendingState.sending);
   }
 
   void sendingFailed() {
     // reset the state;
-    state = state.copyWith(sendBtnVisible: true, allowEdit: true);
+    state = state.copyWith(sendingState: SendingState.preparing);
```
kumarpalsinh25, line 61: `SendingState.preparing` for sendingFailed ?

app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
```diff
@@ -1,72 +1,68 @@
...
   }
 
   void messageSent() {
     // reset the state;
-    state = const ChatInputState();
+    state = state.copyWith(sendingState: SendingState.preparing);
```
kumarpalsinh25, line 66: `SendingState.preparing` for messageSent?

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -58,38 +58,17 @@ class BubbleBuilder extends ConsumerWidget {
...
-                        } else {
-                          chatInputNotifier.showReplyView(true);
-                          chatInputNotifier.setRepliedToMessage(message);
-                          chatInputNotifier.setReplyWidget(child);
-                        }
-                      },
+                onLeftSwipe: (DragUpdateDetails details) {
+                  FocusScope.of(context)
+                      .requestFocus(chatInputFocusState.state);
+                  chatInputNotifier.setReplyToMessage(message);
```
kumarpalsinh25, line 64: What about `isAuthor` condition here?

app/lib/l10n/app_en.arb
```diff
@@ -199,6 +199,8 @@
   "@convertingToActerSpace": {},
   "copyUsername": "Copy username",
   "@copyUsername": {},
+  "copyMessage": "Copy",
```
kumarpalsinh25, line 202:
```suggestion
  "copy": "Copy",
```

Better to make it generic as it can be useful at other place also

app/lib/features/chat/widgets/message_actions.dart
```diff
@@ -0,0 +1,278 @@
...
+import 'package:flutter_chat_types/flutter_chat_types.dart';
+import 'package:flutter_easyloading/flutter_easyloading.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:html/parser.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat::message_actions');
+
+class MessageActions extends ConsumerWidget {
```
kumarpalsinh25, line 19: 👏

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -2,17 +2,15 @@ import 'dart:io';
 
 import 'package:acter/common/models/types.dart';
```
kumarpalsinh25, line 3: Overall changes in this class looks to me still if @gtalha07 can review this once then it will be good as he might have better understanding of this class.

kumarpalsinh25 (approved):
Nice Improvements!
Good to go for merge after addressing minor queries mentioned in comments. :)

kumarpalsinh25:
How this will show if message content is too big?

Screenshot 2024-04-09 at 1 08 02 PM

app/lib/l10n/app_en.arb
gnunicorn: I was considering that, but then thought that I could see the the case where this would be a different term than using the general one in other languages so added the specifier "Message" (as in "copy this message" - not as a generic term). I guess one can argue either way...

app/lib/features/chat/widgets/bubble_builder.dart
gnunicorn: why? you can reply to your own message ...

app/lib/features/chat/models/chat_input_state/chat_input_state.dart
kumarpalsinh25: No worries, we can keep it simple for now and add those scenario in future based on the requirements.

app/lib/l10n/app_en.arb
kumarpalsinh25: No problem, understood your thoughts. We can keep this as it is.

gnunicorn:
@kumarpalsinh25 they are currently organised as inline-events. So when you long-press the item to make them appear, they change the flowing behavior of the entire list (see this message pop "up"). So you can just scroll around and the bottom and top will always be reachable. Good catch though as when I trying to do that overlay-flow-version also noticed the designs don't really spell out what is supposed to happen if the element is either at the top, the bottom or covers the entire screen ..


https://github.com/acterglobal/a3/assets/40496/57ef717b-ed31-4a18-a0bd-cdfc693546c1


I think this solution is sufficient for now.

gnunicorn:
I am not merging yet because I noticed that the "edit" doesn't put the text back into the input field now anymore. This is due to `FlutterMentions` being outdated and not allowing us to give in a proper controller for the input and I am seeing if I can hotfix this now or if there is some other way this needs to be addressed (like forking `FlutterMentions` or switching to AppFlowyEditor ...)

gnunicorn:
I've decided to merge as is and start the edit-problem as a new bug fix to be done. Because the only reason it seemed to have worked is because of the very special way that it was set up before hand in which it would actually create a complete rerender of the mentions field and unless I want to re-introduce this performance penalty again, there is no way to make that happen with the existing FlutterMentions. This needs a bit more work than just something I want to attach to this PR at this point.

PR #1594, opened 2024-04-08
Comment Module : UI-UX Changes

kumarpalsinh25:
Comment


app/lib/features/comments/widgets/comment.dart
```diff
@@ -1,28 +1,81 @@
...
+        final profileData = data.profile;
+        return Card(
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            children: [
+              ListTile(
+                leading: ActerAvatar(
+                  mode: DisplayMode.DM,
+                  avatarInfo: AvatarInfo(
+                    uniqueId: manager.roomIdStr(),
```
gnunicorn, line 44: Why `roomIdStr`? That sounds wrong. This should be `comment.sender().toString()` no?

app/lib/features/comments/widgets/comment.dart
```diff
@@ -1,28 +1,81 @@
...
+                    displayName: profileData.displayName ?? manager.roomIdStr(),
+                    avatar: profileData.getAvatarImage(),
+                  ),
+                  size: 18,
+                ),
+                title: Text(
+                  profileData.displayName ?? manager.roomIdStr(),
```
gnunicorn, line 51: here, too. Probably safe to to just grep that once as a variable and just assign it ...
```suggestion
                  profileData.displayName ?? userId,
```

app/lib/features/comments/widgets/comment.dart
```diff
@@ -1,28 +1,81 @@
...
+                  style: Theme.of(context).textTheme.titleSmall,
+                ),
+                subtitle: Text(comment.sender().toString()),
```
gnunicorn, line 54: This only makes sense if there is a display name or we just show this twice..
```suggestion
                subtitle: profileData.displayName == null ? null : Text(comment.sender().toString()) ,
```

gnunicorn (approved):
Looking good, just a minor fix for the UserID.

On the design itself, I think both the avatar + userInfo as well as the empty-state are too large at the moment, take up so much more space compared to the actual comments / actual point of information. But we can also address that later, if that isn't too much of a concern right now ...

PR #1605, opened 2024-04-10
Fix accept of invitation in integration test according to PR#1597

bitfriend:
In PR#1597, `get_spaces_and_chats()` limited rooms into `Joined` state, but `client` calls `invitation_controller.load_invitations()` on first sync callback.
I changed time point of accept of invitation.
And used retry-loop to see if the convo can be available already, according to @gnunicorn hint.


gnunicorn (approved):
Thanks heaps for fixing the tests, @bitfriend :clap: !

PR #1609, opened 2024-04-11
Fix accept of invitation in integration test according to PR#1597

bitfriend:
In PR#1597, `get_spaces_and_chats()` limited rooms into `Joined` state, but `client` calls `invitation_controller.load_invitations()` on first sync callback.
@gnunicorn recommended to use retry-loop to see if the convo can be available already.
I am getting another error `sisko should belong to convo` like [this PR](https://github.com/acterglobal/a3/actions/runs/8643766368/job/23699632406?pr=1580)
Will extend this loop to all points that room is created or invited.


bitfriend:
We are using retry loop, because it takes for timeline to reorder events.

gnunicorn (approved):
looks fine. I only skimmed most of the tests, but don't see why we should stop this from being merged.

PR #1617, opened 2024-04-11
Share event as ical

gnunicorn:
As it says so on the tin: This allows for a very simple ical (no RSVP status or anything) export of an Event for sharing it with external parties.

See it in action:

https://github.com/acterglobal/a3/assets/40496/8bafea0b-1b82-4b77-90f4-94604922ee33


kumarpalsinh25 (approved):
Flutter side code changes looks to good.

On UI-UX side, I would suggest to put share event button on top app bar action button area beside menu icon.

Share screenshot for reference.
Frame 6

gnunicorn:
I've taken your suggestion @kumarpalsinh25 and move the thing into the top bar. Though I think this is a bit too in-your-face, especially when you are scrolling and even the title disappears:

Screenshot_1712920473
Screenshot_1712920478



So I decided to replace it with the generic share icon that then opens a submenu also indicating what I tried to say with the option below: there is more to come. It's a bit sad & weird to have only one item, but I think we need to spell out the "ical"-part of it as it otherwise is confusing what you are sharing ...


https://github.com/acterglobal/a3/assets/40496/d0aba28f-4ccf-4839-93ac-205c38fb776d

kumarpalsinh25:
@gnunicorn 
Understood your thoughts on UX explanation. Looks fine on video screen recording. :)

PR #1618, opened 2024-04-12
Fix the white screen issue when entering a space

kumarpalsinh25:
Due to bad usage of `Slivers`, there were plenty of errors on the view part of the related-spaces classes. 

We really need to optimise code of related-spaces classes. But it was bit time taking so to quick fixed the white screen issue, I have removed `Slivers` from the affected class.

Reference Video:

https://github.com/acterglobal/a3/assets/51526480/b7d510f2-cb8b-4fe5-99c9-5becd732155e

fixes,
https://github.com/acterglobal/a3-meta/issues/152

Side notes:
- There were some `SliverGrid` need to remove for which will show single item on desktop view instead of grid.
- Need to optimise code of related-spaces to get out of this type of errors.

### DIFF
--- a/app/lib/features/space/pages/overview_page.dart
+++ b/app/lib/features/space/pages/overview_page.dart
@@ -30,7 +30,8 @@ class ActerSpaceChecker extends ConsumerWidget {
     final expCheck = expectation ?? (a) => a != null;
     return appSettings.when(
       data: (data) => expCheck(data) ? child : const SizedBox.shrink(),
-      error: (error, stackTrace) => Text(L10n.of(context).failedToLoadSpace(error)),
+      error: (error, stackTrace) =>
+          Text(L10n.of(context).failedToLoadSpace(error)),
       loading: () => const SizedBox.shrink(),
     );
   }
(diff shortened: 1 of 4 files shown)


gnunicorn (approved):
I am okay with this as we focus on mobile rather than desktop, but what does it look like on desktop now? Probably ugly ... should we at least center the items in a max-width-contrained-box in that case?

kumarpalsinh25:
@gnunicorn 

Yes it will look bit weird on desktop. 
Screenshot 2024-04-12 at 8 25 47 PM

But we really need to optimise the code of the event related-space classes. Events and members page don't have such bad cod so doesn't have this kind of issue.

Screenshot 2024-04-12 at 8 27 21 PM

After optimisation, we can bring back grid view in related spaces also. :)

kumarpalsinh25:
@gnunicorn 
Merging this PR to main as spaces are just not usable to users due to white screen.

Created new issue for https://github.com/acterglobal/a3/issues/1619 for code quality optimisation of the related space classes

PR #1623, opened 2024-04-14
Better public search

gnunicorn:
What started out as trying to get the app to not crash when searching the public directory ended up as a major refactor (from one file into its own feature-folder), several clean ups and a better integration in the quick jump. Each commit has a distinct atomic change set but overall, we have the following tasks finished:

- [x] Fix public search crash by cleaning up the search widget
- [x] Showing the Room Details for rooms we know of (because we are part of them)
- [x] Improved UI and design for the search-section of the public search
- [x] Using better type-ahead-widget for search selection
- [x] Adding more default servers for convenience
- [x] Make the Uri-String-parsing-widget reusable as `MaybeDirectRoomActionWidget`:
   - [x] Put that into the quick-jump section, too
   - [x] add support for matrix:/-uri-schemes
   - [x] show locally known rooms rather than attempting to join them again
   - [x] Integrate it into the quick-jump-search-bar
- [x] Improved searchbar for quick-jump
- [x] Integrate Join-Space into quick-jump: when no space is found a button is shown directly taking the input into the search form.

Watch it yourself:

https://github.com/acterglobal/a3/assets/40496/b42e2134-093a-409c-9ab5-7fffc9d04efe

Still to do:
 - [ ] the initial search from the quick-jump is used but not always properly shown in the join-space-searcher
 - [ ] the join-space-page needs a new name, labels and routing as it is generic over space & chat now
 - [ ] the UI of the found-item for matching IDs needs some context or as its render isn't meant for that area
 - [ ] Local room with an alias isn't yet found properly


gnunicorn:
And now with a prettier InlineTextActionButton for the search:
image

app/lib/features/spaces/widgets/public_spaces_selector.dart
```diff
@@ -219,28 +276,19 @@ class PublicSpaceItem extends ConsumerWidget {
...
                   error: (e, s) {
                     _log.severe('loading failed', e, s);
-                    return ActerAvatar(
-                      mode: DisplayMode.Space,
-                      avatarInfo: AvatarInfo(
-                        uniqueId: spaceId,
-                        displayName: spaceId,
-                      ),
-                    );
+                    return fallbackAvatar();
```
gtalha07: What's the reason of using custom widget method for fallback when ```ActerAvatar``` should/will manage its fallback ? If there's a bug, we should open this issue on plugin repo...

native/acter/api.rsh
```diff
@@ -2160,6 +2160,14 @@ object PublicSearchResultItem {
     fn join_rule_str() -> string;
     // fn room_type() -> Option<RoomType>;
     fn room_type_str() -> string;
+
+    /// whether to have avatar
+    fn has_avatar() -> bool;
+
+    /// get the binary data of avatar
+    /// if thumb size is given, avatar thumbnail is returned
```
gtalha07, line 2222: There's been on and off discussion around the avatar thumbnails about their source of generation and how matrix treats thumbnail objects. Does matrix generates media thumbnail on its own on the backend or we have to explicitly upload its version from client ??

app/lib/features/spaces/widgets/public_spaces_selector.dart
```diff
@@ -336,6 +336,86 @@ class PublicSpaceItem extends ConsumerWidget {
...
+    this.onSelectedMatch,
+    this.canMatchAlias = false,
+    this.canMatchId = false,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final searchVal = ref.watch(searchValueProvider);
+    if (onSelectedMatch != null && searchVal?.isNotEmpty == true) {
+      final aliased = RegExp(r'https://matrix.to/#/(?<alias>#.+):(?<server>.+)')
```
gtalha07: As I am reviewing it commit by commit for better analysis, these patterns should exist common-wide as utilities but if its been moved in later commits, that's good. 👍🏻

app/lib/features/spaces/widgets/public_spaces_selector.dart
```diff
@@ -336,6 +336,86 @@ class PublicSpaceItem extends ConsumerWidget {
...
+                icon: const Icon(Atlas.entrance_thin),
+                label: Text(L10n.of(context).tryToJoin),
+              ),
+            ),
+          ),
+        );
+      }
+
+      final id = RegExp(
+        r'https://matrix.to/#/(?<id>![^?]+)(\?via=(?<server_name>[^&]+))?(&via=(?<server_name2>[^&]+))?(&via=(?<server_name3>[^&]+))?',
```
gtalha07: Same as above :)

app/lib/features/public_room_search/providers/public_search_providers.dart
```diff
@@ -0,0 +1,38 @@
...
+  });
+  return controller;
+});
+
+final publicSearchProvider = StateNotifierProvider.autoDispose<
+    PublicSearchNotifier, PagedState<Next?, PublicSearchResultItem>>((ref) {
+  return PublicSearchNotifier(ref);
+});
+
+final searchValueProvider = StateProvider<String?>((ref) => null);
```
gtalha07: Shouldn't these value providers also have ```.autoDispose()``` modifier so they can disposed off automatically when not in listening state rather than manually setting their state above?

app/lib/features/public_room_search/widgets/public_room_item.dart
```diff
@@ -0,0 +1,163 @@
...
+                overflow: TextOverflow.ellipsis,
+              ),
+            ),
+          ),
+        ],
+      ),
+    );
+  }
+
+  ActerAvatar fallbackAvatar() {
```
gtalha07, line 149: Ignore my comment regarding avatar fallback method! ;)

app/lib/features/public_room_search/providers/public_search_providers.dart
```diff
@@ -1,23 +1,17 @@
...
-});
-
 final publicSearchProvider = StateNotifierProvider.autoDispose<
     PublicSearchNotifier, PagedState<Next?, PublicSearchResultItem>>((ref) {
   return PublicSearchNotifier(ref);
 });
 
-final searchValueProvider = StateProvider<String?>((ref) => null);
-final selectedServerProvider = StateProvider<String?>((ref) => null);
+final searchFilterProvider =
```
gtalha07, line 14: Nice! 👍🏻

app/lib/features/public_room_search/widgets/public_room_search.dart
```diff
@@ -90,6 +91,56 @@ class _MaybeQuickActionWidget extends ConsumerWidget {
...
+  const SearchField({super.key});
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() => _SearchFieldState();
+}
+
+class _SearchFieldState extends ConsumerState<SearchField> {
+  bool hasSearchTerm = false;
+
+  final TextEditingController searchTextController = TextEditingController();
```
gtalha07: Nice updated change! 👍🏻

app/lib/features/public_room_search/widgets/maybe_direct_room_action_widget.dart
```diff
@@ -0,0 +1,109 @@
+import 'package:acter/common/utils/rooms.dart';
+import 'package:acter/common/utils/routes.dart';
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+
+final aliasedHttpRegexp =
+    RegExp(r'https://matrix.to/#/(?<alias>#.+):(?<server>.+)');
```
gtalha07, line 17: Probably better off keeping these matchers over utilities ;)

app/lib/features/public_room_search/widgets/maybe_direct_room_action_widget.dart
```diff
@@ -0,0 +1,261 @@
...
+        id.namedGroup('server_name2') ?? '',
+        id.namedGroup('server_name3') ?? '',
+      ].where((e) => e.isNotEmpty).toList();
+      return Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 10),
+        child: renderForRoomId(context, ref, '!$roomId', servers),
+      );
+    }
+
+    return const SizedBox(height: 0);
```
gtalha07, line 239: Alternatively can use ```const SizedBox.shrink()```.

app/lib/features/public_room_search/widgets/public_room_search.dart
```diff
@@ -0,0 +1,332 @@
...
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              Flexible(
+                child: Padding(
+                  padding: const EdgeInsets.symmetric(vertical: 5),
+                  child: ListTile(
+                    leading: const Bone.square(size: 30),
+                    title: Skeletonizer(
+                      child: Text(
+                        'nothing',
```
gtalha07, line 245: Loading texts should also be formatted as l10n strings...

app/lib/features/search/widgets/spaces_builder.dart
```diff
@@ -13,14 +18,53 @@ class SpacesBuilder extends ConsumerWidget {
...
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: List.generate(
+          3,
+          (index) => Container(
+            padding: const EdgeInsets.all(10),
+            child: const Column(
+              children: [
+                Skeletonizer(child: Icon(Icons.abc)),
+                SizedBox(height: 3),
+                Skeletonizer(child: Text('space name')),
```
gtalha07, line 48: Should have l10n based string.

gtalha07 (approved):
Thanks! This is huge and impressive work regarding public search spaces which can be said as stood out its own module. Generally, the PR looks great but because of its nature and complexity, I have tried to examine everything , understanding bits and made some pointers regarding code style, missing l10n localised format for some strings etc. Generally this goes as my approval but it'd be great if some other reviewer can also look on it optionally and put feedback which I may have unintentionally missed. Overall, a solid work! 🎉

gtalha07:
Also, lastly the PR would need to have change logs for end-users ! ;)

app/lib/features/public_room_search/widgets/public_room_search.dart
gnunicorn: no... why? this is in a skeletonizer, it is never shown.... in the latest skeletonizer we finally have the `Bone` class that will allow us to not have these pointless filler anymore... \o/

app/lib/features/spaces/widgets/public_spaces_selector.dart
gnunicorn: just a separation of code.

app/lib/features/public_room_search/providers/public_search_providers.dart
gnunicorn: this is gone in the latest version and will be calculated from the search state object now.

app/lib/features/public_room_search/widgets/public_room_search.dart
gtalha07: sounds plausible if there's no way of it showing up. 👍🏻

Issue #1654, opened 2024-04-19
Unable to scroll sub-spaces list


kumarpalsinh25:
https://github.com/acterglobal/a3/assets/51526480/283d0878-e048-4c31-a9d8-944afe1db4d3

Issue #1603, opened 2024-04-10
Remove "parents" title, if there is no parents in space.


emilvincentz:
![image](https://github.com/acterglobal/a3/assets/40496/01b35288-edc6-49a2-84a3-a802c246ffc9)

emilvincentz:
[Skindergades.pdf](https://github.com/acterglobal/a3/files/14931907/Skindergades.pdf)

PR #1655, opened 2024-04-19
Fixes for space

kumarpalsinh25:
fixes,
 
 - #1603
 - #1654
 - #1644

### DIFF
--- a/app/lib/features/space/widgets/relatest_spaces.dart
+++ b/app/lib/features/space/widgets/relatest_spaces.dart
@@ -95,7 +95,7 @@ class RelatedSpaces extends StatelessWidget {
   }
 
   Widget? renderParentsHeader(BuildContext context) {
-    if (!showParents || (spaces.parents.isEmpty && spaces.mainParent != null)) {
+    if (!showParents || (spaces.parents.isEmpty && spaces.mainParent == null)) {
       return null;
     }
     return Row(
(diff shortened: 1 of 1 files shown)


gtalha07:
mind sharing screenshot of related spaces UI as of this PR version ?

gnunicorn (approved):
I'm always a little scared when I see the GridView and sliverdelegates now, since I find them hard to predict for all cases and they have shown several times to be causing runtime problems for specific circumstances (see white page) - which I keep wondering if that isn't something that a type checker should be able to identify ... 

anyways ... _from what I can see_ I do not see any problems here ...

kumarpalsinh25:
> mind sharing screenshot of related spaces UI of this PR version ?


Screenshot 2024-04-19 at 12 36 08 PM

kumarpalsinh25:
> I'm always a little scared when I see the GridView and sliverdelegates now, since I find them hard to predict for all cases and they have shown several times to be causing runtime problems for specific circumstances (see white page) - which I keep wondering if that isn't something that a type checker should be able to identify ...
> 
> anyways ... _from what I can see_ I do not see any problems here ...


Yes, agree with you. I am also afraid `Slivers` comes into the use-case but on the current PR instead of `SliverGrid`, I have used `GridView` and tested code with multiple use-cases and haven't seen any errors so far..

`Slivers` can be useful in certain use-cases but relevant code should be good enough with all the scrolling use-cases.

Hope in future when we refactor `Related-spaces` and `Sub-spaces` code then there might not be any chances of weird errors that we have seen in past.

PR #1656, opened 2024-04-19
chat settings typing notice and user configuration

### LINKED ISSUE #811: Reactivate typing feature.
Now we have somewhat migrated the chat infrastructure to riverpod state handlers, we should extend it to typing state and UI handlers.

gtalha07:
fixes #811.

### Resurrects typing notifications to chat but with user setting controlled.
Followed by the PR, this includes removing the unnecessary UI code for ```TypeIndicator``` and instead use ```flutter_chat_ui``` plugin ```TypeIndicatorOptions``` for typing notification UI. On the data layer, a ```StateNotifier``` is being used to directly configure and control typing notifications stream based on user preference. On new user accounts, this setting is automatically configured to be enabled. 

### Preview:

(Desktop) : 
https://github.com/acterglobal/a3/assets/68579938/d8832d05-3c42-4545-bcc9-ff635325f92f
(Mobile) :
https://github.com/acterglobal/a3/assets/68579938/a359caec-9075-4126-b8e8-e35622a87b59


native/core/src/events/settings/user.rs
```diff
@@ -16,6 +16,7 @@ pub enum AutoDownload {
 pub struct AppChatSettings {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub auto_download: Option<AutoDownload>,
+    pub typing_notice: Option<bool>,
```
gnunicorn, line 19:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub typing_notice: Option<bool>,
```

or old settings will not be deserialized properly

native/core/src/events/settings/user.rs
```diff
@@ -39,10 +40,34 @@ impl ActerUserAppSettingsContentBuilder {
         if let Some(chat) = &mut self.chat {
             chat.auto_download = Some(auto_d);
         } else {
+            let typing_notice = self
+                .chat
+                .clone()
+                .expect("No typing notification setting found")
+                .typing_notice;
```
gnunicorn, line 47: the else here means there is no `self.chat` ... this will never work.
```suggestion
```

native/core/src/events/settings/user.rs
```diff
@@ -39,10 +40,34 @@ impl ActerUserAppSettingsContentBuilder {
...
             self.chat = Some(AppChatSettings {
                 auto_download: Some(auto_d),
+                typing_notice,
```
gnunicorn, line 50: instead do:

```suggestion
                ..Default::default()
```

native/core/src/events/settings/user.rs
```diff
@@ -39,10 +40,34 @@ impl ActerUserAppSettingsContentBuilder {
...
+
+    pub fn typing_notice(&mut self, value: bool) -> Result<&mut Self, ParseError> {
+        if let Some(chat) = &mut self.chat {
+            chat.typing_notice = Some(value);
+        } else {
+            let auto_d = self
+                .chat
+                .clone()
+                .expect("No auto chat download setting found")
+                .auto_download;
```
gnunicorn, line 64: same as above

native/core/src/events/settings/user.rs
gtalha07: Ohh righttt...

native/core/src/events/settings/user.rs
gtalha07: Thanks for pointing it out sooner than later!

.changes/811-chat-typing-feature.md
```diff
@@ -0,0 +1 @@
+- [feat.] Chat now supports typing notifications and inform you who's been typing. This setting is enabled by default but can be turned off in App Settings -> Chat Settings.
```
gnunicorn, line 1:
```suggestion
- [feat.] Chat now shows typing notifications to inform you who's typing. This setting is enabled by default but can be turned off in App Settings -> Chat Settings.
```

app/lib/common/widgets/chat/convo_card.dart
```diff
@@ -443,15 +460,37 @@ class _SubtitleWidget extends ConsumerWidget {
...
+    BuildContext context,
+    List<UserId> userIds,
+    WidgetRef ref,
+  ) {
+    final textStyle = Theme.of(context).textTheme.bodySmall!;
+    if (userIds.length == 1) {
+      final userName = simplifyUserId(userIds[0].toString());
+      return Padding(
+        padding: const EdgeInsets.symmetric(vertical: 10),
+        child: Text('$userName is typing...', style: textStyle),
```
gnunicorn, line 473: we probably want these strings all translatable.

app/lib/features/chat/providers/notifiers/chat_typing_notifier.dart
```diff
@@ -0,0 +1,44 @@
...
+    });
+    if (enabled == null) {
+      // default settings not available, enable it automatically.
+      final updater = userAppSetttings.updateBuilder();
+      updater.typingNotice(true);
+      await updater.send();
+      configure(true);
+    } else {
+      configure(enabled);
+    }
```
gnunicorn, line 31: we should _not_ overwrite the settings, we should just consider 'null' to be doing what we do by default. In particular these type of side-effects should not be happening when you initialize a notifier!
```suggestion
```

app/lib/features/chat/providers/notifiers/chat_typing_notifier.dart
```diff
@@ -0,0 +1,44 @@
...
+
+class ChatTypingEventStateNotifier extends StateNotifier<TypingEvent?> {
+  final Ref ref;
+  ChatTypingEventStateNotifier(this.ref) : super(null) {
+    _init();
+  }
+  StreamSubscription<TypingEvent?>? _poller;
+  void _init() async {
+    final userAppSetttings = await ref.read(userAppSettingsProvider.future);
+    final enabled = userAppSetttings.typingNotice();
```
gnunicorn, line 18:
```suggestion
```

we should do this with a `ref.watch` from the provider, not here.

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -106,3 +107,9 @@ final chatMentionsProvider =
   }
   return mentionRecords;
 });
+
+// state of typing notice setting
+final chatTypingEventStateProvider = StateNotifierProvider.autoDispose<
+    ChatTypingEventStateNotifier, TypingEvent?>(
+  (ref) => ChatTypingEventStateNotifier(ref),
```
gnunicorn, line 114: we should be checking here whether it is enabled via a `ref.watch` to make sure we update it properly

app/lib/features/settings/pages/chat_settings_page.dart
```diff
@@ -65,6 +66,61 @@ class ChatSettingsPage extends ConsumerWidget {
...
+                  // ignore: use_build_context_synchronously
+                  L10n.of(context).settingsSubmittingSuccess,
+                  toastPosition: EasyLoadingToastPosition.bottom,
+                );
+              } catch (error, stackTrace) {
+                _log.severe('Failure submitting settings', error, stackTrace);
+                EasyLoading.showError(
+                  // ignore: use_build_context_synchronously
+                  L10n.of(context).settingsSubmittingFailed(error),
+                );
```
gnunicorn, line 101: :+1:

app/lib/features/settings/pages/chat_settings_page.dart
```diff
@@ -65,6 +66,61 @@ class ChatSettingsPage extends ConsumerWidget {
...
+            onToggle: (newVal) async {
+              EasyLoading.show(status: L10n.of(context).settingsSubmitting);
+              try {
+                final notifier =
+                    ref.read(chatTypingEventStateProvider.notifier);
+                final updater = settings.updateBuilder();
+                updater.typingNotice(newVal);
+                await updater.send();
+                // refresh state
+                notifier.configure(newVal);
```
gnunicorn, line 90: this should not be here but happen reactively through an `ref.watch`!
```suggestion
```

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -106,3 +107,9 @@ final chatMentionsProvider =
...
+);
```
gnunicorn, line 115: Looking at it after some understanding, can't the entire notifier be slimmed down to a stream provider?

```suggestion
// state of typing notice setting
final chatTypingEventStateProvider = StreamProvider<TypingEvent?>((ref) async* {
  // Connect to an API using sockets, and decode the output
      final userAppSetttings = await ref.watch(userAppSettingsProvider.future);
    if (!userAppSetttings.typingNotice()) {
       return null;
    }
  await for (final event in client.typingEventRx()!) {
    yield event;
  }
});
```

native/core/src/events/settings/user.rs
```diff
@@ -41,8 +43,22 @@ impl ActerUserAppSettingsContentBuilder {
...
+        Ok(self)
+    }
+
+    pub fn typing_notice(&mut self, value: bool) -> Result<&mut Self, ParseError> {
+        if let Some(chat) = &mut self.chat {
+            chat.typing_notice = Some(value);
+        } else {
+            self.chat = Some(AppChatSettings {
+                auto_download: Default::default(),
+                typing_notice: Some(value),
```
gnunicorn, line 58: :+1:

gnunicorn (approved):
Just a request to use l10n, a potential improvement on slimming down the riverpod provider and the language of the changeset message.

Great job, mate!

app/lib/features/chat/providers/notifiers/chat_typing_notifier.dart
gtalha07: So that means, initially these settings should not have any set value OR you're proposing this isn't a great time/area to have it set up?

app/lib/features/chat/providers/chat_providers.dart
gtalha07: I'll try it this way. The initial non-included similar version has another provider aka ```StreamProvider``` which was observable for ```StateNotifier``` but has a problem of still being subscribed to receive events...

app/lib/features/chat/providers/notifiers/chat_typing_notifier.dart
gnunicorn: we shouldn't set this _unless_ the user interacts with it. and unitl then interpret the `null` as `true`.

app/lib/features/chat/providers/chat_providers.dart
gnunicorn: @gtalha07 minor update to deal with `null` correctly.

kumarpalsinh25 (approved):
Nice work! 👏

Issue #1540, opened 2024-03-23
space-internal navigation not obvious


JMEFKT:
depending on the mobile screen size, one can't see that there is more in the space overview next to pins & events & scrolling sideways isn't intuitive = feedback from users

![share_8220122628607237845](https://github.com/acterglobal/a3/assets/116185441/21b20f27-8f15-4872-9c5b-25bcf1992662)

PR #1659, opened 2024-04-19
Reworking the Space and Chat routing, including Spacebar navigation, for resilience and better UX

### LINKED ISSUE #1470: Updates -> Space -> Chat -> back navigation closes app
weird stuff with the navigation, reported by an external user. confirmed. Instead of closing the app, this should clearly navigate back to the spaces->chat-page.

gnunicorn:
1. Hardening of leaving space, externalized leave-space-dialog, fixes https://github.com/acterglobal/a3-meta/issues/195:

https://github.com/acterglobal/a3/assets/40496/4ce23427-32a0-4af2-9d2f-f2b3eb3b009c

and a bit an improved designed and UX followed:

https://github.com/acterglobal/a3/assets/40496/090255e8-2644-4f9a-ba85-98ccf04ee49b

2. Make the space top a proper AppBar to remove the issue that the icons are sometimes unavailable in Android.

3. Improved upon the `goToChat`-util to make sure we are also switching to the proper branch. Fixes #1470 (see below)

4. Expand upon the `goTo`-util to make the same work for the spaces. Allow for the same behavior of first switching the branch and thus keeping space-related stuff always within the home-branch.

https://github.com/acterglobal/a3/assets/40496/8b5778b1-fa87-4892-8c9f-648013d945ac

6. Expanding the goto-principle to QuickJump, removing the outdated navigateTo-calblack action, simplifying the code.

7. Refactoring of the Space navigation, moving away from the complex tab-bar & controller to s a simpler `Row`, which we can manage ourselves nicely, so we can compact the hidden items behind a more obvious popup-menu. It also switches out the last item in the menu for the active one if it would otherwise be hidden in the menu. Addresses #1540, fixes https://github.com/acterglobal/a3-meta/issues/192

See:

https://github.com/acterglobal/a3/assets/40496/06a92bb8-7dec-41e0-b589-6f9eff657553

Also for resizing:

https://github.com/acterglobal/a3/assets/40496/6cdd8879-5872-47b7-951a-3cd280c07dc5


app/lib/features/space/widgets/top_nav.dart
```diff
@@ -26,110 +23,192 @@ class TopNavBar extends ConsumerStatefulWidget {
...
-                labelStyle: useCols
-                    ? Theme.of(context).textTheme.labelSmall
-                    : Theme.of(context).textTheme.bodySmall,
-                labelColor: Colors.white,
-                onTap: (index) {
-                  final target = tabs[index].target;
-                  context.goNamed(
-                    target,
-                    pathParameters: {'spaceId': widget.spaceId},
+  Widget buildContainer(BuildContext context, Widget child) {
```
gtalha07, line 46: I understand its usage but this seems like generic component which can be extended across other UI loading/error states, can we put this under ```common``` section if there's a possibility of its reusability?

gtalha07 (approved):
nice QoL changes and code structuring! Approved 🎉

app/lib/features/space/widgets/top_nav.dart
gnunicorn: I did that in the past and now we have many of those that are effectively used once ... I decided to change strategy and only generalize within a component for now and only later look up if they is common grounds to generalize further. for now, I'd like to keep it this way.

app/lib/router/router.dart
```diff
@@ -170,27 +168,6 @@ List<StatefulShellBranch> mobileShellBranches(Ref ref) {
   ];
 }
 
-List<StatefulShellBranch> desktopShellBranches(Ref ref) {
```
kumarpalsinh25, line 173: I doubt on this changes to make common branches structure as we have different indexing for tabs in mobile and desktop.

app/lib/router/utils.dart
```diff
@@ -0,0 +1,90 @@
...
+        ShellBranch.chatsShell => chatTabNavKey,
+        ShellBranch.activitiesShell => activitiesTabNavKey,
+        ShellBranch.quickJumpShell => searchTabNavKey,
+      };
+}
+
+/// Make sure we are on the right navigation branch
+/// this needs to be run _before_ you `go` or `push` the route
+/// or we only switch in the UI and no really in the active shell
+bool navigateOnRightBranch(
```
kumarpalsinh25, line 28: Not really understanding changes of this code.

app/lib/router/router.dart
gnunicorn: the tabs shouldn't not be relying the internal branch-index order to just be matching up, it doesn't make any sense. there is an enum now specifying each index in a typed fashion.

PR #1661, opened 2024-04-19
A first stab at a unified ToolKit: Colors & Buttons

gnunicorn:
This is a first attempt in streamlining our Acter Toolkit with a focus on Buttons and cleaning up the Colors.

This defines:
 - the default action button is a simple `OutlinedButton`, use that everywhere for cancel and regular stuff
 - if you have a primary action use the newly added `ActerPrimaryActionButton`, which already applies the proper theme using the primary-color
 - similarly if this is a dangerous action use the `ActerDangerActionButton` to clearify that
 - if you have a simple clickable text, use the default `TextButton` - it is not elevated, not outlined and otherwise clean.
 - if you have an inlined action, so e.g. in a Text flow, use the new `ActerInlineTextButton`, which underlines the text nicely so the user understand they can click something.

Furthermore:
 - This streamlines the usage of colors by going through the Theme primarily. Further colors are specified as an extension to the ColorScheme
 - this also applies to further special themedata (e.g. for the buttons mentioned above): they all come directly from the Theme, no more hardcoding of colors or themedata via direct imports
 - this also applies for the chatTheme used by the Flyer Chat UI -> comes from the Theme

In implementation:
 - Buttons with local colors are now gone, they are all replaced with the semantic buttons from above
 - Inlined Colors have been replaced with theme-based lookups (as much as feasible)
 - The inlined text button is used in an example to unify and clean up the Comments Section (align it with the attachments section and clean both up a little, too).

This is what this looks like now:

https://github.com/acterglobal/a3/assets/40496/50e52cb4-7238-4dc3-8564-f029dd658178

This might be easier to review commit-by-commit, at least each one contains more or less exactly one specific changeset.


app/lib/common/themes/app_theme.dart
```diff
@@ -39,257 +40,32 @@ class EmojiConfig {
...
-  Color get tertiary2 => const Color(0xFFFFC333);
+  Color get primary => const Color(0xFF9CCAFF);
+  Color get tertiary => const Color(0xFFFFC333);
   Color get tertiary3 => const Color(0xFF3AE3E0);
   Color get neutral => const Color(0xFF121212);
   Color get neutral2 => const Color(0xFF2F2F2F);
   Color get neutral3 => const Color(0xFF5D5D5D);
   Color get neutral4 => const Color(0xFF898989);
   Color get neutral5 => const Color(0xFFB7B7B7);
   Color get neutral6 => const Color(0xFFE5E5E5);
```
kumarpalsinh25, line 52: We can also get rid of unwanted `neutral` color sets and use only default `neutral` color based on the requirement.

app/lib/common/themes/acter_theme.dart
```diff
@@ -40,9 +40,9 @@ class ActerTheme {
     dividerTheme: dividerTheme,
     dialogTheme: dialogTheme,
     bottomSheetTheme: bottomSheetTheme,
-    elevatedButtonTheme: elevatedButtonTheme,
+    elevatedButtonTheme: elevatedButtonTheme(colorScheme),
```
kumarpalsinh25, line 54: Very nice!

app/lib/common/themes/app_theme.dart
```diff
@@ -39,257 +40,32 @@ class EmojiConfig {
...
+  Color get textHighlight => secondary;
+  Color get textButtonColor => yellowColor;
+
+  // states
+  Color get success => secondary;
 
-  Color get badgeUnread => const Color(0xFF67A24A);
-  Color get badgeImportant => const Color(0xFFFFC333);
+  // specific widgets
+  Color get badgeUnread => secondary;
```
kumarpalsinh25, line 63: Where is `secondary` color declaration? Are we considering default `ThemeData` secondary color?

app/lib/common/themes/app_theme.dart
```diff
@@ -39,257 +40,32 @@ class EmojiConfig {
...
-        selectedLabelTextStyle: const TextStyle(
-          color: Colors.white,
-          fontSize: 12,
-          fontWeight: FontWeight.w400,
-        ),
-        selectedIconTheme: const IconThemeData(color: Colors.white, size: 18),
-        unselectedIconTheme: const IconThemeData(color: Colors.white, size: 18),
-      ),
-    );
-  }
```
kumarpalsinh25, line 294: Biggg move towards unified theme !! 👏

app/lib/common/themes/app_theme.dart
```diff
@@ -39,257 +40,32 @@ class EmojiConfig {
...
+  Color get tasksBG => primary;
+  Color get tasksFG => primaryContainer;
+  Color get taskOverdueBG => errorContainer;
```
kumarpalsinh25, line 70: Where is `errorContainer` color declaration? Are we considering default `ThemeData errorContainer` color?

app/lib/common/themes/acter_theme.dart
```diff
@@ -11,6 +12,10 @@ import 'package:acter/common/themes/components/progress_indicator_theme.dart';
 import 'package:acter/common/themes/components/text_theme.dart';
 import 'package:flutter/material.dart';
 
+extension ActerChatThemeExtension on ThemeData {
```
kumarpalsinh25, line 15: 👏

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -66,11 +66,13 @@ class _TextMessageBuilderConsumerState
     if (matches.isEmpty) {
       return LinkPreview(
         metadataTitleStyle: userId == authorId
-            ? const ActerChatTheme().sentMessageLinkTitleTextStyle
-            : const ActerChatTheme().receivedMessageLinkTitleTextStyle,
+            ? Theme.of(context).chatTheme.sentMessageLinkTitleTextStyle
+            : Theme.of(context).chatTheme.receivedMessageLinkTitleTextStyle,
```
kumarpalsinh25, line 70: 👏

app/lib/common/themes/acter_theme.dart
```diff
@@ -14,6 +14,9 @@ import 'package:flutter/material.dart';
 
 extension ActerChatThemeExtension on ThemeData {
   ActerChatTheme get chatTheme => const ActerChatTheme();
+
+  ElevatedButtonThemeData get dangerButtonTheme =>
```
kumarpalsinh25, line 18: 👏

app/lib/common/toolkit/buttons/danger_action_button.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:acter/common/themes/acter_theme.dart';
+import 'package:flutter/material.dart';
+import 'dart:math' as math;
+import 'dart:ui' show lerpDouble;
+
+class ActerDangerActionButton extends ElevatedButton {
```
kumarpalsinh25, line 11: Do we require separate component as `ActerDangerActionButton` instead of using it same as `ActerPrimaryActionButton` with different button theme?

app/lib/common/toolkit/buttons/inline_text_button.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:acter/common/themes/acter_theme.dart';
+import 'package:flutter/material.dart';
+import 'dart:math' as math;
+import 'dart:ui' show lerpDouble;
+
+class ActerInlineTextButton extends TextButton {
```
kumarpalsinh25, line 12: Do we require separate class for `ActerInlineTextButton`? How about using default `TextButton` with our required theme set?

app/lib/features/comments/widgets/comments_list.dart
```diff
@@ -30,6 +37,16 @@ class CommentsList extends ConsumerWidget {
         );
   }
 
+  Widget createComment() {
+    return Padding(
+      padding: const EdgeInsets.symmetric(horizontal: 16),
+      child: CreateCommentWidget(
+        manager: widget.manager,
+        onClose: () => setState(() => editorOpened = false),
```
kumarpalsinh25, line 45: 👏

kumarpalsinh25 (approved):
Well done!
Nicely handle different aspects of theme class.

app/lib/common/themes/app_theme.dart
gnunicorn: generally yes, but many of them seem to be used in the ChatTheme and I wasn't really clear what I was touching or how to better infer the data, so I left this untouched for now. Ideally I want the entire chatTheme also to become a `chatThemeMaker` that given the `colorScheme` is just generated on the fly without hardcoded colors.

app/lib/common/themes/app_theme.dart
gnunicorn: yes, this directly comes from the regular ThemeData. That's effectivly what the get does, just creates an alias for the more semantic `badgeUnread` color...

app/lib/common/themes/app_theme.dart
gnunicorn: yes, these all directly use the existing colorschemes now. See above for a more in-depth explanation.

app/lib/common/toolkit/buttons/danger_action_button.dart
gnunicorn: IMHO, I think we do want this. First, because of semantic reasons: I don't understand why so many toolkits moved away from semantic naming things (like the famous original "Bootstrap" did) and instead went for purely function naming again, but I think it is way better to think of our toolkit in terms of higher level abstractions. So we want a special ActerDangerActionButton to be used whenever what is happening is potentially dangerous and the user should be warned about it. 

And that might move into more than just a difference in color: for e.g. for the nuke-button we have that really nice behavior that you have to _long hold_ the button to activate it and I am considering making that the default way to activate this button. To give the severity of the action the attention it deserves. And that would make this button not only look different but also behave different.

Last, but not least, we might decide to not use the ElevatedButton as the baseline but one more astray from the PrimaryButton. All of these things considered, I think making this an explicit type is better in terms of creating a proper toolkit.

app/lib/common/toolkit/buttons/inline_text_button.dart
gnunicorn: Because there are _two_ ways in which the TextButton is used and they differ in design. Take the first-run-feature-introduction-wizzard and its "Previous", "Next" and "Skip" buttons and then the underlined button as used here in this comments creation above. We can't have a unified design for both of them because we want them to look different.

I opted for keeping the current TextButton as is and instead create a custom version with underlining for this more specific case (though I intend to spread that one a bit more widely in the future). That's why this is also opting for different theme internally.

PR #1671, opened 2024-04-23
Refactor permission checking

bitfriend:
Will replace the existing fns with following fns.
- `room.can_user_ban()`
- `room.can_user_invite()`
- `room.can_user_kick()`
- `room.can_user_redact_other()`
- `room.can_user_redact_own()`
- `room.can_user_send_message()`
- `room.can_user_send_state()`
- `room.can_user_trigger_room_notification()`


native/acter/src/api/attachments.rs
```diff
@@ -476,9 +480,11 @@ impl AttachmentsManager {
 
         RUNTIME
             .spawn(async move {
-                trace!("before redacting attachment");
+                let permitted = room.can_user_redact_own(&my_id).await?;
+                if !permitted {
+                    bail!("No permissions to redact your message in this room");
+                }
```
gnunicorn, line 486: if we are already adding this here, we need to make the proper check:
if the event is by _this_ user we need to check for `can_user_redact_own` and if it is not we need to check for `can_user_redact_other`.

This basically applies to all _redacts_ ...

native/acter/src/api/news.rs
```diff
@@ -634,8 +637,14 @@ impl NewsEntryDraft {
         trace!("starting send");
         let client = self.client.clone();
         let room = self.room.clone();
+        let my_id = room
+            .client()
```
gnunicorn, line 641: we already have client
```suggestion
        let my_id = client
```

native/acter/src/api/pins.rs
```diff
@@ -370,11 +370,23 @@ impl PinDraft {
...
+            .to_owned();
         let content = self.content.build()?;
+
         RUNTIME
             .spawn(async move {
-                let resp = room.send(content).await?;
-                Ok(resp.event_id)
+                let permitted = room
+                    .can_user_send_message(&my_id, MessageLikeEventType::RoomMessage)
+                    .await?;
```
gnunicorn, line 384: hmm.... not super sure about this one ... the goal is to have type-specific power-levels for our own objects, too ... this implies that anyone can send them ... while this is fine for now, it isn't actually accurate...

native/acter/src/api/room.rs
```diff
@@ -591,8 +619,13 @@ impl Room {
         {
             bail!("No permissions to add parent to room");
         }
-        let room = self.room.clone();
         let client = self.core.client().clone();
+        let room = self.room.clone();
+        let my_id = room
+            .client()
```
gnunicorn, line 625: already have client
```suggestion
        let my_id = client
```

native/acter/src/api/room.rs
```diff
@@ -644,6 +688,10 @@ impl Room {
                     }
                     SyncOrStrippedState::Sync(ev) => ev.event_id().to_owned(),
                 };
+                let permitted = room.can_user_redact_own(&my_id).await?;
```
gnunicorn, line 691: needs the same redact check as above.

native/acter/src/api/settings/space.rs
```diff
@@ -182,9 +182,20 @@ impl Room {
 
         let client = self.room.client().clone();
         let room = self.room.clone();
+        let my_id = room
+            .client()
+            .user_id()
```
gnunicorn, line 187:
```suggestion
        let my_id = client
            .user_id()
```

using it that often makes you wonder if we shouldn't just add that to `self`, no?

gnunicorn (approved):
but a few minor code improvements.

native/acter/src/api/attachments.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/3ec8b833f13bc2974c6a896e7223b64068b081b3

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/3ec8b833f13bc2974c6a896e7223b64068b081b3

native/acter/src/api/news.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/1a56d9a426d303deef8dfdcdb08c5c870f3a6716

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/1a56d9a426d303deef8dfdcdb08c5c870f3a6716

native/acter/src/api/settings/space.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/1a56d9a426d303deef8dfdcdb08c5c870f3a6716

native/acter/src/api/pins.rs
bitfriend: For now, we have no ways to support power levels separately for pin, news or attachment.
This issue will be fixed in another PR.

native/acter/src/api/attachments.rs
```diff
@@ -476,9 +472,17 @@ impl AttachmentsManager {
...
         RUNTIME
             .spawn(async move {
-                trace!("before redacting attachment");
+                let evt = room.event(&event_id).await?;
+                let event_content = evt.event.deserialize_as::<RoomMessageEvent>()?;
+                let permitted = if event_content.sender() == my_id {
+                    room.can_user_redact_own(&my_id).await?
+                } else {
+                    room.can_user_redact_other(&my_id).await?
+                };
```
gnunicorn, line 481: nice!

native/acter/src/api/client.rs
```diff
@@ -333,29 +334,37 @@ impl Client {
...
     }
 
-    pub async fn room_by_alias_typed(&self, room_alias: &OwnedRoomAliasId) -> Result<Room> {
-        for r in self.core.client().rooms() {
+    pub fn room_by_id(&self, room_id: &RoomId) -> Result<SdkRoom> {
+        self.core
+            .client()
+            .get_room(room_id)
+            .context("Room not found")
+    }
```
gnunicorn, line 349: okay this becomes very confusing in naming now ... we should at least put some doc comment on each of them then, no?

also the idea of `_typed` here was to separate `string`-based ffi-API from internally rust-typed API ... 

but possibly something for a future PR then... also with uniffi we might get rid of all of that finally...!

native/acter/src/api/room.rs
```diff
@@ -996,10 +1018,17 @@ impl Room {
...
+        let client = self.core.client().clone();
         let mode = new_mode.and_then(|s| notification_mode_from_input(&s));
+
         RUNTIME
             .spawn(async move {
-                let notification_settings = room.client().notification_settings().await;
+                let permitted = room.can_user_trigger_room_notification(&my_id).await?;
+                if !permitted {
+                    bail!("No permissions to change notification mode in this room");
+                }
```
gnunicorn, line 1030: this is not the right check. every user can always change its notifications mode, as that is only for them.
```suggestion
```

gnunicorn (approved):
Just one faulty permission check that needs to be removed before merging.

Side note: there are many naming- and formatting changes in here that are of no consequence. For future PRs it'd be nice to have them separated out either as an external PR or into a separate commit and be told _which things really need a review_ and which can be ignored. That makes reviewing a lot easier and less straining ...

native/acter/src/api/client.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/7f4d5a8d4b9c3d0a33c890ac86f912adec53e44b

bitfriend: Appended description to each `***_typed()` fn in https://github.com/acterglobal/a3/pull/1671/commits/291683c29afa91edda98414778cc4a6727d15c92

native/acter/src/api/room.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/1671/commits/df6c69236f58d8d1058a76c8d7a814351729b036

Issue #1416, opened 2024-02-19
Expand Video & Image preview size, allow to pinch-to-zoom


emilvincentz:
On IOS 

https://github.com/acterglobal/a3/assets/77483554/20613aac-3a3d-4c2d-bfb2-6909e52fd124

Also images shows long title.

https://github.com/acterglobal/a3/assets/77483554/29f5665f-dc9c-418b-adee-fd9deddb0baa

gnunicorn:
would be nice to be able to zoom in as well.

Issue #1586, opened 2024-04-04
Handling of downloaded media is unsafe


gnunicorn:
Currently when a user wants to access some data, e.g. an image, in the [chat we ask the SDK to give us `mediaPath`](https://github.com/acterglobal/a3/blob/ddef5c2e2f861e1652f82ffd9556609ff5e770cd/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart#L30). If that fails, we can download it by doing:

https://github.com/acterglobal/a3/blob/ddef5c2e2f861e1652f82ffd9556609ff5e770cd/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart#L62-L74

Now, the attentive reader probably asks itself: if we have to provide a specific path for the download and that is a _temporary_ path ... how does `mediaPath` not require that parameter? Or even more specifically, what is the path we get out of that `mediaPath`? And the reader might become suspicious that we'd be storing and returning said _temporary_ path in `mediaPath`.

And the reader would be correct:

https://github.com/acterglobal/a3/blob/ddef5c2e2f861e1652f82ffd9556609ff5e770cd/native/acter/src/api/room.rs#L1514-L1518

This is unsafe for several reasons:
1. the data contained in download could be encrypted data. As we clearly are not clearing the file after reading it (or the entire `mediaPath`-thing wouldn't work in the first place), that means we are keeping data in plain at rest. That is not okay.
2. As this is a temporary path, it can be cleared at any point. Storing this path is unsafe as it could have been cleared at any time. Such a path should never be stored on the client side.

Issue #1250, opened 2023-12-27
Implement image thumbnail in chat media messages


kumarpalsinh25:
We want on the SDK:
- ability to add path of a thumbnail to a video/image we upload
- blurhash support ?
- ability to use the thumbnails for video & image

We want this at:
- chat images & videos
- any attachments
- image &video msg type, e.g. in News

gnunicorn:
with #1587 this is not that urgent anymore.

PR #1673, opened 2024-04-24
Chat multi-media improvements

kumarpalsinh25:
This PR contains serval improvements in chat multi-media.

1) Video chat message:
- Now chat message will display video thumb image instead of entire video player in chat message itself which was leading to performance issue sometimes.
- Now we will have video message view with relevant aspect ratio.

https://github.com/acterglobal/a3/assets/51526480/01d650a1-16d5-4c9e-8606-a0ec20e9f354

2) Full screen video message:
- Full screen video message will utilise entire screen based on the aspect ratio.
- Fixes minor title text overlapping issue.

https://github.com/acterglobal/a3/assets/51526480/b6ea62c1-9683-4b1a-9dd3-c662846d2c68

3) Support of zoom image:
- Now on the full screen image view, user can zoom in and out the image.

https://github.com/acterglobal/a3/assets/51526480/c68a5504-989c-4809-90ea-67142dcd74ca

fixes, #1416

### DIFF
--- a/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
+++ b/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
@@ -100,6 +100,9 @@ class MediaChatNotifier extends StateNotifier<MediaChatState> {
     }
   }
 
+  //FIXME : This is temporarily solution for media thumb management which lead to security issue.
+  // Reference https://github.com/acterglobal/a3/issues/1586
+  // Reference https://github.com/acterglobal/a3/issues/1250
   static Future<File?> getThumbnailData(String mediaPath) async {
     try {
       final tempDir = await getTemporaryDirectory();
--- a/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
+++ b/app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
@@ -5,9 +5,11 @@ import 'package:acter/common/providers/chat_providers.dart';
 import 'package:acter/features/chat/models/media_chat_state/media_chat_state.dart';
 import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
+import 'package:fc_native_video_thumbnail/fc_native_video_thumbnail.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:riverpod/riverpod.dart';
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
 
 final _log = Logger('a3::chat::media_chat_notifier');
 
@@ -33,8 +35,10 @@ class MediaChatNotifier extends StateNotifier<MediaChatState> {
         //Get media path if already downloaded
         final mediaPath = await _convo!.mediaPath(messageInfo.messageId, false);
         if (mediaPath.text() != null) {
+          final videoThumbnailFile = await getThumbnailData(mediaPath.text()!);
           state = state.copyWith(
             mediaFile: File(mediaPath.text()!),
+            videoThumbnailFile: videoThumbnailFile,
             mediaChatLoadingState: const MediaChatLoadingState.loaded(),
           );
         } else {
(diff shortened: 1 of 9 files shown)


app/lib/features/chat/widgets/image_message_builder.dart
```diff
@@ -131,9 +130,9 @@ class ImageMessageBuilder extends ConsumerWidget {
             ? BorderRadius.circular(6)
             : BorderRadius.circular(15),
         child: ConstrainedBox(
-          constraints: BoxConstraints(
-            maxHeight: isReplyContent ? size.height * 0.2 : size.height * 0.3,
```
gtalha07, line 135: This would probably create some confusion for end-user as image UI should appear shrinked as a original message of reply to make replied content more prominent as we see in common chat messengers...

gtalha07 (approved):
Great improvements overall! But I would like to bring back sizing condition for replied original message content as they wouldn't be much distinctive from normal message and shadows the original intent!

gnunicorn:
what does this look like on mobile / portrait mode? is that pinch to zoom? How is sizing of videos working for landscape videos? Can we zoom videos as well?

app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
```diff
@@ -93,4 +99,36 @@ class MediaChatNotifier extends StateNotifier<MediaChatState> {
...
+        width: 300,
+        height: 300,
+        keepAspectRatio: true,
+        format: 'jpeg',
+        quality: 90,
+      );
+
+      if (thumbnailGenerated) {
+        return destFile;
+      }
```
gnunicorn, line 127: ehm ... while I appreciate the attempt to just do things on the flutter side, taking any data, transforming it and then putting it on the filesystem is not an acceptable solution for us as it means we are potentially storing end-to-end-encrypted information in plain on the device (known as unencrypted at rest). This is highly insecure and puts our users at risk. As layed out in https://github.com/acterglobal/a3/issues/1586 as well.

I am _okay_ with temporarily accepting this PR if a FIXME linking to the above issue is added and a comment is left on that issue pointing to this additional code base to be fixed as well.

app/lib/features/chat/widgets/video_message_builder.dart
```diff
@@ -116,24 +116,72 @@ class VideoMessageBuilder extends ConsumerWidget {
...
+          context: context,
+          barrierDismissible: false,
+          useRootNavigator: false,
+          builder: (ctx) => VideoDialog(
+            title: message.name,
+            videoFile: mediaState.mediaFile!,
+          ),
+        );
+      },
+      child: ClipRRect(
```
gnunicorn, line 131: I feel like this is nesting deep enough now to be moved into its own private Widget in this file for lower indentation levels and better legibility.

gnunicorn (approved):
I have a security/privacy related concern with this, but once addressed this can be merged. Well done!

app/lib/features/chat/widgets/image_message_builder.dart
kumarpalsinh25: Good catch. Let me do the needful here.

kumarpalsinh25:
> what does this look like on mobile / portrait mode? is that pinch to zoom? How is sizing of videos working for landscape videos? Can we zoom videos as well?



1) Mobile/portrait mode view:
- Attaching screen recording for the reference.

2) Sizing of videos working for landscape videos:
- Video screen recording showing reference for both portrait and landscape mode.

3) Can we zoom videos as well?
- No, we can't zoom video at the moment. Zooming feature is for image only. Video screen recording showing the same.

Attached Screen Recoding:
https://github.com/acterglobal/a3/assets/51526480/2f56d88c-91c8-463f-97b8-82fe2fa67ace

app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
kumarpalsinh25: Yes that's true and agree with your thoughts.

Once we implement thumb management from rust sdk (https://github.com/acterglobal/a3/issues/1250), this code will change this is temporarily solution to avoid unwanted performance and sizing issue on chat messages. I will required FIXME not here so that we can take care of it in future.

Probably `getTemporaryDirectory` will not be directly accessible to end users from file explore so want be that much security issue but ya surely in future we need to avoid this.

PR #1533, opened 2024-03-20
Implement Kick & ban of users from any room

gnunicorn:
Via implementing the further actions of the member-info-drawer:

https://github.com/acterglobal/a3/assets/40496/ceddb20b-bf07-43ec-bf2f-039e3a6c8bb7

Best to be reviewed on commit at a time, as they contain:

1. a refactor of member related widgets into its own feature with several files. This contains a few but minor changes as well. 
   - moving of the power level changing from the member-list-entry further-actions to the member-info-drawer
   - the room-specific info in the member-info-drawer
   - cleaning up the member-list-entry a bit 
3. implementation of the Kick dialog
4. implementation of the kick & ban, which is mostly a copy-paste of the kick 

Fixes #1234 .

Issue #1526, opened 2024-03-18
user can't change password.


gnunicorn:
but should be able to!

PR #1676, opened 2024-04-25
Change password

kumarpalsinh25:
fixes, #1526 and #1533 

Reference screenshot:

https://github.com/acterglobal/a3/assets/51526480/5493d663-c3cc-4451-8000-c0b64d6a4bc2


app/lib/features/settings/pages/change_password.dart
```diff
@@ -0,0 +1,218 @@
...
+      ],
+    );
+  }
+
+  void _changePassword(BuildContext context) async {
+    if (!formKey.currentState!.validate()) return;
+    EasyLoading.show(status: L10n.of(context).changingYourPassword);
+    try {
+      final client = ref.read(alwaysClientProvider);
+      await client.changePassword(oldPassword.text, newPassword.text);
```
gnunicorn, line 195: err... minor error in consistency. you are always checking the confirmation to be `.trim()` but now you are submitting them without `.trim()`... should be the same for all (though probably not a problem in practice because of the `\s` denial regexp):

```suggestion
      await client.changePassword(oldPassword.text.trim(), newPassword.text.trim());
```

app/lib/features/settings/pages/change_password.dart
```diff
@@ -0,0 +1,218 @@
...
+              onPressed: () {
+                setState(() {
+                  _oldPasswordVisible = !_oldPasswordVisible;
+                });
+              },
+            ),
+          ),
+          inputFormatters: [
+            FilteringTextInputFormatter.deny(RegExp(r'\s')),
+          ],
```
gnunicorn, line 102: I see we also have this denial in the registration and login pages, but I can't find any acceptable spec-reason to have deny them - just this SHOULD https://spec.matrix.org/v1.10/client-server-api/#notes-on-password-management . Which means users _could have_ included spaces in their previous password from different clients and we'd deny them being changed ...?!? At least for the `oldPassword`, we should not enforce this.

```suggestion
```


I think the only/main reason we had that added in the login and registration was that people would add spaces at the end of their password and not realize that and then couldn't login ... But it stands to reason we might want to catch that as a warning on our password fields (like showing that you have capslock on and have a white space at the beginning and/or end) rather than outright denying them being entered. But that is a future thing. I'll make a ticket for fixing them properly (it's a good starter ticket).

gnunicorn (approved):
Looking good. Just two minor things to apply before merging.

PR #1687, opened 2024-04-29
Edit chat Profile

kumarpalsinh25:


app/lib/features/chat/pages/edit_chat_page.dart
```diff
@@ -0,0 +1,218 @@
...
+    );
+  }
+
+  Future<void> _save() async {
+    if (chatNameController.text.trim().isEmpty) return;
+    final convo = await ref.read(chatProvider(widget.roomId).future);
+    if (!mounted) return;
+    EasyLoading.show(status: L10n.of(context).updatingChat);
+    try {
+      // await convo.setName(chatNameController.text.trim());
```
kumarpalsinh25, line 198: @gnunicorn @gtalha07 @bitfriend 
Can we have method line `convo.setName() `to change chat title data?

Similar to what we have in Edit Space (`await space.setName(title)`).

gtalha07: @kumarpalsinh25 I can look on that first basis, will get ready by today 👍🏻!

.changes/1687-edit-chat-profile.md
```diff
@@ -0,0 +1 @@
+- [new] : Now user can update chat profile details (Name, Topic & Avatar) if they have required access.
```
gnunicorn, line 1:
```suggestion
- [new]: You can now update chat profile details (Name, Topic & Avatar) for chats you are admin of
```

app/lib/features/chat/pages/edit_chat_page.dart
```diff
@@ -0,0 +1,218 @@
...
+    final convoData = await ref.read(chatProvider(widget.roomId).future);
+    if (profileData.hasAvatar()) {
+      Directory appDocDir = await getApplicationDocumentsDirectory();
+      final dirPath = p.join(appDocDir.path, 'dir');
+      await Directory(dirPath).create(recursive: true);
+      String filePath = p.join(appDocDir.path, '${widget.roomId}.jpg');
+      final imageFile = File(filePath);
+      imageFile.writeAsBytes(profileData.avatar!.asTypedList());
+      selectedUserAvatar.value = imageFile;
+    }
```
gnunicorn, line 56: why is this not just using the regular `ActerAvatar` facilities?

app/lib/features/chat/pages/edit_chat_page.dart
```diff
@@ -0,0 +1,218 @@
...
+          textInputType: TextInputType.multiline,
+          controller: chatDescriptionController,
+          maxLines: 10,
+        ),
+      ],
+    );
+  }
+
+  Future<void> _save() async {
+    if (chatNameController.text.trim().isEmpty) return;
```
gnunicorn, line 193: yeah, no ... you can set the description and name independently of one another ... forcing a name and always updating both isn't quite right...

app/lib/features/chat/pages/edit_chat_page.dart
```diff
@@ -0,0 +1,218 @@
...
+    final convo = await ref.read(chatProvider(widget.roomId).future);
+    if (!mounted) return;
+    EasyLoading.show(status: L10n.of(context).updatingChat);
+    try {
+      await convo.setName(chatNameController.text.trim());
+      await convo.setTopic(chatDescriptionController.text.trim());
+      if (selectedUserAvatar.value != null) {
+        await convo.uploadAvatar(selectedUserAvatar.value!.path);
+      }
+      EasyLoading.dismiss();
```
gnunicorn, line 203: each one of them is a separate matrix room event, if they didn't change, we should be sending the event as that just produces noise ...

gnunicorn (changes requested):
I am pondering if this is the best way to deal with it. For once the permissions are not as linked as this flow suggest (you might have the ability to change the description as an admin but not the title), secondly I wonder if this needs an entire new page ... while on other places we try to move into a more inline-editing-flow ...

last but not least I am confused by the way the avatar is rendered ....

app/lib/features/chat/pages/edit_chat_page.dart
```diff
@@ -0,0 +1,218 @@
...
+      if (!mounted) return;
+      Navigator.of(context).pop();
+    } catch (e, st) {
+      _log.severe('Failed to edit chat', e, st);
+      EasyLoading.dismiss();
+    }
+  }
+
+  Widget _saveButton() {
+    return ActerPrimaryActionButton(
```
gtalha07, line 213: Is this documented somewhere explaining its need and purpose? So we know when to use it and not just use simple ```Elevated``` or ```Outlined```?

gtalha07 (commented):
LGTM after other feedback is addressed!

app/lib/features/chat/pages/edit_chat_page.dart
gnunicorn: not yet.

kumarpalsinh25:
@gnunicorn 

> For once the permissions are not as linked as this flow suggest (you might have the ability to change the description as an admin but not the title), secondly I wonder if this needs an entire new page ... while on other places we try to move into a more inline-editing-flow ...

Agree with you. Good notice. Let me do require changes here.


> last but not least I am confused by the way the avatar is rendered ....

My Bad! Let me do the corrections here :)

app/lib/features/chat/pages/edit_chat_page.dart
kumarpalsinh25: This due to flow. When user select avatar, not uploading it directly and showing updated avatar from local file path and It will be uploaded upon click of Save button.

But let me do needful changes on the flow with inline approach so that this won't be the case.

kumarpalsinh25:
@gnunicorn 
I have did required changes on the code to make "setRoomName" and "setRoomDescription" as independent widgets which we can open based on the user permissions.

Share video recording for new flow reference.


https://github.com/acterglobal/a3/assets/51526480/fa6fa416-b49d-4704-8960-9afe8f9a3d33

app/lib/common/widgets/chat/edit_room_description_sheet.dart
```diff
@@ -0,0 +1,95 @@
...
+    super.key,
+    required this.description,
+    required this.roomId,
+  });
+
+  final _descriptionController = TextEditingController();
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    _descriptionController.text = description;
```
gnunicorn, line 43: this is a side-effect and shouldn't be done in `build` but in `didUpdateDependencies` of a stateful object no? Using a stateless object with a local texteditingcontroller is considered bad style, isn't it? (as it has an internal state but that is hidden to the rendering tree)...

I'd rather you'd make that a proper statefulconsumer and set that not in `build` but have it some of the state-update functions...

app/lib/common/widgets/chat/edit_room_description_sheet.dart
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      EasyLoading.show(status: L10n.of(context).updateDescription);
+      context.pop();
+      final convo = await ref.read(chatProvider(roomId).future);
+      await convo.setTopic(_descriptionController.text.trim());
+      EasyLoading.dismiss();
+    } catch (e, st) {
+      _log.severe('Failed to edit chat description', e, st);
+      EasyLoading.dismiss();
+    }
```
gnunicorn, line 93: this is not the proper flow.

1. let's check for actual differences first (we have the `description` on the widget param) and not submit if it isn't any different. Let's just close the screen in that case.
2. we should show the error to the user as well, so they know it _didn't_ succeed and
3. we should not close the box upon error, we might be throwing away long crafted content and frustrate the user. better to keep it open until we succeeded

```suggestion
    final newDesc = _descriptionController.text.trim();
    if (newDesc == description.trim()) {
      context.pop();
      return; // no changes to submit
    }
    
    try {
      EasyLoading.show(status: L10n.of(context).updateDescription);
      final convo = await ref.read(chatProvider(roomId).future);
      await convo.setTopic(newDesc);
      EasyLoading.dismiss();
      context.pop();
    } catch (error, st) {
      _log.severe('Failed to edit chat description', error, st);
      EasyLoading.showError(L10n.of(context).updateDescriptionFailed(error);
    }
```

app/lib/common/widgets/chat/edit_room_name_sheet.dart
```diff
@@ -0,0 +1,93 @@
...
+              ),
+            ],
+          ),
+        ],
+      ),
+    );
+  }
+
+  Future<void> _editName(BuildContext context, WidgetRef ref) async {
+    try {
```
gnunicorn, line 82: same as above

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -121,21 +149,75 @@ class _RoomProfilePageState extends ConsumerState<RoomProfilePage> {
...
+      EasyLoading.show(status: L10n.of(context).uploadAvatar);
+      final convo = await ref.read(chatProvider(widget.roomId).future);
+      final file = result.files.first;
+      if (file.path != null) await convo.uploadAvatar(file.path!);
+      // close loading
+      EasyLoading.dismiss();
+    } catch (e, st) {
+      _log.severe('Failed to edit chat', e, st);
+      EasyLoading.dismiss();
+    }
```
gnunicorn, line 169: we should show the failure. as said above

gnunicorn (approved):
but the submission code needs an update for all of them - we could fail without informing the user and that is just confusing.

PR #1691, opened 2024-04-29
show proper room member invite/join UI states

### LINKED ISSUE #1107: Stuck on loading when inviting user to chat
I wanted to invite a user to a chat, but it was stuck on loading.
However later on if figured out it prob was because she was already in the chat.
Ideally we want to show a message that the user has alreayd joined, instead of endless loading

(update 8th Novemember)
It unfortunately also seems to happen although the user is not in the room.
What happens:
I click "create chat". -> I type user name in search field -> the search result list is stuck on loading -> if i wait long enough, below image shows.

I reopen the app several times -> works again.

Image

gtalha07:
- [x] fix #1107.
- [x] minor user invite/join state button improvements.
- [x] Replace static strings with localised l10n.
- [x] add changelogs.

Demo: 

1. User invited/join state live changes and show proper reads.

https://github.com/acterglobal/a3/assets/68579938/59ed2d8e-8780-4c4d-a115-29b8d66c6d4e

2. If direct invitation sent to unknown userId, it shows proper error UI now.

https://github.com/acterglobal/a3/assets/68579938/9a632704-5406-477f-9332-77d9ca290768


emilvincentz:
Looks good @gtalha07!
However a few  questions: 
1. How come the user @talha:acter.global does not come up in the public directory search, with the profile picture and username? instead it shows "directly invite"?
2. Can a user search by display name?
3. can the user currently input an email? or do we not have this flow yet?

gtalha07:
> Looks good @gtalha07! However a few questions:
> 
> 1. How come the user @talha:acter.global does not come up in the public directory search, with the profile picture and username? instead it shows "directly invite"?
 The direct invite search expects fully qualified username .i.e. starting with initials `@` so having user input this will infer that 
  it is a known information and so we can directly send invitation over ```userId```. The other case suggests that you don't 
  have ```userId``` information which you can directly search it via ```displayName``` that'll show relative results in 
  public directory.
     
  Screenshot 2024-04-30 at 2 15 00 PM
  
> 2. Can a user search by display name?
   Yes, the search works just like any other search bar. This will come up with all the relative results. 
> 3. can the user currently input an email? or do we not have this flow yet?
   In meaning of just like direct invite, no. We don't currently support this flow yet.

app/lib/common/dialogs/invite_to_room_dialog.dart
```diff
@@ -158,6 +159,7 @@ class _InviteToRoomDialogState extends ConsumerState<InviteToRoomDialog>
     final room = ref.watch(briefRoomItemWithMembershipProvider(roomId));
     final invited =
         ref.watch(roomInvitedMembersProvider(roomId)).valueOrNull ?? [];
+    final joined = ref.watch(membersIdsProvider(roomId)).valueOrNull ?? [];
```
gnunicorn, line 162: may I suggest that we move this watch into the `UserStateButton`? That would remove the flicker of refreshing the entire page when we invited someone or their state changed ...

app/lib/common/widgets/user_builder.dart
```diff
@@ -49,6 +61,7 @@ class UserBuilder extends ConsumerWidget {
     final room = ref.watch(briefRoomItemWithMembershipProvider(roomId));
     final invited =
         ref.watch(roomInvitedMembersProvider(roomId)).valueOrNull ?? [];
+    final joined = ref.watch(membersIdsProvider(roomId)).valueOrNull ?? [];
```
gnunicorn, line 64: would also avoid doing that twice if we moved it into the `UserStateButton`...

gnunicorn (approved):
just a minor request to remove UX flickering...

app/lib/common/dialogs/invite_to_room_dialog.dart
gtalha07: Good point! fixed by 08838aa

app/lib/common/widgets/user_builder.dart
gtalha07: fixed by 08838aa

PR #1699, opened 2024-04-30
Room hierarchy join button fixes

gnunicorn:
- Fixes chat-room join space button update issue: https://github.com/acterglobal/a3-meta/issues/197
- Fixes routing after join-chat having no back-button: https://github.com/acterglobal/a3-meta/issues/196

This is what it does now:

https://github.com/acterglobal/a3/assets/40496/3d172c9a-a052-4910-b8a4-05e16ede1d76


app/lib/features/space/widgets/relatest_spaces.dart
```diff
@@ -214,8 +212,8 @@ class RelatedSpaces extends StatelessWidget {
         firstPageKey: const Next(isStart: true),
         provider: remoteSpaceHierarchyProvider(spaces),
         itemBuilder: (context, item, index) => SpaceHierarchyCard(
-          key: Key('subspace-list-item-${item.roomIdStr()}'),
```
gtalha07, line 217: Are we sure this doesn't break the UI tests???

app/lib/features/space/widgets/relatest_spaces.dart
```diff
@@ -284,7 +282,6 @@ class RelatedSpaces extends StatelessWidget {
           itemBuilder: (context, index) {
             final space = spaces.otherRelations[index];
             return SpaceCard(
-              key: Key('subspace-list-item-${space.getRoomIdStr()}'),
```
gtalha07, line 287: Here too, is this intended for ```KeyFinder``` in  integration suite???

gtalha07 (commented):
LGTM overall, but please address the widget key removal as it might affect the integration test suite!

app/lib/common/widgets/room/room_hierarchy_join_button.dart
```diff
@@ -0,0 +1,81 @@
...
+                viaServerName,
+                forward,
+              );
+            },
+            child: Text(L10n.of(context).join),
+          ),
+        );
+      case 'public':
+        return Tooltip(
+          message: L10n.of(context).youNeedBeInvitedToJoinThisRoom,
```
kumarpalsinh25, line 59: For public room also need to be invited?

app/lib/common/widgets/room/room_hierarchy_join_button.dart
```diff
@@ -0,0 +1,81 @@
...
+      );
+    }
+    switch (joinRule) {
+      case 'private':
+      case 'invite':
+        return Tooltip(
+          message: L10n.of(context).youNeedBeInvitedToJoinThisRoom,
+          child: Chip(label: Text(L10n.of(context).private)),
+        );
+      case 'restricted':
```
kumarpalsinh25, line 40: So this `restricted` is members only permission where you need to be invited.

We try to call Join API and it will return error if user is not invited, correct?

kumarpalsinh25 (approved):
LGTM!

Few minor queries need to be address before merge.

1) Joining Button Queries:
- Added comment on relevant code

2) Back button issue:
- Can you help with code file/line which was causing this issue and it is fixed with this PR?

app/lib/common/widgets/room/room_hierarchy_join_button.dart
gnunicorn: no, this is wrong. you can just join it ...

app/lib/common/widgets/room/room_hierarchy_join_button.dart
gnunicorn: > So this restricted is members only permission where you need to be invited.

no. restricted and you can see it means you can join it.

gnunicorn:
> Can you help with code file/line which was causing this issue and it is fixed with this PR?

yes, sure. 

the error was using this `goNamed` here:
https://github.com/acterglobal/a3/pull/1699/files#diff-92ae72068d2bb89e0d7b37dc00de5fa49e659067b8eb3ce27d6774b6f997d96fL114-L117

and is replaced with the more robust `goToChat` here:

https://github.com/acterglobal/a3/pull/1699/files#diff-92ae72068d2bb89e0d7b37dc00de5fa49e659067b8eb3ce27d6774b6f997d96fR123-R127

and `goToSpace` here:

https://github.com/acterglobal/a3/pull/1699/files#diff-ec7392b445291b807dba9130c907cc65694852826835f7fa26f8516e073df1d2R130-R133

PR #1701, opened 2024-04-30
 check connectivity before sending message

### LINKED ISSUE #1544: Error when sending message
When i click "send" it just looks like this.

Image

gtalha07:
- fixes #1544.

https://github.com/acterglobal/a3/assets/68579938/b8f3916f-1132-4bce-ac35-78a0b9e64bb0


gnunicorn (approved):
code wise this is fine ... but it feels like it is just shipping around the state issue that we still have somewhere in the providers ....

PR #1720, opened 2024-05-06
revert send message error and fix timeline event state error overflow

gtalha07:
- This will revert #1701 as this seems to not check whether server is reachable.
- Also this fixes the UI overflow issue on timeline message error state.

Screenshot 2024-05-06 at 8 06 09 PM 
Screenshot 2024-05-06 at 8 07 36 PM


gnunicorn (approved):
I guess that means we have to reopen https://github.com/acterglobal/a3/issues/1544 !?!

gnunicorn:
ah, already happened..!

Issue #1686, opened 2024-04-28
Accepting chat invite or space invite, doesnt direct user to the accepted chat/space


emilvincentz:
The problem with this, is that you accept the chat/space and want to afterwards view it. But you dont always remember the name of the chat /space. Also as the chat doesnt always appear on the top.
Ive observed user clicking on the accept button, and then response being "oh, where did it go?"
As such we should direct the user to the accepted chat/space. the same as on creating a space or chat.

kumarpalsinh25:
https://github.com/acterglobal/a3/assets/51526480/d669f911-02b6-4f17-a7a4-782722226af0

gnunicorn:
@kumarpalsinh25 https://github.com/acterglobal/a3/blob/cc40faff170d9360c15434d9a7ca3af1a6907cbe/native/acter/api.rsh#L2428 .

PR #1724, opened 2024-05-08
Go to space or chat after accepting invitation

kumarpalsinh25:
fixes, #1686

https://github.com/acterglobal/a3/assets/51526480/2efe4d80-1bb0-4620-bfcf-8ac7a7b89c9a

### DIFF
--- a/app/lib/features/activities/widgets/invitation_card.dart
+++ b/app/lib/features/activities/widgets/invitation_card.dart
@@ -3,6 +3,7 @@ import 'package:acter/common/themes/app_theme.dart';
 
 import 'package:acter/common/toolkit/buttons/primary_action_button.dart';
 import 'package:acter/features/activities/providers/invitations_providers.dart';
+import 'package:acter/router/utils.dart';
 import 'package:acter_avatar/acter_avatar.dart';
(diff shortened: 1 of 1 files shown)


gtalha07:
Nice fixes! Although I see one possible suggested improvement for user-experience as seen in demo video:
- Can we delay routing or add some check before space/chat is properly loaded? Accessing route before we have any data implies that it's broken and also opens up a doorway for a scenario where it actually fails to fetch due to internet problems/server issue where in that case, we'd want nice error UI.

gnunicorn:
> Can we delay routing or add some check before space/chat is properly loaded? Accessing route before we have any data implies that it's broken and also opens up a doorway for a scenario where it actually fails to fetch due to internet problems/server issue where in that case, we'd want nice error UI or prevent routing.

Arguably we want a better loading state in that case with some skeletonizer fields, no?

gtalha07:
> > Can we delay routing or add some check before space/chat is properly loaded? Accessing route before we have any data implies that it's broken and also opens up a doorway for a scenario where it actually fails to fetch due to internet problems/server issue where in that case, we'd want nice error UI or prevent routing.
> 
> Arguably we want a better loading state in that case with some skeletonizer fields, no?

Yes, precisely!!!

Issue #1700, opened 2024-04-30
When linking subspaces, it doesent show straight  away.


emilvincentz:

emilvincentz:
also, sometimes when linking to a sub space it shows up as a recommended space

PR #1729, opened 2024-05-09
Fixes : Linking subspaces doesn't show straight away

kumarpalsinh25:
- Remove unused code from the space provider
- fixes, #1700 

Also during fixes of #1700, I come to know that code related to update join rule (`await room.setJoinRule(update);`) not working at all and never get back from `await` state. Added `//FIXME` comment over there.

Reference video:

https://github.com/acterglobal/a3/assets/51526480/4656eaf3-02e3-4a19-9593-7267688caed2

### DIFF
--- a/app/lib/features/space/sheets/link_room_sheet.dart
+++ b/app/lib/features/space/sheets/link_room_sheet.dart
@@ -383,6 +383,7 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
         for (final roomId in currentRooms) {
           update.addRoom(roomId);
         }
+        //FIXME : Below is not working at all.
         await room.setJoinRule(update);
       }
     }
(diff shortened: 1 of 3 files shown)


app/lib/features/space/sheets/link_room_sheet.dart
```diff
@@ -383,6 +383,7 @@ class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
         for (final roomId in currentRooms) {
           update.addRoom(roomId);
         }
+        //FIXME : Below is not working at all.
         await room.setJoinRule(update);
```
kumarpalsinh25, line 387: @gnunicorn 
Based on my testing, `await room.setJoinRule(update);` is not working. Can you test it once from your side?

gnunicorn: added and pushed an integration test proofing this works as expected. Unless you can tell me a specific case where this doesn't I'd mark this resolved. Only note is that this isn't instantaneous but needs a sync, so it might take a second...

kumarpalsinh25: Ok thanks

kumarpalsinh25:
@gnunicorn 
Can you resolve conflicts of the rust file to proceed this PR for merge?

PR #1734, opened 2024-05-10
Calendar event location apis

gtalha07:
- Adds ```CalendarEventInfo``` object and methods for ```EventLocation```.
- ```CalendarEventDraft``` now supports physical/virtual location setters for location fields.
- ```CalendarEventUpdateBuilder``` have unset/reset option for locations.
- ```CalendarEvent``` now have various methods for fetching ```EventLocation``` data.


native/acter/src/api/calendar_events.rs
```diff
@@ -265,6 +267,36 @@ impl CalendarEvent {
...
+                result.push(EventLocationInfo::new(loc));
+            }
+        }
+        result
+    }
+
+    pub fn locations(&self) -> Vec<EventLocationInfo> {
+        let calendar_event = self.clone();
+        let locations = calendar_event.inner.locations();
+        locations.iter().map(EventLocationInfo::new).collect()
```
gnunicorn, line 297: see, you already did it!

native/acter/src/api/calendar_events.rs
```diff
@@ -348,6 +380,56 @@ impl CalendarEventDraft {
...
     }
 
+    pub fn physical_location(
+        &mut self,
+        name: String,
+        description: String,
+        description_html: Option<String>,
+        cooridnates: String,
+        uri: Option<String>,
+    ) -> Result<()> {
```
gnunicorn, line 390: fine for now, but I'd like this to be a builder-pattern in the future rather than additional methods ...

native/core/src/events/calendar.rs
```diff
@@ -56,6 +56,100 @@ pub enum EventLocation {
...
+    pub location_type: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub description: Option<TextMessageEventContent>,
+    pub icon: Option<Icon>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub coordinates: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub uri: Option<String>,
```
gnunicorn, line 70: why copy this over rather than just have the `EventLocation` as an inner type? Why is this type needed at all actually?

gtalha07:
> It's okay and we can push this through as is if we need this urgently, but I think there is room for improvement, in particularly around the EventLocationInfo type (which I do not understand why that is needed)

Particularly, it's been written this way as I feel this makes api structure more flexible and convenient to consume. But it can be subjective and arguable. Also I see it's easier to write object which have simpler structure on ```api.rsh``` rather than complex enums.

native/core/src/events/calendar.rs
gtalha07: Good thinking, this is a nice improvement and will reduce redundancy. I'll change it! 👍🏻

Issue #1746, opened 2024-05-13
MacOS release doesn't keep the session between app restart 


gnunicorn:
Starting up, logging in, syncing, all good. starting up again: nothing.

Video:

https://github.com/acterglobal/a3/assets/40496/c03cea90-cf39-4c2e-b2de-4842f85ee91d

Issue #1728, opened 2024-05-09
MacOS opens with black screen


emilvincentz:

emilvincentz:
![Image](https://github.com/acterglobal/a3/assets/77483554/001b08a5-e0af-4bb0-9df6-62aec6af5270)

gnunicorn:
can you start it from the commandline and paste the output here please?!?

emilvincentz:
@gnunicorn 
Last login: Sun May 12 17:05:14 on ttys000
/Applications/Acter.app/Contents/MacOS/Acter ; exit;
emil@Matildes-Air ~ % /Applications/Acter.app/Contents/MacOS/Acter ; exit;
flutter: log settings: acter::api::client=trace,acter::api::spaces=trace,acter_core::executor=trace,acter_core=debug,acter=warn,a3=warn,warn
flutter: logs will be found in /Users/emil/Library/Containers/global.acter.a3/Data/Library/Caches/global.acter.a3
flutter: handle error - address: 105553128788720 ; len: 33
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: expected value at line 1 column 1

2024-05-12 17:05:35.474 Acter[35593:1673430] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.

gnunicorn:
> [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: expected value at line 1 column 1


now that is bad.

emilvincentz:
Logs from trying to:
1. run the app from the command line to overwrite the log settings by running /Applications/Acter.app/Contents/MacOS/Acter settings set rust-log trace
2. run the app again from the command line and copy-paste the output into the issue . it might be a lot ...

Last login: Thu May 23 16:20:16 on ttys000
emil@Matildes-Air ~ % /Applications/Acter.app/Contents/MacOS/Acter settings set rust-log trace
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Invalid argument(s): Option key is mandatory.
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 34766, tid: 6125481984, name io.flutter.ui
os: macos arch: arm64 comp: no sim: no
build_id: 'b09aafd842763feab274505d08c78f21'
[...]

PR #1747, opened 2024-05-13
Update secure storage

gnunicorn:
Updated to latest version of secure storage, added a fix for #1746 , hopefully also solves the startup freeze issue reported in #1728  - gotta see.


gnunicorn:
confirmed to work on windows without problems, too

PR #1748, opened 2024-05-13
New super invites info system

gnunicorn:
Implement the new super invites info for tokens added in 0.8.2 .

Includes a UI test case update, too:

https://github.com/acterglobal/a3/assets/40496/0ae90ac3-9ecd-404f-b60b-831ff1db9bde


app/lib/features/super_invites/dialogs/redeem_dialog.dart
```diff
@@ -0,0 +1,157 @@
...
+      child: Card(
+        child: ListTile(
+          key: redeemInfoKey,
+          title: Text(
+            L10n.of(context).superInvitedBy(
+              displayName != null ? '$displayName ($userId)' : userId,
+            ),
+          ),
+          subtitle: Text(
+            L10n.of(context).superInvitedTo(info.roomsCount()),
```
kumarpalsinh25, line 105: Is there any possibility to show room type bifurcation. Something Like `Abc user invites to you join 1 Space and 2 Chats`?

kumarpalsinh25 (approved):
Flutter side changes looks good to me,

app/lib/features/super_invites/dialogs/redeem_dialog.dart
gnunicorn: not at this point - we are not exposing any details about the rooms other than the total count from the server yet.

Issue #1750, opened 2024-05-14
New Update : App getting crashed on Image/Video posting


kumarpalsinh25:
https://github.com/acterglobal/a3/assets/51526480/61651a84-8dc5-4465-b611-473d2b7ce67b

gnunicorn:
can reproduce.

error comes from the recently changed `__MsgDraft_into_news_slide_draft` call:

```
 #15 pc 00000000033e8426  /data/app/~~9PMOwpFzlGMbLiduYvsdpg==/global.acter.a3-ivCgfl7oUElj6rDdB4cAzQ==/lib/x86_64/libacter.so
      #16 pc 0000000002001116  /data/app/~~9PMOwpFzlGMbLiduYvsdpg==/global.acter.a3-ivCgfl7oUElj6rDdB4cAzQ==/lib/x86_64/libacter.so
      #17 pc 0000000002b20b28  /data/app/~~9PMOwpFzlGMbLiduYvsdpg==/global.acter.a3-ivCgfl7oUElj6rDdB4cAzQ==/lib/x86_64/libacter.so (__MsgDraft_into_news_slide_draft+24)
      #18 pc 0000000000006bca  [anon:dart-code]
Lost connection to device.

```

investigating the reasons.

PR #1765, opened 2024-05-15
Fixing #1750: clone before mutating, simplify news conversion code

gnunicorn:
as it says on the tin.

Evidence this fixes it for all three major points of failure:

https://github.com/acterglobal/a3/assets/40496/dbe7e195-32f0-4a31-9269-499bcaba156b


bitfriend (approved):
Looks good to me

PR #1770, opened 2024-05-16
remove unnecessary params of `EventLocation` enum

gtalha07:


native/core/src/events/calendar.rs
```diff
@@ -22,34 +22,18 @@ pub enum EventLocation {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         name: Option<String>,
 
-        /// further description to this location
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        description: Option<TextMessageEventContent>,
-
```
gnunicorn, line 28: Sorry, this is needed (for future usage). Consider that the address of a physical location might need some extra information, e.g. some directions within a complex that can't be easily done in a one-liner...

native/core/src/events/calendar.rs
```diff
@@ -22,34 +22,18 @@ pub enum EventLocation {
...
         #[serde(default, skip_serializing_if = "Option::is_none")]
         icon: Option<Icon>,
 
         /// A `geo:` URI [RFC5870] for the location.
         #[serde(default, skip_serializing_if = "Option::is_none")]
         coordinates: Option<String>,
-
-        /// further Link
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        uri: Option<String>,
```
gnunicorn, line 39: this could be a link to an external website of the location or a link to google maps... even physical locations have virtual representation now

native/core/src/events/calendar.rs
```diff
@@ -22,34 +22,18 @@ pub enum EventLocation {
...
         uri: String,
 
-        /// Optional name of this virtual location
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        name: Option<String>,
-
-        /// further description for virtual location
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        description: Option<TextMessageEventContent>,
-
```
gnunicorn, line 52: same as above. we need these fields even if we don't support them in the UI yet.

gnunicorn (changes requested):
Nope. Even if the UI doesn't support them yet, they have a use and we should not remove these fields.

Issue #1763, opened 2024-05-15
Show description of chat at beginning of history


gnunicorn:
When joining a fresh chat, at the beginning of history we want to show the chat description prominently in the timeline.

- [ ] Ideally the admin can also quickly edit it from right there
- [ ] and invite new users.

Additionally we want to 
- [ ] track whether a user already entered a chat room
- [ ] and if they haven't yet and there is a description, we want to prominently show the description once

PR #1771, opened 2024-05-16
Chat room fixes

gnunicorn:
When wanting to working on some stuff for #1763, I noticed some improvements to be done on the chat:

- [x] Simplify chat room design to just use columns
- [x] Move message filtering and sorting into provider
- [x] fix typing notifications:
   - [x] settings was ignored when sending typing notifications
   - [x] have one typing notification stream per room - previously the stream meant we'd only be holding _the latest_ typing notification, which means we'd only ever show the typing for _one_ room max (despite having the state per room).

### DIFF
--- a/.changes/1771-fix-typing-notifications.md
+++ b/.changes/1771-fix-typing-notifications.md
@@ -1,0 +1,2 @@
+- [fix] do not send typing notifications if the user disabled that in the settings
+- [fix] only one room could be receiving typing state at a time despite multiple rooms actually being in that state.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,6 +58,7 @@ dependencies = [
  "ruma-common",
  "ruma-events",
  "sanitize-filename-reader-friendly",
+ "scc",
  "serde",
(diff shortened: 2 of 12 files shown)


.changes/1771-fix-typing-notifications.md
```diff
@@ -0,0 +1,2 @@
+- [fix] do not send typing notifications if the user disabled that in the settings
+- [fix] only one room could be receiving typing state at a time despite multiple rooms actually being in that state.
```
gtalha07, line 2: Does that mean we don't get notified about typing states if happening across multiple rooms? Or they would be queued up?

gtalha07 (approved):
LGTM!

.changes/1771-fix-typing-notifications.md
gnunicorn: > Does that mean we don't get notified about typing states if happening across multiple rooms?

Yes, previously only one room would be shown as typing on the flutter side ever. I noticed in the code ...

Issue #1662, opened 2024-04-21
bubbles at the start of a chat


gnunicorn:
At the beginning of each chat we have these weird tiny bubbles without content:
![grafik](https://github.com/acterglobal/a3/assets/40496/d4455368-c29c-4553-a074-e69609a913bf)

I suspect these show-indicate some state events that we simply have no proper way of showing, but they should a) be shown and possible to understand and b) grouped together into a single event that can then be expanded, similar to how element allows to combine several status events.

PR #1773, opened 2024-05-18
Improving Chat design: more muted states design, show topic at the beginning of the history 

### LINKED ISSUE #1763: Show description of chat at beginning of history

gnunicorn:
This improves the Chat room in several manners. Commits are atomic about one aspect at a time, so that might be an easier way to review than looking at the entire changeset at a time.

- Show Encryption info message and chat room topic at the beginning of the room history, fixes #1763 . Includes Editing facilities
- Upgrade of the matrix-rust-sdk to provide more robust back-pagination facilities
- Fixes in the state management of back pagination as that wasn't always happen properly, in particular if the room had very few messages
- Refactoring of the chat room message widget into separate files
- New facilities to filter the messages to show already on the provider side, fixing #1662 .
- Several improvements on rendering the different custom messages:
  - membership changes: Show the display name or "you" if it was you; use L10n rather than hard-coded strings; Use a more muted style to be less pungent
  - A new format for encrypted messages with a new tap-for-more-drawer that we can allow for quicker action buttons in the future (e.g. verify device, verify with other user, provide encryption key)

In the real world this looks like the following now:

https://github.com/acterglobal/a3/assets/40496/1c3caa35-8f65-4d08-bf8e-a140250caae7


app/lib/features/chat/utils.dart
```diff
@@ -6,12 +6,58 @@ import 'package:flutter/material.dart';
...
+      // no support yet
+      return false;
+    case 'm.sticker':
+    case 'm.room.redaction':
+    case 'm.room.encrypted':
+      // supported
+      return true;
+    case 'm.room.message':
+      // special case of supporting locations
+      return message.metadata?['msgType'] == 'm.location';
```
kumarpalsinh25, line 60: Not getting understanding of this condition.

```
case 'm.room.message':
      // special case of supporting locations
      return message.metadata?['msgType'] == 'm.location';
```

app/lib/features/chat/widgets/messages/location_message.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_chat_types/flutter_chat_types.dart';
+
+class LocationMessageWidget extends StatelessWidget {
```
kumarpalsinh25, line 4: Ahh, I see. We have support of the location message as well. I never used/see this in acter.

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
```diff
@@ -58,7 +58,7 @@ class ChatRoomNotifier extends StateNotifier<ChatRoomState> {
         state = state.copyWith(
           loading: const ChatRoomLoadingState.loading(),
         );
-        final hasMore = await timeline.paginateBackwards(20);
+        final hasMore = !await timeline.paginateBackwards(20);
```
kumarpalsinh25, line 61: Not getting idea about this condition changes.

kumarpalsinh25 (approved):
Niceee Improvement!!

Added few minor queries but flutter code looks good to me.

app/lib/features/chat/utils.dart
gnunicorn: it's what we had before, just slightly refactored. I can't tell you the _meaning_ of each of them, just tried to not break them.

app/lib/features/chat/widgets/messages/location_message.dart
gnunicorn: yeah, dunno if we ever implemented sending it...

app/lib/features/chat/providers/notifiers/chat_room_notifier.dart
gnunicorn: I've upgraded the SDK to simplify the code, as a result the API has been updated to not indicate that there is more but that it reached the end. see
https://github.com/acterglobal/a3/pull/1773/files#diff-12c0480bc493ffbb94ce49b75db168b6aec7504019bcbdea587c1192ffe429c2L1128

As a result the state assumptions in the provider for `hasMore` are = `! reachedEnd`. This part of the same commit, it changed together.

Issue #845, opened 2023-08-10
Activity indicator on chat overview


gnunicorn:
There is no activity indicator on chat overview, making it difficult to figure out which ones are unread.

Even just a dot would be a good start.

![Image](https://user-images.githubusercontent.com/77483554/259409029-ebb8da81-d766-4eb3-be29-023bd82803b8.png)

gnunicorn:
Read/Unread-marker are not currently implement since the internal state refactor (it was down-prioritised to fix it)

gnunicorn:
let's check if the new matrix sdk allows for that https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/index.html

PR #1782, opened 2024-05-23
Chat read marker support

gnunicorn:
Attempting to implement #845 behind a new labs-feature (for testing).

Okay, after reading up on it, I think what we are doing here in this PR is considered _correct_ right now, so I bring this up to review now. That said, it seems to still not work super reliably for some reason, but I'd like to merge this for now and see what really happens in practice.

That said, what even is in this PR:

- An update on the notification counting rust API to use what comes from the sync
- A generalization of the Urgency-marker of the activity icon so we can reuse it on the chat badges (and some simplifications around that in the menu UI)
- that marker added to the chat when we have unread messages
- the virtual read-marker as a custom system Message in the chat UI
- a feature to mark the latest chat message as and submit the read receipt for it (probably still buggy for when new messages comes in but you don't scroll - open for better ideas how to manage that nicely)
- a split of the chat input providers from the chat to avoid re-renders just because the read-marker has moved...

See it in action:

https://github.com/acterglobal/a3/assets/40496/1f9fc058-0489-4fd4-ab00-ce9ac506db02

indicator shows as long as at least one is unread:

https://github.com/acterglobal/a3/assets/40496/e16cf491-3cb5-45d6-8935-2f172d5f30c2


gtalha07:
Any screenshots or previews possible ?

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -204,3 +217,50 @@ final chatTypingEventProvider = StreamProvider.autoDispose
...
+
+    return UrgencyBadge.none;
+  }
+  final chats = ref.watch(chatsProvider);
+  if (chats.isEmpty) {
+    return UrgencyBadge.none;
+  }
+  UrgencyBadge currentBadge = UrgencyBadge.none;
+
+  for (final chat in chats) {
```
gtalha07, line 252: Yes, this would seem not a good optimal approach, I am facing similar case in fetching space details for multiple parents. I initially used ```.map()``` approach but somehow the provider is not functioning correctly and only returns one value .i.e. I sense it's because of ```FutureProviders``` bug in terms of dealing with future ```ref.watch()```.  Suggestion would be to use map along with awaiting all futures concurrently and returning first non-null data:
```suggestion
  final badges = await Future.wait(
  chats.map((chat) async {
    final unreadCounter = ref.watch(unreadCountersProvider(chat.getRoomIdStr()).future);
    if (unreadCounter.$1 > 0) {
      return UrgencyBadge.important;
    } else if (unreadCounter.$2 > 0) {
      return UrgencyBadge.unread;
    } else {
      return null;
    }
  }),
);

return badges.firstWhere((badge) => badge!= null, orElse: () => null);
```

app/pubspec.yaml
```diff
@@ -37,6 +37,7 @@ dependencies:
   skeletonizer: ^1.1.1
   flutter_chat_ui: ^1.6.9
   flutter_chat_types: ^3.3.2
+  scroll_to_index: ^3.0.1
```
gtalha07, line 40: I believe ```flutter_chat_ui``` support scroll to unread [feature](https://pub.dev/documentation/flutter_chat_ui/latest/flutter_chat_ui/ScrollToUnreadOptions-class.html), if its used for that case specifically...

gtalha07 (commented):
I've left a few remarks specifically some loop improvements regarding badge checks which can addressed. Also it'd be nice if we can preview screenshots, to see how it turns in actual UI before I can give any explicit approval!

app/lib/common/widgets/chat/convo_with_profile_card.dart
```diff
@@ -76,4 +77,33 @@ class ConvoWithProfileCard extends ConsumerWidget {
       },
     );
   }
+
+  Widget avatarWithIndicator(BuildContext context, WidgetRef ref) {
```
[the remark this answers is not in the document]
gnunicorn: yeah, we'll see ... and figure it out when we get there ;) .

app/lib/features/chat/providers/chat_providers.dart
gnunicorn: not sure I agree ... this would poll _all_ providers regardless while mine might be _slower_ but would stop on the first match, which considering that they are cached shouldn't be that much of a difference. I think a better solution is to put all that SDK side, but I leave that for a future implementation I think...

app/pubspec.yaml
gnunicorn: It comes with the flutter_chat_ui but needed to add it so I can connect to the scroll controller as the chat ui doesn't expose the type...

gnunicorn:
@gtalha07 as requested I added some screencasts to the description showing it in action

gtalha07 (approved):
Generally this looks fine and feature addition, but I have few suggestions to have much improved for the user readability and recognition of it (not much urgent for this PR as it goes in experimental flag):

- I had assumed this shows up as count marker for unread message on far right side of the chat .i.e. similar design for element. Is it possible that we can use that kind of UI for it? Not saying it should be exactly same, the positioning can be done suitably but count can be more expressive.
- I have some concerns regarding marker color, as on avatar lighter color notes, this almost becomes invisible (to put it bluntly) or very obscure to actually recognize it. It'd be better if we can put this marker somewhere in background which is of static color .i.e. room title right side could be one option but I'd put that choice in your hands :)

Generally I think this is okay to go for now, hence approved!

gnunicorn:
@gtalha07 yeah, this needs some product and design work before we can move it off the labs flag. The problem with the "numbers" you are referring to is that the ones we get from the API are in itself inconsistent right now. Not sure where that comes from but I think it might be because matrix-sdk is moving to sliding-sync and we don't yet support that...

PR #1784, opened 2024-05-24
Chat emoji design fixes

gnunicorn:
Improved design of the emoji bar in chat (left shows desktop, right show mobile):

Bildschirmfoto_2024-05-24_14-42-09

Wrapped in card for design, only shows a number when more than 1 and highlight if we clicked it in a less aggressive style. Also includes a few minor code style improvements. For reviewing it is probably easiest to do them commit by commit as only the commit [602208c](https://github.com/acterglobal/a3/pull/1784/commits/602208cecd32e4bef5aef0684582c04b588f94bf) contains the actual changes, while the other are just code clean ups.

fixes https://github.com/acterglobal/a3-meta/issues/238 .


app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -59,12 +57,16 @@ class BubbleBuilder extends ConsumerWidget {
             ? child
             : SwipeTo(
                 onLeftSwipe: (DragUpdateDetails details) {
-                  chatInputNotifier.setReplyToMessage(message);
```
gtalha07, line 62: Does this approach accounts as a bug ?

gnunicorn: probably. the problem is that I do not for sure know what `watch` on a `notifier` even does... but generally we should reduce all `watch` to the specific minimums required and the recommendation for notifier is to `ref.read` when you need them so I changed it to that.

gtalha07 (approved):
LGTM!

app/lib/features/chat/widgets/bubble_builder.dart
gtalha07: Yeah, I'm also unsure about ```watch``` behaviour on ```notifier``` but to be on safe-side, I think this is good enough 👍🏻 !

PR #1792, opened 2024-05-28
multi-parent avatars for spaces

### LINKED ISSUE #1798: ```SpaceRelations``` multi parents and child's bug.
There seems to issue with how ```SpaceRelations``` work in regards of linking more than one space as child. 

### Reproducible Steps:

1. Create two spaces and subspaces in each of them as child. Let's say `Root Space A` and `Root Space B` with subspaces `Subspace A` and `Subspace B`.
2. Now go in any of root spaces and choose `Link Another Existing Space`. Let's say I go in `Root Space B` and link `Subspace A` with it. Now `Root Space B` should have both child's linked.

###  Expected Behaviour:
- `Root Space B` should show both subspaces in its spaces section. 
- `Subspace A` should show both parent avatars of `Root Space A` and `Root Space B`.

### Actual Behaviour:
- `Root Space B` subspaces section UI becomes empty.
- `Subspace A` doesn't show any parent avatars.

3. Now un-link the `Subspace A` from `Root Space B`. This would show `Root Space B` subspaces section again and old parent avatar of `Subspace A` which is `Root Space A`.

gtalha07:
- Fixes #1798.
- adds the multi-avatar parent spaces support in Spaces Overview UI. From the ```acter_avatar``` [update](https://pub.dev/packages/acter_avatar/changelog#2005).

### Feature Preview:
**(Mobile)** -> Link as sub-space and update avatars

https://github.com/acterglobal/a3/assets/68579938/36ed4dd0-9d9c-4e11-a7fc-4bce0d9dbb52

**(Desktop)** -> Link and de-link as sub-space and update avatars

https://github.com/acterglobal/a3/assets/68579938/86f9d506-a070-4c5f-aa85-58fe00cef8cf


emilvincentz:
Nice Talha.
How does this look like for chat? and how does it look like when there is more than 3 linked spaces?

app/lib/common/widgets/spaces/space_with_profile_card.dart
```diff
@@ -107,14 +106,16 @@ class SpaceWithProfileCard extends StatelessWidget {
...
         ),
-        parentBadges: showParent && parentRoomId != null
-            ? [
-                AvatarInfo(
-                  uniqueId: parentRoomId!,
-                  displayName: parentProfile?.displayName ?? parentRoomId,
-                  avatar: parentProfile?.getAvatarImage(),
-                ),
-              ]
+        parentBadges: showParents && (parents != null && parents!.isEmpty)
```
gnunicorn, line 109:
```suggestion
        parentBadges: showParents && parents?.isEmpty != false
```

app/lib/features/invite_members/widgets/space_member_invite_card.dart
```diff
@@ -48,8 +48,24 @@ class SpaceMemberInviteCard extends ConsumerWidget {
...
+          final roomId = e.space.getRoomIdStr();
+          final displayName = e.profile.displayName ?? roomId;
+          final avatar = e.profile.getAvatarImage();
+          return AvatarInfo(
+            uniqueId: roomId,
+            displayName: displayName,
+            avatar: avatar,
+          );
+        }).toList();
+      }
```
gnunicorn, line 67: so ... this happens so often, it screams like that should be its own provider, don't you think? then we could just have a single `parentBadges: ref.watch(parentAvatarInfos(roomId)),` which is a quite a bit nicer and

app/pubspec.yaml
```diff
@@ -57,7 +57,10 @@ dependencies:
       url: https://github.com/acterglobal/flutter_matrix_html.git
   flutter_slidable: ^3.0.0
   quds_popup_menu: ^0.0.1+1
-  acter_avatar: 2.0.0+5
+  acter_avatar:
+    git:
+      url: https://github.com/acterglobal/acter-avatar.git
+      ref: main
```
gnunicorn, line 63: hmm... shouldn't we release that as a new version?

app/lib/features/invite_members/pages/invite_space_members.dart
```diff
@@ -63,28 +63,31 @@ class _InviteSpaceMembersConsumerState
...
-    if (parentSpace == null) return const SizedBox.shrink();
+    if (parentSpaces == null && parentSpaces!.isEmpty) {
+      return const SizedBox.shrink();
+    }
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
-      children: [
+      children: <Widget>[
         Text(L10n.of(context).parentSpace),
```
gnunicorn, line 77: this should probably also be updated ...
```suggestion
        Text(parentSpaces.length > 1 ? L10n.of(context).parentSpaces : L10n.of(context).parentSpace),
```

app/lib/features/invite_members/pages/invite_space_members.dart
```diff
@@ -63,28 +63,31 @@ class _InviteSpaceMembersConsumerState
...
-          onChanged: (value) {
-            if (selectedSpaces.contains(parentSpace.space)) {
-              selectedSpaces.remove(parentSpace.space);
-            } else {
-              selectedSpaces.add(parentSpace.space);
-            }
-            setState(() {});
-          },
-        ),
+        for (var parent in parentSpaces)
```
gnunicorn, line 78:
```suggestion
        for (final parent in parentSpaces)
```

app/lib/common/providers/room_providers.dart
```diff
@@ -169,32 +169,41 @@ final spaceRelationsProvider = FutureProvider.autoDispose
   return await room.spaceRelations();
 });
 
-/// Get the canonical parent of the space. Errors if the space isn't found. Stays up
+/// Get the canonical parents of the space. Errors if the space isn't found. Stays up
 /// to date with underlying client data if a space was found.
-final canonicalParentProvider = FutureProvider.autoDispose
-    .family<SpaceWithProfileData?, String>((ref, roomId) async {
+final canonicalParentsProvider = FutureProvider.autoDispose
+    .family<List<SpaceWithProfileData>, String>((ref, roomId) async {
```
gnunicorn, line 175: okay, naming things ... there is only one _canonical_ parent. If there are more then that's just `parentsProvider` (drop the `canonical`).

app/lib/common/providers/room_providers.dart
```diff
@@ -169,32 +169,41 @@ final spaceRelationsProvider = FutureProvider.autoDispose
...
+
+    for (final parent in allParents) {
+      final parentSpace = await ref
+          .watch(maybeSpaceProvider(parent.roomId().toString()).future);
+      if (parentSpace == null) {
+        continue;
+      }
+      final profile =
+          await ref.watch(spaceProfileDataProvider(parentSpace).future);
+      final SpaceWithProfileData data = (space: parentSpace, profile: profile);
```
gnunicorn, line 199: this isn't really safe for two reasons: if we don't know the space itself, you are just dropping it -- we haven't really agreed this is the right thing to do in general but at the very least it should at least say this here with a comment to make clear we are making this assumption of just skipping it.

secondly, if any _one_ parent fails to load the avatar or info, the entire list won't show and all we have to go for figuring this out is a warning in the logs. Even if a profile picture fails to load we should show the fallback space icon and log the error.

I think it'd be best if we refactored the providers to directly generate the AvatarInfo and update it upon successful loading of the data rather than silently dropping it.

gnunicorn (changes requested):
Generally looks good. But there is a fault in the logic of the providers that needs to be fixed before we can merge it.

app/lib/features/invite_members/widgets/space_member_invite_card.dart
gtalha07: Yeah I think there's a room for simplification here! 👍🏻

gtalha07:
> Nice Talha. How does this look like for chat? and how does it look like when there is more than 3 linked spaces?

For chats, the support isn't added yet. The changes are just for spaces. For more than 3 linked spaces, it should show as count indicator as below:

Screenshot 2024-06-20 at 12 14 55 PM

app/pubspec.yaml
gtalha07: Yeah, I think I'd probably do that now. Couldn't manage time before few days...

app/lib/common/providers/room_providers.dart
gtalha07: I'll see how it can be changed regarding the proposal you've suggested. I agree it's better to safely load list regardless of internal data availability of item.

app/lib/common/widgets/spaces/space_with_profile_card.dart
gtalha07: This'd lead to null error. This validates it... 
```
showParents && (parents?.isEmpty ?? false)
```

app/lib/common/providers/room_providers.dart
```diff
@@ -198,6 +199,70 @@ final canonicalParentProvider = FutureProvider.autoDispose
...
+      final parentSpace = await ref
+          .watch(maybeSpaceProvider(parent.roomId().toString()).future);
+      if (parentSpace == null) {
+        // not found, skipping it
+        continue;
+      }
+      final profile =
+          await ref.watch(spaceProfileDataProvider(parentSpace).future);
+      final SpaceWithProfileData data = (space: parentSpace, profile: profile);
+      parentList.add(data);
```
gnunicorn, line 231: this logic is still flawed for the same reason I brought up in the last review...

gnunicorn (approved):

PR #1806, opened 2024-06-03
Visibility and Accessibility settings for spaces

kumarpalsinh25:
- Implementation UI of Visibility and Accessibility settings for Space via ```joinRuleStr```. Can be set ```Private```, ```Public``` and ```Limited```.
- Implemented methods for linking/de-linking spaces under ```Limited``` Visibility flag for filtering space content.
- Minor UI fix: space title overflow in space settings.
**_(Can be best reviewed per commit)_**

### Preview:
1. Selecting Visibility parameter along with space overview chip updates.

https://github.com/acterglobal/a3/assets/68579938/44bbc134-b45c-4ed1-87c1-5a15afb86a71

2. Selecting spaces to link under ```Limited``` Visibility flag.

https://github.com/acterglobal/a3/assets/68579938/7f7f3a57-ce00-4f7c-8f82-b7fe8b6d2198


app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -174,13 +221,25 @@ class _VisibilityAccessibilityPageState
     );
   }
 
-  void removeSpace(SpaceItem spaceItem) {
+  Future<void> removeSpace(Space? space, SpaceItem spaceItem) async {
+    final room =
+        await ref.read(maybeRoomProvider(space!.getRoomIdStr()).future);
+    if (room != null) {
+      if (await space.isChildSpaceOf(spaceItem.roomId)) {
+        await room.removeParentRoom(spaceItem.roomId, null);
```
kumarpalsinh25, line 229: So addition and removal of spaces for Limited permission is all about adding child and removing child?

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -202,4 +266,45 @@ class _VisibilityAccessibilityPageState
...
+          break;
+        case RoomVisibility.SpaceVisible:
+          update.joinRule('restricted');
+          if (roomId != null) {
+            if (!(await space!.isChildSpaceOf(roomId))) {
+              await room!.addParentRoom(roomId, true);
+              break;
+            } else {
+              await space.addChildRoom(roomId);
+              break;
```
kumarpalsinh25, line 296: Same as above comment, limited permission is all about adding child and removing child?

kumarpalsinh25 (commented):
I can see videos of `Visibility and Accessibility settings` of spaces, can you also share videos of Chat/Group also for this settings?

Also share video of creation flow of space and chat with this settings.

emilvincentz:
Great work, here is my feedback:

### Proposed changes:
Change the sub description to: 

Private (invite only)
**Only invited people can join.**

Public
**Anyone can find and join.**

Limited
**Anyone in selected spaces can find and join.** 



### Question:

- How does this look like for chats? 

- Is Access and nesting the same? (Referring to 00:30 in the second video)
What I mean, is that what if Space A is private, but then someone create a subspace under space A (called space B) will the settings for Space A then auto switch to “limited”? 
Further of the admins then want to make Space B private (invite only), does this mean that Space B no longer shows as a subspace / is nested under space A?

- Do we still need to show the small red and blue icons next to the chip update?

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -0,0 +1,310 @@
...
+            }
+          }
+          break;
+      }
+
+      await room!.setJoinRule(update);
+      EasyLoading.dismiss();
```
gnunicorn, line 303: Ideally as said above the spacesList is coming from a provider and thus when updated we don't actually mutate local state but are refreshed once that comes down the pipe ...

app/lib/features/spaces/pages/create_space_page.dart
```diff
@@ -38,6 +45,32 @@ class _CreateSpacePageConsumerState extends ConsumerState<CreateSpacePage> {
...
+      ref.listenManual(selectedSpaceIdProvider, (previous, next) {
+        if (next != null) {
+          ref
+              .read(_selectedVisibilityProvider.notifier)
+              .update((state) => RoomVisibility.SpaceVisible);
+        } else {
+          ref
+              .read(_selectedVisibilityProvider.notifier)
+              .update((state) => RoomVisibility.Private);
+        }
```
gnunicorn, line 72: as above.
```suggestion
          ref
              .read(_selectedVisibilityProvider.notifier)
              .update((state) => next != null ? RoomVisibility.SpaceVisible : RoomVisibility.Private);
```

app/lib/features/spaces/pages/create_space_page.dart
```diff
@@ -176,6 +211,76 @@ class _CreateSpacePageConsumerState extends ConsumerState<CreateSpacePage> {
...
+          subtitle: L10n.of(context).limitedVisibilitySubtitle,
+          isShowRadio: false,
+        );
+      default:
+        return RoomVisibilityItem(
+          iconData: Icons.lock,
+          title: L10n.of(context).private,
+          subtitle: L10n.of(context).privateVisibilitySubtitle,
+          isShowRadio: false,
+        );
```
gnunicorn, line 280: as said above we probably want to show an error here as there are some `knock_...` we do not yet support...

app/lib/l10n/app_en.arb
```diff
@@ -2166,6 +2166,17 @@
...
+  "visibilityTitle": "Visibility",
+  "visibilitySubtitle": "Select who can join this space.",
+  "visibilityNoPermission": "You don't have necessary permissions to change this space visiblity",
+  "public": "Public",
+  "publicVisibilitySubtitle": "Open space for anyone, best for Invite only, best for yourself or teams communities",
+  "privateVisibilitySubtitle": "Invite only, best for yourself or teams",
+  "limited": "Limited",
+  "limitedVisibilitySubtitle": "Anyone in current space and selected parent space can find and join",
+  "visibilityAndAccessibility": "Visibility and Accessibility",
+  "spaceWithAccess": "Space with access",
```
gnunicorn, line 2179: as per @emilvincentz's request for changes.

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -0,0 +1,310 @@
...
+        case RoomVisibility.SpaceVisible:
+          update.joinRule('restricted');
+          if (roomId != null) {
+            if (!(await space!.isChildSpaceOf(roomId))) {
+              await room!.addParentRoom(roomId, true);
+              break;
+            } else {
+              await space.addChildRoom(roomId);
+              break;
+            }
```
gnunicorn, line 297: As per comment above. When changing the list of spaces for the join rule we must provide the entire new list (of each room) through the update builder. And we do not want to touch the parent-child-relationship, as we might not even have the permissions to do so.

so, assuming we are passed a `newSpaceIds`, which has already been properly updated prior to this function being called, we should instead:
```suggestion
            for (final roomId in (newSpaceIds ?? [])) {
              update.addRoom(roomId);
            }
```

And if we now also did the same for removing: just took the room out of that list and submitted it rh

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -0,0 +1,310 @@
...
+  Future<void> updateSpaceVisibility(
+    RoomVisibility value,
+    String spaceId, {
+    String? roomId,
+  }) async {
+    try {
+      EasyLoading.show(status: 'Updating space settings', dismissOnTap: false);
+      final sdk = await ref.read(sdkProvider.future);
+      final room = await ref.read(maybeRoomProvider(spaceId).future);
+      final space = await ref.read(maybeSpaceProvider(spaceId).future);
```
gnunicorn, line 279: See below for why these are needed:
```suggestion
  Future<void> updateSpaceVisibility(
    RoomVisibility value, {
    List<String>? newSpaceIds,
  }) async {
    try {
      EasyLoading.show(status: 'Updating space settings', dismissOnTap: false);
      final sdk = await ref.read(sdkProvider.future);
```

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  Future<void> removeSpace(Space? space, SpaceItem spaceItem) async {
+    final room =
+        await ref.read(maybeRoomProvider(space!.getRoomIdStr()).future);
+    if (room != null) {
+      if (await space.isChildSpaceOf(spaceItem.roomId)) {
+        await room.removeParentRoom(spaceItem.roomId, null);
+      } else {
+        await space.removeChildRoom(spaceItem.roomId, null);
```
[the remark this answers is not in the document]
gnunicorn: or we just remove the item from the list locally and reuse the update function as outlined in the comment below.

gnunicorn (changes requested):
It looks good, but unfortunately there are a few mistakes in here that need to be addressed. While one is a comment on design and thus taste, the way the list of spaces under `restricted` is being updated is wrong and needs to be changed (I have written that up pretty extensively, if you need more hints or want me to do it, ping me). The other comments are also doing (slightly) incorrect things but if you want this in this release, could be overlooked _if_ there are corresponding github issues created outlining them. But the restricted-rooms-list- update _must_ be changed before this can be merged.

gnunicorn:
I've updated the PR to resolve the comments I had left before. In particular the updates are:
 - use a proper provider for the restricted roomIds and render a widget per each, remove local state generation
 - update the list of restricted rooms rather than the child relationship
 - more muted visibility chip
 - additional permission check for this specific type
 - disabled UI if the permissions are lacking
 - show the rooms even if we are lacking the permissions to change them (unreported bug)
 - generalize the providers and the widget to be reusable for chat rooms, too
 - use it for chat rooms, too
 - update the texts in according to @emilvincentz's suggestion

See it in action:


https://github.com/acterglobal/a3/assets/40496/ac278597-e8d7-414e-afa1-68a76c4ee473

gnunicorn (approved):
acceptable as per my changes.

PR #1812, opened 2024-06-04
Bookmarking spaces

gnunicorn:
This adds bookmarking support for spaces:

https://github.com/acterglobal/a3/assets/40496/43c0f08b-11c5-4276-9507-952528dd18eb

If you bookmarked any spaces, your space listing on the home pages will only show those. You can toggle the bookmark of any spaces on its space page and that is directly reflected.

Additionally this PR does:
 - make the "show all N spaces" to a less intrusive acter text inline button
 - renames the convo favoriting into bookmarking (first commit)
 - fixes a context.go issue on the space-listings page (which would lead to missing the `<-` button, as still visible in the video above)

### DIFF
--- a/app/lib/features/home/widgets/my_spaces_section.dart
+++ b/app/lib/features/home/widgets/my_spaces_section.dart
@@ -1,11 +1,13 @@
 import 'package:acter/common/providers/space_providers.dart';
 import 'package:acter/common/themes/app_theme.dart';
+import 'package:acter/common/toolkit/buttons/inline_text_button.dart';
 
 import 'package:acter/common/toolkit/buttons/primary_action_button.dart';
 import 'package:acter/common/tutorial_dialogs/space_overview_tutorials/create_or_join_space_tutorials.dart';
 import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/widgets/spaces/space_card.dart';
 import 'package:acter/features/home/data/keys.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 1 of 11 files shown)


app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -18,68 +20,122 @@ class MySpacesSection extends ConsumerWidget {
...
+    if (spaces.isEmpty) {
+      return const _NoSpacesWidget();
+    }
 
     int spacesLimit =
         (limit != null && spaces.length > limit!) ? limit! : spaces.length;
+    return _RenderSpacesSection(
+      spaces: spaces,
+      limit: spacesLimit,
+      showAll: spacesLimit != spaces.length,
```
kumarpalsinh25, line 53:
```suggestion
      showAll: spacesLimit <= spaces.length,
```

app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -18,68 +20,122 @@ class MySpacesSection extends ConsumerWidget {
...
+      showActions: spacesLimit == spaces.length,
```
kumarpalsinh25, line 54:
```suggestion
      showActions: spacesLimit <= spaces.length,
```

app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -18,68 +20,122 @@ class MySpacesSection extends ConsumerWidget {
...
+  }
+}
+
+class _RenderSpacesSection extends ConsumerWidget {
+  final int limit;
+  final List<Space> spaces;
+  final Widget title;
+  final bool showAll;
+  final bool showActions;
+  final int showAllCounter;
```
kumarpalsinh25, line 74: Instead of `showAll` and `showActions` param, we can use param like `bookmarkOnly` bool param. 

And also `showAllCounter` may not required because we can get it from spaces list directly which we are already passing in param.

app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -18,68 +20,122 @@ class MySpacesSection extends ConsumerWidget {
...
+                  },
+                  child: Text(
+                    L10n.of(context).seeAllMySpaces(showAllCounter),
                   ),
-                ],
+                ),
               ),
+            ],
+          ),
+        if (showActions)
```
kumarpalsinh25, line 116: As per my above comment, this can be manage by `bookmarkOnly` param.

app/lib/features/home/widgets/my_spaces_section.dart
```diff
@@ -18,68 +20,122 @@ class MySpacesSection extends ConsumerWidget {
...
-                              ),
-                              const SizedBox(height: 10),
-                              ActerPrimaryActionButton(
-                                onPressed: () => context.pushNamed(
-                                    Routes.searchPublicDirectory.name,),
-                                child: Text(L10n.of(context).joinSpace),
-                              ),
-                            ],
-                          ),
+        if (showAll)
```
kumarpalsinh25, line 99: As per my above comment, this can be manage by `bookmarkOnly` param.

kumarpalsinh25 (approved):
Nicee Feature addition!
Flutter side code looks good to me. Added few minor suggestion.

app/lib/features/home/widgets/my_spaces_section.dart
gnunicorn: This isn't about the bookmarked ones but also if we have a longer total list than the one we are currently showing. I'd rather keep them explicit in their nature rather than imply that from context.

Issue #1808, opened 2024-06-04
Task Feature Refactoring


kumarpalsinh25:
Reference:
https://balsamiq.cloud/sj2u1jx/plw5law/rE1AC

- [x] Refactor Add Task List and Add Task Item Bottom Sheet design
- [x] Refactor View Task List and Task Items
- [x] Refactor Task List Details Page and Task Item Detail Page
- [x] Refactor Edit and Delete Task List and Item

PR #1813, opened 2024-06-04
Refactoring of Add Task List and Add Task Item

kumarpalsinh25:
Ref #1808

Reference VIdeo:

https://github.com/acterglobal/a3/assets/51526480/573ec1f8-3d48-47b0-8df4-bc31c6962326


app/lib/features/tasks/widgets/task_list_card.dart
```diff
@@ -134,19 +138,17 @@ class _TaskListCardState extends ConsumerState<TaskListCard> {
...
                     children.add(
-                      Padding(
+                      Container(
+                        alignment: Alignment.centerLeft,
                         padding: const EdgeInsets.symmetric(
                           horizontal: 24,
                           vertical: 8,
                         ),
-                        child: OutlinedButton(
+                        child: ActerInlineTextButton(
```
gnunicorn, line 147: :+1:

app/lib/features/tasks/sheets/create_update_task_item.dart
```diff
@@ -0,0 +1,228 @@
...
+            hintText: L10n.of(context).dueDate,
+            suffixIcon: IconButton(
+              onPressed: selectDueDate,
+              icon: const Icon(Icons.calendar_month),
+            ),
+          ),
+          onTap: selectDueDate,
+          controller: _taskDueDateController,
+        ),
+      ],
```
gnunicorn, line 173: when changing the tasks due date the most common way to think about it that something is due "today" or "tomorrow" (and patterns alike). That's why for changing the due-date we have a custom date picker that allows for those to easily be picked. I understand the want to not reuse that (rather ugly) widget but we need to keep that use case in mind. Maybe by adding two `ActerInlineTextButtons` for quick setting? (the following is not real applicable code, just to translate an idea)
```suggestion
        TextFormField(
          readOnly: true,
          decoration: InputDecoration(
            hintText: L10n.of(context).dueDate,
            suffixIcon: IconButton(
              onPressed: selectDueDate,
              icon: const Icon(Icons.calendar_month),
            ),
          ),
          onTap: selectDueDate,
          controller: _taskDueDateController,
        ),
        Row(children:[
          ActerInlineTextButton(child: Text(L10n.of(context).today, onPressed: () => setState(() => { selectedDate = DateTime.today() }), 
           ActerInlineTextButton(child: Text(L10n.of(context).tomorrow, onPressed: () => setState(() => { selectedDate = DateTime.tomorrow() }),
           ])
      ],
```

app/lib/features/tasks/sheets/create_update_task_item.dart
```diff
@@ -0,0 +1,228 @@
...
+    final date = await showDatePicker(
+      context: context,
+      initialDate: selectedDate,
+      firstDate: DateTime.now(),
+      lastDate: DateTime.now().addYears(1),
+    );
+    if (date == null || !mounted) return;
+    selectedDate = date;
+    _taskDueDateController.text = taskDueDateFormat(date);
+    setState(() {});
```
gnunicorn, line 187: seen this empty set-state pattern a few times, but shouldn't this rather be:
```suggestion
    setState(() {
      selectedDate = date;
      _taskDueDateController.text = taskDueDateFormat(date);
    });
```
?

app/lib/features/tasks/sheets/create_update_task_item.dart
```diff
@@ -0,0 +1,228 @@
...
+    final taskDraft = widget.taskList.taskBuilder();
+    taskDraft.title(_taskNameController.text);
+    if (_taskDescriptionController.text.isNotEmpty) {
+      taskDraft.descriptionText(_taskDescriptionController.text);
+    }
+    if (selectedDate != null) {
+      taskDraft.dueDate(
+        selectedDate!.year,
+        selectedDate!.month,
+        selectedDate!.day,
```
gnunicorn, line 209: these `!` shouldn't be necessary due to your check above.
```suggestion
        selectedDate.year,
        selectedDate.month,
        selectedDate.day,
```

app/lib/features/tasks/sheets/create_update_task_item.dart
```diff
@@ -0,0 +1,228 @@
...
+      );
+    }
+    try {
+      await taskDraft.send();
+      EasyLoading.dismiss();
+      if (!mounted) return;
+      if (widget.cancel != null) widget.cancel!();
```
gnunicorn, line 216: why are you calling `cancel` on a successful submission? that seems unexpected...

app/lib/features/tasks/sheets/create_update_task_list.dart
```diff
@@ -0,0 +1,199 @@
...
+            Text(
+              L10n.of(context).description,
+              style: Theme.of(context).textTheme.bodySmall,
+            ),
+            MdEditorWithPreview(
+              key: CreateUpdateTaskList.descKey,
+              onChanged: (String? value) {
+                _descriptionText.value = value ?? '';
+              },
+            ),
```
gnunicorn, line 140: we are phasing out the MDEditorWithPreview in favor of the AppFlowyEditor. We should use that one here instead.

app/lib/features/tasks/sheets/create_update_task_item.dart
```diff
@@ -0,0 +1,228 @@
...
+        ),
+        const SizedBox(height: 5),
+        TextFormField(
+          decoration: InputDecoration(
+            hintText: L10n.of(context).description,
+          ),
+          minLines: 4,
+          maxLines: 4,
+          controller: _taskDescriptionController,
+        ),
```
gnunicorn, line 147: isn't this a fully markup supporting field? We should use the appflowy editor here...

gnunicorn (approved):
Great work. Just a few usability improvements

app/lib/features/tasks/sheets/create_update_task_item.dart
kumarpalsinh25: Added options as below.

Screenshot 2024-06-05 at 6 35 29 PM

app/lib/features/tasks/sheets/create_update_task_item.dart
kumarpalsinh25: Need to keep it as it variable is global and giving syntax error.

Screenshot 2024-06-05 at 6 36 50 PM

app/lib/features/tasks/sheets/create_update_task_item.dart
kumarpalsinh25: This is to give call back to outer screen and close inline input box.

app/lib/features/tasks/sheets/create_update_task_list.dart
kumarpalsinh25: Can we do this on separate PR as I first I would like to focus on Refactoring task UI-UX to quickly get out of lab feature and then improve individual smaller pieces on quick PRs.

Side-note:
- Not sure but probably also need to replace `taskListDraft.descriptionMarkdown` with `taskListDraft.descriptionHtml(plaintext, htmltext)`.

app/lib/features/tasks/sheets/create_update_task_item.dart
kumarpalsinh25: Same as above.

app/lib/features/tasks/sheets/create_update_task_item.dart
gnunicorn: odd. looks like a false-positive of your IDE

PR #1826, opened 2024-06-07
Task Module : List page and Details Page Refactoring

kumarpalsinh25:
This PR contains several changes for the improvement of the task feature as shown in below reference video.

https://github.com/acterglobal/a3/assets/51526480/2bb4d451-8bf7-4493-8435-69340fe4a71f

Key points
- Task list UI improvement
- Task Item UI improvement
- Task list details page UI improvement
- Task item detail page UI improvement
- Get task feature out of lab (https://github.com/acterglobal/a3/pull/1826/commits/83b7d717fef86b77076e087babfdf4107eaa39b3)

Reference:
[#1808](https://github.com/acterglobal/a3/issues/1808)


gnunicorn:
UI Feedback:

I feel that the items on the home screen are somewhat of an improvement but leave much open:image

1. as a card with all the spaces, they are pretty large now, not really easy to quickly glance over - try it with like 5 or 10 items ... I am afraid this becomes a lot of stuff quickly ... but we might be able to condense it nicely enough (see below)
2. showing my own icon is kinda pointless as they are assigned to me ...
3. The task and List order is kinda confusing. What is what here? I'd put the list name inline with the due-date
4. it misses the space, which is important context.

Re 1, 3 and 4. I'd put them all into one line in the order of due, space name, listname (with the lattest being a text-inline-button to link to the list). For 2: I'd only show an icon if there is more than just myself assigned - then I'd show all assigned (including myself).

---

I find the task details list misleading:
image

There are three items in a list (which takes up a lot of spaces), that are fundamentally different and act different despite looking sort-of the same:
1. that list takes up a lot of space, especially compared to the information it contains
2. the `>` implies some navigation, which on the list you do get, but on the due-date it is an edit... that doesn't make sense IMHO.
3. Also navigating to the list doesn't really make sense in the first place, I think
4. adding a description has been removed to the edit only? well, okay, I guess... though I'd rather have that inline there

I'd suggest the following changes:
1. remove the listname from the list of items, put it right below the task title. It isn't interacted with the same way, it is just a metadata information at this point (like who created it and when would be - wondering if it makes sense to add these as well?)
2. remove the titles from the list. The icons and the value together are indicators enough. If they are empty add a muted quick action as an inline text button that explains its purpose (e.g. in case of the due date, add a `add due date` button as the title).
3. use the compact view for the lists
4. Muted actions ("remove myself") to make them take less attention. In general maybe using less bright colors for the entire texts here makes sense.
5. additional `add description` inline text button quick action above the list..

gnunicorn:
@kumarpalsinh25 ready for review?

kumarpalsinh25:
> @kumarpalsinh25 ready for review?

Yes, there are few know feedback are pending which I would to cover up in right after this PR. Just to not make this PR not so big and cover up feedback points in back-to-back later PRs.

gnunicorn:
@kumarpalsinh25 could you fix the merge conflicts before I do the review please?

kumarpalsinh25:
> @kumarpalsinh25 could you fix the merge conflicts before I do the review please?

Yes sure

kumarpalsinh25:
@gnunicorn 
Resolved conflicts and branch is rebased with main.

app/lib/features/settings/pages/labs_page.dart
```diff
@@ -66,16 +66,6 @@ class SettingsLabsPage extends ConsumerWidget {
             SettingsSection(
               title: Text(L10n.of(context).apps),
               tiles: [
-                SettingsTile.switchTile(
-                  key: tasksLabSwitch,
```
gnunicorn, line 70: hmpf.... I am not really happy about this approach of just removing the labs flag. IMHO this isn't even in the state to even have the labs-flag default-on yet, we are still missing several basics and there are still going to be major changes ... the point of the labs is to make clear that exactly that is the case to the user: default off for unstable features, default-on for features we consider complete and are just doing minor bug fixes on, removing the labs once we consider this stable. This is anything but stable yet.

gnunicorn (approved):
I have glanced over most of it and it seems fine, but I didn't do a too in-depth analysis. Only request I have is that this stay behind the default-off labs feature for now as this is clearly still unstable and worked on a lot. Taking it out of labs is clearly too early as this is way to immature.

app/lib/features/settings/pages/labs_page.dart
kumarpalsinh25: Yupp, agree with you. Let me roll-back it to lab feature again.

PR #1828, opened 2024-06-07
new due date picker via bottom drawer

gnunicorn:
Remove the ugly omni-datepicker and instead use a bottom drawer to allow for quick due actions and selecting a custom date by using the regular date picker.

- [ ] make unset work, too

https://github.com/acterglobal/a3/assets/40496/1ec7e431-1395-42aa-8ebc-6127453e4960


kumarpalsinh25 (approved):
Nice improvements! 

Minor suggestion: (Can be done later also)
- Postpone and Postpone 2 days can be combine menu with default 1 day selection. User can set number of days from his side baed on the his/her choice.

Issue #1838, opened 2024-06-18
Generalised the flow of the Change Title


kumarpalsinh25:
- Tap on any title to change it based on the permission
- Three dot menu option will also have Edit Title menu based on the permission

PR #1839, opened 2024-06-18
Generalised the flow of the Change Title

kumarpalsinh25:
- Tap on any title to change it based on the permission
- Three dot menu option will also have Edit Title menu based on the permission

fixes, #1838 

1) Space Name:

https://github.com/acterglobal/a3/assets/51526480/a73fed34-81dd-4e30-b02b-7223d680108f

2) Chat Name:

https://github.com/acterglobal/a3/assets/51526480/d4a5edb3-1d2f-4cf9-a377-6bebdfe87586

3) Event Name:

https://github.com/acterglobal/a3/assets/51526480/3f628aef-eb1d-49b1-b85d-f64e282e84e9

4) Pin Name:

https://github.com/acterglobal/a3/assets/51526480/4ee4b63d-9f9c-4779-ae8c-1a0ec696f865

5) Task List and Task Item Name:

https://github.com/acterglobal/a3/assets/51526480/79c66d25-6a8b-4f70-a361-0cd8489eadc6


app/lib/features/pins/providers/notifiers/edit_state_notifier.dart
```diff
@@ -41,6 +44,22 @@ class PinEditNotifier extends StateNotifier<PinEditState> {
 
   void setEditMode(bool editMode) => state = state.copyWith(editMode: editMode);
 
+  Future<void> editTitle(BuildContext context) async {
```
gnunicorn, line 47: arguably shouldn't be on the notifier, should it ?!? I prefer the other pattern where this just a function on the widget ... (feels like `BuildContext` are generally not to be used in notifiers ...)

app/lib/features/chat/pages/room_profile_page.dart
```diff
@@ -112,44 +143,48 @@ class _RoomProfilePageState extends ConsumerState<RoomProfilePage> {
...
+  void showEditNameBottomSheet(ProfileData? convoProfile) {
+    showEditTitleBottomSheet(
+      context: context,
+      bottomSheetTitle: L10n.of(context).editName,
+      titleValue: convoProfile?.displayName ?? '',
+      onSave: (newName) => _editName(newName),
+    );
+  }
+
+  Future<void> _editName(String newName) async {
```
gnunicorn, line 172:
```suggestion
      onSave: _saveName,
    );
  }

  Future<void> _saveName(String newName) async {
```

naming things...

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -251,11 +255,18 @@ class _EventDetailPageConsumerState extends ConsumerState<EventDetailPage> {
...
-                  calendarEvent.title(),
-                  maxLines: 2,
-                  overflow: TextOverflow.ellipsis,
-                  style: Theme.of(context).textTheme.titleMedium,
+                GestureDetector(
+                  onTap: () {
+                    if (canPostEvent) {
+                      showEditEventTitleBottomSheet(calendarEvent.title());
+                    }
+                  },
```
gnunicorn, line 263: considering that we also want longPress -> copy (next), shouldn't we use `InkWell` for these?

gnunicorn (approved):
Nice. thanks. Remarks but none is hard required before a merge.

app/lib/features/pins/providers/notifiers/edit_state_notifier.dart
kumarpalsinh25: Good notice. Updated code on latest commit. :)

app/lib/features/chat/pages/room_profile_page.dart
kumarpalsinh25: Updated naming things on latest commit as suggested.

app/lib/features/events/pages/event_details_page.dart
kumarpalsinh25: Probably it will create any issue. 
Let share one demo with below screen recording for reference. 


https://github.com/acterglobal/a3/assets/51526480/c239748f-e14d-4aa5-aa35-6c05d84d4592

PR #1840, opened 2024-06-18
Auto-update of room notification change

bitfriend:
`element` app changed notification setting of `#matrix-dev:matrix.org` room, but `acter` doesn't update that change automatically.

https://github.com/acterglobal/a3/assets/48786056/cf696914-188e-4c5b-8f28-330e98b48905

Will fix this issue.


bitfriend:
Okay, understood.
Then will discard this PR.

PR #1870, opened 2024-06-21
Activities screen shows unconfirmed Emails

### LINKED ISSUE #1861: When instering email at signup, its not clear that you have to straight away verify email, otherwhise link expires

gnunicorn:
Fixes #1861 through two measures:
1. A more forceful message upon the registration
2. An additional widget on the Activities screen to nudge users to confirm their email addresses

The latter is proven with a test as well:

https://github.com/acterglobal/a3/assets/40496/eb83711b-8a34-4855-b0f3-00a50ee1728c

### DIFF
--- a/app/lib/common/providers/common_providers.dart
+++ b/app/lib/common/providers/common_providers.dart
@@ -28,6 +28,18 @@ Future<ProfileData> getProfileData(Account account) async {
   return ProfileData(displayName.text(), avatar.data());
 }
 
+final genericUpdatesStream =
+    StreamProvider.family<int, String>((ref, key) async* {
+  final client = ref.watch(alwaysClientProvider);
+  int counter = 0; // to ensure the value updates
+
+  // ignore: unused_local_variable
+  await for (final value in client.subscribeStream(key)) {
+    yield counter;
+    counter += 1;
+  }
+});
+
 final myUserIdStrProvider = StateProvider(
   (ref) => ref.watch(
     alwaysClientProvider.select((client) => client.userId().toString()),
@@ -67,6 +79,8 @@ class EmailAddresses {
 
 final emailAddressesProvider = FutureProvider((ref) async {
   final client = ref.watch(alwaysClientProvider);
+  // ensure we are updated if the upgrade comes down the wire.
+  ref.watch(genericUpdatesStream('global.acter.dev.three_pid'));
   final threePidManager = client.threePidManager();
   final confirmed =
       asDartStringList(await threePidManager.confirmedEmailAddresses());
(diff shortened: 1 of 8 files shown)


app/lib/features/activities/pages/activities_page.dart
```diff
@@ -153,35 +154,51 @@ class ActivitiesPage extends ConsumerWidget {
...
 
     if (ref.watch(featuresProvider).isActive(LabsFeature.encryptionBackup)) {
       final backups = renderBackupSection(context, ref);
       if (backups != null) {
-        renderEmptyState = false;
         children.add(backups);
       }
     }
+    final hasUnconfirmedEmails = ref.watch(hasUnconfirmedEmailAddresses);
+    print('has unconfirmed addrs: $hasUnconfirmedEmails');
```
gtalha07, line 191: We probably don't want print lines and use logging...

app/lib/features/onboarding/pages/link_email_page.dart
```diff
@@ -105,6 +106,7 @@ class LinkEmailPage extends ConsumerWidget {
     EasyLoading.show(status: L10n.of(context).linkingEmailAddress);
     try {
       await manager.requestTokenViaEmail(emailController.text.trim());
+      ref.invalidate(emailAddressesProvider);
```
gtalha07, line 109: I believe we were on a way to refactor ```ref.invalidate``` in this specific case and use proper object updates underlying client. Can you brief why it's needed?

gtalha07 (commented):
Looks good generally. But there's a usage of ```ref.invalidate``` for data that resides on client. Needs to be addressed!

app/lib/features/onboarding/pages/link_email_page.dart
gnunicorn: This is needed. While we do want to get rid of `ref.invalidate` for all data that is coming through `sync`, the email address providing isn't managed through sync but has a manual request scheme it needs. Those need to be actively refreshed unfortunately and that is happening here.

app/lib/common/providers/common_providers.dart
```diff
@@ -67,6 +79,8 @@ class EmailAddresses {
 
 final emailAddressesProvider = FutureProvider((ref) async {
   final client = ref.watch(alwaysClientProvider);
+  // ensure we are updated if the upgrade comes down the wire.
+  ref.watch(genericUpdatesStream('global.acter.dev.three_pid'));
```
kumarpalsinh25, line 83: What is `global.acter.dev.three_pid` ? Is that matrix third-party id from acter home server?

kumarpalsinh25 (approved):
Useful Feature!

app/lib/common/providers/common_providers.dart
gnunicorn: No, it's the internal event name for where we store the email address. This isn't super great like this but something we can tackle once we have Emil support by dart-uniffi

Issue #1841, opened 2024-06-18
Completion of task module


kumarpalsinh25:
- Delete Task List & Task Item
- Report Task List & Task Item
- Inline due date 
- General UI-UX improvements

PR #1871, opened 2024-06-21
Task Feature : General UI-UX Improvements

kumarpalsinh25:
Ref #1841,

1) My Task : Home Screen
- Now we have compact view of my task item list
- Task List have better representation by leading icon

Screenshot 2024-06-22 at 12 24 48 AM

2) Task Item Detail Page:

https://github.com/acterglobal/a3/assets/51526480/362841e4-5ef8-4c93-81c6-c4d32ad8e902

### DIFF
--- a/app/lib/features/tasks/pages/task_item_detail_page.dart
+++ b/app/lib/features/tasks/pages/task_item_detail_page.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 
 import 'package:acter/common/providers/room_providers.dart';
 import 'package:acter/common/toolkit/buttons/inline_text_button.dart';
-import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/utils/utils.dart';
 import 'package:acter/common/widgets/edit_title_sheet.dart';
 import 'package:acter/common/widgets/edit_html_description_sheet.dart';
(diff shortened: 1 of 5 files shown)


app/lib/features/tasks/pages/task_item_detail_page.dart
```diff
@@ -353,11 +356,7 @@ class TaskItemDetailPage extends ConsumerWidget {
     return ListTile(
       dense: true,
       leading: const Icon(Atlas.business_man_thin),
-      title: Text(
-        L10n.of(context).assignment,
-        style: Theme.of(context).textTheme.bodyMedium,
-      ),
-      subtitle: task.isAssignedToMe()
+      title: task.isAssignedToMe()
```
gnunicorn, line 359: Something I might have overlooked before but whey do we not show the assigned if it isn't the user themselves? This check seems wrong...

gnunicorn (approved):
Lgtm but one thing looks like a bug that was already in even before

app/lib/features/tasks/pages/task_item_detail_page.dart
kumarpalsinh25: Ahh I see,
I have did required changes in the task details page as below.

Screenshot 2024-06-25 at 10 09 49 AM

kumarpalsinh25: How ever I find some additional issue on assignment which I reported here.
https://github.com/acterglobal/a3/issues/1877

Issue #1760, opened 2024-05-14
Not able to scroll up or down on invite code generator


emilvincentz:
https://github.com/acterglobal/a3/assets/77483554/c64389de-b839-490a-8eca-862296e6aa17

PR #1876, opened 2024-06-24
Invite code improvements

### LINKED ISSUE #1874: Invite Code improvements
- [ ] autogenerated code using space display name as prefix (lowerCased?) and shorten
 - [ ] link to manage the token from the invite box

gnunicorn:
Fixes #1874, #1760 .

Generate the automatic codes based on the room display name + a few random numbers to make it easier to use. Replace the inactivation feature with a `Manage` link:

https://github.com/acterglobal/a3/assets/40496/26875585-c5f8-46a5-95c7-14207959f015

Technically, this additionally a) refactored the token generation part into a separate widget; b) updates the super invites update widget to fix several UX glitches in there (but only make it pretty to a very limited degree).


app/lib/features/invite_members/widgets/invite_code_ui.dart
```diff
@@ -0,0 +1,142 @@
...
+
+      int end = 5;
+      if (prefix.isEmpty) {
+        end = 8;
+      } else if (prefix.length > 8) {
+        prefix = prefix.substring(0, 8);
+        end = 3;
+      } else if (prefix.length > 4) {
+        end = 3;
+      }
```
kumarpalsinh25, line 118: How about putting this logic on new code generation provider?

app/lib/features/invite_members/widgets/invite_code_ui.dart
```diff
@@ -0,0 +1,142 @@
...
+            ),
+          ),
+          const SizedBox(height: 10),
+          Align(
+            alignment: Alignment.centerRight,
+            child: ActerInlineTextButton(
+              onPressed: () async {
+                final token =
+                    await ref.read(superInviteTokenProvider(inviteCode).future);
+                if (!context.mounted) return;
```
kumarpalsinh25, line 65: Not quit understand the usage of 
`final token = await ref.read(superInviteTokenProvider(inviteCode).future);`

Can I have more information about this?

kumarpalsinh25 (approved):
Looks good to me.

app/lib/features/invite_members/widgets/invite_code_ui.dart
gnunicorn: needed for routing. the target expect `extra: SuperInvitesToken` at the moment and I didn't want to refactor routing just for that at the moment (though it is bad routing). Instead I fetch the token (we already have anyways) from the provider via its code and pass it along.

kumarpalsinh25: `inviteCode` isn't work with route param?

app/lib/features/invite_members/widgets/invite_code_ui.dart
gnunicorn: I opted for keeping the code generator dumb for now as given that it can have multiple rooms entered and doesn't (at this point) actually mess with any refs or something like it, I'd like to see more usage patterns before generalizing this. But yeah, if we see this taking place at more than this location we should probably move it out.

PR #1878, opened 2024-06-25
remove ```.autoDispose``` modifier from parent avatar providers to avoid avatar flickering

gtalha07:
- Removes ```.autoDispose``` modifier from ```parentAvatarInfosProvider``` and its internal providers as a fix for constant avatar flickering and updates.

Preview:

(Before Avatar Flickering)

https://github.com/acterglobal/a3/assets/68579938/af7cc4dd-3c2b-43ed-a54f-470e64108a7d

(After fix)

https://github.com/acterglobal/a3/assets/68579938/5697c3f5-74fd-41d0-8e14-a58a4565df88

### DIFF
--- a/app/lib/common/providers/room_providers.dart
+++ b/app/lib/common/providers/room_providers.dart
@@ -164,8 +164,8 @@ final roomSearchedChatsProvider =
 
 /// If the room exists, this returns its space relations
 /// Stays up to date with underlying client data if a room was found.
-final spaceRelationsProvider = FutureProvider.autoDispose
-    .family<SpaceRelations?, String>((ref, roomId) async {
+final spaceRelationsProvider =
+    FutureProvider.family<SpaceRelations?, String>((ref, roomId) async {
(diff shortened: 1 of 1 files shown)


app/lib/common/providers/room_providers.dart
```diff
@@ -259,7 +259,7 @@ final _roomAvatarProvider = FutureProvider.autoDispose
 
 /// Provide the AvatarInfo for each room. Update internally accordingly
 final roomAvatarInfoProvider =
-    StateProvider.autoDispose.family<AvatarInfo, String>((ref, roomId) {
+    StateProvider.family<AvatarInfo, String>((ref, roomId) {
```
gnunicorn, line 262: side note that this doesn't need to be a `State` Provider either:
```suggestion
    Provider.family<AvatarInfo, String>((ref, roomId) {
```

gnunicorn (approved):
lgtm

Issue #1788, opened 2024-05-28
Os Error: too many files open + Error sending Request


emilvincentz:
The app has started to crash + not loading spaces or images in chat properly + bug report also not working.

![Image](https://github.com/acterglobal/a3/assets/77483554/cee629b3-7fc8-4ea3-b65e-f35984e3a99d)

![Image](https://github.com/acterglobal/a3/assets/77483554/79665d0d-3ab5-4624-9b11-15602d7b5949)

![Image](https://github.com/acterglobal/a3/assets/77483554/8de1ea51-f4dc-4d88-99a6-01ddcb601940)

emilvincentz:
Might also be connected to this loading errror:


![Image](https://github.com/acterglobal/a3/assets/77483554/7574e368-e4c8-4466-910f-c294ef602272)

emilvincentz:
maybe related to this? 
error loading image in chat #1261

gnunicorn:
looks like we are leaking some file descriptor(s) ... there is plenty of reasons why that could be and maybe we just want to "resolves" this problem for now by reacting on that particular error code with prompting the user to restart the app ... it is hard for us to investigate that at the moment ..

emilvincentz:
This now happens all the time and doesn't help with restarting the app

emilvincentz:
![image](https://github.com/acterglobal/a3/assets/77483554/0bb32465-4d60-4587-9618-8767c7b4d5ee)

![image](https://github.com/acterglobal/a3/assets/77483554/8960b3a8-aae9-4315-86cc-e3b5a934ba0e)

emilvincentz:
well the last one Im not 100% sure of. All the loading error just started appearing at ones, along with the crashes and fail to bug report too, so thought they might be all connected, but maybe not

emilvincentz:
It seems that they somehow are related. At least the two main messages that keeps appearing is the "Error sending request" and "Too many files open".
Below you can see that error messages changes between these two error messages, compared to the ones above.
Also when trying to bug report and error message shows, the app freezes.



![Image](https://github.com/acterglobal/a3/assets/77483554/e5e1f284-2a89-4f18-bfc5-d258416e953e)
![Image](https://github.com/acterglobal/a3/assets/77483554/931dc72e-7e3e-49ff-bf28-5d8835a47086)
![Image](https://github.com/acterglobal/a3/assets/77483554/f436b653-d0ea-4506-ab4f-b31b5289fec0)

[...]

PR #1880, opened 2024-06-25
Limited Media Cache

gnunicorn:
New MediaCache Wrapper Limiting Queue to ensure we do not run into too-many-files-open-errors, like in #1788 . Limits concurrent file accesses to 50 on iOS and 100 for everyone else.


gtalha07 (approved):
Looks good to me!!

PR #1887, opened 2024-06-26
Take Tasks out of the LabsFeature

gnunicorn:
Noticed that the Dashboard was missing my Tasks but also still had the task-list feature in the main navigation. Decided to drop the charade and just take the task feature out of labs. Removed all instances of it being tested for and cleaned the code accordingly. No other changes.


gtalha07 (approved):
lgtm!

PR #1889, opened 2024-06-26
set macos minimum deployment target

### LINKED ISSUE #1888: Support MACOSX_DEPLOYMENT_TARGET=12
MacOS 12 is still supported by Apple, so we want to support it. However, the Apple Store build says we are only supporting MacOS 13.1 and above. Which is odd, as I can't find that no where defined in our code base, there everything goes to 10.14 or 12 as the minimum... Either way , we _do_ want to support MacOS 12.

gtalha07:
- Fixes #1888. Sets the minimum deployment target of MacOS upto 10.14.


gnunicorn (approved):

gnunicorn:
Let's see if that allows us to release for older mac ...

PR #1926, opened 2024-07-08
Generic pins list page

kumarpalsinh25:
Big improvements in pin list pages as below.

- Generic Pin List page (App Pin List and Space Pin List)
- Search bar option in pin list
- Pins providers clean-up and optimisations
- General code clean-up

Reference Video:

https://github.com/acterglobal/a3/assets/51526480/17e66ab8-1e3f-4162-bc37-6878e95880c4


app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+    } else {
+      return const SizedBox.shrink();
+    }
+  }
+
+  Widget _buildBody() {
+    AsyncValue<List<ActerPin>> pinList;
+
+    final searchValue = ref.watch(searchValueProvider) ?? '';
+    hasSearchTerm = searchValue.isNotEmpty;
```
gnunicorn, line 84: let's not have side-effects in the render-function please.

also ... why not just make the bool a getter function that just evaluates? like `get hasSearchTerm -> (ref.watch(searchValueProvider) ?? '').isNotEmpty`

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        _buildSearchBar(),
+        Expanded(
+          child: pinList.when(
+            data: (pins) => _buildPinsList(pins),
+            error: (error, stack) =>
+                Center(child: Text(L10n.of(context).loadingFailed(error))),
+            loading: () => Center(child: Text(L10n.of(context).loading)),
```
gnunicorn, line 105: Skeleton pls?!?

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+        ],
+      ),
+    );
+  }
+
+  Widget _buildPinsEmptyState() {
+    bool canAdd = true;
+    if (widget.spaceId != null && !hasSearchTerm) {
+      final membership = ref.watch(roomMembershipProvider(widget.spaceId!));
+      canAdd = membership.requireValue!.canString('CanPostPin');
```
gnunicorn, line 165:
```suggestion
      canAdd = membership.valueOrNull?.canString('CanPostPin') == true;
```

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+    return Center(
+      heightFactor: 1,
+      child: EmptyState(
+        title: hasSearchTerm
+            ? L10n.of(context).noMatchingPinsFound
+            : L10n.of(context).noPinsAvailableYet,
+        subtitle: L10n.of(context).noPinsAvailableDescription,
+        image: 'assets/images/empty_pin.svg',
+        primaryButton: canAdd && !hasSearchTerm
+            ? ActerPrimaryActionButton(
```
gnunicorn, line 176: should use the same `hasSpaceWithPermissionProvider`

gnunicorn (approved):
Nice one.

just a few things before merge: please replace the `requireValue!` cases (I provided suggestions), make sure that we only show the `+` button(s) in case `hasSpaceWithPermissionProvider` indicates there are actually spaces to create pins within and let's remove that `hasSearchTerm = ...`-side-effect of the render function.

The other remarks change at your own peril.

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -47,25 +26,41 @@ class AsyncPinNotifier
...
+    return await _getPinList(client);
+  }
+
+  Future<List<ActerPin>> _getPinList(Client client) async {
+    //GET ALL PINS
+    if (arg == null) {
+      return (await client.pins()).toList(); // this might throw internally
+    }
+    //GET SPACE PINS
+    else {
```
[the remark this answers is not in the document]
kumarpalsinh25: Will do this.

app/lib/features/pins/providers/notifiers/pins_notifiers.dart
```diff
@@ -47,25 +26,41 @@ class AsyncPinNotifier
...
+  Future<List<ActerPin>> build(String? arg) async {
     final client = ref.watch(alwaysClientProvider);
-    final spaceId = arg.getRoomId();
-    _listener =
-        client.subscribeStream('$spaceId::pins'); // keep it resident in memory
+
+    //GET ALL PINS
+    if (arg == null) {
+      _listener = client.subscribeStream('pins');
+    }
```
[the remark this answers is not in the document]
kumarpalsinh25: Wanted to generalise list provider as many of the things are same in code. Also purpose is the same like if we pass the spaceId then it will give list data based space else all pins list.

Keeping this for now, but happy to separate this if you thing that this is not so good idea.

app/lib/common/widgets/add_button_with_can_permission.dart
```diff
@@ -1,33 +1,45 @@
...
-        color: Theme.of(context).colorScheme.surface,
-        onPressed: onPressed,
+    if (canPermissionParam == null) return _buildIconButton(context);
+
+    final membership =
+        ref.watch(roomMembershipProvider(canPermissionParam!.spaceId));
+
+    bool canAdd = membership.requireValue!.canString(
+      canPermissionParam!.canString,
+    );
```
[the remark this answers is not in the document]
kumarpalsinh25: Sorry, my bad! You already mentioned about his earlier also. Will do the needful changes.

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+        children: [
+          Text(L10n.of(context).pins),
+          _buildSpaceName(),
+        ],
+      ),
+      actions: [
+        AddButtonWithCanPermission(
+          canPermissionParam: widget.spaceId != null
+              ? (spaceId: widget.spaceId!, canString: 'CanPostPin')
+              : null,
```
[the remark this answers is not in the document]
kumarpalsinh25: Gotcha! Will do the needful here. Good notice! Thanks.

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+          onPressed: () => context.pushNamed(
+            Routes.actionAddPin.name,
+            queryParameters:
+                widget.spaceId != null ? {'spaceId': widget.spaceId} : {},
```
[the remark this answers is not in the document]
kumarpalsinh25: This is the error it is showing when I do don't pass empty queryParams.
https://github.com/acterglobal/a3/assets/51526480/985e03df-7528-408e-b584-4e22a945a0b0

Still let me cross check entire param hierarchy and try to fix above error without passing empty queryParam.

app/lib/features/pins/pages/pins_list_page.dart
```diff
@@ -0,0 +1,188 @@
...
+
+  AppBar _buildAppBar() {
+    return AppBar(
+      centerTitle: false,
+      title: Column(
+        mainAxisAlignment: MainAxisAlignment.start,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text(L10n.of(context).pins),
+          _buildSpaceName(),
```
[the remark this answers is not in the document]
kumarpalsinh25: Ok sure, let me do needful changes.

app/lib/features/pins/pages/pins_list_page.dart
kumarpalsinh25: >  have side-effects in the render-function please

Sorry but getting your this point, can you explain it bit more?

> why not just make the bool a getter function that just evaluates

Yes, that's good idea.

app/lib/features/pins/pages/pins_list_page.dart
kumarpalsinh25: Yes, for sure.

app/lib/features/pins/pages/pins_list_page.dart
kumarpalsinh25: Sure

app/lib/features/pins/pages/pins_list_page.dart
gnunicorn: > Sorry but getting your this point, can you explain it bit more?

The render function is meant to be executed because of state changes. Thus, it should _not_ change state while executing. Setting variables that are then set the same way for longer than the render execution takes (e.g. in this case object variables) is such a state change that shouldn't be happening within the rendering function. 

Reasoning behind that is that you want to have a clean `event -> state change -> render` path but if the state can be, and is, modified by the `render` you make an unclean internal mismatch which makes it harder to reason about the code and when thing change or what state things are in at a certain point in time. To avoid such debugging difficulties, `render` functions are expected to be _side effect free_, as-in, they don't change the state, they just render it.

gnunicorn: (which is the paradigm that flutter generally puts forward and that riverpod made even more stringent. so if that happens _now_ it is most certainly just bad style ;) )

app/lib/features/pins/pages/pins_list_page.dart
kumarpalsinh25: Done!

Screenshot 2024-07-10 at 9 34 48 PM

app/lib/features/pins/pages/pins_list_page.dart
kumarpalsinh25: Thanks @gnunicorn for bit of explain but still not sure I am fully understanding this thing.
However I did some changes in this commit. Can you please check and confirm if it is good or not.
https://github.com/acterglobal/a3/pull/1926/commits/8c8f42dde527740405e7053d6ad368cacdb11358

app/lib/features/pins/pages/pins_list_page.dart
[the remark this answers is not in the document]
kumarpalsinh25: Fixed above issue by below commit.
https://github.com/acterglobal/a3/pull/1926/commits/0a2e1de9f5b867dd751229179337594fa4690dac

app/lib/features/pins/pages/pins_list_page.dart
gnunicorn: yes, this is all good. in essence: do not change object variables while in `render`. your update does that now by removing that variable entirely.

kumarpalsinh25: Ahh, now I got your point. Thanks.

PR #1931, opened 2024-07-09
refactor chat mentions widget

gtalha07:
This attempts to remap the existing ```flutter_mentions``` widget to ```auto_complete_trigger``` along with reactive updates to avatars by cleaning up mention providers and instead directly feed ```avatarInfo``` to keep up-to-date with underlying avatar data, essentially optimising the list result even if its big .i.e. we're talking over +100 members.

The reason for switching this plugin to a more simplified version ```auto_complete_trigger``` is that it directly nests the ```TextEditingController```, giving us more autonomy and thereby and more control over tweaking stuff behind where ```flutter_tagger``` lacks. 

### What are the next steps??
To keep PR short and concise for quick review, this doesn't bring any changes or improvements to current mention UI. However these course of features are set to be iterated on our version of ```auto_complete_trigger```.

- Bringing styles on trigger text .i.e. overriding ```TextEditingController``` to make `TextSpans` on demand.
- Up-down arrows functionality to select items and hit enter to add it (Desktop).
- Hashtag trigger for room mentions / Emoji text triggers.

Here's a preview which shows how the list reactively updates it underlying ```AvatarInfo``` data upon mention search.

https://github.com/user-attachments/assets/925628f0-21ab-4de4-b7ca-f77ea4c8e6ea

### DIFF
--- a/.changes/1853-chat-mentions.md
+++ b/.changes/1853-chat-mentions.md
@@ -1,1 +1,2 @@
 - Chat mentions are much more optimized and efficient in bringing up users even on large rooms.
+- Swipe-to-reply/edit feature is now disabled on deleted/encrypted messages as intended with chat message actions.
--- a/app/lib/features/chat/widgets/mention_profile_builder.dart
+++ b/app/lib/features/chat/widgets/mention_profile_builder.dart
@@ -33,7 +33,7 @@ class MentionProfileBuilder extends ConsumerWidget {
       ),
       error: (error, st) => ErrorWidget(L10n.of(context).failedToLoad(error)),
       data: (data) {
-        final users = data.fold<Map<String, String?>>({}, (map, uId) {
+        final users = data.fold<Map<String, String>>({}, (map, uId) {
           if (uId != userId) {
             final displayName = ref
                 .watch(
@@ -76,9 +76,9 @@ class MentionProfileBuilder extends ConsumerWidget {
                     final autocomplete = MultiTriggerAutocomplete.of(ctx);
                     ref
                         .read(chatInputProvider.notifier)
-                        .addMention(displayName ?? '', id);
+                        .addMention(displayName, id);
                     return autocomplete.acceptAutocompleteOption(
-                      displayName ?? id.substring(1),
+                      displayName.isNotEmpty ? displayName : id.substring(1),
                     );
                   },
                   leading: Consumer(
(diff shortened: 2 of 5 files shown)


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -1040,6 +898,126 @@ class _TextInputWidget extends ConsumerWidget {
       chatInputNotifier.emojiPickerVisible(false);
     }
   }
+
+  @override
+  Widget build(BuildContext context) {
+    ref.listen(chatInputProvider(widget.roomId), (prev, next) {
```
gnunicorn, line 904: this was mostly a move, correct?

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -1040,6 +898,126 @@ class _TextInputWidget extends ConsumerWidget {
...
+        chatFocus.requestFocus();
+      }
+    });
+    return CallbackShortcuts(
+      bindings: <ShortcutActivator, VoidCallback>{
+        const SingleActivator(LogicalKeyboardKey.enter): () {
+          widget.onSendButtonPressed();
+        },
+      },
+      child: MultiTriggerAutocomplete(
```
gnunicorn, line 924: only this part actually update, am I right?

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -54,19 +57,19 @@ class BubbleBuilder extends ConsumerWidget {
         isMemberEvent
             ? child
             : SwipeTo(
-                onRightSwipe: (DragUpdateDetails details) {
-                  ref
-                      .read(chatInputProvider(roomId).notifier)
-                      .setReplyToMessage(message);
-                },
+                onRightSwipe: redactedOrEncrypted
```
gnunicorn, line 60: you should probably also mention this new feature in the description...

gnunicorn (approved):
Looking good. But some improvements in the usage of the consumers ... though it stands to reason, this can be applied in a follow-up-PR soon, too.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: Yep.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: Yes indeed.

PR #1932, opened 2024-07-09
fix chat list ordering

gtalha07:
This checks up the sorting on ```ConvoDiff``` and re-arranges based on timestamp both for chats and space related chats.

### Before:

Screenshot 2024-07-09 at 11 54 20 AM

### After:

Screenshot 2024-07-09 at 12 07 24 PM

### DIFF
--- a/app/lib/common/providers/chat_providers.dart
+++ b/app/lib/common/providers/chat_providers.dart
@@ -26,7 +26,7 @@ final _convosProvider =
 
 /// Provider that sorts up list based on latest timestamp from [_convosProvider].
 final chatsProvider = StateProvider<List<Convo>>((ref) {
-  final convos = ref.watch(_convosProvider);
+  final convos = List.of(ref.watch(_convosProvider));
   convos.sort((a, b) => b.latestMessageTs().compareTo(a.latestMessageTs()));
   return convos;
 });
@@ -43,9 +43,10 @@ final chatProvider =
 
 final relatedChatsProvider =
     FutureProvider.family<List<Convo>, String>((ref, spaceId) async {
-  final chats =
-      (await ref.watch(spaceRelationsOverviewProvider(spaceId).future))
-          .knownChats;
+  final chats = List.of(
+    (await ref.watch(spaceRelationsOverviewProvider(spaceId).future))
+        .knownChats,
+  );
   chats.sort((a, b) => b.latestMessageTs().compareTo(a.latestMessageTs()));
   return chats;
 });
--- a/app/lib/common/providers/chat_providers.dart
+++ b/app/lib/common/providers/chat_providers.dart
@@ -4,7 +4,6 @@ import 'package:acter/features/home/providers/client_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:riverpod/riverpod.dart';
 
-/// Provider the profile data of a the given space, keeps up to date with underlying client
 final convoProvider =
     AsyncNotifierProvider.family<AsyncConvoNotifier, Convo?, Convo>(
   () => AsyncConvoNotifier(),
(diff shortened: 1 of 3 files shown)


app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -86,54 +86,66 @@ class ChatRoomsListNotifier extends StateNotifier<List<Convo>> {
...
+        _sortList(newList);
         state = newList;
         break;
       case 'Remove':
         final index = diff.index()!;
         final newList = listCopy();
         newList.removeAt(index);
+        _sortList(newList);
         state = newList;
         break;
```
gnunicorn, line 124: I like the idea, but this _won't work_. The `ConvoDiff` is a sorted list using indixes, you are changing the list that the next diff is going to be applied on. This is particuarly problematic for  cases like this `remove` here, where an item is supposed to be removed based on the index. Because you resorted the list before, we will be removing the wrong item.

app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -86,54 +86,66 @@ class ChatRoomsListNotifier extends StateNotifier<List<Convo>> {
 
   List<Convo> listCopy() => List.from(state, growable: true);
 
+  void _sortList(List<Convo> list) {
+    list.sort((a, b) => b.latestMessageTs().compareTo(b.latestMessageTs()));
+  }
+
```
gnunicorn, line 92: May I suggest to not do this here but instead put a provider in between this notifier and the actual list to be displayed and make that resort the list instead? This way we always have the proper indexed list (as the rust-sdk requires) and we can always resort on the UI for the proper displaying.

gnunicorn (changes requested):
Put in the comments why we can't do this.

app/lib/common/providers/notifiers/chat_notifiers.dart
gtalha07: Sounds good.

app/lib/common/providers/chat_providers.dart
```diff
@@ -18,13 +17,20 @@ final latestMessageProvider =
...
   final client = ref.watch(alwaysClientProvider);
   return ChatRoomsListNotifier(ref: ref, client: client);
 });
 
+/// Provider that sorts up list based on latest timestamp from [_convosProvider].
+final chatsProvider = StateProvider<List<Convo>>((ref) {
+  final convos = ref.watch(_convosProvider);
+  convos.sort((a, b) => b.latestMessageTs().compareTo(a.latestMessageTs()));
+  return convos;
+});
```
gnunicorn, line 32: are we sure this isn't mutating the original list? I am never 100% sure about that identity in dart...
```suggestion
final chatsProvider = StateProvider<List<Convo>>((ref) {
  final convos = List.of(ref.watch(_convosProvider));
  convos.sort((a, b) => b.latestMessageTs().compareTo(a.latestMessageTs()));
  return convos;
});
```

app/lib/common/providers/chat_providers.dart
```diff
@@ -37,8 +43,11 @@ final chatProvider =
 
 final relatedChatsProvider =
     FutureProvider.family<List<Convo>, String>((ref, spaceId) async {
-  return (await ref.watch(spaceRelationsOverviewProvider(spaceId).future))
-      .knownChats;
+  final chats =
```
gnunicorn, line 46: same as above.

gnunicorn (approved):
just one list identity query.

app/lib/common/providers/chat_providers.dart
gtalha07: Yeah it kinda does, they are mutable by default, sorry will fix this!

PR #1939, opened 2024-07-11
Refactor queryHierarchy to pull the entire hierarchy at once.

### LINKED ISSUE #1938: Subspaces do not show in Parent space if not joined
Our UX test shows that:
1. creating a space
2. creating a subpspace for it (which by default will be "limited" visible)
4. have a super invite for the parent space
5. a newly registered user sees the parent spaces, but not the subspace (despite that it can also join it):

image

gnunicorn:
Fix not displaying related rooms, fixes #1938 .

There is still a bit of an odd case when we have some in the internal list but zero to show (e.g. they are all private). Without make a server requests we won't know that, but I'd like to avoid having the server request sent and delay the menu-building even further just to display that screen... It's a pretty rare case though and one I'd ignore for now....

See it live: Related chats & spaces showing up in the actual spaces list (the one with the room id is an old crap thing without any data, that's why it shows up like that), even more in the sublisting:

https://github.com/acterglobal/a3/assets/40496/8837c347-375a-4f40-9619-ebc6e1fea91b


bitfriend (approved):
Looks good to me

app/lib/features/space/widgets/space_sections/chats_section.dart
```diff
@@ -35,9 +41,15 @@ class ChatsSection extends ConsumerWidget {
...
+  Widget buildChatsSectionUI(
+    BuildContext context,
+    WidgetRef ref,
+    List<Convo> chats,
+    bool hasMore,
+  ) {
     int chatsLimit = (chats.length > limit) ? limit : chats.length;
-    bool isShowSeeAllButton = chats.length > chatsLimit;
+    int moreCount = limit > chats.length ? limit - chats.length : 0;
+    bool isShowSeeAllButton = (chats.length > chatsLimit) || hasMore;
```
kumarpalsinh25, line 52: This logic looks pretty complicated but fine for now.

app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -69,15 +68,15 @@ final tabsProvider =
     }
   }
 
-  final spacesList = await ref.watch(relatedSpacesProvider(spaceId).future);
-  if (spacesList.isNotEmpty) {
+  final hasSpaces = await ref.watch(hasSubSpacesProvider(spaceId).future);
+  if (hasSpaces) {
```
kumarpalsinh25, line 72: Look good for fine but need think in future for proper solution.

kumarpalsinh25 (approved):
As per current requirement code look good and considering that in future we need to come up with some proper solution.

app/lib/features/space/providers/space_navbar_provider.dart
gnunicorn: I have an idea about that already, expect an update soon.

PR #1940, opened 2024-07-12
Generalised avatar view and change flow

kumarpalsinh25:
Make avatar view and change flow generic and cover different use-case as below.

### 1) Avatar Empty - With Change Permission - Direct change options:

https://github.com/user-attachments/assets/ffb1f842-eafb-4489-8746-1e5526269753

### 2) Avatar Available - With Change Permission - Full View + Change:

https://github.com/user-attachments/assets/65046f1d-490e-427b-847b-417f3f73e7d0

### 3) Avatar Available - No Permission - Full View Only:

https://github.com/user-attachments/assets/980847c1-1c2a-48c1-be81-c639d5769948

### 4) Avatar Empty - No Permission - No Full screen View:

https://github.com/user-attachments/assets/a88a1b09-92f2-4de3-a751-27719cb95e5f

### 5) Zoom option:

https://github.com/user-attachments/assets/16ef586f-0d38-41c0-aa04-06f42191e53f

Side-note:
- 
Full screen avatar is still not support for DM and Group as it need to manage with different scenario.

### DIFF
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -192,9 +192,9 @@ Future<void> openAvatar(
   WidgetRef ref,
   String roomId,
 ) async {
-  final membership = await ref.watch(roomMembershipProvider(roomId).future);
+  final membership = await ref.read(roomMembershipProvider(roomId).future);
   final canUpdateAvatar = membership?.canString('CanUpdateAvatar') == true;
-  final avatarInfo = ref.watch(roomAvatarInfoProvider(roomId));
+  final avatarInfo = ref.read(roomAvatarInfoProvider(roomId));
 
   if (avatarInfo.avatar != null && context.mounted) {
     //Open avatar in full screen if avatar data available
@@ -202,9 +202,7 @@ Future<void> openAvatar(
       Routes.fullScreenAvatar.name,
       queryParameters: {'roomId': roomId},
     );
-  } else if (avatarInfo.avatar == null &&
-      canUpdateAvatar &&
-      context.mounted) {
+  } else if (avatarInfo.avatar == null && canUpdateAvatar && context.mounted) {
     //Change avatar if avatar is null and have relevant permission
     uploadAvatar(ref, context, roomId);
   }
@@ -230,7 +228,12 @@ Future<void> uploadAvatar(
     EasyLoading.dismiss();
   } catch (e, st) {
     _log.severe('Failed to upload avatar', e, st);
-    EasyLoading.dismiss();
+    if (context.mounted) {
+      EasyLoading.showError(
+        L10n.of(context).failedToUploadAvatar(e),
+        duration: const Duration(seconds: 3),
+      );
+    }
   }
 }
 
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -3,8 +3,11 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:math';
+import 'package:acter/common/providers/room_providers.dart';
+import 'package:acter/common/utils/routes.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
+import 'package:file_picker/file_picker.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 import 'package:flutter_easyloading/flutter_easyloading.dart';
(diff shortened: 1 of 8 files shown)


app/lib/common/utils/utils.dart
```diff
@@ -184,6 +187,53 @@ String randomString() {
...
 }
 
+Future<void> openAvatar(
+  BuildContext context,
+  WidgetRef ref,
+  String roomId,
+) async {
+  final membership = await ref.watch(roomMembershipProvider(roomId).future);
+  final canUpdateAvatar = membership?.canString('CanUpdateAvatar') == true;
+  final avatarInfo = ref.watch(roomAvatarInfoProvider(roomId));
```
gnunicorn, line 197: always using `ref.read` in function calls, as we are not actually setting up any watcher and data isn't being executed if things changed. Using `read` clarifies that this is the case.

```suggestion
  final membership = await ref.read(roomMembershipProvider(roomId).future);
  final canUpdateAvatar = membership?.canString('CanUpdateAvatar') == true;
  final avatarInfo = ref.read(roomAvatarInfoProvider(roomId));
```

app/lib/common/utils/utils.dart
```diff
@@ -184,6 +187,53 @@ String randomString() {
...
+    uploadAvatar(ref, context, roomId);
+  }
+}
+
+Future<void> uploadAvatar(
+  WidgetRef ref,
+  BuildContext context,
+  String roomId,
+) async {
+  final room = await ref.read(maybeRoomProvider(roomId).future);
```
gnunicorn, line 218: :+1:

app/lib/common/utils/utils.dart
```diff
@@ -184,6 +187,53 @@ String randomString() {
...
+  try {
+    if (!context.mounted) return;
+    EasyLoading.show(status: L10n.of(context).avatarUploading);
+    final file = result.files.first;
+    if (file.path != null) await room.uploadAvatar(file.path!);
+    // close loading
+    EasyLoading.dismiss();
+  } catch (e, st) {
+    _log.severe('Failed to upload avatar', e, st);
+    EasyLoading.dismiss();
```
gnunicorn, line 233: `EasyLoading.showError` please. Tell the user if we couldn't complete it or they assume it worked and well, it didn't ;) .

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
```diff
@@ -0,0 +1,61 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Scaffold(
+      appBar: _buildAppBar(context, ref),
+      body: _buildBody(context, ref),
+    );
+  }
+
+  AppBar _buildAppBar(BuildContext context, WidgetRef ref) {
+    final membership = ref.watch(roomMembershipProvider(roomId)).requireValue;
```
gnunicorn, line 22:
```suggestion
    final membership = ref.watch(roomMembershipProvider(roomId)).valueOrNull;
```

I feel like you should create a custom rewrite in your editor that replace `requireValue` with `valueOrNull` on typing ;) .

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
```diff
@@ -0,0 +1,61 @@
...
+      leading: IconButton(
+        onPressed: context.pop,
+        icon: const Icon(Icons.close),
+      ),
+      actions: [
+        if (membership?.canString('CanUpdateAvatar') == true)
+          IconButton(
+            onPressed: () => uploadAvatar(ref, context, roomId),
+            icon: const Icon(Icons.edit_outlined),
+          ),
```
gnunicorn, line 33: When seeing the videos I was wondering about resetting the avatar ... like, just removing it... but we don't even have the rust API ready for that yet, so ... nothing here for now ;) .

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
```diff
@@ -0,0 +1,61 @@
...
+    return AppBar(
+      leading: IconButton(
+        onPressed: context.pop,
+        icon: const Icon(Icons.close),
+      ),
+      actions: [
+        if (membership?.canString('CanUpdateAvatar') == true)
+          IconButton(
+            onPressed: () => uploadAvatar(ref, context, roomId),
+            icon: const Icon(Icons.edit_outlined),
```
gnunicorn, line 32: Should we use the upload icon here? the pencil implies I _edit the image itself_, rather than replace it...

gnunicorn (approved):
nice work. some minor requireValue stuff that needs to be fixed first ;)

gtalha07 (approved):
The code looks fine to me, as well as the new UI improvements. But I have few suggestions for improved user-experience in regards of design flow. Specifically I see (from screen videos), when proceeding direct-change avatar flow, there should be confirmation sheet before opening picker which can show options .i.e. ```Change/Upload Avatar``` and ```Cancel```. This is good extra step for user and provides more secure and intended way of changing important data. Overall, this goes as my approval but if it's something trivial that can be added right away before merge, I would encourage but I'd leave that up to you!

gnunicorn:
>  Specifically I see (from screen videos), when proceeding direct-change avatar flow, there should be confirmation sheet before opening picker which can show options .i.e. Change/Upload Avatar and Cancel. This is good extra step for user and provides more secure and intended way of changing important data. 

Why? If you didn't intend to do that, you just cancel/navigate back... I don't think this extra step is helpful unless there was an actual "option" here ... e.g. "Select File", "Take Picture with Camera" . Otherwise it is clicked wrong once but learnt immediately through that. I'd not do that extra step, as it is also not harmful in any way.

gtalha07:
> > Specifically I see (from screen videos), when proceeding direct-change avatar flow, there should be confirmation sheet before opening picker which can show options .i.e. Change/Upload Avatar and Cancel. This is good extra step for user and provides more secure and intended way of changing important data.
> 
> Why? If you didn't intend to do that, you just cancel/navigate back... I don't think this extra step is helpful unless there was an actual "option" here ... e.g. "Select File", "Take Picture with Camera" . Otherwise it is clicked wrong once but learnt immediately through that. I'd not do that extra step, as it is also not harmful in any way.

Well, I didn't mean to put in harmful sense, but that way, it could be more informational about what this interaction actually means. Because we have different actions on avatars currently, i.e. tapping on parent badge leads to space but tapping on the navigated space avatar opens up picker. Unless we plan on documenting this somewhere else?? Anyways just my thoughts...

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
kumarpalsinh25: Oh no, Again!! :(

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
kumarpalsinh25: I also thought on the same during this PR development, but kept it for later.

app/lib/common/widgets/avatar/full_screen_avatar_page.dart
kumarpalsinh25: Yes sure

PR #1942, opened 2024-07-12
Fix empty spaces nav

gnunicorn:
Addition to [#1939](https://github.com/acterglobal/a3/pull/1939#discussion_r1675352728), fixing the bug that the `Spaces` and `Chats` entries showed despite not having any children of that type, if all known children of that type turned out to be inaccessible to us:

**See it in Action**:
Compare to the screenrecording of the PR above and you'll notice the spaces section doesn't show up empty anymore.

https://github.com/user-attachments/assets/6a5fd993-0d8e-4dac-b852-28b3255d2bd9

**How it works**
Without delaying the navigation for a full remote-path check the new notifiers now return true quickly if there are items and if the indicator is set that there might be more, initially returns _false_ and only after a remote request done in the background confirmed, we do have other spaces/chats we could join, updates to `true`. If that never comes back with a list of at least one item, it stays on `false` and the UI never falsely switches to showing an empty section in the first place. All, without delaying the first render of the navigation \o/ .

**To the reviewer**
Ignore the first commit, that is only an update of the freshly generated `_ffi`-file.


app/lib/common/providers/notifiers/relations_notifier.dart
```diff
@@ -0,0 +1,40 @@
...
+class HasSubSpacesNotifier extends FamilyAsyncNotifier<bool, String> {
+  @override
+  FutureOr<bool> build(String spaceId) async {
+    final relatedSpaces =
+        await ref.watch(spaceRelationsOverviewProvider(spaceId).future);
+    if (relatedSpaces.knownSubspaces.isNotEmpty) {
+      return true; // we have subspaces and know it
+    }
+    if (relatedSpaces.hasMoreSubspaces) {
+      // there might be some, but we need to confirm remotely. We do that without blocking
```
kumarpalsinh25, line 15: 👍🏻👍🏻

kumarpalsinh25 (approved):
Nice work!! 👏🏻

PR #1944, opened 2024-07-14
Space categories backoffice

gnunicorn:
This brings a `Category`-feature to Acter spaces: saved as a state event, admins can now configure a list of categories (identified via its `id`) that consists of a title, an icon and a list of `entries` for various space sections, e.g. chats or spaces. The list of entries are plain strings for compatibility.

This PR comes with a full e2e integration test showing that it works when there are no categories yet, can be updated and `id`s are properly replaced (at the right position). Due to the fact that our ffi-gen doesn't support `Vec<String>`-params from the dart-side, the Api is a bit odd. But should provide all the main features we need from it.


gtalha07 (approved):
LGTM!

PR #1945, opened 2024-07-14
Implement user bookmarks

gnunicorn:
Rust SDK Api to manage bookmarks - list or eventIds per type - as a global user config, synced across devices. Simple toggle per item through `add('pins', eventId)` and `remove('pins', eventId)`. Double adding is not possible, keys are de-duplicated before submitting. Items are added to the end, so essentially it is a "newest last" list of bookmarks (might make sense to reverse this for displaying them to the user).

To stay up to date, `client.subscribeStream('global.acter.bookmarks')`: this will trigger after any update is received.

Contains e2ee tests, of course.


bitfriend (approved):
Looks good to me

PR #1949, opened 2024-07-15
Keep avatar binary in flutters resident memory

### LINKED ISSUE #1937: Random Avatar Issues
There are random issues in avatar as below.

1) Weird avatar view:
ima_9a9c164

2) Wrong avatar showing:
ima_2984c21

3) Avatar not loading at all:
ima_e9c3703

gnunicorn:
While working on the android crash problem I noticed the following in remark in the generated ffi for `FfiBufferUint8`
```dart
  /// Returns a typed view of this buffer.
  /// Note: The lifetime of this view is tied to the lifetime of this FfiBufferUint8. This Uint8List must not live longer than the creating FfiBufferUint8
  Uint8List asTypedList() {
    final buffer = _box.borrow();
    final addressRaw = _api._ffiBufferAddress(buffer);
    final size = _api._ffiBufferSize(buffer) ~/ 1;
    return ffi.Pointer<ffi.Uint8>.fromAddress(addressRaw).asTypedList(size);
  }
```

Which means that if we ever where to discard the surrounding `FfiBufferUint8` the underlying memory space we have a view on might not be reused for other things on the rust side. This would explain very well the issues we have been seeing in #1937 and the freezes of freezes when going to any page that has an avatar on it (e.g. chat or updates).

And as it turns out we are usually doing the following to fetch an avatar:

```rust
  final avatar = await account.avatar(thumbSize);
  final avatarData  = avatar.data();
  if (avatarData != null) {
    return MemoryImage(avatarData.asTypedList());
  }
```

As we are pulling the `FfiBufferUint8` by calling `data()` and discard it before even returning the function. As a result the view we have handed to `MemoryImage` might or might actually still be valid. Sometimes it works, often times it does not. 

The fix is rather simple: make a copy of the binary in the flutter side to keep the data in resident memory there regardless of what rust does via `Uint8List.fromList(`. (Until we have uniffi-dart where we have a better memory management model). And that is the fix this PR does. And one fix where we called `.data()` more than once (last commit).

Fixes #1937, https://github.com/acterglobal/a3-meta/issues/368

### DIFF
--- a/app/lib/common/providers/common_providers.dart
+++ b/app/lib/common/providers/common_providers.dart
@@ -75,8 +75,9 @@ final _accountAvatarProvider =
   final account = ref.watch(accountProvider);
   final thumbSize = sdk.api.newThumbSize(48, 48);
   final avatar = await account.avatar(thumbSize);
-  if (avatar.data() != null) {
-    return MemoryImage(Uint8List.fromList(avatar.data()!.asTypedList()));
+  final avatarData = avatar.data();
+  if (avatarData != null) {
+    return MemoryImage(Uint8List.fromList(avatarData.asTypedList()));
   }
   return null;
 });
--- a/app/lib/common/providers/common_providers.dart
+++ b/app/lib/common/providers/common_providers.dart
@@ -75,6 +75,8 @@ final _accountAvatarProvider =
   final account = ref.watch(accountProvider);
   final thumbSize = sdk.api.newThumbSize(48, 48);
   final avatar = await account.avatar(thumbSize);
+  // Only call data() once as it will consume the value and any subsequent
+  // call will come back with `null`.
   final avatarData = avatar.data();
   if (avatarData != null) {
     return MemoryImage(Uint8List.fromList(avatarData.asTypedList()));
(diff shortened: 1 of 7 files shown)


app/lib/common/providers/common_providers.dart
```diff
@@ -73,8 +75,9 @@ final _accountAvatarProvider =
   final account = ref.watch(accountProvider);
   final thumbSize = sdk.api.newThumbSize(48, 48);
   final avatar = await account.avatar(thumbSize);
-  if (avatar.data() != null) {
-    return MemoryImage(avatar.data()!.asTypedList());
```
gtalha07, line 77: can possibly add comment here too so in case anyone forgets why its intended this way ;)

gtalha07 (approved):
Looks good!

app/lib/common/providers/common_providers.dart
```diff
@@ -73,8 +75,9 @@ final _accountAvatarProvider =
...
+  final avatarData = avatar.data();
```
gnunicorn, line 78:
```suggestion
  // Only call data() once as it will consume the value and any subsequent
  // call will come back with `null`.
  final avatarData = avatar.data();
```

Issue #1239, opened 2023-12-20
Logout/Remove session not working


kumarpalsinh25:
Tested with...
- iPhone 12
- iOS 17.1
- 

https://github.com/acterglobal/a3/assets/51526480/fd1b3778-2ee6-4006-986f-bea60dd250f7

gtalha07:
Same [thoughts](https://github.com/acterglobal/a3/issues/1240#issuecomment-1864417496), the functionality isn't there. We need to remove these buttons.

kumarpalsinh25:
> Same https://github.com/acterglobal/a3/issues/1240#issuecomment-1864417496, the functionality isn't there. We need to remove these buttons.

#gnunicorn 
You mentioned that this is not expected behavior. Any thoughts on this?

kumarpalsinh25:
> Same https://github.com/acterglobal/a3/issues/1240#issuecomment-1864417496, the functionality isn't there. We need to remove these buttons.

@gnunicorn 
You mentioned that this is not expected behavior. Any thoughts on this?

gnunicorn:
pretty sure this was functional once. this seems like a regression we need should fix and add a ui-test-case for.

PR #1950, opened 2024-07-15
Fix bug in force logout

bitfriend:
Will fix [this issue](https://github.com/acterglobal/a3/issues/1239)


bitfriend:
As you see, `ffi-gen` didn't implement `Vec<String>` yet and `as FfiListFfiString` failed.
image
So I replaced `delete_devices()` API with `delete_device()` just taking a single device id.

bitfriend:
This is result video.

https://github.com/user-attachments/assets/09780fc1-aec0-471f-be1c-d73ee36e4483

gnunicorn (approved):

Issue #1948, opened 2024-07-15
crashes on start on Android


gnunicorn:
Installing works fine, but starting the App the Logo appears shortly in full screen then it just disappears. Fails on Android 13/Galaxy A03 core .

gnunicorn:
Seen it live, it instigated that we had seen this before. Trying locally to verify that it isn't the bug we had fixed in the past where a never SDK api was needed, API Level 33 (which is Android 13) actually works fine.

gnunicorn:
But it brought up another idea: when running the App in a 32bit environment, because our FFI-gen just hard ignores 32bit support and just hard assumes 64bit platforms, this is a behaviour we have seen before.

Some investigation reveals that despite the target phone having an Arm64, the vendors can opt-in to using a 32bit firmware, which would be trying to use our arm-v7 target. And according to some sources on the internet that `core UI`-series does exactly that. Which would mean, it would try to use our arm-v7 target for which our ffi-gen is incompatible and has the wrong internal length ...

gnunicorn:
I can confirm it is crashing when trying to run it on a 32bit x82 emulator in the same manner and the following log:

```
"da": 819 untranslated message(s).
"de": 43 untranslated message(s).
"pl": 157 untranslated message(s).
"pt": 1145 untranslated message(s).
To see a detailed report, use the untranslated-messages-file
option in the l10n.yaml file:
untranslated-messages-file: desiredFileName.txt
<other option>: <other selection>

This will generate a JSON format file containing all messages that
need to be translated.
Launching lib/main.dart on sdk gphone x86 in debug mode...
"da": 819 untranslated message(s).
"de": 43 untranslated message(s).
"pl": 157 untranslated message(s).
"pt": 1145 untranslated message(s).
To see a detailed report, use the untranslated-messages-file
option in the l10n.yaml file:
untranslated-messages-file: desiredFileName.txt
<other option>: <other selection>

This will generate a JSON format file containing all messages that
need to be translated.
✓ Built build/app/outputs/flutter-apk/app-debug.apk
Connecting to VM Service at ws://127.0.0.1:40585/I5jJ0rtqGAY=/ws
D/ProfileInstaller(10666): Installing profile for global.acter.a3
I/FirebaseApp(10666): Device unlocked: initializing all Firebase APIs for app [DEFAULT]
[...]

PR #1951, opened 2024-07-15
Fix 32bits support

gnunicorn:
Builds on https://github.com/acterglobal/ffi-gen/pull/13 as a potential solution for #1948 . Needs more and proper testing. This is on top of #1949 to ensure this isn't a bug we are facing here, too.


gnunicorn:
superseeded by #1952

PR #1952, opened 2024-07-15
Fix 32bits support

gnunicorn:
Builds on https://github.com/acterglobal/ffi-gen/pull/13 as a potential solution for https://github.com/acterglobal/a3/issues/1948 . Needs more and proper testing. This is on top of https://github.com/acterglobal/a3/pull/1949 to ensure this isn't a bug we are facing here, too.


gnunicorn:
Okay, found a fix for the Apple builds too and the demos successfully build for this now. I am still a bit scared about the nature of this change as it could have unexpected side-effects elsewhere and otherwise fine platforms. Thus I'd suggest we merge this only after the weekly release tomorrow and test it via nightly builds. And if things to well, we can push another Android release any time...

bitfriend (approved):
Looks good to me.

PR #1953, opened 2024-07-16
Generic Task List Page

kumarpalsinh25:
### Improvements in task list pages as below.

- Generic Task List page (App Task List and Space Task List)
- Search bar option in task list
- Task providers clean-up and optimisations
- General code clean-up

### Reference Video:

https://github.com/user-attachments/assets/02e87c2e-d46a-44f5-bade-7e7ef51833a9


gnunicorn:
does it search tasks, too? Or only lists?

app/lib/features/tasks/providers/notifiers.dart
```diff
@@ -86,3 +90,41 @@ class TaskNotifier extends FamilyAsyncNotifier<Task, Task> {
...
+    final tasksList = (await client.taskLists()).toList();
+    // this might throw internally
+
+    if (arg == null) {
+      return tasksList;
+    } else {
+      //GET SPACE TASKS LIST
+      return tasksList
+          .where((t) => t.spaceIdStr() == arg!)
+          .toList(); // this might throw internally
```
gnunicorn, line 127: why would that throw internally?
```suggestion
          .toList();
```

app/lib/features/tasks/providers/notifiers.dart
```diff
@@ -86,3 +90,41 @@ class TaskNotifier extends FamilyAsyncNotifier<Task, Task> {
...
+    });
+    ref.onDispose(() => _poller.cancel());
+
+    return await _getTasksList(client);
+  }
+
+  Future<List<TaskList>> _getTasksList(Client client) async {
+    //GET ALL TASKS LIST
+    final tasksList = (await client.taskLists()).toList();
+    // this might throw internally
```
gnunicorn, line 119: so... this is really not ideal as for two reasons: 1. we are loading _all tasks lists_ for every spaceId and the global one, holding them all in memory despite only needing them to be existing once and then filtered for whatever space we need, and 2. upon an update of `tasks` being issues by the client, _all_ of these notifiers fire and refresh again, despite only a subset actually needing to be refreshed.

app/lib/features/tasks/providers/notifiers.dart
```diff
@@ -86,3 +90,41 @@ class TaskNotifier extends FamilyAsyncNotifier<Task, Task> {
...
+    if (arg == null) {
+      return tasksList;
+    } else {
+      //GET SPACE TASKS LIST
+      return tasksList
+          .where((t) => t.spaceIdStr() == arg!)
+          .toList(); // this might throw internally
+    }
+  }
+}
```
gnunicorn, line 130: I'd recommend we split this up:

```suggestion
//List of all task list
//If space Id given then it will return list of space task list
class AsyncAllTaskListsNotifier
    extends AsyncNotifier<List<TaskList>> {
  late Stream<bool> _listener;
  late StreamSubscription<bool> _poller;

  @override
  Future<List<TaskList>> build() async {
    final client = ref.watch(alwaysClientProvider);

    //GET ALL TASKS LIST
    _listener = client.subscribeStream('tasks');

    _poller = _listener.listen((e) async {
      state = await AsyncValue.guard(() => _getTasksList(client));
    });
    ref.onDispose(() => _poller.cancel());

    return await _getTasksList(client);
  }

  Future<List<TaskList>> _getTasksList(Client client) async {
    //GET ALL TASKS LIST
    return (await client.taskLists()).toList();
}
```

(see below)

app/lib/features/tasks/providers/tasklists_providers.dart
```diff
@@ -0,0 +1,29 @@
...
+//Single Task List Item based on the task list id
+final taskListItemProvider =
+    AsyncNotifierProvider.family<TaskListItemNotifier, TaskList, String>(
+  () => TaskListItemNotifier(),
+);
+
+//List of all task list
+//If space Id given then it will return list of space task list
+final tasksListsProvider = AsyncNotifierProvider.family<AsyncTaskListsNotifier,
+    List<TaskList>, String?>(() => AsyncTaskListsNotifier());
```
gnunicorn, line 14: this way we only hold one in-memory list and only one signaling provider to the underlying SDK.

```suggestion
//List of all task list
//If space Id given then it will return list of space task list
final allTasksListsProvider = AsyncNotifierProvider<AsyncTaskListsNotifier,
    List<TaskList>>(() => AsyncAllTaskListsNotifier());
    
final taskListProvider = FutureProvider.family<List<TaskList>, String?>((ref, spaceId) async {
  final allTaskLists = await ref.watch(allTasksListsProvider.future);
  if (spaceId == null) {
    return allTaskLists;
  } else {
    return tasksList
      .where((t) => t.spaceIdStr() == spaceId!)
      .toList(); // this might throw internally
   }
});
```

app/lib/features/tasks/providers/tasklists_providers.dart
```diff
@@ -0,0 +1,29 @@
...
+typedef TasksListSearchParams = ({String? spaceId, String searchText});
+
+final tasksListSearchProvider = FutureProvider.autoDispose
+    .family<List<TaskList>, TasksListSearchParams>((ref, params) async {
+  final tasksList = await ref.watch(tasksListsProvider(params.spaceId).future);
+  List<TaskList> filteredTaskList = [];
+  for (final task in tasksList) {
+    if (task.name().toLowerCase().contains(params.searchText.toLowerCase())) {
+      filteredTaskList.add(task);
+    }
```
gnunicorn, line 26: we should also filter the inner tasks ... people would expect that if they searched for a task-item, they'd also see that show up regardless of whether the word is in the tasklist...

app/lib/features/tasks/pages/tasks_list_page.dart
```diff
@@ -0,0 +1,215 @@
...
+
+    if (searchValue.isNotEmpty) {
+      tasksList = ref.watch(
+        tasksListSearchProvider(
+          (spaceId: widget.spaceId, searchText: searchValue),
+        ),
+      );
+    } else {
+      tasksList = ref.watch(tasksListsProvider(widget.spaceId));
+    }
```
gnunicorn, line 113: we could also handle that by just having the entire list returned if `searchValue.isEmpty` on the search provider ... making this more straight forward and moving the "logic" into the provider (where logic ideally resides)

gnunicorn (approved):
Looking good. But the providers need some improvement before we can merge this. I've added suggestions. Let me know if they need clarification. Once applied, feel free to merge.

app/lib/features/tasks/providers/notifiers.dart
kumarpalsinh25: I see, understood the potential problem of above approach. Did the require changes as per below comment suggestions.

app/lib/features/tasks/providers/notifiers.dart
kumarpalsinh25: I copied code from pins notifiers and comment was there. I don't know if it is applicable to tasks module or not but this comment is also there in events notifiers.

app/lib/features/tasks/providers/tasklists_providers.dart
kumarpalsinh25: That good thing. Added task items in search criteria.

PR #1970, opened 2024-07-22
Refresh chats & subspaces per action, remove requireValue

gnunicorn:
- Refreshing subspaces and chats via the press of a button:

https://github.com/user-attachments/assets/b3bca8e4-e61c-4181-bfa8-76d18a1d17f6

- Remove all remaining `requireValue`. They are bad and we are seeing them breaking builds in the wild!


gtalha07 (approved):
LGTM!

PR #1974, opened 2024-07-23
General Fixes

kumarpalsinh25:


app/lib/common/widgets/chat/convo_card.dart
```diff
@@ -218,29 +218,21 @@ class _SubtitleWidget extends ConsumerWidget {
...
                   ),
                 ),
                 Flexible(
                   child: Html(
-                    padding: const EdgeInsets.symmetric(vertical: 10),
                     // ignore: unnecessary_string_interpolations
                     data: '''$body''',
                     maxLines: 1,
-                    defaultTextStyle: const TextStyle(
-                      overflow: TextOverflow.ellipsis,
```
gtalha07, line 240: Wouldn't this throw UI overflow errors if it can't adjust text based on resizing or screen width?

app/lib/common/widgets/chat/convo_card.dart
```diff
@@ -261,28 +253,21 @@ class _SubtitleWidget extends ConsumerWidget {
...
               ),
             ),
             Flexible(
               child: Html(
-                padding: const EdgeInsets.symmetric(vertical: 10),
                 // ignore: unnecessary_string_interpolations
                 data: '''$body''',
                 maxLines: 1,
-                defaultTextStyle: const TextStyle(
-                  overflow: TextOverflow.ellipsis,
```
gtalha07, line 283: Same goes here. The overflow manages the text part based on screen width and its removed...

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -427,6 +427,7 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
       ),
       builder: (ctx) => AttachmentOptions(
         onTapCamera: () async {
+          Navigator.of(context).pop();
```
gtalha07, line 430: Not sure if this'd work the same in case of desktop. Hence it wasn't added here. On desktop, attachments open as dialog so this might pop up the chat page. I'd like to be more affirm that it doesn't get broken by attaching screenshot video (Desktop).

gtalha07 (changes requested):
Great fixes. But over some parts, I see instances where it could lead up to text overflows and abnormal route pop-off which I have pointed out. Would be very nice if you can also attach screenshot of desktop version where it is tested out to be correctly working.

Also I think we might wanna remove Members count text from the DM room page also as you described in 4th point.

app/lib/common/widgets/chat/convo_card.dart
kumarpalsinh25: Resolved by https://github.com/acterglobal/a3/pull/1974/commits/6b0537a78daba6b6bac6f53a5538ee6f71bbb4dd

app/lib/common/widgets/chat/convo_card.dart
kumarpalsinh25: Resolved by https://github.com/acterglobal/a3/pull/1974/commits/6b0537a78daba6b6bac6f53a5538ee6f71bbb4dd

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Won't create any issue on desktop. Here is the video recording of the same.

https://github.com/user-attachments/assets/f9f67ff2-4185-4b0f-b0a4-f5c601ac4349

gtalha07: Not a fan of bottom-sheets for desktop UI but I guess that doesn't break it. 👍🏻

gtalha07 (approved):

Issue #1915, opened 2024-07-02
Failed to edit space: No such file (Os error 2)


emilvincentz:
I wanted to change the name, but says error (but is changing anyway)

https://github.com/acterglobal/a3/assets/77483554/b1a4e7a6-a27c-4ddc-9f39-11ebd914e4a9

emilvincentz:
https://github.com/acterglobal/a3/assets/77483554/69da2920-7abd-4f60-99db-4ffbdba6fb9b

PR #1975, opened 2024-07-23
Get rid of edit space screen and manage it with inline option

kumarpalsinh25:
fixes,
https://github.com/acterglobal/a3-meta/issues/403
https://github.com/acterglobal/a3/issues/1915

https://github.com/user-attachments/assets/dac26e14-4fd9-4dfc-ad2e-39fcaafcdff6


emilvincentz:
looks like a good and quick fix

kumarpalsinh25:
> looks like a good and quick fix

Not for quick fix but actually this is the UX pattern we have follow in the entire app.

PR #1976, opened 2024-07-23
Fixing Subchat listing

gnunicorn:
This fixes missing the chats-section when you were not in any of the subchats, as well as several errors in trying to list them up. It also improves the performance by dropping several unnecessary fetching of entire objects where roomIds only will do.

See that it works in the previously broken demo space:

https://github.com/user-attachments/assets/deea3a3a-f4a7-4066-ad3b-4ff2f17b47e2

There are several fixes in this coming together. I'll explain them with comments on the PR review to make it easier to understand why the changes were made the way they were.

### DIFF
--- a/app/lib/common/providers/space_providers.dart
+++ b/app/lib/common/providers/space_providers.dart
@@ -1,4 +1,3 @@
-import 'package:acter/common/providers/chat_providers.dart';
 import 'package:acter/common/providers/notifiers/relations_notifier.dart';
 import 'package:acter/common/providers/notifiers/space_notifiers.dart';
 import 'package:acter/common/providers/room_providers.dart';
@@ -108,11 +107,10 @@ class SpaceItem {
 }
 
 class SpaceRelationsOverview {
-  bool hasMoreSubspaces;
-  bool hasMoreChats;
+  bool hasMore;
   SpaceRelations rel;
-  List<Space> knownSubspaces;
-  List<Convo> knownChats;
+  List<String> knownSubspaces;
+  List<String> knownChats;
   List<String> suggestedIds;
   Space? mainParent;
   List<Space> parents;
@@ -126,8 +124,7 @@ class SpaceRelationsOverview {
     required this.mainParent,
     required this.parents,
     required this.otherRelations,
-    required this.hasMoreSubspaces,
-    required this.hasMoreChats,
+    required this.hasMore,
   });
 }
 
@@ -231,37 +228,27 @@ final spaceRelationsOverviewProvider =
   if (relatedSpaces == null) {
     throw SpaceNotFound;
   }
-  bool hasMoreSubspaces = false;
-  bool hasMoreChats = false;
-  final List<Space> knownSubspaces = [];
-  final List<Convo> knownChats = [];
+  bool hasMore = false;
+  final List<String> knownSubspaces = [];
+  final List<String> knownChats = [];
   final List<String> suggested = [];
   List<Space> otherRelated = [];
-  for (final related in relatedSpaces.children()) {
+  final children = relatedSpaces.children();
+  for (final related in children) {
     String targetType = related.targetType();
     final roomId = related.roomId().toString();
     if (related.suggested()) {
       suggested.add(roomId);
     }
-    if (targetType == 'ChatRoom') {
-      try {
-        final chat = await ref.watch(chatProvider(roomId).future);
-        knownChats.add(chat);
-      } catch (e) {
-        hasMoreChats = true;
-      }
+    if (targetType == 'Unknown') {
+      // this one is not found locally
+      hasMore = true;
+    } else if (targetType == 'ChatRoom') {
+      // we know this as a chat room
+      knownChats.add(roomId);
     } else {
-      try {
-        final space = await ref.watch(spaceProvider(roomId).future);
-        final isChildSpaceOf = await space.isChildSpaceOf(spaceId);
-        if (isChildSpaceOf) {
-          knownSubspaces.add(space);
-        } else {
-          otherRelated.add(space);
-        }
-      } catch (e) {
-        hasMoreSubspaces = true;
-      }
+      // this must be some space.
+      knownSubspaces.add(roomId);
     }
   }
   List<Space> parents = [];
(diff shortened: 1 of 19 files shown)


app/lib/common/providers/space_providers.dart
```diff
@@ -233,38 +228,27 @@ final spaceRelationsOverviewProvider =
...
-        knownChats.add(chat);
-      } catch (e) {
-        hasMoreChats = true;
-      }
+    if (targetType == 'Unknown') {
+      // this one is not found locally
+      hasMore = true;
+    } else if (targetType == 'ChatRoom') {
+      // we know this as a chat room
+      knownChats.add(roomId);
```
gnunicorn, line 248: This is a first part of the fix:
`ChatRoom` and `Space` are only given  `targetTypes` when we can _load_ them rooms and check them for it. For the remote rooms, this isn't the case, hence the `Unknown` flag. Previously, however, we'd only flag the `hasMoreChat` if we had tried and failed to read a room. 

The fix here is to not guess the type, but generalize that `hasMore` can be either chats or spaces until we have fetched them from remote.

app/lib/common/providers/space_providers.dart
```diff
@@ -233,38 +228,27 @@ final spaceRelationsOverviewProvider =
...
   }
-  final membership = await ref.watch(roomMembershipProvider(spaceId).future);
-  bool hasMoreSubspaces = false;
-  bool hasMoreChats = false;
-  final List<Space> knownSubspaces = [];
-  final List<Convo> knownChats = [];
+  bool hasMore = false;
+  final List<String> knownSubspaces = [];
+  final List<String> knownChats = [];
   final List<String> suggested = [];
```
gnunicorn, line 234: this is an important simplification for performance: by using the strings and just relying on the underlying rust having checked properly, we can remove an entire ref.watch call stack and the necessary memory problems.

app/lib/common/widgets/chat/convo_with_avatar_card.dart
```diff
@@ -70,7 +68,14 @@ class ConvoWithAvatarInfoCard extends ConsumerWidget {
                 onFocusChange: onFocusChange,
                 onLongPress: onLongPress,
                 leading: avatarWithIndicator(context, ref),
-                title: buildDisplayName(context, ref),
+                title: Text(
+                  avatarInfo.displayName ?? roomId,
+                  style: Theme.of(context)
```
gnunicorn, line 73: When given a roomInfo we'd unfortunately try to fetch a chat name for an object we don't have a local representation for. Thus failing and just showing `roomId`s. Instead we are now using the given `AvatarInfo` that is pushed in and always has the resolved displayName regardless of the origin object type.

app/lib/features/space/widgets/space_sections/chats_section.dart
```diff
@@ -44,12 +44,34 @@ class ChatsSection extends ConsumerWidget {
   Widget buildChatsSectionUI(
     BuildContext context,
     WidgetRef ref,
-    List<Convo> chats,
-    bool hasMore,
+    List<String> chats,
   ) {
-    int chatsLimit = (chats.length > limit) ? limit : chats.length;
-    int moreCount = limit > chats.length ? limit - chats.length : 0;
-    bool isShowSeeAllButton = (chats.length > chatsLimit) || hasMore;
```
gnunicorn, line 52: this was another source for a bug, where we'd attempt to figure out the specifics of when to show more items, we'd overestimate (because we didn't take into account whether any were actually resolved remotely).

The basic logic goes like this:
1. show up to `limit` items (usually 3)
2. fill up with known chats
3. if there are more, cut at 3, show the "See all" button
4. if there are spaces left, check whether we have some from remote
5. and if so, fill them up. but also take into account that the total should stay at 3 and only if there are more do we want to show the `Show all` button.

The new check is a lot more expressive and checks the actual remote data we can fetch. And updates accordingly. So we might only show 2 for a moment until we learn that there is another few remote ones.

This should also properly fix the range-errors we still have seen some times if we knew about further rooms but weren't able to load them from remote. Because we disregard that number entirely now.

app/lib/features/space/widgets/space_sections/chats_section.dart
```diff
@@ -62,25 +84,33 @@ class ChatsSection extends ConsumerWidget {
             pathParameters: {'spaceId': spaceId},
           ),
         ),
-        chatsListUI(chats, chatsLimit),
-        if (moreCount > 0 && hasMore) renderFurther(context, ref, moreCount),
+        chatsListUI(ref, chats, chatsLimit),
+        if (renderRemote) renderFurther(context, ref, moreCount),
```
gnunicorn, line 88: just one flag now.

native/acter/src/api/room.rs
```diff
@@ -545,27 +545,33 @@ impl SpaceRelations {
...
+            .map(|c| c.room_id())
+            .collect::<Vec<_>>();
         let room_id = self.room.room_id().to_owned();
         RUNTIME
             .spawn(async move {
-                let relations = c.space_relations(&room).await?.children.iter().filter(|c| c.suggested()).map(|c| c.room_id()).collect::<Vec<_>>();
                 let mut next : Option<String> = Some("".to_owned());
                 let mut rooms = Vec::new();
                 while next.is_some() {
-                    let request = assign!(get_hierarchy::v1::Request::new(room_id.clone()), { from: next.clone(), max_depth: Some(1u32.into()) });
```
gnunicorn, line 556: okay, this is the last part of the actual fix: we were not really always properly loading all subsequent children, because we limited the `max_depth=1`. As this is a tree-like-structure, the _parent_ we are asking for is the root level of 1. Thus, we need to ask for root-level of `2` to actually get all the proper children information. I am not quite sure how this has worked before to be honest ... it puzzles me a bit...

native/acter/src/api/room.rs
```diff
@@ -545,27 +545,33 @@ impl SpaceRelations {
...
+            .core
+            .children
+            .iter()
+            .filter(|c| c.suggested())
+            .map(|c| c.room_id())
+            .collect::<Vec<_>>();
         let room_id = self.room.room_id().to_owned();
         RUNTIME
             .spawn(async move {
-                let relations = c.space_relations(&room).await?.children.iter().filter(|c| c.suggested()).map(|c| c.room_id()).collect::<Vec<_>>();
```
gnunicorn, line 552: This is a pure optimization when I realized we didn't have to make that call again but already have it under `self.core`...

gtalha07 (approved):
Thanks for the effort providing explanation of the changes. Looks fine!

PR #1986, opened 2024-07-24
Desktop Settings-Routing fixes

### LINKED ISSUE #1906: can't navigate back from space -> settings
resize to phone-size, then go to any space -> settings and the back-button on the top left is missing on desktop.

gnunicorn:
Remove several context.go and make them use `pushReplacement` to switch between settings screens. Cleaning up a lot of `isDesktop` and using proper isLeadingImplied for AppBar to make the routing smooth across breakpoints (imagine someone rotating their iPad). Did some more code cleaning around the routing on the way, like replacing local size-checks with the `isLargeScreen` helper to unify the behavior and simplify the chat-columns-routing.

Fixed for Settings, Space Settings:

https://github.com/user-attachments/assets/1aeb7390-9328-477a-bddf-a809bf669325

Fixed for the partly weird Chat settings, too:

https://github.com/user-attachments/assets/10d3219b-9aa7-44f4-bbec-4ee4c7bd32bc

In an attempt to fix `GoError: There is nothing to pop`, this also refactors `context.pop` into a new `context.closeDialog` helper and uses that everywhere. If nothing else, it should give us a bit more context where we are when doing things wrong.

Fixes #1906 .

### DIFF
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -70,8 +70,9 @@ extension RefDebounceExtension on Ref {
 
 const largeScreenBreakPoint = 770;
 
-bool isLargeScreen(BuildContext context) {
-  return MediaQuery.of(context).size.width >= largeScreenBreakPoint;
+extension ActerContextUtils on BuildContext {
+  bool get isLargeScreen =>
+      MediaQuery.of(this).size.width >= largeScreenBreakPoint;
 }
 
 DateTime kFirstDay = DateTime.utc(2010, 10, 16);
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -17,6 +17,7 @@ import 'package:intl/intl.dart';
 import 'package:jiffy/jiffy.dart';
 import 'package:go_router/go_router.dart';
 import 'package:logging/logging.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
 import 'package:url_launcher/url_launcher.dart';
 
 final _log = Logger('a3::common::util');
@@ -73,6 +74,15 @@ const largeScreenBreakPoint = 770;
 extension ActerContextUtils on BuildContext {
   bool get isLargeScreen =>
       MediaQuery.of(this).size.width >= largeScreenBreakPoint;
+
+  Future<void> closeDialog<T>([T? result]) async {
+    try {
+      Navigator.of(this).pop(result);
+    } catch (error, stackTrace) {
+      _log.severe('Routing failed', error, stackTrace);
+      await Sentry.captureException(error, stackTrace: stackTrace);
+    }
+  }
 }
 
 DateTime kFirstDay = DateTime.utc(2010, 10, 16);
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -75,13 +75,22 @@ extension ActerContextUtils on BuildContext {
   bool get isLargeScreen =>
       MediaQuery.of(this).size.width >= largeScreenBreakPoint;
 
-  Future<void> closeDialog<T>([T? result]) async {
+  NavigatorState closeDialog({Routes? orGo, bool ignorePopFailure = false}) {
+    final navigator = Navigator.of(this);
     try {
-      Navigator.of(this).pop(result);
+      if (orGo != null && !navigator.canPop()) {
+        navigator.pushReplacementNamed(orGo.name);
+      } else {
+        navigator.pop();
+      }
     } catch (error, stackTrace) {
-      _log.severe('Routing failed', error, stackTrace);
-      await Sentry.captureException(error, stackTrace: stackTrace);
+      if (!ignorePopFailure) {
+        // track if we failed to pop so we can fix that.
+        _log.severe('Routing failed', error, stackTrace);
+        Sentry.captureException(error, stackTrace: stackTrace);
+      }
     }
+    return navigator;
   }
 }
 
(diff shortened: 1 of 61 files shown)


app/lib/features/search/widgets/spaces_builder.dart
```diff
@@ -78,7 +79,7 @@ class SpacesBuilder extends ConsumerWidget {
                     ),
                   ),
                   onTap: () {
-                    if (context.canPop()) context.pop();
+                    context.closeDialog(ignorePopFailure: true);
```
gtalha07, line 82: What does that mean? Is there any probable case where it can fail? What will happen after that?

gtalha07 (approved):

app/lib/features/search/widgets/spaces_builder.dart
gnunicorn: it means we attempt to pop, but if it doesn't succeed this won't log the error. which is the same behavior as the previous if did. There are a few cases where it is not clear from context whether the canPop() check is only there as a precaution or expected that there won't be any popping ... for those I added that ignore.

app/lib/common/utils/utils.dart
```diff
@@ -70,8 +71,27 @@ extension RefDebounceExtension on Ref {
 
 const largeScreenBreakPoint = 770;
 
-bool isLargeScreen(BuildContext context) {
-  return MediaQuery.of(context).size.width >= largeScreenBreakPoint;
+extension ActerContextUtils on BuildContext {
+  bool get isLargeScreen =>
+      MediaQuery.of(this).size.width >= largeScreenBreakPoint;
+
+  NavigatorState closeDialog({Routes? orGo, bool ignorePopFailure = false}) {
```
kumarpalsinh25, line 78: Not understanding the reason for `ignorePopFailure`. Why we need to ignore that?

app/lib/features/search/widgets/quick_jump.dart
```diff
@@ -146,8 +146,7 @@ class _QuickJumpState extends ConsumerState<QuickJump> {
   }
 
   void routeTo(Routes route) {
-    if (context.canPop()) context.pop();
-    context.pushNamed(route.name);
+    context.closeDialog(ignorePopFailure: true).pushNamed(route.name);
```
kumarpalsinh25, line 149: As commented above, not understanding the reason for `ignorePopFailure`

app/lib/common/utils/utils.dart
kumarpalsinh25, line 78: Won't be `andGo`?
Mean close dialog and go to specific route?

kumarpalsinh25 (approved):
Overall code looks good. Very good improvement in routing stuff. May this PR resolve many route issues with dynamic screen sizes. Well done!

Added few comment which need to be answered before merge.

app/lib/common/utils/utils.dart
gnunicorn: No. it means close the dialog if possible, OR if there is nothing to pop go there.

app/lib/common/utils/utils.dart
```diff
@@ -70,8 +71,27 @@ extension RefDebounceExtension on Ref {
...
+    final navigator = Navigator.of(this);
+    try {
+      if (orGo != null && !navigator.canPop()) {
+        navigator.pushReplacementNamed(orGo.name);
+      } else {
+        navigator.pop();
```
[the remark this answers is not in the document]
gnunicorn: I honestly don't really understand why and when this would be needed. Are we sure _it is needed_ and not just the common way replacement of a broken `context.pop`?

But the `closeDialog` doesn't replace any `Navigator.of(context, rootNavigator: true).pop()` at this point either way, so ...

app/lib/common/utils/utils.dart
gnunicorn: there is at least one case where from the context of the current code it isn't clear if that can be issues without there being a place to `pop` to and thus it does `context.canPop() ? context.pop() : context.goNamed(...` . `orGo` This covers that pattern.

app/lib/common/utils/utils.dart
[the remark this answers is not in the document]
gnunicorn: Noticed how even the cookbook example does not use `Navigator.of(context, rootNavigator: true).pop()` when inside the `showDialog` (despite that using `useRootNavigator:true` by default) - https://docs.flutter.dev/cookbook/effects/nested-nav . IF the right `context` is used (the one provided in the build function for the `showDialog`) this should work just fine.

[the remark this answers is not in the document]
kumarpalsinh25: > I honestly don't really understand why and when this would be needed. Are we sure it is needed and not just the common way replacement of a broken context.pop?

Don't exactly remember but in some of the cases while closing dialogs with `context.pop` , it actually close screen open behind it and dialog keeps remain open. So in those cases `rootNavigator` help in closing dialog first.

Let me share one of the screen recoding for reference.

https://github.com/user-attachments/assets/a3131716-0716-4de2-98d4-2eebc1b088b2

In Above video, on click of close button it actually closing up the dialog first, but not exactly remember many time I see the issue like dialog keeps remain open on click of close button also where context.pop() written.

> But the closeDialog doesn't replace any Navigator.of(context, rootNavigator: true).pop() at this point either way, so ...

👍

app/lib/common/utils/utils.dart
kumarpalsinh25: Yes I have read it bit confused over there.

kumarpalsinh25:
@gnunicorn 
I have added my comments but non of things now blocking PR Merge.

app/lib/common/utils/utils.dart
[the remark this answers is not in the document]
gnunicorn: @kumarpalsinh25 but that isn't due to the `context.pop` but instead that we are putting the dialog onto the wrong navigator. the back-button pops on the latest, but the showDialogs are plugged to the root.

This fixes that: https://github.com/acterglobal/a3/pull/1997#issue-2430426341

PR #1987, opened 2024-07-24
extend textController to support triggerStyles

gtalha07:
This brings the ```auto_complete_trigger``` to be included in our main repo source and a new feature to be added in regards of trigger characters. As integrating in the whole package brings large-set of files (mostly which are from base project ), the only relevant and most interesting to see are which I have pointed out over few, can be reviewed and looked upon.

### Preview on Desktop: 

https://github.com/user-attachments/assets/402ea8fe-dbec-4c51-8755-65332b980f25


app/packages/multi_trigger_autocomplete/lib/src/autocomplete_controller.dart
```diff
@@ -0,0 +1,53 @@
+import 'package:flutter/material.dart';
+
+class AutocompleteTextController extends TextEditingController {
```
gtalha07, line 3: This is the main class which extends ```TextEditingController``` to allow us build custom text spans based on trigger characters. ```triggerStyles``` is passed down here from main widget ```MultiTriggerAutoComplete```.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -196,7 +196,29 @@ class _ChatInput extends ConsumerStatefulWidget {
 }
 
 class __ChatInputState extends ConsumerState<_ChatInput> {
-  TextEditingController controller = TextEditingController();
+  late AutocompleteTextController controller;
```
gtalha07, line 199: This now replaces with extended controller ```AutocompleteTextController``` which takes ```triggerStyles```.

app/packages/multi_trigger_autocomplete/lib/src/autocomplete_trigger.dart
```diff
@@ -0,0 +1,118 @@
+import 'package:flutter/material.dart';
+import 'package:multi_trigger_autocomplete/multi_trigger_autocomplete.dart';
+
+/// The type of the [AutocompleteTrigger] callback which returns a [Widget] that
+/// displays the specified [options].
+typedef AutocompleteTriggerOptionsViewBuilder = Widget Function(
+  BuildContext context,
+  AutocompleteQuery autocompleteQuery,
+  AutocompleteTextController textEditingController,
```
gtalha07, line 9: just an update to replace ```TextEditingController``` to ```AutocompleteTextController```.

app/packages/multi_trigger_autocomplete/lib/src/multi_trigger_autocomplete.dart
```diff
@@ -0,0 +1,452 @@
...
+    this.optionsWidthFactor = 1.0,
+    this.debounceDuration = const Duration(milliseconds: 300),
+  });
+
+  /// The triggers that trigger autocomplete.
+  final Iterable<AutocompleteTrigger> autocompleteTriggers;
+
+  /// A optional map of trigger symbols to their corresponding TextStyles.
+  /// This allows for custom styling of different triggers
+  final Map<String, TextStyle>? triggerStyles;
```
gtalha07, line 105: Added param to take trigger word optional style.

app/packages/multi_trigger_autocomplete/lib/src/multi_trigger_autocomplete.dart
```diff
@@ -0,0 +1,452 @@
...
+  ///
+  /// ** See code in examples/api/lib/widgets/autocomplete/raw_autocomplete.focus_node.0.dart **
+  /// {@end-tool}
+  /// {@endtemplate}
+  ///
+  /// required for [AutocompleteTextController].
+  final FocusNode focusNode;
+
+  /// Extended version of [TextEditingController]. This is required for
+  final AutocompleteTextController textEditingController;
```
gtalha07, line 141: This is now required along with ```FocusNode```. And removed ```initialValue``` option.

app/packages/multi_trigger_autocomplete/test/multi_trigger_autocomplete_test.dart
```diff
@@ -0,0 +1,341 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:multi_trigger_autocomplete/multi_trigger_autocomplete.dart';
+
+void main() {
```
gtalha07, line 6: Here is just an update of taking required parameters along with ```AutocompleteTextController```.

app/pubspec.yaml
```diff
@@ -59,6 +59,8 @@ dependencies:
   acter_avatar: ^2.0.1+1
   flutter_easyloading: ^3.0.5
   appflowy_editor: ^3.0.0
+  multi_trigger_autocomplete:
```
gtalha07, line 62: Replaced mainstream with sub-project dependency.

gnunicorn:
@gtalha07 can you elaborate a bit more on the rational for this? Do I understand correctly that in order to do our wanted changes, we have to create a fork? These are not things the upstream project would want or expect in their version? Do you consider the project unmaintained?

gnunicorn:
another side note: it would have been nice if you had one commit to just import the project so that only the changes made could be reviewed easily. This is really not feasible to review now as all files and all lines are marked as new, so I don't know what you've changed ...

gnunicorn (changes requested):
It's really hard to understand what should be reviewed and what should not. Can I ask that you change the history of what to be done here to:
 1. one commit that imports the existing project code, with superfluous stuff removed (the gifs)
 2. one commit to change the name and imports into our code base as is
 3. one or several other commits on top that do the changes on that code base and our app as necessary.

That makes for a cleaner importing of the project as well as for something I can actually review as I know what to look at ...

(you can also combine 1 and 2 if you prefer)

gtalha07:
> @gtalha07 can you elaborate a bit more on the rational for this? Do I understand correctly that in order to do our wanted changes, we have to create a fork? These are not things the upstream project would want or expect in their version? Do you consider the project unmaintained?



> @gtalha07 can you elaborate a bit more on the rational for this? Do I understand correctly that in order to do our wanted changes, we have to create a fork? These are not things the upstream project would want or expect in their version? Do you consider the project unmaintained?

The fork is just there so it's easier to follow how the package is modified which I think might turn out completely different, more aligning to our requirements rather than plugin intended features. I don't think the changes followed here and after, would come up as extended version but rather , completely revised version of it with different way of handling more of the core stuff. Just as an example that I removed some params .i.e. ```initialValue``` and making some required .i.e. ```TextEditingController``` because without those change-set, the styling won't work. I don't consider project unmaintained though. But the way it's structured, it's hard to expect that without major refactoring, it'd come up near close to what I envisioned...

Still I'm open to feedback of how we can structure this version so it doesn't conflict or make it more overshadowed than original but come out as wholly new standalone, based on whole different mechanism but can provide way of extensify features.

gnunicorn:
@gtalha07  Thanks for explaining the why. I'm okay with maintaining a fork in the repo. Looking forward to a more cleaned up version to do the same :) .

PR #1996, opened 2024-07-25
Fix UI bug - '_dependents.isEmpty': is not true

bitfriend:
When navigating to chat room page, we get this error sometimes.
`'_dependents.isEmpty': is not true`
image
We face something like this screenshot, but error message is `'_dependents.isEmpty': is not true` not `'child == _child': is not true`.
It seems to be related with widget key.
https://stackoverflow.com/a/71085742
Will fix non-unique keys thoughout flutter code.


.changes/1996-_dependents.isEmpty-is-not-true.md
```diff
@@ -0,0 +1 @@
+- Fix UI bug - '_dependents.isEmpty': is not true. We can see this error in red screen at chat room page.
```
gnunicorn, line 1: this isn't a useful message to an end-user. better leaving the file empty in this case IMHO.
```suggestion
```

gnunicorn (approved):

PR #1997, opened 2024-07-25
Additional Router Fixes

gnunicorn:
follow up of #1986 , which lead to me more learning about how we are actually doing things wrong:

1. this removes the newly added method and replaces all usage of it and `Navigator.of(..).pop` with the more correctly used `Navigator(context).pop(...)` - which is shorter and better
2. Fixes several places where there was a confusing usage of `ctx` and `context`, ensure that `context` is always the most inner context and thus correctly pops wherever we are
3. Fixes the back-routing issue reported in that PR by kumar by putting dialogs _not_ at the rootNavigator. Specifically, outsources two commonly used ones into a single use that fixes that problem: for redactions and reporting.

Before:

https://github.com/user-attachments/assets/318d75c9-7483-4309-b313-dfe172a159eb

After:

https://github.com/user-attachments/assets/6c1cdf5c-86a7-4fd5-80a2-bf75465f923a

### DIFF
--- a/app/lib/common/actions/redact_content.dart
+++ b/app/lib/common/actions/redact_content.dart
@@ -14,27 +14,53 @@ import 'package:logging/logging.dart';
 
 final _log = Logger('a3::common::redact');
 
+Future<bool> openRedactContentDialog(
+  BuildContext context, {
+  final String? title,
+  final String? description,
+  required final String eventId,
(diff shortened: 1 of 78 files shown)


kumarpalsinh25 (approved):
Changes looks good to me.

PR #1998, opened 2024-07-25
Fix spaces chats screen.

gnunicorn:
Follow up to #1976 :
Also fix the spaces chats screen by using the same feature as the overview widget. Reduces the usage of a custom scroll view and slivers by using the same listviews as for the overview widgets.

Fixes https://github.com/acterglobal/a3-meta/issues/422 .


gtalha07 (approved):
Looks good!

PR #2002, opened 2024-07-26
extend ```multi_trigger_auto_complete``` under ```acter_trigger_auto_complete```.

gtalha07:
This brings the auto_complete_trigger to be included in our main repo source and a new feature to be added in regards of trigger characters. As integrating in the whole package brings large-set of files (mostly which are from base project ), should be __***BEST REVIEWED PER COMMIT***__.

### PREVIEW:

https://github.com/user-attachments/assets/fc61e669-93d6-469e-adf8-cf36143564e1


app/packages/acter_trigger_auto_complete/asset/emoji_demo.gif
gnunicorn, line 1: Can't we collapse the removing of the demo gifs and this one into one? Thus way we are not adding the `gifs` in the first place.

This commit already contains other changes to the code itself like the renaming and the changelog, let's also remove the gifs already. I'd prefer they never made it into the repo ;) .

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -196,33 +196,55 @@ class _ChatInput extends ConsumerStatefulWidget {
...
 
 class __ChatInputState extends ConsumerState<_ChatInput> {
-  TextEditingController controller = TextEditingController();
+  late ActerTextController textController;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    _setController();
+  }
```
gnunicorn, line 205: shouldn't this be a `didWidgetChange`?

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -196,33 +196,55 @@ class _ChatInput extends ConsumerStatefulWidget {
...
+        color: Theme.of(context).colorScheme.onSecondary,
+        height: 0.5,
+        background: Paint()
+          ..color = Theme.of(context).colorScheme.secondary
+          ..strokeWidth = 10
+          ..strokeJoin = StrokeJoin.round
+          ..style = PaintingStyle.stroke,
+      ),
+    };
+    textController = ActerTextController(triggerStyles: triggerStyles);
```
gnunicorn, line 219: why not do that in the class directly?

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:flutter/material.dart';
+
+/// Extended [TextEditingController] which takes up trigger styles as optional.
+/// Styles the trigger inputs based on map for mentions, hashtags or emojis etc.
+/// If provided empty, will
```
gnunicorn, line 5: will ...?

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
```diff
@@ -0,0 +1,55 @@
...
+class ActerTextController extends TextEditingController {
+  ActerTextController({String? text, this.triggerStyles}) : super(text: text);
```
gnunicorn, line 7: I think calling this `ActerTextController` is a bit weird. either call it `ActerTriggerAutoCompleteTextController` or just `TriggerAutocompleteTextController`?!?

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
```diff
@@ -25,7 +25,7 @@ class ActerTextController extends TextEditingController {
       final trigger = entry.key;
       final triggerStyle = entry.value;
 
-      final regex = RegExp('\\$trigger\\S+(?: \\S+)*');
+      final regex = RegExp('\\$trigger\\S+(?: \\S+)*(?=\\s)');
```
gnunicorn, line 28: I'd really like some unit tests for the regexp to understand better what to expect from it... and also understand changes made and why...

gnunicorn (changes requested):
Thanks heaps. In particular for cleaning up the history, that makes it really easy to follow what needs reviews and what doesn't. Unfortunately this history still leaves the gifs and other pointless stuff in the history of our repo (which I'd like to avoid), so I have to ask you to squash the first three commits into one and rebase the rest on top of that ... Other than that, just minor inquiries to be considered.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: That'd mean we are taking away the customization support from the plugin, which is what it needs to offer...

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
gtalha07: Sorry, missed the comment here. Needs correction! 👍🏻

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
gtalha07: First one sounds better with more meaningful to plugin. Will update it!

app/packages/acter_trigger_auto_complete/lib/src/acter_text_controller.dart
gtalha07: Ahh should've it documented explaining why specifically it's used. Also I will think of some basic tests to include hereafter.

gnunicorn (approved):

PR #2003, opened 2024-07-26
Boosting Chat performance

gnunicorn:
This improves the performance of the chat section. See it in action in debug mode on my large account:

https://github.com/user-attachments/assets/7bfb7b6e-d7ca-4037-943f-f66e5e1b2b62

Compare this to how the performance is currently on `main`. Don't mind the red-screens, these appear as rendering the list under debug on my system also used to take 100% on all CPUs (it does not anymore with these changes).

https://github.com/user-attachments/assets/6cd018a9-8be4-4890-a551-9d0acfd0ffb1

Notice also how the changes reduces the flicker a lot. In the background you can see the nightly on the same list as you see on the front. Notice how things flicker in there when we change the order of rooms, compared to the list in the front running these changes:

https://github.com/user-attachments/assets/e17dd24b-f080-4781-b78c-eec39b7f3a13

There are several parts to this improvements, probably best to review commit-by-commit to understand what is happening. In particular the following things to lead to that speed boost:

1. Replacing the `convo`-object in the state that is passed around with bland roomId-strings and manage the rest via riverpod providers: 6f48b7bec225695645c4f0c8846964e6283d556f , 8c4e621beab8cb13e5221faf0a15421b16570c25 . Also reducing the same usage in providers. This massively improves performance for two reasons (I believe): first, it reduces the overhead of having the convo object passed around, but more importantly it reduces the state to a string that is simple to compare and with specific providers taking care of the actual updates, reduces the actual repaints that need to happen. Lastly, by passing only the string-list over we are also just faster in processing it in and off itself.
2. Using a locally managed animated list reduces the rerenders that happen on the chat room itself massively, avoiding changes where they are not necessary and making the entire experience smoother when thing "move" to the top. refs aae59777115d75fad596f06fefdb78ff8e691e7f . Notice this smoother experience:

https://github.com/user-attachments/assets/270f5df5-db2f-4234-b33b-caceb250e2b1

3. Reduce more flickers by having the roomAvatarInfo only updated once the reloading is done, refs 91f03c126c5b990ea5a372d020f4893d19afced4 . Before, any time _anything_ on the room/convo changed or lead to an update trigger, we'd also switch to the fallback for a second, loaded the new Avatar and then displayed that. Leading to an ugly "flicker", even sometimes showing of the fallback avatar. 
4. Switch the chat room list to a SliverList, removing the `shrink: true`-flag 0b60303e7f6ff055bb0e01f02c9072e2ee75ebbd . This _massively_ improves scrolling performance as we do not have multiple scroll controller but a unified one that is being used for the entire screen. This also reduces the pre-renders the internal list does to figure out the size and all the other bad stuff associated with `shrink: true`. This is really massive. 
5. Add a loading skeleton hack for when the first messages are being processed. Doesn't actually improve performance, but gives the impression it does ;) . cc5d27f6ab5873337a2f7d97a05e0944f5780602

On the way, this also updates the matrix-rust-sdk ( f72a67a65f1f9710313dd91126712a03ab39e277 ) and the skeletonizer for their latest features and support. It also adds a diffing utility for managing the animated list model.

Fixes https://github.com/acterglobal/a3-meta/issues/428, https://github.com/acterglobal/a3-meta/issues/414 by using the `waitForConvo` method for all chats now.


app/lib/common/providers/notifiers/room_notifiers.dart
```diff
@@ -41,3 +44,82 @@ class AsyncMaybeRoomNotifier extends FamilyAsyncNotifier<Room?, String> {
...
+    );
+    _displayNameListener = ref.listen(
+      roomDisplayNameProvider(roomId),
+      (previous, next) => _maybeUpdate(roomId),
+    );
+
+    ref.onDispose(() {
+      _displayNameListener.close();
+      _avatarListener.close();
+    });
```
gnunicorn, line 77: `ref.watch` (what we had before) comes with the nasty side-effect of returning us `null` for `valueOrNull` on any type of `loading` taking place. That is pretty annoying as it a) means we are updating the internal object several times (once for loading to start and once for when the new data is available), but also that we produce annoying flickerings because we are resetting the internal value to `null` rather than from the oldValue -> newValue directly.

Using these listeners and the `maybeUpdate` which gradually only updates the features that _have new data availalbe_ (as in, aren't loading/refreshing/revalidating anymore) means we directly switch from oldValue -> newValue and do updates only in that case at all. Reducing the flickering in the UI that is currently caused on every room-update-trigger.

app/lib/common/widgets/chat/convo_card.dart
```diff
@@ -37,56 +40,35 @@ class ConvoCard extends ConsumerStatefulWidget {
...
-      showSelectedIndication: widget.showSelectedIndication,
-      subtitle: latestMsg != null
-          ? _SubtitleWidget(
-              room: widget.room,
-              latestMessage: latestMsg,
-            )
-          : const SizedBox.shrink(),
-      trailing: widget.trailing ?? renderTrailing(context),
+      onTap: onTap,
+      title: titleIsLoading ? Skeletonizer(child: Text(roomId)) : null,
```
gnunicorn, line 54: notice how we are here showing a `Skeletonizer` when the `displayName` isn't available yet. this reduces the times we see `roomId`s shown in the UI for chat altogether.

gnunicorn (commented):
Just some minor comments to give more context to specific changes.

bitfriend:
Congrats, @gnunicorn 🎉.
Your above videos showed the great improvement in speed 💪.

gnunicorn:
probably also fixing https://github.com/acterglobal/a3-meta/issues/395 and https://github.com/acterglobal/a3-meta/issues/394

bitfriend:
https://github.com/acterglobal/a3/commit/8c4e621beab8cb13e5221faf0a15421b16570c25 reduced the loading time of chat list page greatly.
Good job.
`retry` argument of `convo_with_rety()` seems to be needless.

bitfriend:
[Replacing convo with string](https://github.com/acterglobal/a3/commit/6f48b7bec225695645c4f0c8846964e6283d556f) reduced time for comparison.
Great discovery.

bitfriend (approved):
Great discovery.
It reduced loading time of list significantly.

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -180,7 +171,8 @@ final filteredChatsProvider =
 final isRoomEncryptedProvider =
     FutureProvider.family<bool, String>((ref, roomId) async {
   final convo = await ref.watch(chatProvider(roomId).future);
-  return await convo.isEncrypted();
+  // FIXME: unify this over all rooms.
+  return (await convo?.isEncrypted()) == true;
```
gtalha07, line 175: Oh makes sense, ignore my above comment ;)

gtalha07 (approved):
Nice thinking out of the box and bringing performance boost!

app/lib/common/widgets/spaces/space_info.dart
```diff
@@ -12,11 +12,6 @@ final isActerSpaceForSpace =
   return await space.isActerSpace();
 });
 
-final isEncryptedForSpace =
```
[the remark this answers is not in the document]
gnunicorn: it wasn't used anywhere

PR #2005, opened 2024-07-28
Attachment Fixes

### LINKED ISSUE #1999: Chat Attachments are not working
After recent router related PR, chat attachment dialog gets close and but not able to post any attachment.

https://github.com/user-attachments/assets/5bbd3f55-b593-4caf-90a3-d7aea4ddc4b5

gnunicorn:
Refactors the Attachments to:
1. Fixes #1999 which was caused by us popping the same context, we then checked of being mounted (which it then isn't anymore): 1529668387c5be831f12d330669333ce2c0cc098
2. use the same flow, widgets and process for attachments and chat attachments, reducing code and error surface. Also simplifying the process itself and reducing the code base of almost duplicate code: b566842f1f19ba2cf5731b882f70380e45dcc519
3. replace the attachment redaction code with the generic content redaction code, fixing the error seen at the end of the first screencast:

It works now:

https://github.com/user-attachments/assets/1310e0be-e2d8-404e-af80-07bea64d01cf

And the fix for the error at the end:

https://github.com/user-attachments/assets/20fa2885-699e-40be-b64b-8e66a9dc4b3a


kumarpalsinh25 (commented):
Overall changes looks good to me.

Still I would request @gtalha07 to review this once as he has worked on this prior and have better idea on this.

gtalha07 (approved):
Looking good!

PR #2006, opened 2024-07-28
Invite Users, refresh space overview

gnunicorn:
Two minor UX improvements (and one filename typo fix):
1. When inviting from several spaces, we now show a progress indicator while loading, c0690d5b75f00ac8f474e08c10f97874d4c7a240  fixes https://github.com/acterglobal/a3-meta/issues/418

https://github.com/user-attachments/assets/9edde2f4-d39b-4f72-9a7d-bc4378c6ddb6

2. We still have some issues with the chat/spaces list not always refreshing properly. To get around that, with 73a3bdf823acdd12ea6082b108412e7d16edf8d4 you can now "pull-to-refresh" and we'd reevaluate the lowest space and room providers in the stack, leading to a data refresh:

https://github.com/user-attachments/assets/58989e57-cc66-49a0-a6ca-4549819d445a

PR #2007, opened 2024-07-28
Closing rooms

### LINKED ISSUE #1934: Closing Space & Chat

gnunicorn:
Fixes #1934 by adding a new feature available to all admins in their own space and room to "close" that room:

Example with a space:

https://github.com/user-attachments/assets/3254a4fe-74a7-4b3f-a185-5887633c7dbe

Example with a chat:

https://github.com/user-attachments/assets/03ba01d4-45f2-4247-bc28-99aafe6ac0f8

As described in the text to the user, when the user issues this (long-press protected) feature, it will:
1. kick all other users with powerlevels lower than theirs
2. remove the child from all parents, where the the user has permissions to issue that
3. set the join-level to "private" and leaves the room. 

To the reviewers: that was built as part of #2006 at first but then I realised it is big enough in its own right and issued a separate PR for it. But that means it still has the history of the other PR in it. For you only reviewing dfbaf01e6b608e74453433ed8ef0b63e3301a70f is relevant though.


gtalha07 (approved):
LGTM!

kumarpalsinh25 (approved):
Nice Work!

PR #2012, opened 2024-07-30
Chat/Spaces section calculation fixes the XIth 

gnunicorn:
Okay, I had it now. Another weird edge case of the [chats/spaces section in the overview not behaving properly](https://github.com/acterglobal/a3-meta/issues/435). Screws this, I am adding tests!

This brings:
1. a refactor of the complex code to calculate the configuration of a space/chat section into its own helper function, reused in both sections: 796604e083f7d04e158cced5af60382454104742
2. tests for that helper function: 796604e083f7d04e158cced5af60382454104742
3. a new tests that shows that bug and then a fix to it, as its own commit 796604e083f7d04e158cced5af60382454104742
4. running `flutter tests` for unit tests in `app` as a CI job 56b45f3cebd2f1e006f4682d55106acaebceb37a

Note: the CI is currently failing due to other old test not passing. I'll put that and some other clean up into a follow up.

### DIFF
--- a/app/test/features/spaces/space_section_calculator_tests.dart
+++ b/app/test/features/spaces/space_section_calculator_tests.dart
@@ -1,0 +1,71 @@
+import 'package:acter/features/space/widgets/related/util.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('Related Widget Calculations', () {
+    test('too many local', () {
+      final config = calculateSectionConfig(
+        limit: 3,
+        localListLen: 5,
+        remoteListLen: 0,
+      );
+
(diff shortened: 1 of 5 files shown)


app/test/features/spaces/space_section_calculator_tests.dart
```diff
@@ -0,0 +1,84 @@
+import 'package:acter/features/space/widgets/related/util.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
gtalha07, line 4: 🎉

gtalha07 (approved):

Issue #2024, opened 2024-08-02
[tracking] Post mentions upgrade fallout


gnunicorn:
We've identified a few problems around the chat text bar since we merged the new mentions feature that need to be dealt with (no particular order):

- [x] #2022
- [x] #2021
- [x] #1989
- [x] #1988
- [x] https://github.com/acterglobal/a3-meta/issues/444
- [x] https://github.com/acterglobal/a3-meta/issues/441
- [x] https://github.com/acterglobal/a3-meta/issues/440
- [x] https://github.com/acterglobal/a3-meta/issues/448
- [x] https://github.com/acterglobal/a3/issues/2040
- [x] https://github.com/acterglobal/a3/issues/2049
- [x] https://github.com/acterglobal/a3-meta/issues/457

gtalha07:
Closing milestone as completed!

PR #2026, opened 2024-08-02
Post mention upgrade fixes

gtalha07:
- Tracking #2024.

### DIFF
--- a/app/lib/features/chat/widgets/custom_input.dart
+++ b/app/lib/features/chat/widgets/custom_input.dart
@@ -717,6 +717,20 @@ class _TextInputWidgetConsumerState extends ConsumerState<_TextInputWidget> {
     }
   }
 
+  // adds new line
+  void _insertNewLine() {
+    final TextSelection selection = widget.controller.selection;
+    if (selection.isValid) {
+      final String text = widget.controller.text;
+      final int start = selection.start;
+      final newText = text.replaceRange(start, selection.end, '\n');
+      widget.controller.value = TextEditingValue(
+        text: newText,
+        selection: TextSelection.collapsed(offset: start + 1),
+      );
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
     ref.listen(chatInputProvider, (prev, next) {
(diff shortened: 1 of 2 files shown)


app/packages/acter_trigger_auto_complete/lib/src/acter_trigger_autocomplete.dart
```diff
@@ -377,13 +377,13 @@ class MultiTriggerAutocompleteState extends State<MultiTriggerAutocomplete> {
 
   @override
   void dispose() {
-    _textEditingController.removeListener(_onChangedField);
-    if (widget.textEditingController == null) {
+    if (!mounted) {
```
gtalha07, line 380: @gnunicorn, I think we don't want dispose off controller here at all. As doing interaction on chat screen causes this to dispose which I'm not sure why...

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -236,6 +237,14 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
       baseOffset: cursorPos + emojiLength,
       extentOffset: cursorPos + emojiLength,
     );
+
+    // Ensure we keep the cursor up
+    // frame delay to keep focus connected with keyboard.
+    if (!chatFocus.hasFocus) {
+      Future.delayed(const Duration(milliseconds: 100), () {
```
gnunicorn, line 244: if we want this to happen for a "frame delay" shouldn't this be done in a `WidgetsBinding.instance.addPostFrameCallback((Duration duration) {` rather than a `Future.delayed`?!?

gnunicorn (approved):
From the code I have little idea whether this fixes the issues, but I suppose you did test this ...

it would be great if we had widget tests and/or other automatic tests for this, as it is a pretty vital component in our system with many features and states and it is easy to break some of them unknowingly ...

app/lib/features/chat/widgets/custom_input.dart
gtalha07: Yeah I suppose it can be more sensible to use it! ;)

Issue #1935, opened 2024-07-10
Push notifications does not redirect to the correct page/chat.


emilvincentz:
Experienced by several user from ActionAid Italy

gnunicorn:
where does it go? Does it only open the app?

emilvincentz:
It opens the app, but instead of going to the specific chat, it goes to the front page

emilvincentz:
several users mentioned this as well, causing connfusion.

PR #2028, opened 2024-08-04
Desktop notifications support

gnunicorn:
This:
- fixes the dart analyzers for the plugins ecda3c7d7bddbccd41cf32b62366630ad9f273ef
- refactors the notifications code into a separate plugin, fe257cf7ebb1b4ba9a1d152269623e1cfa7faa4c
- adds NTFY support to allow for Desktop Notifications support (while the app is running)
- add separate windows notification support via `windows_notifications` in the plugin
- refactor into platform specific files for better upcoming maintainability
- Refactors the router and passes the initial payload if were started by clicking on a notification, fixes https://github.com/acterglobal/a3/issues/1935

On Windows:

https://github.com/user-attachments/assets/7f133053-4e2d-428f-8a4f-c5f62c76093c

On MacOS:
Notice that Quicktime is putting the screen into no-notification-mode that's why I have to navigate to the notification center for the video. But that is a video-recording-only behavior:

https://github.com/user-attachments/assets/57b19f79-0a49-4f92-ba5a-7e6492d4f565

Tasks:
 - [x] lacks windows support (maybe putting that into a follow-up though)
 - [ ] NTFY needs to restart on failure
 - [ ] NTFY should be behind an access token


gtalha07 (approved):
Looks good!

gnunicorn:
@gtalha07 I've added some more router refactoring so I asked for another review. It's enough if you only review those changes. They are about the last bullet point in the description.

gnunicorn:
@gtalha07 naah, I'm good for now. I think I'll postpone the other stuff to later to not make this blow up even more ...

packages/acter_notifify/test/acter_notifify_test.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:flutter_test/flutter_test.dart';
+
+
+void main() {
+  test('import smoketest', () {
+    // doesn't do anything
```
gtalha07, line 6: Are these intended to be done later?

gtalha07 (approved):
Thanks. The changes , particularly the routing one which I've looked more closely, seems good and much better code readable now. Do you think we can plan including tests for it, ensuring it doesn't break along the way? Not in this PR but ticket can be opened.

PR #2038, opened 2024-08-05
Enven support

gnunicorn:
This adds [`enven`](https://pub.dev/packages/enven) for simplified configuration options. See [it's docs for details](https://pub.dev/packages/enven).

tl;dr: we now have a `.env` file in the `app` and you can overwrite any value in it by adding another `.env.dev` file locally for your setup (do not commit it). Don't forget to run `dart run enven` (in `app`) initially as well as when you change either of these files to create the  `app/lib/config/env.g.dart` file that contains the generated `Env` with the `const`s on it.

For the reviewer: this also updates the workflows to use that new system to create the `.env.production` overwrite (rather than `--dart-define` as we used to do) and some minor fixes to the workflows. It also required a bit of rewriting and rewiring of the initialization procedure of the rust-package to allow passing the `const`s in properly, hence that number of changes.

### DIFF
--- a/app/integration_test/support/mail.dart
+++ b/app/integration_test/support/mail.dart
@@ -2,11 +2,11 @@
 
 import 'dart:convert';
 import 'package:acter/common/utils/constants.dart';
+import 'package:acter/config/env.g.dart';
 import 'package:acter/features/home/data/keys.dart';
 import 'package:acter/features/search/model/keys.dart';
 import 'package:acter/features/settings/widgets/email_address_card.dart';
 import 'package:acter/features/settings/widgets/settings_menu.dart';
-import 'package:acter_flutter_sdk/acter_flutter_sdk.dart';
 import 'package:convenient_test_dev/convenient_test_dev.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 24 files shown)


app/integration_test/support/mail.dart
```diff
@@ -111,7 +111,7 @@ extension ActerMail on ConvenientTest {
   }) async {
     var link = await getLinkInLatestEmail(emailAddr, contains: contains);
     if (replaceWithHomeserver) {
-      final homeserver = Uri.parse(defaultServerUrl);
+      final homeserver = Uri.parse(Env.defaultHomeserverUrl);
```
bitfriend, line 114: All occurrences of `String.fromEnvironment()` or `bool.fromEnvironment()` should be replaced with `enven`?

gnunicorn: Yes, they are not built with the environment anymore but from `Env` now.

gnunicorn:
Proof in the demo build that the `.env.production` was used properly for the build (notice the `version` string)
image

bitfriend (approved):
Great solution.
Confirmed it works on my windows.
Looks good to me.

PR #2039, opened 2024-08-05
Quickjump routing fixes

gnunicorn:
We neglected to look after whether we are in the dialog or page screen mode in the updates of QuickJump. As a result issues like https://github.com/acterglobal/a3-meta/issues/449 occurred. This PR fixes that and navigation from quick jump works again fine

**on mobile**

https://github.com/user-attachments/assets/77d46f81-bc33-444e-b9ee-cec3f41e5073

and **on desktop**

https://github.com/user-attachments/assets/d9ce34bc-403f-47de-83d3-dca887f0cc55

Fixes https://github.com/acterglobal/a3-meta/issues/449

### DIFF
--- a/.changes/meta-449.md
+++ b/.changes/meta-449.md
@@ -1,0 +1,1 @@
+- Fix bug in quick jump navigation, where it used to go black rather than properly jump sometimes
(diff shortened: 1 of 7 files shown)


kumarpalsinh25 (approved):
LGTM!

PR #2042, opened 2024-08-06
Links on attachments

gnunicorn:
This is the upcoming API for links in attachments. Also contains a small fix to expose `.name` on the attachment itself so we can show it. We might have to add an additional component later to allow for changing/updating that.

//cc @kumarpalsinh25

### DIFF
--- a/native/core/src/events/attachments.rs
+++ b/native/core/src/events/attachments.rs
@@ -115,7 +115,7 @@ pub enum AttachmentContent {
     File(FileMessageEventContent),
     /// A location attachment.
     Location(LocationMessageEventContent),
-    /// A location attachment.
+    /// A link attachment.
     Link(LinkAttachmentContent),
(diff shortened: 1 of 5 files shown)


native/core/src/events/attachments.rs
```diff
@@ -73,19 +92,31 @@ impl FallbackAttachmentContent {
...
-    /// An audio message.
+    /// An audio attachment.
     Audio(AudioMessageEventContent),
-    /// A file message
+    /// A file attachment
     File(FileMessageEventContent),
-    /// A location message.
+    /// A location attachment.
     Location(LocationMessageEventContent),
+    /// A location attachment.
```
bitfriend, line 118: You might want to use the following comment.
```suggestion
    /// A link attachment.
```

bitfriend (approved):
Looks good to me.

PR #2050, opened 2024-08-08
Fix shake detection

gnunicorn:
By including a newer version of the shake_detector plugin and fixing it up to use more accurate and better detection mode. 

We might still want to increase the threshold though, have to check this in the upcoming nightlies.

Reviewers note: the first two commits are `git subtree` of just the upstream version. it is enough to review the last two commit 7ddf5cc680bacea0cba293b99b831c53779b48d9 and 4004e5156e43d37d8698c99132854dae300cb3a3 .

Fixes https://github.com/acterglobal/a3-meta/issues/281 and https://github.com/acterglobal/a3-meta/issues/453


kumarpalsinh25 (approved):
LGTM!

Issue #2040, opened 2024-08-06
selecting text doesn't work in chat editor


gnunicorn:
I can't move the cursor for dome reason:

https://github.com/user-attachments/assets/76953857-4e4c-4d85-a7b2-839908480c69

gtalha07:
@gnunicorn I've tried reproducing this case, both on iOS and android. It seems to move cursor correctly when tapping sub-parts. See the video recording:

https://github.com/user-attachments/assets/d31cb7b3-8b99-42ec-be9a-4b0172a3c880

However, I seem to discover another bug, which keeps shifting cursor to end of text as you can see in video. That shouldn't happen...

gnunicorn:
@gtalha07 that was on the last release from last week. maybe your recent changes already fixed that? I will try again with the latest release from today.

gnunicorn:
It's fixed in the latest release from today

Issue #456, opened 2023-01-24
news image glitching: reverse colors


gnunicorn:
When resizing my linux desktop effektio app from widescreen to portrait mode, it often comes to glitches in the news image as seen below. Resizing the window over a certain point fixes it again, but until I do that it stays this way - also applies to the other images that follow.

![grafik](https://user-images.githubusercontent.com/40496/214274034-02369465-fd8f-4675-8349-0262ddaedfa9.png)

gtalha07:
It is because of flag we have set ```debugInvertOversizedImages = true``` and implies that we need to optimize the decoded image resolution as its taking more memory than expected.

PR #2051, opened 2024-08-08
Various chat input fixes

### LINKED ISSUE #2049: Edit message bugs + line jump
When you swipe to edit message, the old text is not displayed in editor.

Also, when re-editing the message, the initial message is shown. See video below.

Lastly, jump to new line to make a paragraph does not show properly. 
Eg. Typing this:
Line 1
Line 2 

Line 4

Is shown like this:

Line 1
Line 2 
Line 4

https://github.com/user-attachments/assets/08106dd9-eac7-4f92-8abf-431f9efcb0db

gtalha07:
### Fixes the following issues by order of commit:
- Fix for cursor jumping at end of text length when try selection editing. bf623b1 (Discovered in [#2040](https://github.com/acterglobal/a3/issues/2040#issuecomment-2275657319)).
- Optimize suggestion container height and require at least one alphabet to trigger suggestions. cda65d4
- Fix #2049. Commits -> 5fef663 && 8687e81  | ```flutter_matrix_html``` [fix](https://github.com/acterglobal/flutter_matrix_html/commit/23de151fefc012b25649321742cc2c3c2dca6b76)
- Fix meta issue [#456](https://github.com/acterglobal/a3-meta/issues/456). b67835d


app/lib/features/chat/widgets/pill_builder.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:acter/common/toolkit/buttons/room_chip.dart';
+import 'package:acter/common/toolkit/buttons/user_chip.dart';
+import 'package:flutter/material.dart';
+
+Widget pillBuilder({
```
gnunicorn, line 5: let's make this an actual widget please.

app/lib/features/chat/utils.dart
```diff
@@ -8,239 +10,292 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
-    case 'm.sticker':
-    case 'm.room.redaction':
-    case 'm.room.encrypted':
-      // supported
-      return true;
-    case 'm.room.message':
-      // special case of supporting locations
-      return message.metadata?['msgType'] == 'm.location';
-  }
+class ChatUtils {
```
gnunicorn, line 15: why wrap all that into a class with static functions? What's the benefit of that?

app/lib/features/chat/utils.dart
gnunicorn, line 1: I assume that aside from wrapping these into a class nothing changed - the diff is impossible to read - and will skim over the changed.

app/lib/features/chat/widgets/bubble_builder.dart
```diff
@@ -21,7 +21,9 @@ import 'package:swipe_to/swipe_to.dart';
 
 final _log = Logger('a3::chat::bubble_builder');
 
-class BubbleBuilder extends ConsumerWidget {
+// keep this widget stateful consumer
+// otherwise doesn't hold correct message state
+class BubbleBuilder extends ConsumerStatefulWidget {
```
gnunicorn, line 26: but ... does that mean our internal provider _doesn't_ have the correct state? Does that mean that once I leave the room and re-enter it, it will show the wrong state? Why would we make the _widget_ be the single source of truth about state if the we used _riverpod_ for state management?

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -685,10 +681,21 @@ class _TextInputWidgetConsumerState extends ConsumerState<_TextInputWidget> {
...
   void initState() {
     super.initState();
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      widget.controller.text = ref.read(
-        chatInputProvider.select((value) => value.message),
-      );
+    ref.listenManual(
+        chatInputProvider.select((state) => state.selectedMessageState),
+        (prev, next) {
+      if (next == SelectedMessageState.edit) {
```
gnunicorn, line 687: there was a comment here explaining what is happening. Why did you remove that?

app/lib/features/chat/widgets/mention_profile_builder.dart
```diff
@@ -54,55 +54,52 @@ class MentionProfileBuilder extends ConsumerWidget {
...
-                  onTap: () {
-                    final autocomplete = MultiTriggerAutocomplete.of(context);
-                    ref
-                        .read(chatInputProvider.notifier)
-                        .addMention(displayName, id);
-                    return autocomplete.acceptAutocompleteOption(
-                      displayName.isNotEmpty ? displayName : id.substring(1),
+        return Card(
+          margin: const EdgeInsets.symmetric(horizontal: 8),
+          elevation: 2,
```
gnunicorn, line 59: what's the reasoning for removing the sized box?

app/lib/features/chat/widgets/text_message_builder.dart
```diff
@@ -131,43 +130,48 @@ class _TextWidget extends ConsumerWidget {
...
-                            color: isNotice
-                                ? Theme.of(context)
-                                    .colorScheme
-                                    .onSurface
-                                    .withOpacity(0.5)
-                                : null,
-                          ),
-                  maxLines: isReply ? 3 : null,
+        enlargeEmoji
+            ? Text(
```
gnunicorn, line 137: so... you removed `messageWidth` being used, but it is still part of the widget's _required_ parameters ... why?

and doesn't this means messages now occupy the entire width? I don't think we want that..

app/lib/features/member/dialogs/show_member_info_drawer.dart
```diff
@@ -10,18 +11,30 @@ Future<void> showMemberInfoDrawer({
...
+          builder: (context) => AlertDialog(
+            content: MemberInfoDrawer(
+              key: key,
+              roomId: roomId,
+              memberId: memberId,
+              isShowActions: isShowActions,
+            ),
+          ),
+        )
+      : await showModalBottomSheet(
```
gnunicorn, line 26: what is the reasoning behind this? why didn't say we want user-dialogs to be center dialogs on larger screens ... is that fixing anything?

gnunicorn (changes requested):
There are several changes in this PR that are hard to follow and while the commits are linked to specific fixes the _why_ of the specific changes isn't explained: e.g. Why are we switching from the provider-based messaging to having it on the text controller again? What does it fix and what about the potential fallout of this change? Why put the helper functions into a class? To name a few.

It is hard to judge whether these changes are necessary, actually fix the thing and moreover don't have other side-effects.

This is a particuarly complex piece of UI-code, there are many things interconnected and interwoven and a lot can go wrong when changing things. Unfortunately we lack tests for this and  thus every change needs to be carefully considered and all its side-effects analyzed and looked at. 

For the future may I recommend that if a single commit fixes a single thing, make that single commit its own PR. That way it is easier to review and discuss potential problems, than in a collection of things. And if things built upon each other, that's fine to, just put a reference in the second PR saying that it is built on top of the other. But right now, I can't sign off on this as I understand what the reasons are for most changes, it's hard to understand what its implications are and whether they have been addressed.

app/lib/features/chat/utils.dart
gtalha07: This is just an attempt to improve methods identification. While I see we have many utilities now for each features, this is becoming confusing as most of them follow the same functional naming but hard to judge where it's actually intended and used. Having class wrapper under them is I think a good way for recognition and readability.

app/lib/features/chat/utils.dart
gtalha07: Yes, nothing changed as such aside from it.

app/lib/features/chat/widgets/custom_input.dart
gtalha07: I've skimmed it down to only listen ```SelectedMessageState``` events because keeping ```SelectedMessage``` is causing inconsistent frequent re-renders which resets the controller which is what we've experienced in past to avoid mixing text editing controllers with riverpod. I will bring back the comment with updated action it does.

app/lib/features/chat/widgets/mention_profile_builder.dart
gtalha07: The ```SizedBox``` was acting as constraint to limit the height of suggestions list or it was thought to be. But it came to discovery that it is ineffective of resizing layout of ```Overlay``` (Portal) of internal mentions plugin which seems to have its mechanism of sizing using ```widthFactor``` and ```heightFactor```. In cda65d4, you can see I've extended plugin to include ```heightFactor``` param to adjust suggestion height.

app/lib/features/member/dialogs/show_member_info_drawer.dart
gtalha07: I thought of it as UX-improvement as bottomsheet aren't usually well suited on larger screens, especially on large screenWidth having split views...

app/lib/features/chat/widgets/text_message_builder.dart
gtalha07: Oh I think this was unintentional. Apologies! I will undo it.

app/lib/features/chat/widgets/bubble_builder.dart
gnunicorn: but the question is why? Why would that come about?

app/lib/features/chat/providers/notifiers/chat_input_notifier.dart
```diff
@@ -23,45 +21,10 @@ class ChatInputNotifier extends StateNotifier<ChatInputState> {
     );
   }
 
-  void updateMessage(String value) {
-    state = state.copyWith(message: value);
-  }
-
   void setEditMessage(Message message) {
```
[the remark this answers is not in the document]
gnunicorn: but... if we _removed_ it from the provider then we won't be holding the proper state anymore on re-entry now, do we? Like the messages are gone that we typed ... that is clearly a regression to what we had before, no?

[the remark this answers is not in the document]
gtalha07: They aren't being added anyway as we aren't holding state of typed messages in ```ActerTriggerTextEditingController```. This change is only affecting the part ```onChanged:``` of text field where it's re-evaluating every time as message is composed, which sounds bad as the controller part resets in result of riverpod updates and widget re-rendering. I think it's much more safe to handle this differently using ```ComposerDraft``` and using that to update controller. It's already in the [works](https://github.com/acterglobal/a3/pull/2056)

Issue #1924, opened 2024-07-05
switch to using `ComposerDraft` from the matrix-sdk for chat message compsing (rather than riverpod)


gnunicorn:
As one problem with the chat input field is clearly caused by our combination of riverpod provider and the controllers messing with each other, we might want to look at the [ComposerDraft the matrix-sdk people have recently introduced](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.ComposerDraft.html#). This has the same function of allowing to keep the composing state of a chat message around when the user leaves the room, but with the additional advantage that it is also stored in the local database between restarts ... pretty neat!

gtalha07:
Completed in #2056.

PR #2056, opened 2024-08-12
compose draft state for the room message

gtalha07:
- Addition and expose ```ComposeDraft``` methods over ffi to get ```ComposerDraft``` for storing composed room messages. refs #1924.

### DIFF
--- a/native/acter/src/api/convo.rs
+++ b/native/acter/src/api/convo.rs
@@ -2,13 +2,13 @@ use acter_core::{statics::default_acter_convo_states, Error};
 use anyhow::{bail, Context, Result};
 use derive_builder::Builder;
 use futures::stream::{Stream, StreamExt};
-use matrix_sdk::{executor::JoinHandle, RoomMemberships};
+use matrix_sdk::{executor::JoinHandle, ComposerDraft, ComposerDraftType, RoomMemberships};
 use matrix_sdk_ui::{timeline::RoomExt, Timeline};
 use ruma::assign;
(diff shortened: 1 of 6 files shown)


native/acter/src/api/convo.rs
```diff
@@ -313,6 +313,99 @@ impl Convo {
...
+
+            ("reply", id) => {
+                if let Some(id) = id {
+                    ComposerDraftType::Reply {
+                        event_id: OwnedEventId::try_from(id)?,
+                    }
+                } else {
+                    bail!("Invalid event id or not found");
+                }
+            }
```
gnunicorn, line 379: can be simplified in the match...
```suggestion
            ("edit", Some(id)) => ComposerDraftType::Edit {
                        event_id: OwnedEventId::try_from(id)?,
                    },
            ("reply", Some(id)) => ComposerDraftType::Reply {
                        event_id: OwnedEventId::try_from(id)?,
                    },
            ("reply", _)  | ("edit", _) => bail!("Invalid event id");
```

native/acter/src/api/convo.rs
```diff
@@ -313,6 +313,99 @@ impl Convo {
...
+            _ => bail!("Invalid draft type"),
```
gnunicorn, line 380:
```suggestion
            (draft_type, _) => bail!("Invalid draft type {draft_type}"),
```

gnunicorn (approved):
this only contains the rust side for the composer, so it doesn't yet fix the riverpod replacement issue. But aside from some code improvement this is just fine.

PR #2066, opened 2024-08-15
Unit Tests for custom chat input

gnunicorn:
Showing at least two UX errors already, which are skipped at the moment:
 - putting it into reply-mode while typing makes the cursor skip
 - clearing the text doesn't actually hide the send-button - which indicates the state updates aren't linked up properly.

### DIFF
--- a/app/lib/features/chat/widgets/custom_input.dart
+++ b/app/lib/features/chat/widgets/custom_input.dart
@@ -44,7 +44,16 @@ final _allowEdit = StateProvider.family<bool, String>(
   ),
 );
 
+final canSendProvider = FutureProvider.family<bool?, String>(
+  (ref, roomId) async {
+    final membership = ref.watch(roomMembershipProvider(roomId));
+    return membership.valueOrNull?.canString('CanSendChatMessages');
+  },
+);
+
 class CustomChatInput extends ConsumerWidget {
+  static const noAccessKey = Key('custom-chat-no-access');
+  static const loadingKey = Key('custom-chat-loading');
   final String roomId;
   final void Function(bool)? onTyping;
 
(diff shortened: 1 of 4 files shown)


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -44,20 +44,25 @@ final _allowEdit = StateProvider.family<bool, String>(
...
 
+final canSendProvider = FutureProvider.family<bool?, String>(
+  (ref, roomId) async {
+    final membership = ref.watch(roomMembershipProvider(roomId));
+    return membership.valueOrNull?.canString('CanSendChatMessages');
+  },
+);
+
 class CustomChatInput extends ConsumerWidget {
+  static const noAccessKey = Key('custom-chat-no-access');
```
gtalha07, line 55: I'm okay with having keys here for now but for future reference, let's keep these under separate files where it's easier to read up and link ;)

app/test/features/chat/custom_input.dart
```diff
@@ -0,0 +1,222 @@
...
+    },
+  );
+
+  group('Custom Chat Input - Edit States', () {
+    final overrides = [
+      canSendProvider.overrideWith((ref, roomId) => true),
+      isRoomEncryptedProvider.overrideWith((ref, roomId) => true),
+    ];
+
+    testWidgets(
```
gtalha07, line 87: TextEditingControllers value along with riverpod provider didn't worked for us before, so better is to listen controller which act as ```Stream``` under and use ```notifyListeners``` method. I'll add a fix and use this test to verify! 👍🏻

gtalha07 (approved):
Thanks! a good starting point to effectively remove incorrect state updates.

app/lib/features/chat/widgets/custom_input.dart
gnunicorn: I disagree, having them as static on the main Objects makes it immediately clear where they are used to be for and nicely namespaces them. We had them in other files and they started to rod, meaning that we collected Keys that were never used and didn't serve any purpose. This way we keep them close to where they are used and through the classname their use is totally clear. I'd prefer to do it this way.

PR #2069, opened 2024-08-15
[Chat Input ] Edit message cursor jump fix

gtalha07:
- Fixes[ #meta-issue-468](https://github.com/acterglobal/a3-meta/issues/468). The changes include removing ```message``` string we were storing in ```chatInputProvider``` for tracking send button updates. This combination with ```TextEditingController``` is problematic as what've seen from the past where the states aren't properly synced in between.  The ```ValueListener``` over controller to listen controller changes directly, provides more robustness to send button state updates. Another side-effect of riverpod  with ```TextEditingController``` is cursor jumps when editing message in selection .i.e. controller state resets upon riverpod ```updateMessage()``` calls which also shifts cursor to the end of selection.

https://github.com/user-attachments/assets/9f8c222c-a2f1-479f-8788-17d60792b3e4

### DIFF
--- a/app/lib/features/chat/utils.dart
+++ b/app/lib/features/chat/utils.dart
@@ -1,5 +1,6 @@
 import 'package:acter/common/providers/room_providers.dart';
 import 'package:acter/common/toolkit/buttons/primary_action_button.dart';
+import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter/features/room/actions/join_room.dart';
 import 'package:acter/router/utils.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
@@ -8,6 +9,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:html/dom.dart' as html;
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
+import 'package:html/parser.dart';
 
 //Check for mentioned user link
 final mentionedUserLinkRegex = RegExp(
@@ -243,4 +245,33 @@ String prepareMsg(MsgContent? content) {
     matrixLinks,
     (match) => '<a href="${match.group(0)}">${match.group(0)}</a>',
   );
+}
+
+String parseEditMsg(WidgetRef ref) {
+  // shouldn't ever happen to be null if state is edit or reply
+  final message = ref.read(chatInputProvider).selectedMessage!;
+
+  if (message is types.TextMessage) {
+    // Parse String Data to HTML document
+    final document = parse(message.text);
+
+    if (document.body != null) {
+      // Get message data
+      String msg = message.text.trim();
+
+      // Get list of 'A Tags' values
+      final aTagElementList = document.getElementsByTagName('a');
+
+      for (final aTagElement in aTagElementList) {
+        final userMentionMessageData =
+            parseUserMentionMessage(msg, aTagElement);
+        msg = userMentionMessageData.parsedMessage;
+      }
+
+      // Parse data
+      final messageDocument = parse(msg);
+      return messageDocument.body?.text ?? '';
+    }
+  }
+  return '';
 }
--- a/app/lib/features/chat/utils.dart
+++ b/app/lib/features/chat/utils.dart
@@ -1,6 +1,5 @@
 import 'package:acter/common/providers/room_providers.dart';
 import 'package:acter/common/toolkit/buttons/primary_action_button.dart';
-import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter/features/room/actions/join_room.dart';
 import 'package:acter/router/utils.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
@@ -247,10 +246,7 @@ String prepareMsg(MsgContent? content) {
   );
 }
 
-String parseEditMsg(WidgetRef ref) {
-  // shouldn't ever happen to be null if state is edit or reply
-  final message = ref.read(chatInputProvider).selectedMessage!;
-
+String parseEditMsg(types.Message message) {
   if (message is types.TextMessage) {
     // Parse String Data to HTML document
     final document = parse(message.text);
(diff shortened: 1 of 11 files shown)


gnunicorn:
@gtalha07 in the video the top bar (most notably the user count) flickers. why is that? is that new?

app/test/features/chat/custom_input.dart
```diff
@@ -170,19 +216,21 @@ void main() {
       (tester) async {
         await tester.pumpWidget(
           ProviderScope(
-            overrides: overrides,
+            overrides: [
+              sdkProvider.overrideWith((ref) => MockActerSdk()),
+              ...overrides,
```
gnunicorn, line 221: nice! but do we really need to mock the ActerSDK itself? Isn't it enough to mock a higher level provider instead?

app/test/features/chat/custom_input.dart
```diff
@@ -212,10 +260,60 @@ void main() {
         await tester.pump();
         await tester.enterTextWithoutReplace(find.byType(TextField), 't');
 
-        await tester.pump();
+        // This test is timing out due to a pending timer.
+        // See MultiTriggerAutocompleteState._onChangedField in:
+        // acter_trigger_autocomplete.dart:279
+        // put 300ms delay as (debounceTimerDuration)
+        await tester.pump(Durations.medium2);
         expect(controller.text, 'testing code'); // <- cursor moved
```
gnunicorn, line 268:
```suggestion
        expect(controller.text, 'testing code');
```

app/test/features/chat/custom_input.dart
```diff
@@ -212,10 +260,60 @@ void main() {
...
+
+        final element = tester.element(find.byType(CustomChatInput));
+        final container = ProviderScope.containerOf(element);
+
+        final TextField textField = tester.widget(find.byType(TextField));
+        final controller = textField.controller!;
+
+        // initial state should be empty
+        assert(controller.text.trim().isEmpty, true);
+
```
gnunicorn, line 300: Can I also has a test that confirms if it wasn't empty and we route to another chat it is not the same and come back, it holds the previous state for that same room? That's the main concern I have with this changeset and a test proofing it works would relieve my worry.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -692,10 +705,23 @@ class _TextInputWidgetConsumerState extends ConsumerState<_TextInputWidget> {
...
+        // frame delay to keep focus connected with keyboard.
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          widget.chatFocus.requestFocus();
+        });
+      } else if (next == SelectedMessageState.replyTo) {
+        // frame delay to keep focus connected with keyboard..
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          widget.chatFocus.requestFocus();
+        });
+      }
```
gnunicorn, line 724: isn't it easier to just wrap the entire part in the PostFrameCallback? It's also more legible...
```suggestion
        WidgetsBinding.instance.addPostFrameCallback((_) {
      if (next == SelectedMessageState.edit) {
        // a new message has been selected to be edited or switched from reply
        // to edit, force refresh the inner text controller to reflect that
        widget.controller.text = parseEditMsg(ref);
        // frame delay to keep focus connected with keyboard.
          widget.chatFocus.requestFocus();
      } else if (next == SelectedMessageState.replyTo) {
        // frame delay to keep focus connected with keyboard..
          widget.chatFocus.requestFocus();
      }
        });
```

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -199,11 +194,19 @@ class _ChatInput extends ConsumerStatefulWidget {
 class __ChatInputState extends ConsumerState<_ChatInput> {
   late ActerTriggerAutoCompleteTextController textController;
   final FocusNode chatFocus = FocusNode();
+  final ValueNotifier<bool> _isInputEmptyNotifier = ValueNotifier(true);
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
     _setController();
+    textController.addListener(_updateInputState);
```
gnunicorn, line 203: shouldn't we do this within `_setController`?

app/lib/features/chat/utils.dart
```diff
@@ -244,3 +246,32 @@ String prepareMsg(MsgContent? content) {
     (match) => '<a href="${match.group(0)}">${match.group(0)}</a>',
   );
 }
+
+String parseEditMsg(WidgetRef ref) {
+  // shouldn't ever happen to be null if state is edit or reply
+  final message = ref.read(chatInputProvider).selectedMessage!;
```
gnunicorn, line 252: may I suggest a change of API here...

```suggestion
String parseEditMsg(Message message) {
```

and also, as this is then nicely easily input-output-based ... can we have unit tests for this confirming it does what we want? Should be rather straight forward then.

gnunicorn (approved):
\o/ tests and fixing them!

Just a few minor requests to changes in the code. Most concerned I am about adding one more test about re-entry. If that can be added and proven, go ahead and merge this :) .

gtalha07:
> @gtalha07 in the video the top bar (most notably the user count) flickers when you type. why is that? is that new?

It's because of typing notice call which goes when user starts typing on field. This has been as it is before.

app/test/features/chat/custom_input.dart
gtalha07: The log points out to the last trace of sdk which is chained with other providers in the tree. The lib missing error only seems to resolve with this override not the other ones...

app/test/features/chat/custom_input.dart
gtalha07: The new refactored mentions code (with text controller) isn't storing any previous state of room when switching. As I wanted to switch this mechanism with ```compose draft``` in the follower PRs. Do you think we want to switch it back to riverpod version of state storing for now? And migrate that to ```composeDraft``` in separate PR ? OR wait ? I will test in both cases...

app/lib/features/chat/widgets/custom_input.dart
gtalha07: It can be. Thanks for pointer!

gtalha07: Unfortunately, it throws error because ```ref``` update call. We have to keep it as before as ref isn't valid in next frame...
```
════════ Exception caught by scheduler library ═════════════════════════════════
Bad state: Cannot use "ref" after the widget was disposed.

```

app/test/features/chat/custom_input.dart
gnunicorn: well, could you add a test and have it skipped for now? and we'll fix it in the composedraft one? but then this second one needs to come soon .... before the next release ideally ...

PR #2071, opened 2024-08-16
Fix filename support for video, audio and image draft

gnunicorn:
include any provided `.filename` in Audio, Image and Video content draft


bitfriend (approved):
Looks good to me

PR #2072, opened 2024-08-16
Device Calendar Sync 

gnunicorn:
This PR brings syncing of all events not-denied to your device calendar on Android & iOS. See it in Action (the other app is the emulator Default Calendar App from Google):

https://github.com/user-attachments/assets/543db5aa-a6d3-400c-8888-86944c53e59f

It is behind a feature flag, that you can hard toggle to see and remove the calendar:

https://github.com/user-attachments/assets/e10f6bd8-3519-4ad9-a232-e82a2a647d25

And on iOS (emulator) as well:

https://github.com/user-attachments/assets/095f8ade-ce42-4bb6-b923-62d94ff96323

The following caveats apply:
 - It's behind a default-off feature flag
 - will only sync events you with rsvp status unknown, yes and maybe
 - Every event is added with a reminder of 10mins before (maybe we only want to do that for anything that is in `yes`?)
 - have this and the notifications on brings two annoying popup requests at the beginning
 - if the user refuses the permission we won't ask again.
 - there is no background sync, it needs the app to run (occasionally)
 - all rsvp status and event details changes are observed via riverpod and synced immediately (as the video above shows)
 - however, edits from the calendar by the user will be ignored  
 - [x] needs to be tested on an actual iOS

_To the reviewer_
Next to the actual feature the following refactors were necessary:
1. 8ba4b5855cdea98f0b260f1602fffca94843434e refactors the labFlag as it wasn't actually loading on startup, or at least wasn't fast enough for us to notice in loading. This refactors this to allow us to have a provider we can actually await for as well as make sure that the data was submitted before going for initialize or it would fail. That it works shows the video above.
2. 863e40e1d48b20a3ed723d4af68ec1d63eff90e5 refactors the `home_shell` into an `app_shell` that is moved to `config` as it is about setting up the general infrastructure. Aside from moving the file and outsourcing the LogoutWidget (which is otherwise features wise untouched), this moves the `_init` process to give us more executive control, like in this case make sure we are not asking for multiple permissions at the same time but after one another (notifications & calendar permission).


app/lib/features/calendar_sync/providers/events_to_sync_provider.dart
```diff
@@ -0,0 +1,32 @@
...
+    final rsvpStatus = myRsvpStatus.statusStr();
+    if (rsvpStatus != 'no') {
+      // we sync all that aren't denied yet
+      final event = await ref.watch(
+        calendarEventProvider(eventId).future,
+      ); // ensure we are listening to updates of the events themselves
+      toSync.add((event: event, rsvp: rsvpStatus));
+    }
+  }
+  return toSync;
```
kumarpalsinh25, line 31: How about using only single `for` loop?

```suggestion
   final List<EventAndRsvp> toSync = [];

  for (final event in allEventList) {
    final eventType = getEventType(event);
    if (eventType == EventFilters.upcoming ||
        eventType == EventFilters.ongoing) {
      final eventId = event.eventId().toString();
      final myRsvpStatus =
          await ref.watch(myRsvpStatusProvider(eventId).future);
      final rsvpStatus = myRsvpStatus.statusStr();
      if (rsvpStatus != 'no') {
        // we sync all that aren't denied yet
        final event = await ref.watch(
          calendarEventProvider(eventId).future,
        ); // ensure we are listening to updates of the events themselves
        toSync.add((event: event, rsvp: rsvpStatus));
      }
    }
  }
  return toSync;
```

app/lib/features/calendar_sync/calendar_sync.dart
```diff
@@ -0,0 +1,283 @@
...
+        return;
+      }
+      // FIXME: we probably want to debounce this ...
+      await _refreshCalendar(calendarId, next.valueOrNull ?? []);
+    },
+    fireImmediately: true,
+  );
+}
+
+Future<void> _refreshCalendar(
```
kumarpalsinh25, line 105: Skipping review of `_refreshCalendar` as not getting much idea of the code.

app/lib/features/calendar_sync/calendar_sync.dart
```diff
@@ -0,0 +1,283 @@
...
+Future<List<String>> _findActerCalendars() async {
+  // confirm this key exists.
+  final calendars = _logError(
+    await deviceCalendar.retrieveCalendars(),
+    'Failed to load calendars',
+  );
+  if (calendars == null) {
+    return [];
+  }
+  if (Platform.isAndroid) {
```
kumarpalsinh25, line 218: Any reason for this platform specific condition?

kumarpalsinh25 (approved):
Very existed about this feature!
Great Job! Code overall looks good to me.

app/lib/features/calendar_sync/providers/events_to_sync_provider.dart
gnunicorn: why? for code readability?

because the first is an iterator. in terms of execution this is _one_ for loop. And wherever possible I prefer to use iterators ... It's just that a) we are not sure about the behavior of ref.watch in iterator callbacks and b) it doesn't work if the result is an await/future at the moment ...

app/lib/features/calendar_sync/calendar_sync.dart
gnunicorn: had to update this slightly for iOS support. But nothing major. Also limiting our deleting to _local_ calendars so we don't accidentally delete others ... like one they might have named "Acter" in their google calendar.

app/lib/features/calendar_sync/calendar_sync.dart
gnunicorn: a shame, as this is the main part of the code ;)

gnunicorn:
I take it.

PR #2073, opened 2024-08-17
Add severe log to missed parts

bitfriend:
- Add severe log to error case of stream listener
- Add severe log to error case of async provider


app/lib/common/widgets/event/event_selector_drawer.dart
```diff
@@ -26,52 +26,50 @@ Future<String?> selectEventDrawer({
       builder: (context, ref, child) {
         final calEventsLoader = ref.watch(allEventListProvider(spaceId));
         return calEventsLoader.when(
-          data: (calEvents) {
```
gnunicorn: I trust this hasn't changed in actual code execution...

app/lib/features/chat/widgets/create_chat.dart
```diff
@@ -706,18 +706,16 @@ class _UserWidget extends ConsumerWidget {
               style: Theme.of(context).textTheme.labelMedium,
             ),
       leading: avatarLoader.when(
-        data: (avatar) {
```
gnunicorn: I'd still like to avoid having to review _several_ files with just minor code style changes that are unrelated to the actual PR ... please put those in a separate PR, mark that as `minor` and just merge it, there is no reason for anyone actually reviewing that so its a waste of time...

gnunicorn (approved):
there is literally _one_ log.severe and everything else is just code style. this _doesn't_ need anyone's review, please mark PRs with insignificant changes like this as `minor` and put them on auto-merge ...

Issue #2075, opened 2024-08-19
Linux nightly fails


gnunicorn:
[Fails since today](https://github.com/acterglobal/a3/actions/runs/10445834290/job/28922367343#step:20:143).

```
Got dependencies!
105 packages have newer versions incompatible with dependency constraints.
Try `flutter pub outdated` for more information.

Building Linux application...                                   
CMake Error at flutter/ephemeral/.plugin_symlinks/tray_manager/linux/CMakeLists.txt:32 (message):


  The `tray_manager` package requires ayatana-appindicator3-0.1 or
  appindicator3-0.1.  See
  https://github.com/leanflutter/tray_manager#linux-requirements


Unable to generate build files
Error: Process completed with exit code 1.
```

gnunicorn:
fixed by §2076

PR #2076, opened 2024-08-19
Add tray-icon-support for linux build

gnunicorn:
Fixing #2075 .


gnunicorn:
confirmed that the linux build works via the demo-builder.

PR #2077, opened 2024-08-19
load and save composer draft for chat room input text

gtalha07:
Fixes [meta#448](https://github.com/acterglobal/a3-meta/issues/448). Implements ```ComposerDraft``` for saving composed message state of room in an improved way .i.e. can be restored in between restarts and saves reply, edit states of message. See it in action:

Desktop: (Shows different states .i.e. text, edit/reply state storing between two rooms)

https://github.com/user-attachments/assets/4047b75f-ba17-44e2-a882-1d5770d9f478

Mobile: (Shows different states storing between restarts)

https://github.com/user-attachments/assets/f3e7a45b-94b5-4061-bcc5-e5b72b2f0441


app/test/features/chat/custom_input_test.dart
```diff
@@ -376,110 +392,70 @@ void main() {
...
-              ),
-            ),
-          ),
-        );
+        await switchToRoom('roomId-1');
 
-        // not visible
-        expect(find.byKey(CustomChatInput.noAccessKey), findsNothing);
-        expect(find.byKey(CustomChatInput.loadingKey), findsNothing);
+        await tester.enterText(find.byType(TextField), 'Hello Room 1');
```
gnunicorn, line 415: shouldn't this _check_ if the value was attempted to save?!?

app/test/features/chat/custom_input_test.dart
```diff
@@ -376,110 +392,70 @@ void main() {
...
 
-        // switched to new room, initial state should be empty
-        assert(controller.text.trim().isEmpty, true);
+        await tester.pump(Durations.medium2);
+        roomDrafts['roomId-1']!.setPlainText('Hello Room 1');
```
gnunicorn, line 418: we just seem to write it, no checks.

app/test/features/chat/custom_input_test.dart
```diff
@@ -376,110 +392,70 @@ void main() {
...
 
-        // not visible
-        expect(find.byKey(CustomChatInput.noAccessKey), findsNothing);
-        expect(find.byKey(CustomChatInput.loadingKey), findsNothing);
+        // switch back to room
+        await switchToRoom('roomId-1');
+        // load fake draft
+        final roomDraft1 = roomDrafts['roomId-1']!.plainText();
+        // copy draft over textfield
+        await tester.enterText(find.byType(TextField), roomDraft1);
```
gnunicorn, line 432: how is that loading the draft if we just enter it? We should expect that without entering it!
```suggestion
        final roomDraft1 = roomDrafts['roomId-1']!.plainText();
```

app/test/features/chat/custom_input_test.dart
```diff
@@ -376,110 +392,70 @@ void main() {
...
-        expect(find.byKey(CustomChatInput.loadingKey), findsNothing);
+        // load fake draft
+        final roomDraft2 = roomDrafts['roomId-2']!.plainText();
+        // copy draft over textfield
+        await tester.enterText(find.byType(TextField), roomDraft2);
+
+        await tester.pump(Durations.medium2);
+        // verify copied draft is same as written
+        final TextField textField2 = tester.widget(find.byType(TextField));
+        expect(textField2.controller!.text, roomDraft2);
```
gnunicorn, line 451: of course it, is, we just entered it... but it should be _without_ us entering it, no?

app/test/helpers/mocks.dart
```diff
@@ -0,0 +1,57 @@
...
+
+  @override
+  Future<Convo> convoWithRetry(String roomId, [int attempt = 0]) async {
+    return MockConvo();
+  }
+}
+
+/// Mocked version of Convo
+class MockConvo extends Mock implements Convo {
+  @override
```
gnunicorn, line 28: how about providing a value to expect here to ensure the _right_ thing is going to be sent...?

gnunicorn (changes requested):
The test seem to be mostly checking itself, not the actual functionality and I think the wiring can be improved upon as well. comments inlined.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -204,12 +214,13 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
 
   @override
   void dispose() {
+    textController.dispose();
     _isInputEmptyNotifier.dispose();
     super.dispose();
   }
 
   void _setController() {
-    final Map<String, TextStyle> triggerStyles = {
```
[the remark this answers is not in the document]
gtalha07: should be `final`.

app/test/features/chat/custom_input_test.dart
```diff
@@ -106,15 +120,15 @@ void main() {
 
         await tester.enterText(find.byType(TextField), 'testing code');
 
-        await tester.pump();
+        await tester.pump(Durations.medium2);
```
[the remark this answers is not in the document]
gtalha07: Yes, the test throw exceptions of pending timers. I believe it's because of `debounceDelay ` that we also need to mimic that behaviour. I will try to test more and see if I can minimize these delays without failing the tests.

gnunicorn (approved):
lgtm

PR #2078, opened 2024-08-19
Fix chat search

gnunicorn:
This was a pretty nasty bug, which came about for specific situations of state updates between render cycles. The issue stems from the [inner `AnimatedList` manager not properly updating the state if the context is currently missing](https://github.com/acterglobal/a3/pull/2078/files#r1723184044). That can happen when the redraw causes a different instance of the widget to be taken (as we have a new list chats ids) but using the same existing state - which didn't properly update because the context wasn't there. Thus the core fix is to properly reset the inner states in those cases and cause a full redraw of the widget state to have the proper state shown.

This is pretty hard to cause reliably, but as I chose to do this test-driven it was pretty easy to cause after everything has been made testable. The tests were failing before and are working fine now.

image

Thus most other changes in these files are for:
1. slight refactoring of the providers to allow them to be mocked for tests reliably
2. simplification of the widgets code to make it easier to read and debug
3. adding keys and minimal UX fixes to allow for a very controlled environment

More specific explanations inline.

fixes https://github.com/acterglobal/a3-meta/issues/464 . fixes https://github.com/acterglobal/a3-meta/issues/475


app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -122,23 +127,39 @@ class _AnimatedChatsList extends StatefulWidget {
...
+    if (_listKey.currentState == null) {
+      _log.fine('no state, hard reset');
+      // we can ignore the diffing as we aren't live, just reset
+      setState(() {
+        _reset();
+      });
+      return;
+    } else {
+      refreshList();
+    }
```
gnunicorn, line 156: This is the core fix.

app/lib/common/providers/common_providers.dart
```diff
@@ -27,18 +27,33 @@ final genericUpdatesStream =
   }
```
gnunicorn, line 27: these are all made to have easy providers to mock rather than inline-providers

app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -15,7 +15,7 @@ class AsyncConvoNotifier extends FamilyAsyncNotifier<Convo?, String> {
   late StreamSubscription<bool> _poller;
 
   @override
-  FutureOr<Convo> build(String arg) async {
+  FutureOr<Convo?> build(String arg) async {
```
gnunicorn, line 18: is already the existing type but by allowing `null` here, we can easily mock this provider.

app/lib/common/providers/room_providers.dart
```diff
@@ -326,6 +326,12 @@ final isDirectChatProvider =
   return convo?.isDm() == true && members.length == 2;
 });
 
+final isConvoBookmarked =
```
gnunicorn, line 329: room bookmarking would have been nicer but that would have required rust-sdk changes which I wanted to avoid for now. But it's easy to follow up on later with having this provider now.

app/lib/features/chat/models/room_list_filter_state/room_list_filter_state.dart
```diff
@@ -8,16 +7,17 @@ part 'room_list_filter_state.freezed.dart';
 Future<bool> roomListFilterStateAppliesToRoom(
   RoomListFilterState state,
   Ref ref,
-  Convo convo,
+  String convoId,
```
gnunicorn, line 10: in order to allow for proper mocking I had to make sure we used providers here rather than a convo-object I can't directly mock.

but in general the plan for all model/provider code was to go from a direct convo-interaction object to higher-level providers anyways, soo ...

app/lib/features/chat/models/room_list_filter_state/room_list_filter_state.dart
```diff
@@ -26,16 +26,12 @@ Future<bool> roomListFilterStateAppliesToRoom(
...
     if (convoId.toLowerCase().contains(searchTerm)) {
       return true;
     }
-    final avatarInfo = ref.read(roomAvatarInfoProvider(convoId));
-    if (avatarInfo.displayName != null &&
-        avatarInfo.displayName!.toLowerCase().contains(searchTerm)) {
-      return true;
-    }
-    return false;
+    final displayName = await ref.read(roomDisplayNameProvider(convoId).future);
```
gnunicorn, line 32: don't load the avatar if all we need is the displayname...

app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -149,18 +149,18 @@ final timelineStreamProvider =
 });
 
 final filteredChatsProvider =
```
gnunicorn, line 151: swiching to roomIds

app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -15,95 +15,100 @@ import 'package:logging/logging.dart';
 
 final _log = Logger('a3::chat::chats_list');
 
-class ChatsList extends ConsumerStatefulWidget {
```
gnunicorn, line 18: unnecessary stateful widget

app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -15,95 +15,100 @@ import 'package:logging/logging.dart';
...
-              return SliverToBoxAdapter(
-                child: Center(
-                  heightFactor: 10,
-                  child: Text(L10n.of(context).searchingFailed(e)),
-                ),
-              );
-            },
-            skipLoadingOnReload: true,
-          );
+  Widget build(BuildContext context, WidgetRef ref) {
```
gnunicorn, line 24: other than my comment this below is just splitting up the large build into neat separate widgets that are easy to comprehend.

app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -15,95 +15,100 @@ import 'package:logging/logging.dart';
...
-            subtitle: L10n.of(context).getInTouchWithOtherChangeMakers,
-            image: 'assets/images/empty_chat.svg',
-            primaryButton: ActerPrimaryActionButton(
-              onPressed: () async => context.pushNamed(
-                Routes.createChat.name,
-              ),
-              child: Text(L10n.of(context).sendDM),
+      },
+      skipLoadingOnReload: true,
+      skipLoadingOnRefresh: true,
```
gnunicorn, line 72: the only actual change was adding this line to avoid switching to the loading animation even if for just one frame...

app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -122,23 +127,39 @@ class _AnimatedChatsList extends StatefulWidget {
 }
 
 class __AnimatedChatsListState extends State<_AnimatedChatsList> {
-  final GlobalKey<AnimatedListState> _listKey = GlobalKey<AnimatedListState>();
```
gnunicorn, line 125: this is where the actual bug was. keeping the _listKey around but effectively out-of-sync with the actual list meant we were rendering old state and didn't realize it.

app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -193,93 +202,23 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
...
-    final client = ref.watch(alwaysClientProvider);
-    final hasFilters = ref.watch(hasRoomFilters);
     return PageStorage(
       bucket: bucketGlobal,
       child: CustomScrollView(
         controller: controller,
         key: const PageStorageKey<String>('convo-list'),
         physics: const BouncingScrollPhysics(),
         slivers: [
-          SliverLayoutBuilder(
```
gnunicorn, line 205: splitting the build into several cleaner functions.

app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -307,6 +246,64 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
...
+            children: [
+              TextButton(
+                key: RoomsListWidget.closeSearchActionButtonKey,
+                onPressed: () {
+                  setState(() {
+                    _isSearchVisible = false;
+                  });
+                },
+                child: Text(L10n.of(context).close),
+              ),
```
gnunicorn, line 266: here is an actual code change: we used to have different buttons ("clear" and "close") here depending on whether the user had added any search. I always found that slightly confusing, especially since we have a secondary "clear" just in the search-bar itself. Thus I removed that "clear" behavior from here and replaced it with a "close" that has always the same behavior of only closing ...

app/test/helpers/test_util.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+extension PumpUntilFound on WidgetTester {
```
gnunicorn, line 5: for reasons `pumpAndSettle` times out for us, so we have these helper functions, especially pumping the ProviderScope widget is super useful

kumarpalsinh25 (approved):
Changes looks good to me.

PR #2082, opened 2024-08-20
Pin List Item UI-UX Improvements

kumarpalsinh25:
https://github.com/user-attachments/assets/3eec0ad7-34a6-46e5-a932-eb795bc073be


gnunicorn:
@kumarpalsinh25 isn't that a bit to small and tricky if you used your fingers/non-pointer-devices? Instead I see that the user could easily accidentally tap that with their finger though they intend to tap the entire field/text entry and be routed where they don't want to go ... could be a bit confusing ...

Issue #1914, opened 2024-07-02
Show parent in space


gnunicorn:
in the upcoming design there is no way to know the details of what parents spaces exist or what other rooms people can join from. We want to do the following for those:

- when clicking the parent-avatars we want to show a bottom drawer with the list of those spaces (everywhere)
- when clicking the "limited" badge, we show a bottom drawer with the spaces whose members can join

kumarpalsinh25:
Let's connect on call for this and finalised quickly.

emilvincentz:
sounds good, ping me whenever:)

PR #2086, opened 2024-08-21
Space with access list for Limited Visibility

kumarpalsinh25:
Ref [#1914](https://github.com/acterglobal/a3/issues/1914)

Video:

https://github.com/user-attachments/assets/6aa3a952-3dd9-4481-ad69-8b28c6698dea

### DIFF
--- a/app/lib/common/widgets/visibility/visibility_chip.dart
+++ b/app/lib/common/widgets/visibility/visibility_chip.dart
@@ -21,15 +21,16 @@ class VisibilityChip extends ConsumerWidget {
     return spaceVisibility.when(
       data: (visibility) {
         return GestureDetector(
-            onTap: () {
-              if (visibility != RoomVisibility.SpaceVisible) return;
-              showLimitedSpaceList(
-                context,
-                ref,
-                roomId,
-              );
-            },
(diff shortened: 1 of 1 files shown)


app/lib/common/actions/show_limited_space_list.dart
```diff
@@ -0,0 +1,54 @@
...
+    context: context,
+    constraints: const BoxConstraints(maxHeight: 450),
+    isScrollControlled: true,
+    builder: (context) {
+      return _buildLimitedSpaceListUI(context, ref, limitedSpaceIds);
+    },
+  );
+}
+
+Widget _buildLimitedSpaceListUI(
```
gnunicorn, line 26: May I suggest to put this into `ConsumerWidget` (with a `this.roomId`), then the `ref.watch` could be moved here and it would actually update properly on changes, too..

Also, it is a bit nicer ;) .

gnunicorn (approved):
minor structure suggestion

app/lib/common/actions/show_limited_space_list.dart
kumarpalsinh25: Sure

PR #2094, opened 2024-08-23
Fixes : Space Details - Add Task Inline option not working

kumarpalsinh25:
Fixes, https://github.com/acterglobal/a3-meta/issues/408

After bit of debugging and research, finally able to fixes this issue. Was pretty tough to final this issue.

Issue Video:

https://github.com/user-attachments/assets/396b5ffc-461d-4f6f-b55d-e3530d179d1c

Video after fixes:

https://github.com/user-attachments/assets/34991c8d-f60d-49cc-abc9-5dc3dc4568df

### DIFF
--- a/app/lib/features/tasks/widgets/task_items_list_widget.dart
+++ b/app/lib/features/tasks/widgets/task_items_list_widget.dart
@@ -14,21 +14,27 @@ import 'package:logging/logging.dart';
 
 final _log = Logger('a3::tasks::widgets::list');
 
-class TaskItemsListWidget extends ConsumerWidget {
+class TaskItemsListWidget extends ConsumerStatefulWidget {
   final TaskList taskList;
   final bool showCompletedTask;
 
-  TaskItemsListWidget({
+  const TaskItemsListWidget({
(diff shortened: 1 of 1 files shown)


gnunicorn:
@kumarpalsinh25 kudos for finding it ... looking at how we changed things, I was wondering if we wanted to replace this (by now pretty uniquely different) widget with the typical bottom-up drawer instead .... as we have several other issues on it, like the way that submitting an item isn't clear ...

isn't blocking this from merge though.

kumarpalsinh25:
> @kumarpalsinh25 kudos for finding it ... looking at how we changed things, I was wondering if we wanted to replace this (by now pretty uniquely different) widget with the typical bottom-up drawer instead .... as we have several other issues on it, like the way that submitting an item isn't clear ...
> 
> isn't blocking this from merge though.

Yes planning to the same. On scrolling, this inline doesn't seems to be good UI-UX. Will do bottom-up view on later PR along with other improvements on Task Module. Merging this PR as of now to get out from current blocker.

PR #2098, opened 2024-08-23
Reduce accepting timeout to 10 seconds or show an error

gnunicorn:
When an error occurs on accepting an invite we automatically retry in the rust side due to a timing bug in synapse. But that also means if there is an _actual reason_ that accepting fails, we will wait until the end of the retry-period before showing it.

And that retry period was set up to be over 2 hours. This change make 5 retries with 2seconds waiting in between before returning for a total of 10 seconds now.

Fixes https://github.com/acterglobal/a3-meta/issues/472 .


bitfriend (approved):
Looks good to me

PR #2089, opened 2024-08-22
Showing better Error messages

gnunicorn:
We have several screens where when one item fails to load (for whatever reason), we show plenty of ugly inline error messages that are almost impossible to do anything with. This is the first part of #2097 , which implements:

- [x] For entire pages, a good looking error messages overlaying the loading content, with (where possible) 
  - [x] retry-button
  - [x] bug reporting-button with support to sent the error and the stack trace
- [x] do something smart about the passed error (e.g. parsing 'not found' and show something less threatening)
- [x] SpaceDetailsPage, with tests
- [x] Pins
   - [x] PinsList, with Tests 
   - [x] PinsDetails, with Tests

Dialog in Action with 404 on not found:

https://github.com/user-attachments/assets/1e4e5b1f-d385-43b5-bcff-82b36ea97496

Sending Error and Stack via Bug Reporter:

https://github.com/user-attachments/assets/de7d0b77-2d28-4ea2-9e33-34f8cb0f648a

This created the following report:
```
New bug reporter test

Number of logs: 0
Application: a3-dev
Labels: 
User-Agent: a3-dev/a3-dev
Version: DEV
description: with description field for more text now!
error: Space !asdgf not found
stack: 

```

To the reviewers:
This introduces a new `ActerErrorDialog` and `ErrorPage` to the toolkit allowing any page to show neat looking errors (rather than ugly inlined `Text('error msg')` ones) with the ability to report it as a bug or retry (by invalidating the provider, usually). For that the PR refactors the bug reporter into its own feature and refactors some of the pages to allow us to have automatic test confirm the error pages are shown and the retry is successfully done. The idea to expand all this over the entire system, including functional tests.

PR #2100, opened 2024-08-25
Inline Error Button with Retry for chat images

gnunicorn:
Images in chat media had a really ugly illegible error message (see https://github.com/acterglobal/a3-meta/issues/455 ). Built on top of #2089 this adds an inline icon instead that upon click opens our new ErrorDialog with the option to retry or report it as a bug.

 - [x] Tests for proof.

Fixes https://github.com/acterglobal/a3-meta/issues/455 .


app/lib/features/chat/providers/notifiers/media_chat_notifier.dart
```diff
@@ -33,14 +33,21 @@ class MediaChatNotifier extends StateNotifier<MediaChatState> {
...
+            mediaFile: File(mediaPath),
+            videoThumbnailFile: null,
             mediaChatLoadingState: const MediaChatLoadingState.loaded(),
           );
+          final videoThumbnailFile = await getThumbnailData(mediaPath);
+          if (videoThumbnailFile != null) {
+            if (state.mediaFile?.path == mediaPath) {
+              state = state.copyWith(videoThumbnailFile: videoThumbnailFile);
+            }
+          }
```
kumarpalsinh25, line 50: Assuming that this changes are for updating change in 2 separate action for file and thumb so that one can not have to wait.

kumarpalsinh25 (approved):
PR specific changes looks good to me.

PR #2101, opened 2024-08-25
Error pages for more sections

gnunicorn:
- [x] Task
  - [x] TaskList List
  - [x] TaskList Details
  - [x] Task Item Details
- [x] Events
  - [x] Events List
  - [x] Event Details

With tests, on top of #2089 of course.

Additionally this refactors the `AsyncRsvpStatusNotifier` into a typed variant that uses the `RsvpStatusTag` as it should be done.


app/pubspec.yaml
```diff
@@ -148,6 +148,7 @@ dependencies:
   open_filex: ^4.5.0
   phosphor_flutter: ^2.1.0
   device_calendar: ^4.3.2
+  quickalert: ^1.1.0
```
gtalha07, line 151: Aren't we using ```EasyLoading``` package to show different data states? Why this one is needed on top??

app/test/features/events/error_pages_test.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'package:acter/features/events/providers/event_providers.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../helpers/error_helpers.dart';
+import '../../helpers/mock_event_providers.dart';
+import '../../helpers/mock_space_providers.dart';
+import '../../helpers/test_util.dart';
+
+void main() {
+  group('Event List Error Pages', () {
```
gtalha07, line 16: Protests! More like pro tests 😄 👏🏻

gtalha07 (approved):
Nice additions. Overall lgtm!

app/pubspec.yaml
gnunicorn: easyloading is transative, dialog-oriented and non-disruptive. quickalert provides further widget for more prominent, larger and more intrusive widgets. They are complementary.

gtalha07: If that is so, do we want to consider unify all alerts be approached according to this package? If it also supports loading/error states?

Issue #1390, opened 2024-02-14
Chat inline emoji's don't show


gnunicorn:
Just clicked on an emoji in the chat editor and it doesn't show up properly. I am on Linux desktop, running latest `main`:

![image](https://github.com/acterglobal/a3/assets/40496/a5256f63-f728-45d6-a0c9-b1326a3f0fe0)

kumarpalsinh25:
Looks like this is fixed by https://github.com/acterglobal/a3/pull/2109

Issue #1387, opened 2024-02-14
Text Editor/preview doesn't show proper emojis


gnunicorn:
![image](https://github.com/acterglobal/a3/assets/40496/e900ba48-e30d-48dd-9654-bb1d26860ca7)

kumarpalsinh25:
@gnunicorn 
Probably this issue should be fixed when we add Appflowy editor in Text News Update.

kumarpalsinh25:
> unfortunately it seems that https://github.com/acterglobal/a3/issues/1384 doesn't fix that yet for me on Linux :( .

hmm.. strange.

kumarpalsinh25:
Probably this is fixed with https://github.com/acterglobal/a3/pull/2109

PR #2109, opened 2024-08-27
Improve the emoji UX

gnunicorn:
on top of #2180 .

Uses the optional widget overwrites to provides a better emoji picker UX, integrate better into the outer system, closing and backspace bugs fixed.

https://github.com/user-attachments/assets/d5eb0e51-1963-4240-a94c-dcbd22390b1a

Fixes:
 - [x] no way to close the picker on desktop ( 3ccba44302e63f10fca2c227ca22e0250b749132 offers backspace on empty, 3ccba44302e63f10fca2c227ca22e0250b749132 implements close buttons)
 - [x] emoji doesn't get added to the msg box (if it didn't have focus ever), 3ccba44302e63f10fca2c227ca22e0250b749132
 - [x] the wrong border color for the emoji picker b879bfde6ad32e98eb142cd71c04ea4c88300281
 - [x] close, backspace and search button as per our UI system, colors and closing from the search right away.


app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -279,14 +279,23 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
...
+      // Right text of cursor position
+      suffixText = textController.text.substring(cursorPos);
 
-    // Get the left text of cursor
-    String prefixText = textController.text.substring(0, cursorPos);
+      // Get the left text of cursor
+      prefixText = textController.text.substring(0, cursorPos);
+    } else {
+      cursorPos = 0;
+    }
```
kumarpalsinh25, line 298: May I know reason for this condition?

app/lib/common/widgets/emoji_picker_widget.dart
```diff
@@ -83,4 +102,140 @@ class EmojiPickerWidget extends StatelessWidget {
...
+                          const EdgeInsets.symmetric(horizontal: 16),
+                    ),
+                  ),
+                ),
+              ],
+            ),
+          ],
+        ),
+      );
+    },);
```
kumarpalsinh25, line 239: Not enough nesting to break it down in smaller functions?

app/lib/common/widgets/emoji_picker_widget.dart
```diff
@@ -83,4 +102,140 @@ class EmojiPickerWidget extends StatelessWidget {
...
+              color: Colors.transparent,
+              child: SizedBox(
+                height: emojiBoxSize + 8.0,
+                child: Row(
+                  children: [
+                    Expanded(
+                      child: ListView.builder(
+                        padding: const EdgeInsets.symmetric(vertical: 4.0),
+                        scrollDirection: Axis.horizontal,
+                        itemCount: results.length,
```
kumarpalsinh25, line 188: from where `results` array coming, facing difficulty to find it on code changes.

app/lib/common/widgets/emoji_picker_widget.dart
```diff
@@ -26,6 +30,33 @@ class EmojiPickerWidget extends StatelessWidget {
...
 
+    final emojiConfig = EmojiViewConfig(
+      backgroundColor: Theme.of(context).colorScheme.surface,
+      columns: cols,
+      emojiSizeMax: EmojiConfig.emojiSizeMax,
+    );
+    final catConfig = CategoryViewConfig(
+      customCategoryView: (config, state, tab, page) =>
+          actionBar(context, emojiConfig, state, tab, page),
+    );
```
kumarpalsinh25, line 41: So we have added custom view for category? Not quick getting how it was working when there was no `actionBar` on prior code?

app/pubspec.yaml
```diff
@@ -94,7 +94,11 @@ dependencies:
   path: ^1.8.3
   url_launcher: ^6.1.10
   file_picker: ^8.0.3
-  emoji_picker_flutter: ^2.1.1
+  emoji_picker_flutter: #^3.0.0
+    # path: ../../emoji_picker_flutter
+    git:
+      url: https://github.com/acterglobal/emoji_picker_flutter
+      ref: main-no-web
```
kumarpalsinh25, line 101: Can I quick summary of changes in this library?

kumarpalsinh25 (approved):
Overall looks good to me but added few minor question which need to be answer before merge.

app/lib/common/widgets/emoji_picker_widget.dart
gnunicorn: this is an `extends SearchViewState<_CustomSearchView>`, so from its [original source](https://github.com/Fintasys/emoji_picker_flutter/blob/0978a59beb504474d334effd2a131d517ec673fd/lib/src/search_view/search_view.dart#L36) .

the entire code is just a copy-paste extended with the x-for-closing

app/lib/common/widgets/emoji_picker_widget.dart
gnunicorn: I kept it close to the original source ... I can split it into two sub-widget though.

app/lib/common/widgets/emoji_picker_widget.dart
gnunicorn: well, it wasn't working. You couldn't close the emoji-picker on desktop.  that's what this PR fixes.

The default category view only supports either having a backspace or a search button. and while the action-bar-provides both action, having two bar is just very ugly and distracting.

that's why I replaced the category view with our own version including all features - search, backspace and close - and disabled the actionbar.

app/lib/features/chat/widgets/custom_input.dart
gnunicorn: well, if the field never had focus, the cursorPos is `-1`. which then fails when you do `.substring(-1)` (new line 292) without executing the remaining code.

With this fix, this now sets the cursor to 0 in that case and as that sets the cursor at the end, all following emojis are set behind it as expected.


Reading through this now, I am not sure this is 100% compatible with @gtalha07 latest changes where we might have set the text code even if it never had focus. Then we'd add the icons at the beginning, though the expectation would probably be to have it at the end. thus, I think, I'll change this code again ...

gtalha07 (approved):
LGTM overall. While I understand the changes fixes the issues of Emoji picker widget, from the screenshot video, I see the emojis pasted on text field appear as outlined. Wouldn't that be visually ugly for the end-user?

gnunicorn:
>  I see the emojis pasted on text field appear as outlined. Wouldn't that be visually ugly for the end-user?

because on linux that is still broken. see #1387 and #1390 .

app/pubspec.yaml
gnunicorn: I removed that change after. See https://github.com/acterglobal/a3/pull/2108 for the final result. Once merged, this will still just be using the released version.

PR #2110, opened 2024-08-27
Autoactivate Space Features

gnunicorn:
https://github.com/user-attachments/assets/f63971c3-801d-4b61-89d5-d399d0566819

If a feature isn't activated for a space, but the viewer has the permissions to update the space settings, we shows the button and offer to activate the feature, if they agree.

To the reviewer:
1. this refactors the Space Tabs to be typed: e6264d9b36131ed8a2d63e31fda3794267c79900
2. then it fixes the space tabs count items problem (see below): e3127b4b6935abf4645a1623af2e2c3132b7ae8c
3. refactoring the space app settings into actions -- making it typed, generalized and it removes a bunch of code: 80908255b13a4c9a73c6cdfadf9c3f54ac1f73ba
4. the actual feature is implemented by reusing these refactored actions and a similar dialog: 654b1b94e9040d9ea2985354612df32c9d506cb0
5. finally, disable all features by default, 1ea810a04422b1d4ea07f97c9000208d07143162

Re 2:
Before (notice the errors on the debug console)

https://github.com/user-attachments/assets/ead68e0a-69dc-47e6-955e-4048a39cefbb

After (notice the absence of errors):

https://github.com/user-attachments/assets/4da188bc-0b82-4515-b0e5-ab3063d12d5d


app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -3,26 +3,17 @@ import 'package:acter/common/providers/space_providers.dart';
...
-  });
+enum TabEntry {
+  overview,
+  pins,
+  tasks,
+  events,
+  chats,
+  spaces,
+  members,
+  actions;
```
kumarpalsinh25, line 16: 👏

app/lib/features/space/actions/activate_feature.dart
```diff
@@ -0,0 +1,280 @@
+import 'package:acter/common/providers/space_providers.dart';
```
kumarpalsinh25, line 1: This action file code looks good in terms of generic function still facing bit difficulty to end-to-end flow. Would suggest to further break down and add some comment for more explanation.

kumarpalsinh25:
One more thing, it will great if we can improvement text description for activation.

Screenshot 2024-08-28 at 11 38 52 AM

gnunicorn:
> One more thing, it will great if we can improvement text description for activation.

Please offer concrete suggestions or just add them via weblate, later.

app/lib/common/widgets/scrollable_list_tab_scroller.dart
```diff
@@ -363,37 +365,51 @@ class DefaultHeaderWidget extends StatefulWidget {
...
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.itemCount != widget.itemCount) {
+      _setController();
+    }
+  }
+
   @override
   void dispose() {
-    _tabController.dispose();
+    _tabController?.dispose();
```
gtalha07, line 398: Yes, about here. Isn't this enough to dispose off controller ? Also I think this [argument](https://github.com/acterglobal/a3/pull/2077#discussion_r1724902072) was raised in my PR, this should be disposed off automatically, no?

app/lib/common/widgets/scrollable_list_tab_scroller.dart
```diff
@@ -363,37 +365,51 @@ class DefaultHeaderWidget extends StatefulWidget {
...
 
   @override
   void initState() {
     super.initState();
+    _setController();
+  }
+
+  void _setController() {
+    _tabController?.dispose();
+    widget.selectedTabIndex.removeListener(externalTabChangeListener);
```
gtalha07, line 379: I think this is unnecessary as these resources are intended to be in ```dispose()``` method. Can you confirm it's needed?

app/lib/features/space/actions/activate_feature.dart
```diff
@@ -0,0 +1,280 @@
...
+  required WidgetRef ref,
+  required String spaceId,
+  required SpaceFeature feature,
+}) async {
+  final lang = L10n.of(context);
+  EasyLoading.showInfo(lang.loading);
+  try {
+    final space = await ref.read(spaceProvider(spaceId).future);
+    final appSettingsAndMembership =
+        await ref.watch(spaceAppSettingsProvider(spaceId).future);
```
gtalha07, line 26: should be ```ref.read()``` for reading value for function block.

app/lib/features/space/actions/update_feature_power_level.dart
```diff
@@ -0,0 +1,249 @@
...
+  final GlobalKey<FormState> _formKey =
+      GlobalKey<FormState>(debugLabel: 'change power level form');
+
+  String? currentMemberStatus;
+  int? customValue;
+
+  @override
+  void initState() {
+    super.initState();
+    setState(() => currentMemberStatus = widget.currentPowerLevelName);
```
gtalha07, line 98: No this should be not here. During widget initialization, ```setState``` cannot be called as it's already building. This should throw exception on widget ```build()```...
It should be wrapped with ```PostFrameCallback```.

gtalha07, line 98:
```suggestion
    WidgetsBinding.instance.addPostFrameCallback((_) {
    setState(() => currentMemberStatus = widget.currentPowerLevelName);
    });
```

app/lib/features/space/providers/space_navbar_provider.dart
gtalha07: Yes, I prefer enum based structures also. Nice!

gtalha07 (changes requested):
Thanks, this took me long time to review as I am unfollowing the recent flow in this section. But I see some incorrect approaches regarding state calls which could lead to widget build exceptions. And we don't have tests right now for it, Or if any tests already cover this scenarios, please point them out. Otherwise please add tests for these as I'm highly concerned this'll break up the UI on re-builds and address the concerns.

app/lib/features/space/actions/update_feature_power_level.dart
gnunicorn: this is a copy-paste from the power-level dialog, where it works just fine. I can fix it, but this isn't throwing any errors.

app/lib/common/widgets/scrollable_list_tab_scroller.dart
gnunicorn: no, that is incorrect. see above.

app/lib/common/widgets/scrollable_list_tab_scroller.dart
```diff
@@ -363,37 +365,51 @@ class DefaultHeaderWidget extends StatefulWidget {
...
+  TabController? _tabController;
 
   @override
   void initState() {
     super.initState();
+    _setController();
+  }
+
+  void _setController() {
+    _tabController?.dispose();
```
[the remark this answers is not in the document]
gnunicorn: we are doing `_setController` in two places: first at init but _also_ if the widget count changed. That is the actual fix. Because we need to reset the controller if the count has changed and for that we need to _discard_ the old controller and remove the previous listener or it will cause problems being out of sync. In that didWidgetUpdate the dispose doesn't take place and as we have no way of figuring out if a `late` was initialized, having the tabController disposed _if_ it is existing is what we want to do  here.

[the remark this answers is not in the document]
gtalha07: Hmm, I understand the implication but I'd probably look for some other safe approach. Fine for now if you're confident.

app/lib/features/space/actions/update_feature_power_level.dart
gtalha07: Having it not scheduled for next frame would have no effect as you can see the comment [here](https://stackoverflow.com/a/53373017). So it'd just work fine without wrapping it with ```setState```.

gtalha07: Then  you can just remove ```setState``` wrapper as it'd update ```currentMemberStatus``` during initial build in ```initState()```.

gtalha07 (approved):

PR #2111, opened 2024-08-27
Miscellaneous fixings

bitfriend:
- Add logger name to log on flutter side
- Fix timezone bug in some occurrences of timestamp
- Use map to remove bang operator from nullable variables


gnunicorn (changes requested):
Please split this into the several separate PRs that contain atomic changes, rather than putting up bundles like this that are hard to review and discuss.

Issue #2103, opened 2024-08-26
Show visibility/accessibility on space and chat widget


emilvincentz:

emilvincentz:
![Image](https://github.com/user-attachments/assets/bdfbc8f0-d611-49f8-9080-8b4cb105f482)

gnunicorn:
Example from WhatsApp Communities
![Uploading Screenshot_20240826-111338.png…]()

Issue #2095, opened 2024-08-23
Text input field in chat is very narrow.


emilvincentz:
With both the message icon, file upload icon, security shield icons and emoji icon, only half of the width is left to see what you're actually typing. 

Suggestion:
Remove the shield icon upon typing 
Remove the emoji icon on mobile.
(At least on iOS no other apps as far as I know is showing the emoji icon, as the keyboard already has emojis) ref to signal

![image](https://github.com/user-attachments/assets/953d95c9-c06b-4ddb-9b66-88c378effda2)

emilvincentz:
What do you think about this @kumarpalsinh25 ?

kumarpalsinh25:
> What do you think about this @kumarpalsinh25 ?

Yes it looks bit narrow due to encryption status icons. Here is my suggestions.

![Chat Input](https://github.com/user-attachments/assets/6cd91803-ad42-4423-a209-87935b1f255c)

emilvincentz:
Looks much better and with more space.
But do we really need the emoji icon? Dont most users have emojis in their keyboard? 
At least that is what I use, as the current Acter emoji selector and emoji search  isn't the best atm. anyway

kumarpalsinh25:
> Looks much better and with more space. But do we really need the emoji icon? Dont most users have emojis in their keyboard? At least that is what I use, as the current Acter emoji selector and emoji search isn't the best atm. anyway

Yes agree. Emoji selector will be useful when we have other options like sticker, custom emoji etc.. Simple emojis are now provided as default on android and iOS keyboards.

emilvincentz:
Cool!

gnunicorn:
> But do we really need the emoji icon? Dont most users have emojis in their keyboard?

We do still need it for Desktop for sure.

>  Simple emojis are now provided as default on android and iOS keyboards.

How far back does support for that go? And can we be sure about that for all Android OS it will be there? We have people running some pretty old Androids.


WhatsApp and Signal both provided that button on my (very recent) Android:
![image](https://github.com/user-attachments/assets/691aaa56-3396-4f7d-8613-62cab8935935)
![image](https://github.com/user-attachments/assets/f1f1df8e-79d8-4e56-b411-6e937e88acc6)

Once you started typing, they make the thing more dense and remove items - which I think we should also be doing:
![image](https://github.com/user-attachments/assets/09ae2a91-3144-4055-b6ee-45659916c5f7)
[...]

PR #2118, opened 2024-08-29
General Fixes : Chat Input UI and Space Visibility Icons

kumarpalsinh25:
1) fixes, https://github.com/acterglobal/a3/issues/2095
commit : https://github.com/acterglobal/a3/pull/2118/commits/d7a80c5c1322256e56202a04fefce93aa0a8c3ec

https://github.com/user-attachments/assets/e1b35859-7946-49c6-bd9f-3dfe13da20de

2) fixes, https://github.com/acterglobal/a3/issues/2103
commit : https://github.com/acterglobal/a3/pull/2118/commits/b386a0ebd04fc0ea5a9646a18832ba74f78d5b75

https://github.com/user-attachments/assets/e542cbb8-0742-49ac-886a-1781340bf7df

### DIFF
--- a/app/lib/common/themes/components/input_decoration_theme.dart
+++ b/app/lib/common/themes/components/input_decoration_theme.dart
@@ -6,7 +6,7 @@ var inputDecorationTheme = InputDecorationTheme(
   filled: true,
   focusedBorder: OutlineInputBorder(
     borderSide: const BorderSide(
-      color: Colors.white54,
+      color: Colors.white70,
     ),
     borderRadius: BorderRadius.circular(12),
   ),
--- a/app/lib/common/themes/components/input_decoration_theme.dart
+++ b/app/lib/common/themes/components/input_decoration_theme.dart
@@ -6,7 +6,7 @@ var inputDecorationTheme = InputDecorationTheme(
   filled: true,
   focusedBorder: OutlineInputBorder(
     borderSide: const BorderSide(
-      color: Colors.white70,
+      color: Colors.white54,
     ),
     borderRadius: BorderRadius.circular(12),
   ),
--- a/app/lib/common/widgets/spaces/space_with_profile_card.dart
+++ b/app/lib/common/widgets/spaces/space_with_profile_card.dart
@@ -129,10 +129,17 @@ class SpaceWithAvatarInfoCard extends StatelessWidget {
         titleTextStyle: titleTextStyle,
         subtitleTextStyle: subtitleTextStyle,
         leadingAndTrailingTextStyle: leadingAndTrailingTextStyle,
-        title: Row(
+        title: Column(
+          mainAxisSize: MainAxisSize.min,
+          mainAxisAlignment: MainAxisAlignment.start,
+          crossAxisAlignment: CrossAxisAlignment.start,
           children: [
-            Expanded(child: Text(title, overflow: TextOverflow.ellipsis)),
-            if (showVisibilityMark) VisibilityChip(roomId: roomId),
+            Text(title, overflow: TextOverflow.ellipsis),
+            if (showVisibilityMark)
+              VisibilityChip(
+                roomId: roomId,
+                useCompactView: true,
+              ),
           ],
         ),
         subtitle: buildSubtitle(context),
--- a/app/lib/common/widgets/spaces/space_with_profile_card.dart
+++ b/app/lib/common/widgets/spaces/space_with_profile_card.dart
@@ -129,19 +129,7 @@ class SpaceWithAvatarInfoCard extends StatelessWidget {
         titleTextStyle: titleTextStyle,
         subtitleTextStyle: subtitleTextStyle,
         leadingAndTrailingTextStyle: leadingAndTrailingTextStyle,
-        title: Column(
-          mainAxisSize: MainAxisSize.min,
-          mainAxisAlignment: MainAxisAlignment.start,
-          crossAxisAlignment: CrossAxisAlignment.start,
-          children: [
-            Text(title, overflow: TextOverflow.ellipsis),
-            if (showVisibilityMark)
-              VisibilityChip(
-                roomId: roomId,
-                useCompactView: true,
-              ),
-          ],
-        ),
+        title: Text(title, overflow: TextOverflow.ellipsis),
         subtitle: buildSubtitle(context),
         leading: avatar,
         trailing: trailing,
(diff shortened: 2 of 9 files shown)


gnunicorn:
UX review (code review follows):
Re 1:
the feature itself looks good, but the new updated but the updated input I am not convinced by:
1. I meant that that rather than having two buttons, we should have the send button be either the attach or send ... With both appearing and disappearing, the text-field jumps around left and right now, that is pretty distracting.
2. I am not convinced that more rectangular shape is that great. It gives the thing a more technical feel with its sharp corners. I know you two feel strongly about things being "more clean", but this is an inherent social place ("writing each other") and I'd rather stay on the "cozy" side with this one. Note: neither WhatsApp nor Signal have such harsh corners there (at least on my android, see [the screenshots in my earlier comment](https://github.com/acterglobal/a3/issues/2095#issuecomment-2306994827) ). Unless you feel strongly about that, I'd suggest going for a more cozy style, probably even dropping the border entirely and keeping a more elongated rounded shape as these other apps do.

Re 2:
I think we agreed that we _don't_ want to reuse the chip, as it is _too much_ and distracting -which the screencast proof to be true IMHO. But rather have the status appear in the subtitle in muted color underneath the name.

app/lib/common/themes/components/input_decoration_theme.dart
```diff
@@ -6,7 +6,7 @@ var inputDecorationTheme = InputDecorationTheme(
   filled: true,
   focusedBorder: OutlineInputBorder(
     borderSide: const BorderSide(
-      color: Colors.white70,
+      color: Colors.white54,
```
gnunicorn, line 9: we are changing the border color of all inputs? I think then we should also check other places and that this is really what we want...

app/lib/common/widgets/spaces/space_with_profile_card.dart
```diff
@@ -123,7 +129,12 @@ class SpaceWithAvatarInfoCard extends StatelessWidget {
...
         subtitleTextStyle: subtitleTextStyle,
         leadingAndTrailingTextStyle: leadingAndTrailingTextStyle,
-        title: Text(title, overflow: TextOverflow.ellipsis),
+        title: Row(
+          children: [
+            Expanded(child: Text(title, overflow: TextOverflow.ellipsis)),
+            if (showVisibilityMark) VisibilityChip(roomId: roomId),
+          ],
+        ),
         subtitle: buildSubtitle(context),
```
gnunicorn, line 138: I think we agreed to put the visibilty statement into the subtitle. the chip is just too much and too distracting there.

app/lib/l10n/app_en.arb
```diff
@@ -2149,5 +2149,6 @@
   "eventNoLongerAvailable": "Event no longer available",
   "@eventNoLongerAvailable": {},
   "eventDeletedOrFailedToLoad": "This may due to event deletion or failed to load",
-  "@eventDeletedOrFailedToLoad": {}
+  "@eventDeletedOrFailedToLoad": {},
+  "chatNotEncrypted": "This chat is not end-to-end-encrypted"
```
gnunicorn, line 2153: Just a side note as I've seen this in two PRs now: adding stuff at the end has the unfortunate side-effect of often causing merge conflicts. I recommend adding stuff in roughly alphabetical position to avoid that.

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -903,35 +906,9 @@ class _TextInputWidgetConsumerState extends ConsumerState<_TextInputWidget> {
             ),
             child: const Icon(Icons.emoji_emotions),
           ),
-          border: OutlineInputBorder(
```
gnunicorn, line 906: Not convinced by removing this design. See UX comment.

gnunicorn (changes requested):

kumarpalsinh25:
@gnunicorn 

REVIEW-1:

> - we should have the send button be either the attach or send.

I was also thinking about the same as it during the development and then left it as it is. But I agree 100% with on this UX suggestion.

- Rounded chat input

Not strong suggestion so reverting it back to old one.

app/lib/common/themes/components/input_decoration_theme.dart
kumarpalsinh25: Yes checked with other inputs as well and felt that this should good so that's why changes this from common theme file instead of making change at component level.

app/lib/common/widgets/spaces/space_with_profile_card.dart
kumarpalsinh25: Yes sorry, I forgot. let me do needful changes and share update.

app/lib/features/chat/widgets/custom_input.dart
kumarpalsinh25: Will do the needful.

kumarpalsinh25:
### REQUESTED CHANGES WORK

1) Chat Input UI:
Commit: https://github.com/acterglobal/a3/pull/2118/commits/086316bcafefe90fea42b96511888013515cf7ff

https://github.com/user-attachments/assets/2cd05ea1-37ba-445c-8536-a6640ccab2e6


2) Space Visibility Indication:
Commit: https://github.com/acterglobal/a3/pull/2118/commits/480b0c115e5efda5be8484d0817c5671073f9361
Note: Used the same Visibility Widget with `useCompactView` as there are many thing that can be re-used.

Screenshot 2024-08-29 at 6 38 55 PM

app/lib/common/widgets/spaces/space_with_profile_card.dart
gnunicorn: ehem... but that means for _suggested_ we now have _three_ rows? ... I think it would be more appropriate to put the visibility chip inside the `buildSubtitle(context)` widget .... just a `Suggested - Private` or something...

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -884,55 +885,41 @@ class _TextInputWidgetConsumerState extends ConsumerState<_TextInputWidget> {
...
+            borderRadius: BorderRadius.circular(30),
           ),
           enabledBorder: OutlineInputBorder(
-            borderRadius: BorderRadius.circular(30),
-            borderSide: BorderSide(
+            borderSide: const BorderSide(
               width: 0.5,
-              style: BorderStyle.solid,
-              color: Theme.of(context).colorScheme.onSurface,
+              color: Colors.white30,
```
gnunicorn, line 915: what is our take on hardcoding colors now? I thought we'd not like to do that because it breaks e.g. dark-mode-switching ...

gnunicorn (changes requested):
UI-wise looks good. But overlooked two minor problems.

gnunicorn (approved):
sorry, was meant to be approved just for the remaining problems I lined out in the review ...

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -371,38 +371,22 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
...
+                      onTyping: widget.onTyping,
                     ),
                   ),
                   ValueListenableBuilder<bool>(
                     valueListenable: _isInputEmptyNotifier,
                     builder: (context, isEmpty, child) {
                       return !isEmpty
                           ? renderSendButton(context, roomId)
-                          : const SizedBox();
+                          : renderAttachmentPinButton();
```
gtalha07, line 389: So we're essentially hiding attachment button when user is typing text? This seems wrong. Attachment icon should always be visible...

gtalha07 (changes requested):
Overall looks good. But I'm not convinced that we are forcibly hiding attachment icon when user types text. This blocks them for having interaction with attachment when field contains text and they have to delete all text to make it visible. Seems illogical.

gtalha07:
I'd probably suggest we can shrink up the text field UI to accomodate attachment icon and send button on right side and send button conditional rendering based on controller text state...

gnunicorn:
> Overall looks good. But I'm not convinced that we are forcibly hiding attachment icon when user types text. This blocks them for having interaction with attachment when field contains text and they have to delete all text to make it visible. Seems illogical.

That was my suggestion as matrix doesn't support doing texts on attachments (unlike whatsapp and signal where the text is then used to annotate the attachment).

I'd assume that rather than removing the text the user will probably finish their text and send the attachment afterwards ...

I'd give that a try and see if people complain.

gnunicorn:
> I'd probably suggest we can shrink up the text field UI to accomodate attachment icon and send button on right side and send button conditional rendering based on controller text state...

actually that "jumping" is one of the things we wanted to combat as it is rather annoying and confusing if the box shrinks just because I started typing.

I could see the suggestions though of upon having replaced the attachment-button with the hide, that we do show the attachment as a suffix in the text-fields (like whatsapp and signal do - see screenshots). But I am no convinced this is super necessary...

gtalha07:
> > Overall looks good. But I'm not convinced that we are forcibly hiding attachment icon when user types text. This blocks them for having interaction with attachment when field contains text and they have to delete all text to make it visible. Seems illogical.
> 
> That was my suggestion as matrix doesn't support doing texts on attachments (unlike whatsapp and signal where the text is then used to annotate the attachment).
> 
> I'd assume that rather than removing the text the user will probably finish their text and send the attachment afterwards ...
> 
> I'd give that a try and see if people complain.

Understandable. But I'd prefer we keep attachment icon visibility out of send button state . Imagine someone write a long text and have to attach something first, they have to delete all the text. And after which they have to save draft somewhere if the text is important and super long. This seems bad UX-experience to me.

Anyways, if you want to give it a go, sure. But I suspect people might complain more than we think...

gtalha07 (approved):

Issue #2087, opened 2024-08-21
Event : Show time in list


kumarpalsinh25:
![Image](https://github.com/user-attachments/assets/be544686-bfdb-4804-8185-de97dd6a9b06)

![Image](https://github.com/user-attachments/assets/802166ce-2e8d-4a5d-981b-b6cab4ce517b)

Need better UI for Date and Time showing.

PR #2120, opened 2024-08-29
General fixes related to Time Format and Event Time UI

kumarpalsinh25:
fixes, https://github.com/acterglobal/a3/issues/2087
fixes, https://github.com/acterglobal/a3-meta/issues/483

### 12 Hour Format:

https://github.com/user-attachments/assets/28b71334-b00b-4490-afd5-6fe6a1f80beb

### 24 Hour Format:

https://github.com/user-attachments/assets/585d8bdb-efee-4257-9211-9ad2a445c130


app/lib/common/utils/utils.dart
```diff
@@ -120,12 +120,22 @@ String getDayFromDate(UtcDateTime utcDateTime) {
...
+}
+
+String jiffyTime(BuildContext context, int timeInterval) {
   final jiffyTime = Jiffy.parseFromMillisecondsSinceEpoch(timeInterval);
   final now = Jiffy.now().startOf(Unit.day);
   if (now.isSame(jiffyTime, unit: Unit.day)) {
-    // (00:00 AM/PM)
-    return jiffyTime.jm;
+    bool is24HoursFormat = MediaQuery.of(context).alwaysUse24HourFormat;
+    return is24HoursFormat ? jiffyTime.Hm : jiffyTime.jm;
```
gnunicorn, line 138: this seems like a rather convoluted way of doing
`DateFormat.jm().format(localDateTime));` (from the `intl` package we already have included anyways) 

Any reason for not just using [`Intl.DateFormat`](https://api.flutter.dev/flutter/intl/DateFormat-class.html)?

app/lib/features/chat/pages/room_page.dart
```diff
@@ -253,7 +254,7 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
           inputPlaceholder: L10n.of(context).message,
           sendButtonAccessibilityLabel: '',
         ),
-        timeFormat: DateFormat.jm(),
+        timeFormat: is24HoursFormat ? DateFormat.Hm() : DateFormat.jm(),
```
gnunicorn, line 257: so, I see... it looks like DateFormat isn't set up to use the right `local` ... shouldn't we rather be fixing that problem then?

gnunicorn (changes requested):
Not convinced this is the right fix. It looks right, but shouldn't we instead be doing a proper `initializeDateFormatting` from `intl` and just use the time system it then formats with?

Reading a bit more of the used `alwaysUse24HourFormat` it indeed seems to not be reliable, as on Android it is only set if the user explictly said so, but not if only the locale would dictate that. Thus for a continental European Android without the user actively changing the setting, we'd still show the wrong time format.

kumarpalsinh25:
@gnunicorn 
I tried `initializeDateFormatting` from `intl` and use the default `DateFormat` behavior earlier also but getting 24 hour format event though I select European region from system while testing.

I am not sure I did any mistake on Testing this feature. I have did correction based on default behavior as suggested and [committed code here](https://github.com/acterglobal/a3/pull/2120/commits/ca8c151d02147aaefaca920b0391b79f3adfee09).

Can you please review it and check it I did anything wrong or it all we want.

kumarpalsinh25:
@gnunicorn 
Also I would to like to request you to run this branch code on your system and see if time is showing in 24 hours or not.

gnunicorn:
@kumarpalsinh25 ahahahah

image


Yeah, it doesn't integrate with our locale/language setting yet. I guess we should just set the system locale?!?

gnunicorn:
@kumarpalsinh25 I've pushed a commit to the branch setting `Intl.defaultLocale` depending on the language the user selected. Leading to the following behavior, which is what we want, I think:


https://github.com/user-attachments/assets/b161e9dc-9504-4482-bb8b-691e9721d5e7

gnunicorn:
and Arabic, too (not as if I can confirm this being correct, but ...)

image

gnunicorn (approved):
as per my change it does what I'd expect, so I'd approve :D

kumarpalsinh25:
> @kumarpalsinh25 ahahahah
> Yeah, it doesn't integrate with our locale/language setting yet. I guess we should just set the system locale?!?

Haven't find anything related this in https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization

Let me check more details about the same.

app/lib/features/settings/providers/notifiers/locale_notifier.dart
```diff
@@ -0,0 +1,38 @@
...
+    final prefInstance = await sharedPrefs();
+    await prefInstance.setString(languagePrefKey, languageCode);
+    _localSet(languageCode);
+  }
+
+  void _localSet(String languageCode) {
+    state = languageCode;
+    Intl.defaultLocale = languageCode;
+  }
+}
```
kumarpalsinh25, line 38: Ah I see, so this was the missing part?

PR #2124, opened 2024-08-29
Fix timezone bug in some occurrences of timestamp

bitfriend:
Timestamp from rust side is in UTC timezone and it should be converted into local timezone.


bitfriend:
This PR was split from https://github.com/acterglobal/a3/pull/2111

PR #2125, opened 2024-08-29
Avoid usage of bang operator for nullable variables

bitfriend:
We should migrate our codebase to null safety.
We can use `extension_nullable` to do it.
And bang operator is one of reasons that red screen appears sometimes.


bitfriend:
Now we have `Let` extension that @gtalha07 made skeleton.
It was implemented in our codebase unlike `extension_nullable`.
The source code repo of `extension_nullable` is not kept now.

app/lib/common/actions/report_content.dart
```diff
@@ -101,10 +101,10 @@ class _ReportContentWidget extends ConsumerWidget {
...
                 ),
                 value: ref.watch(_ignoreUserProvider),
-                onChanged: (value) =>
-                    ref.read(_ignoreUserProvider.notifier).update(
-                          (state) => value!,
-                        ),
+                onChanged: (val) {
+                  if (val == null) throw 'Changed value not available';
+                  ref.read(_ignoreUserProvider.notifier).update((state) => val);
+                },
```
gnunicorn, line 107: why throw? throws in UI code are bad as they are often unrecoverable crashes if they do occur.
```suggestion
                  if (val == null) {
                     _log.severe('Changed value not available';
                     return
                  }
                  ref.read(_ignoreUserProvider.notifier).update((state) => val);
                },
```

gnunicorn (changes requested):
I stopped right there, I see a lot newly added `throw` statements. Throw isn't any better than `!`, so that clearly defeats the point of removing `!`. Please strongly consider if all these `throw`s are wanted and necessary - at least all that can be triggered by `build`-rendering-code are not acceptable, use `_log.severe` instead.

I have not reviewed the other parts as this is already a show stopped and this is a massive changeset - which I am sure it didn't have to be. This could have easily been done in several smaller ones each one accomplishing the goal for a specific file or subset of  files....

app/lib/common/actions/report_content.dart
bitfriend: Understood.
But how to fix the case that return type is not `void` like [this fn](https://github.com/bitfriend/a3/blob/ce6ead4d8e39e186b294ac2511750b0fdfcf861c/app/lib/common/toolkit/buttons/danger_action_button.dart#L83-L88)?

bitfriend:
The goal of this PR was sufficiently achieved by series of `Remove bang operator from nullable variables`.

PR #2133, opened 2024-09-02
Fix default app permissions

### LINKED ISSUE #2131: Space Feature : Default OFF setting Issue
- Due to default space feature off, I was not able to see event tab menu inside space overview screen even though there are some data are there.

gnunicorn:
This revert the part of #2110 which applied the new defaults also to all previously submitted settings incorrectly - c004ef915187b7f24981226e41f54cf495595f8a (no need for real review of this one). Instead, we are now keeping the old behavior for all existing spaces and instead actively submit (and potentially allow the UI to configure) the new default app permissions when creating spaces - implemented in 9164448dbfb823234cb5e17de07574a6d86eddb0 .

Demonstrated in this video, where I first navigate to an existing space where app settings have never been altered showing the expected legacy behavior and then create a new subspace, where the new behavior is applied and all features are turned off (but can be activated in the settings, as well as in the actions section on the bottom of space details page):

https://github.com/user-attachments/assets/1733f445-eabd-45a1-87c2-2eaf7b209717

Fixes #2131 .

### DIFF
--- a/.changes/2131-backwards-compat-settings.md
+++ b/.changes/2131-backwards-compat-settings.md
@@ -1,0 +1,1 @@
+- [Fix] where the app settings of previously created spaces where interpreted under the new defaults incorrectly. Instead we are keeping the old fallback and default behavior and the wire protocol as is and instead create all futures spaces with the explicit having the features turned off now.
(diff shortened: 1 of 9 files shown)


bitfriend (approved):
Looks good to me

gtalha07 (approved):
lgtm

PR #2134, opened 2024-09-02
Windows path separator is different from other platform's one

bitfriend:
Will use `path` package to parse path.
Will not split path into filename and extension manually using `/`.

### DIFF
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -178,19 +178,19 @@ String getHumanReadableFileSize(int bytes) {
 
 String documentTypeFromFileExtension(String fileExtension) {
   switch (fileExtension) {
-    case 'png':
-    case 'jpg':
-    case 'jpeg':
+    case '.png':
+    case '.jpg':
+    case '.jpeg':
       return 'Image';
-    case 'mov':
-    case 'mp4':
+    case '.mov':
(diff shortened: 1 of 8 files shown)


app/lib/common/utils/utils.dart
```diff
@@ -178,19 +178,19 @@ String getHumanReadableFileSize(int bytes) {
 
 String documentTypeFromFileExtension(String fileExtension) {
   switch (fileExtension) {
-    case 'png':
-    case 'jpg':
-    case 'jpeg':
+    case '.png':
```
gtalha07, line 181: Wouldn't this require to check if platform is windows? What about other platforms ? Are you sure this wouldn't break the other ones?

gtalha07 (commented):
Looks good but I don't have knowledge context whether this'd cover other platforms as well (besides windows). So can't give explicit approval...

app/lib/common/utils/utils.dart
bitfriend: I think `path` package will support all platforms including windows.

bitfriend:
`path` supports both of windows and other platforms.
Please read [this document](https://pub.dev/packages/path#how-cross-platform-is-this).

gtalha07 (approved):

PR #2140, opened 2024-09-03
Improve usage of map

bitfriend:
Will remove unnecessary calling of `Map::containsKey()`.

### DIFF
--- a/app/lib/features/chat/pages/room_page.dart
+++ b/app/lib/features/chat/pages/room_page.dart
@@ -194,8 +194,7 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
 
   void showMessageOptions(BuildContext context, types.Message message) {
     if (message is types.CustomMessage) {
-      if (message.metadata!.containsKey('eventType') &&
-          message.metadata!['eventType'] == 'm.room.redaction') {
+      if (message.metadata!['eventType'] == 'm.room.redaction') {
         return;
       }
     }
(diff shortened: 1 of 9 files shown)


app/lib/features/chat/pages/room_page.dart
```diff
@@ -194,8 +194,7 @@ class _ChatRoomConsumerState extends ConsumerState<ChatRoom> {
 
   void showMessageOptions(BuildContext context, types.Message message) {
     if (message is types.CustomMessage) {
-      if (message.metadata!.containsKey('eventType') &&
-          message.metadata!['eventType'] == 'm.room.redaction') {
+      if (message.metadata!['eventType'] == 'm.room.redaction') {
```
gtalha07, line 197: This has probable chance of throwing null assertion if ```message.metadata``` is actually null.
```
if (message.metadata && message.metadata['eventType'] === 'm.room.redaction') {
    return;
}
```

gtalha07 (approved):
Looks good. There's one probable case where having the bang operator (!) can potentially throw out UI error. I have suggested change to avoid using that operator before merge ;)

app/lib/features/chat/pages/room_page.dart
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2140/commits/5f1691c0c57e555cc9471cd566259240c2a5f306

Issue #2127, opened 2024-08-30
Custom Icon for Pins, Task List and Space Categories


kumarpalsinh25:
- Able to pick any icon from the list
- Able to pick background color

![Screenshot 2024-08-30 at 4 22 24 PM](https://github.com/user-attachments/assets/a13a3bfa-0cbf-42dc-bcea-8db52e73498d)

PR #2143, opened 2024-09-04
Acter Icon Picker implementation and Integrate it with Pin & TaskList

kumarpalsinh25:
START PR REVIEW BY FIRST VIEWING BELOW VIDEOS FOR BETTER UNDERSTANDING.

https://github.com/user-attachments/assets/d748f39e-0b64-489b-9e31-d7b6277e2e30

fixes, https://github.com/acterglobal/a3/issues/2127


app/lib/common/widgets/acter_icon_picker/acter_icon_widget.dart
```diff
@@ -0,0 +1,75 @@
+import 'package:acter/common/widgets/acter_icon_picker/picker/acter_icon_picker.dart';
+import 'package:acter/common/widgets/acter_icon_picker/model/acter_icons.dart';
+import 'package:flutter/material.dart';
+
+class ActerIconWidget extends StatefulWidget {
```
kumarpalsinh25, line 5: This class is for generic ActerIcon widget which is used in multiple places like list item, edit/selection mode on creation form etc.. 


1) Used in Pin Create Form:
Screenshot 2024-09-05 at 1 49 47 PM

2) Used in PinList Item:
Screenshot 2024-09-05 at 1 50 53 PM

3) Used in Pin Detail screen:
Screenshot 2024-09-05 at 1 51 27 PM

4) Used in TaskList Create Form:
Screenshot 2024-09-05 at 1 50 06 PM

5) Used in TaskList Item:
Screenshot 2024-09-05 at 1 52 37 PM

6) Used in TaskList Detail screen:
Screenshot 2024-09-05 at 1 52 48 PM

app/lib/common/widgets/acter_icon_picker/model/acter_icons.dart
```diff
@@ -0,0 +1,162 @@
+import 'package:flutter/cupertino.dart';
+import 'package:phosphor_flutter/phosphor_flutter.dart';
+
+enum ActerIcons {
```
kumarpalsinh25, line 4: This contains set of ActorIcons data.

Screenshot 2024-09-05 at 1 55 06 PM

app/lib/common/widgets/acter_icon_picker/model/color_data.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:flutter/material.dart';
+
+List<Color> iconPickerColors = [
```
kumarpalsinh25, line 3: This contains set of ActerIcons Colors.

Screenshot 2024-09-05 at 1 54 51 PM

app/lib/common/widgets/acter_icon_picker/picker/acter_icon_picker.dart
```diff
@@ -0,0 +1,190 @@
...
+      return ActerIconPicker(
+        selectedColor: selectedColor,
+        selectedIcon: selectedIcon,
+        onIconSelection: onIconSelection,
+      );
+    },
+  );
+}
+
+class ActerIconPicker extends StatefulWidget {
```
kumarpalsinh25, line 29: This class is for ActerIcon Picker.
- It is manage as independent Component which can be reused based on the requirement.
- For shake of simplicity, Entire picker states are managed by basic `ValueNotifier` entire state management kept within class and avoided providers here to make it more independent widget.

Screenshot 2024-09-05 at 1 56 03 PM

app/lib/features/pins/Utils/pins_utils.dart
```diff
@@ -86,3 +90,38 @@ Future<void> saveDescription(
     );
   }
 }
+
+Future<void> savePinIcon(
```
kumarpalsinh25, line 94: This is to update Icon Widget from Pin Details Page.

app/lib/features/pins/pages/create_pin_page.dart
```diff
@@ -93,6 +98,16 @@ class _CreatePinConsumerState extends ConsumerState<CreatePinPage> {
                     mainAxisAlignment: MainAxisAlignment.start,
                     crossAxisAlignment: CrossAxisAlignment.stretch,
                     children: <Widget>[
+                      Center(
+                        child: ActerIconWidget(
```
kumarpalsinh25, line 102: This is how generic `ActerIconWidget` used with simplest code.

app/lib/features/pins/pages/create_pin_page.dart
```diff
@@ -304,6 +319,13 @@ class _CreatePinConsumerState extends ConsumerState<CreatePinPage> {
       final pinDraft = space.pinDraft();
       final pinState = ref.read(createPinStateProvider);
 
+      // Pin IconData
```
kumarpalsinh25, line 322: This is to add Pin Icon data white pin creation.

app/lib/features/pins/widgets/pin_list_item_widget.dart
```diff
@@ -44,7 +46,14 @@ class PinListItemWidget extends ConsumerWidget {
           Routes.pin.name,
           pathParameters: {'pinId': pinId},
         ),
-        leading: const PinIcon(),
+        leading: ActerIconWidget(
```
kumarpalsinh25, line 49: Reused the same `ActerIconWidget` in list item as view only widget.

app/lib/features/tasks/pages/task_list_details_page.dart
```diff
@@ -45,53 +50,86 @@ class _TaskListPageState extends ConsumerState<TaskListDetailPage> {
...
-            icon: const Icon(Icons.more_vert),
-            itemBuilder: (context) {
-              return [
-                PopupMenuItem(
-                  onTap: () => showEditDescriptionSheet(tasklist),
-                  child: Text(
-                    L10n.of(context).editDescription,
-                    style: Theme.of(context).textTheme.bodyMedium,
-                  ),
+      data: (tasklist) {
```
kumarpalsinh25, line 53: Just added `ActerIconWidget` in app bar and manage it's edit right based on the membership.

kumarpalsinh25 (commented):
Added few comments which might be helpful during PR review.

gnunicorn:
_UX discussion_:
I've looked at the video several times now and I am unsure about the size of the icons. I think I've mentioned before that I think showing them that much larger in the editing than in the actual view isn't great UX as it gives you the wrong impression on how it will be perceived by the others - but that is negligible, as you will see it soon enough and can change it if you want.

But in the list itself. I know you love the apple lists, but I find those icons pretty intense, with its size that is twice the title and then comes with strong colors. That is rather far away from the "little sprinkles of color" I imagined and proposed.

Take this random screenshot from Notion for comparison
image

Sure, there it's emoji and they are more detailed and colorful, but none is larger than the texts it belongs to and just acts as _a guide_ not as the primary attention grabbing item.

I feel those icons in this UI grab _a lot_ of attention (in particular because everything else is now monochrome) -- too much IMHO, as they _don't guide_ the users UI over the different items but _grab the attention_ and hold onto it, making it actually _harder_ to parse the text next to them ....

app/lib/features/tasks/actions/update_tasklist.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::tasks::update_tasklist_details');
+
+Future<void> updateTaskListIcon(
+    BuildContext context,
+    WidgetRef ref,
+    TaskList taskList,
+    Color color,
+    ActerIcons acterIcon,
```
gnunicorn, line 18: should be optional, too.

gnunicorn (changes requested):
Code-wise very nice. But we need not submit an icon and a color if the user hasn't diverted from the default. That needs to be supported before we merge this to not pollute the data that users submit...

gnunicorn:
> I feel those icons in this UI grab a lot of attention (in particular because everything else is now monochrome) -- too much IMHO, as they don't guide the users UI over the different items but grab the attention and hold onto it, making it actually harder to parse the text next to them ....

I wonder if we might want to use the color as a foreground color of the icon rather than the background of the even larger badge... I think that might be enough to fix these UX problems I have with it.

kumarpalsinh25:
@gnunicorn 
ICON SIZE:
Notion using emojis and we have icons in current stuff which have background circle. Due to background circle it will not look so good if we reduce icon size too small. Still I have reduce icon size at some level. Check below video for the same.


https://github.com/user-attachments/assets/5d5798e5-102b-4f03-a988-ff81def5d6e4

Happy to do further changes if needed. :)

kumarpalsinh25:
> > I feel those icons in this UI grab a lot of attention (in particular because everything else is now monochrome) -- too much IMHO, as they don't guide the users UI over the different items but grab the attention and hold onto it, making it actually harder to parse the text next to them ....
> 
> I wonder if we might want to use the color as a foreground color of the icon rather than the background of the even larger badge... I think that might be enough to fix these UX problems I have with it.


Yes that can be done. 
Screenshot 2024-09-05 at 4 39 43 PM


Probably for this I need to change some colors and selecte some brighter variant. But take a note that in future when we give support for light theme then it might create issue.

gnunicorn:
> Probably for this I need to change some colors and selected some brighter variant.

yeah, the screenshot is a lot easier to parse I think. I prefer that. We might want to change the default color as though and maybe in that case increasing the icon size as we have very thin icons? 

would love to get @emilvincentz 's feedback on this.


> But take a note that in future when we give support for light theme then it might create issue.

Light theme doesn't have priority. But even then I can easily imagine just using the color as a background or derive another background from that foreground to create enough contrast is enough.

kumarpalsinh25:
@gnunicorn 
General PR Feedback Work:

- Name correction:
https://github.com/acterglobal/a3/pull/2143/commits/850a7d9e8c3251df66e1955ef4568504a8c75b17

- Missing language translation:
https://github.com/acterglobal/a3/pull/2143/commits/043793af6456c614a5f87625d9a286e22aa1fa34

- Arrange pin actions into relevant file and folder:
https://github.com/acterglobal/a3/pull/2143/commits/d5088b10419f0370b0d978222f5c6019f2f359d1

- Minor wording correction:
https://github.com/acterglobal/a3/pull/2143/commits/abd2a188b68efdfc17db1feea1f7bfa4ca34affc

- Make icon and color optional by default and don't pass it if user haven't select it.
https://github.com/acterglobal/a3/pull/2143/commits/d4c77a747fae2c5184a1ab345292f3cc50f1f793

kumarpalsinh25:
@gnunicorn 
This how it will look if we remove icon background.


https://github.com/user-attachments/assets/4eeba10b-dd29-4627-ab25-9facc5cc8778

gnunicorn:
> This how it will look if we remove icon background.

I like it. What do you think, @emilvincentz ?

emilvincentz:
Well I dont have the strongest options. 

As I see it:

Pros of without background colour:
- More subtle

Pros with background colour:
-  that you can use the colour indication for e.g. prioritisation or clustering. Like you woul have two objects with same icon, but use the colour to indicate some kind of status or importance. 

However my main and first thought was, that we just want to make sure that it's easy for the user to know what is pins and what is tasks. Right now there is a small chance that the you scroll over it, you maybe dont realize that one list is tasks and the other is pins, as they look like each other.

But overall, I think both looks good. 

But with these things, I would take a quick screen shot and share in the team channel to get their perspective 
or in the future also put it into the product update channel and see if anyone engages;)

kumarpalsinh25:
@emilvincentz @gnunicorn 
Here we go with final version as per given suggestions....
- Remove circle background and applied color on Icon directly
- Did needful changes on icon size accordingly
- Added more brighter (Accent variant) colors for better contrast

Reference Video:

https://github.com/user-attachments/assets/a260e48c-e4f8-4fd2-b88d-507a67b564c0

@gnunicorn 
Can you resolve conflict related to `rendering.rs` and gets this PR to merge?

kumarpalsinh25:
> Just noticed, at pos 0:40 of the video the icon preview switches back to the wrong one ... you might want to fix that before we merge ...

- I also notice this after uploading video in PR Description and already fixed it by [this commit.](https://github.com/acterglobal/a3/pull/2143/commits/5fdc253bd1e9a3e5a3d52fcf2e6ccd8a2f09ac27). Just haven't did new screen recording :)

> and it appears, the unit tests for pins needs fixing now...

I also check this error but not getting much idea, let me check it in more detail.

app/lib/features/pins/pages/create_pin_page.dart
```diff
@@ -100,7 +100,7 @@ class _CreatePinConsumerState extends ConsumerState<CreatePinPage> {
                     children: <Widget>[
                       Center(
                         child: ActerIconWidget(
-                          icon: ActerIcon.pin,
+                          icon: pinIcon ?? ActerIcon.pin,
```
kumarpalsinh25, line 103: @gnunicorn 
This fixes the issue seeing in video at time 00:40.

PR #2145, opened 2024-09-04
Rust dependency upgrades

gnunicorn:
This upgrades the rust dependencies to the latest version. In order to do that properly, this is split into distinct parts:

1. With the latest matrix-sdk the previous cache-wrapper isn't needed anymore. Thus 5b1b20e4a40134ae5eb752bc5683e8c49b318aff deprecates that wrapper and instead copies the functionality into a new crate which we will be using going forward. Code-wise this hasn't changed and all its functionality is as before and should be 100% backwards compatible. But to make it easier to distinguish this is a separate commit
2. To upgrade to the latest matrix-sdk and be sure to use the proper ruma types, the direct `use ruma`-imports need to be removed and minor type fixes as well as the new file-cacher needs to be used. Finally, the cargo dependencies itself are fixed to reflect that change. This is all done in b7cb75f8189d3fda7cff099813c6cd5a428a1cdf
3. Finally 43beb362749a3d5c4d6f34b1b0c5845b367571e9 contains upgrades of other dependencies, most notably the `minijinja` major jump that required some updates to their new type system.


native/cli/Cargo.toml
```diff
@@ -17,8 +17,8 @@ env_logger = "0.10.0"
 matrix-sdk = { workspace = true }
 matrix-sdk-base = { workspace = true }
 matrix-sdk-sqlite = { workspace = true }
-ruma-client-api = { workspace = true }
-ruma-common = { workspace = true }
+# ruma-client-api = { workspace = true }
```
gtalha07, line 20: Are these needed to be commented ?

native/core/Cargo.toml
```diff
@@ -21,19 +21,19 @@ server = []
...
 enum_dispatch = "0.3.12"
 futures = "0.3.30"
 mime_guess = "2.0.4"
 icalendar = { workspace = true }
-ruma = { workspace = true }
-ruma-client-api = { workspace = true }
-ruma-common = { workspace = true }
-ruma-events = { workspace = true }
-ruma-macros = { workspace = true }
+# ruma = { workspace = true }
```
gtalha07, line 32: Why not having it removed if deprecated?

native/file-event-cache/Readme.md
```diff
@@ -0,0 +1,61 @@
+# Matrix Rust Sdk File Event Cache
+
+A secure and privacy preserving file-storage implementation for the Matrix SDK Event Cache.
```
gtalha07, line 3: Nice putting the documentation! 👏🏻

native/core/src/events/common/categories.rs
```diff
@@ -1,30 +1,10 @@
...
-use ruma_macros::EventContent;
+use matrix_sdk_base::ruma::events::macros::EventContent;
 use serde::{Deserialize, Serialize};
 
-/// The possibly redacted form of [`CategoriesEventContent`].
-///
-/// This type is used when it's not obvious whether the content is redacted or not.
-#[derive(Clone, Debug, Deserialize, Serialize)]
-#[allow(clippy::exhaustive_structs)]
-pub struct PossiblyRedactedCategoriesStateEventContent();
```
gtalha07, line 12: We don't require this with new sdk upgrade?

gtalha07 (approved):
LGTM!

native/core/src/events/common/categories.rs
gnunicorn: No, that workaround was needed for an older ruma version, but it works properly now with the latest `ruma_event` macro.

gnunicorn:
I had to add a not insignificant refactoring to the unique-id/event-id usage of the chat system. I believe that it is all right this way, I did check all cases I could think of, but we should be cautious the next few nightly releases...

PR #2151, opened 2024-09-04
chat input UI improvements

gtalha07:
Fix 1: Border radius 
Screenshot 2024-09-04 at 6 54 25 PM

Fix 3: Improve Edit Message UI:
Screenshot 2024-09-04 at 7 03 29 PM

Fix 2: Update Pill UI for reply builder
Screenshot 2024-09-04 at 6 48 32 PM
Screenshot 2024-09-04 at 6 56 36 PM

Along with refined Pill Builder UI code.


app/lib/features/chat/utils.dart
```diff
@@ -376,3 +377,17 @@ Future<void> saveDraft(
     }
   }
 }
+
+Future<void> onMessageLinkTap(Uri uri, BuildContext context) async {
+  final roomId = getRoomIdFromLink(uri);
+
+  ///If link is type of matrix room link
+  if (roomId != null) {
+  }
```
gnunicorn, line 386: I understand this is a copy-paste thing, but could we maybe fix this? 

1. not have the closing `}` of the `if` in a different line than the `else {` (and several lines in between)
2. have an empty `if` just for an else ... that should be if an if checking the opposite
3. do something in the other case. like we want to redirect to the room, don't we?

gnunicorn (approved):
Just a request to fix something within the copy-pasted-code ...

app/lib/features/chat/utils.dart
gtalha07: You mean like this suggested code:
```
Future<void> onMessageLinkTap(Uri uri, BuildContext context) async {
  final roomId = getRoomIdFromLink(uri);

  ///If link is type of matrix room link
  if (roomId != null) {
    goToChat(context, roomId);
  }

  ///If link is other than matrix room link
  ///Then open it on browser
  if (roomId == null) {
    await openLink(uri.toString(), context);
  }
}
```

gnunicorn: else the second part should be `} else { ` but other than that, yeah!

PR #2158, opened 2024-09-06
Permanent Routing

### LINKED ISSUE #894: when opening mobile app, it should take to the last visited screen.
as a regular user, its annoying that you always have to navigate from updates to (most often) chat

gnunicorn:
This tries to fix #894 . As expected the "back" behavior after the restart is a bit of a problem ...


gnunicorn:
superseeded by #2159 .

PR #2161, opened 2024-09-09
Add Link : Fixes issue related to url prefix

kumarpalsinh25:
Fixes https://github.com/acterglobal/a3-meta/issues/510

- Removed predefine `https://` prefix
- Used more generic url validation pattern which can work with custom schema urls also

### DIFF
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -36,8 +36,10 @@ final idMatrixRegexp = RegExp(
   r'matrix:roomid/(?<id>[^?]+)(\?via=(?<server_name>[^&]+))?(&via=(?<server_name2>[^&]+))?(&via=(?<server_name3>[^&]+))?',
 );
 
-final urlValidatorRegexp = RegExp(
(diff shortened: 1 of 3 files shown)


app/lib/common/utils/utils.dart
```diff
@@ -36,8 +36,10 @@ final idMatrixRegexp = RegExp(
   r'matrix:roomid/(?<id>[^?]+)(\?via=(?<server_name>[^&]+))?(&via=(?<server_name2>[^&]+))?(&via=(?<server_name3>[^&]+))?',
 );
 
-final urlValidatorRegexp = RegExp(
```
gnunicorn, line 39: and the regexp grows, yet now tests are nowhere to be found ...

how are we sure we don't break stuff, if we don't have tests for this?

gnunicorn (commented):
I have no idea ... the fix ix primarily about a regexp and has no tests ...

app/test/features/general/url_validation_test.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:acter/common/utils/utils.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
kumarpalsinh25, line 4: @gnunicorn 
Added UNIT TEST for Regex validation. All Test are passed with given regex!

gnunicorn (approved):
neat, thanks!

PR #2162, opened 2024-09-09
Remove ids from category

gnunicorn:
It is not needed, not useful and in the way -> ac107a77c3aab3937984fd83a49f1c3fc651c4ca .

8b2dd2eda115a3e3c027c109f617608cc2a48a37 Also adds a minor commit to add `boxed_local` for the entire rust crate so we don't have to put that in every local occurrence...


gnunicorn:
replaced by #2164 .

PR #2165, opened 2024-09-09
Add update action from space view

gnunicorn:
With the upgrade-permission and auto-selecting the space for the news.

In action:

https://github.com/user-attachments/assets/a3e742cb-fe84-44a1-96f5-d458b7c9ae33

For the reviewers:
- 8bad620cf5ee48fef35353ceed82bc635808a0b2 contains minor fixes I noticed in the other code around that to be more straight-forward and only show the `+` on the dashboard _if_ the user really has spaces they can post to
- df1ebdf64ee334aca0164df73131d65975fa039f contains the actual fix with the initial-selection addition.

### DIFF
--- a/app/lib/features/news/providers/news_post_editor_providers.dart
+++ b/app/lib/features/news/providers/news_post_editor_providers.dart
@@ -40,6 +40,12 @@ class NewsStateNotifier extends StateNotifier<NewsPostState> {
     );
   }
 
+  void setSpaceId(String? spaceId) {
+    state = state.copyWith(
+      newsPostSpaceId: spaceId,
+    );
+  }
+
   Future<void> selectEventToShare(BuildContext context) async {
     final eventId = await selectEventDrawer(
       context: context,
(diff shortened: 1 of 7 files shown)


app/lib/features/news/providers/news_post_editor_providers.dart
```diff
@@ -40,6 +40,12 @@ class NewsStateNotifier extends StateNotifier<NewsPostState> {
     );
   }
 
+  void setSpaceId(String? spaceId) {
+    state = state.copyWith(
```
gtalha07, line 44: Can we add unit test for this case so to ensure it saves up correct spaceId when switching spaces?

gtalha07 (approved):
Looks good!

app/lib/features/news/providers/news_post_editor_providers.dart
gnunicorn: this one-liner you want a test case for?

PR #2166, opened 2024-09-09
Bookmarked spaces first

### LINKED ISSUE #2104: Bookmarked spaces should be displayed first on 'Spaces' page
When you click 'see all pages' the bookmarked spaces should be showing first on the list (this is the case on home page, but not on Spaces list)

gnunicorn:
.. in ...
- the all spaces list
- the sidebar
- the space selectors (e.g. for an event)
- when searching with the quick jump

see yourself:

https://github.com/user-attachments/assets/1cbd2030-bf97-47d9-ac38-d8965accf0d6

fixes #2104

To the reviewer:
It's one part per commit. Aside from ecfd3c9ececf9658d02c397c657a784766b7a6ff which contains a run-by fix to do less redraw by putting the `hasSpaces` into a provider that caches the result properly and d35106b1d86262517565575e205836a2b915641a which removes the jiggle of pins upon changing search values ...

### DIFF
--- a/app/lib/common/providers/space_providers.dart
+++ b/app/lib/common/providers/space_providers.dart
@@ -16,6 +16,9 @@ final spacesProvider =
   return SpaceListNotifier(ref: ref, client: client);
 });
 
+final hasSpacesProvider =
+    Provider((ref) => ref.watch(spacesProvider).isNotEmpty);
+
 final bookmarkedSpacesProvider = Provider(
   (ref) => ref.watch(spacesProvider).where((s) => s.isBookmarked()).toList(),
 );
--- a/app/lib/common/providers/space_providers.dart
+++ b/app/lib/common/providers/space_providers.dart
@@ -160,7 +160,9 @@ final hasSpaceWithPermissionProvider =
 /// Stays up to date with underlying client info
 final _spaceIdAndNames =
     FutureProvider.autoDispose<List<_SpaceIdAndName>>((ref) async {
-  final spaces = ref.watch(spacesProvider);
+  final spaces = ref
+      .watch(bookmarkedSpacesProvider)
+      .followedBy(ref.watch(unbookmarkedSpacesProvider));
   List<_SpaceIdAndName> items = [];
   for (final space in spaces) {
     final roomId = space.getRoomIdStr();
--- a/app/lib/common/providers/space_providers.dart
+++ b/app/lib/common/providers/space_providers.dart
@@ -20,6 +20,10 @@ final bookmarkedSpacesProvider = Provider(
   (ref) => ref.watch(spacesProvider).where((s) => s.isBookmarked()).toList(),
 );
 
+final unbookmarkedSpacesProvider = Provider(
+  (ref) => ref.watch(spacesProvider).where((s) => !s.isBookmarked()).toList(),
+);
+
 /// List of spaces other than current space and it’s parent space
 final otherSpacesForInviteMembersProvider = FutureProvider.autoDispose
     .family<List<Space>, String>((ref, spaceId) async {
--- a/app/lib/features/search/providers/spaces.dart
+++ b/app/lib/features/search/providers/spaces.dart
@@ -20,7 +20,10 @@ class SpaceDetails extends SearchTermDelegate {
 }
 
 List<SpaceDetails> _filterSpaces(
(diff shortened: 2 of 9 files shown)


app/lib/common/providers/space_providers.dart
```diff
@@ -20,6 +20,10 @@ final bookmarkedSpacesProvider = Provider(
   (ref) => ref.watch(spacesProvider).where((s) => s.isBookmarked()).toList(),
 );
 
+final unbookmarkedSpacesProvider = Provider(
+  (ref) => ref.watch(spacesProvider).where((s) => !s.isBookmarked()).toList(),
+);
+
```
kumarpalsinh25, line 29: How about created provider as below which can used everywhere.


```
final sortedSpacesProvider = Provider<List<Space>>((ref) {
  List<Space> sortedSpaceList = [];
  //GET LIST OF BOOKMARKED SPACES
  final bookmarkedSpaces = ref.watch(bookmarkedSpacesProvider);
  sortedSpaceList.addAll(bookmarkedSpaces);

  //GET LIST OF UN-BOOKMARKED SPACES
  final unBookmarkedSpaces = ref.watch(unBookmarkedSpacesProvider);
  sortedSpaceList.addAll(unBookmarkedSpaces);

  //RETURN SORTED SPACES
  return sortedSpaceList;
});
```

app/lib/features/spaces/pages/spaces_page.dart
```diff
@@ -64,29 +61,42 @@ class _SpacesPageState extends ConsumerState<SpacesPage> {
...
+        mainAxisExtent: 100,
+        childAspectRatio: 4,
+      ),
+      itemBuilder: (context, index) {
+        String roomId;
+        if (index < bookmarked.length) {
+          roomId = bookmarked[index].getRoomIdStr();
+        } else {
+          roomId = others[index - bookmarked.length].getRoomIdStr();
+        }
```
kumarpalsinh25, line 90: This can be simplifies if we use separate provider mentioned above.

app/lib/features/search/providers/spaces.dart
```diff
@@ -50,4 +49,16 @@ final AutoDisposeFutureProvider<List<SpaceDetails>> spacesFoundProvider =
...
+
+final AutoDisposeFutureProvider<List<SpaceDetails>> spacesFoundProvider =
+    FutureProvider.autoDispose((ref) async {
+  final searchValue = ref.watch(searchValueProvider).toLowerCase();
+  // filter and sort them separately to keep the bookmarks at the beginning.
+  final allSpaces =
+      _filterSpaces(ref, searchValue, ref.watch(bookmarkedSpacesProvider));
+  allSpaces.addAll(
+    _filterSpaces(ref, searchValue, ref.watch(unbookmarkedSpacesProvider)),
+  );
```
kumarpalsinh25, line 65: If we added provider suggested above then we can do something like this here,
```
  final searchValue = ref.watch(searchValueProvider).toLowerCase();
  final allSpaces =  ref.watch(sortedSpacesProvider);
  ..
  And do perform search
```

kumarpalsinh25 (approved):
Logic wise looks good to me.

Code wise I would suggest to created one separate provider which will return space list with bookmarked items first.

- This will simplify some conditions on ListBuilders as well as get rid of followBy statements.
- Search provider can directly used common sorted provider.

gnunicorn:
@kumarpalsinh25 I am not yet certain about that idea... in the previous place I mixed these groups (suggested users) I put headlines in between to make it easier for the user to understand why the list is sorted in that weird way ... I am still thinking of putting a spacer in between the two lists on the sidebar as well ...

It won't work for the spaces grid though and I am not sure it makes a lot of sense on search screen either. But I didn't want to do too much on the later as it is scheduled to be redone anyways...

kumarpalsinh25:
> @kumarpalsinh25 I am not yet certain about that idea... in the previous place I mixed these groups (suggested users) I put headlines in between to make it easier for the user to understand why the list is sorted in that weird way ... I am still thinking of putting a spacer in between the two lists on the sidebar as well ...
> 
> It won't work for the spaces grid though and I am not sure it makes a lot of sense on search screen either. But I didn't want to do too much on the later as it is scheduled to be redone anyways...

Ok then feel free to merge. All good except that.

PR #2167, opened 2024-09-09
Event: Copy to new

gnunicorn:
First step towards the idea of an event-series: one-click action to create another event exactly like the one you are looking at:

https://github.com/user-attachments/assets/22ad0078-c299-4e1d-9d2d-b94589aca39b

To the reviewer:
- This is achieved by passing the `CalendarEvent` via `extra` and the event-creator taking all the values and setting its local state, hence the slight delay you see in it rendering. 1897be4a46354c7569c20ee918bd1108aa2d88b9
- [7d17138](https://github.com/acterglobal/a3/pull/2167/commits/7d17138732f6294da625c1f44924c0bbc93cf7bd) then contains a fix when you tried to open the time picker on a datetime that is lower than today, which is valid now, as we might have used a long-passed event as the template ...

### DIFF
--- a/.changes/2167-copy-event-as-new.md
+++ b/.changes/2167-copy-event-as-new.md
@@ -1,0 +1,1 @@
+- You can now take any Event you see and use it as a template to create any other similar event. Clicking Actions -> `Copy as New` will open the create event dialog with all fields filled in with the content from the current event for you to edit and submit.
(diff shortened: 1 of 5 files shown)


gtalha07 (approved):
Nice feature addition!

kumarpalsinh25 (approved):
Very Nice!

PR #2175, opened 2024-09-12
Unify Chat and Space Cards

gnunicorn:
while working on #2174 I noticed that we should probably unify the convo and room cards and to fix the problems of their appearance and features being different. This does that, it (in order of commits):

1. Unifies the Hierarchy Cards 
2. Unifies the Space and Convo Cards, move the chat specific convo card into that feature, 
3. run-by moves the spaceInfo into the space-feature (the only place using it) 
5. then another run-by that fixes the design of the super invites a bit...
4. as a result, they are also harder to distinguish, so deals with splitting the now unified design into two separate lists for the super invites

Look how pretty:

https://github.com/user-attachments/assets/197c0ac5-5b1d-41e3-9d3f-17cfa1e76bc8

To the reviewer:
This, naturally is based off #2174, so you can ignore the first commit ... This mostly moves code around and removes unnecessary widgets


app/lib/common/widgets/room/room_hierarchy_card.dart
```diff
@@ -118,38 +114,39 @@ class SpaceHierarchyCard extends ConsumerWidget {
...
+        subtitle: subtitle,
+        onTap: onTap ?? () {},
+        onFocusChange: onFocusChange,
+        onLongPress: onLongPress,
+        avatarSize: avatarSize,
+        contentPadding: contentPadding,
+        shape: shape,
+        showSuggestedMark: showSuggested,
+        trailing: trailing,
+        //  ??
```
gtalha07, line 129: commented!

gtalha07 (approved):
Looks good overall.

gnunicorn:
had to force push to make the rebase work (merge was broken). code wise only removed the commented code and fixed the test, all else is the same...

PR #2187, opened 2024-09-19
Minor joining fixes

gnunicorn:
- f7122e93e1d92ea7b42bfa62dfe6485cdac46578 so that we see the room name when we are invited
- a798aed5d690ab406433748f9782b9beca93878f so that the invite list refreshes upon rejecting an invite
- be97be6d249ac79c5d95d45b4e2fdf36a03ddde1 uses the room id server as a fallback for public rooms that are linked into a space that don't have any particular `via`-parameter set

### DIFF
--- a/app/lib/features/activities/widgets/invitation_card.dart
+++ b/app/lib/features/activities/widgets/invitation_card.dart
@@ -224,6 +224,7 @@ class _InvitationCardState extends ConsumerState<InvitationCard> {
     EasyLoading.show(status: L10n.of(context).rejecting);
     try {
       bool res = await widget.invitation.reject();
+      ref.invalidate(invitationListProvider);
       if (!context.mounted) {
         EasyLoading.dismiss();
         return;
--- a/app/lib/features/activities/widgets/invitation_card.dart
+++ b/app/lib/features/activities/widgets/invitation_card.dart
@@ -13,7 +13,7 @@ import 'package:logging/logging.dart';
 
 final _log = Logger('a3::activities::invitation_card');
 
-class InvitationCard extends ConsumerWidget {
+class InvitationCard extends ConsumerStatefulWidget {
   final Invitation invitation;
 
   const InvitationCard({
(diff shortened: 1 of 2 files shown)


app/lib/features/activities/widgets/invitation_card.dart
```diff
@@ -22,13 +22,33 @@ class InvitationCard extends ConsumerWidget {
...
+  String? roomTitle;
+
+  @override
+  void initState() {
+    super.initState();
+    _fetchTitle();
+  }
+
+  void _fetchTitle() async {
+    final title = await widget.invitation.room().displayName();
```
kumarpalsinh25, line 38: Why to use stateful widget and setting title first on initialisation?

`invitation.room().displayName();` doesn't work on ListTile title where we are actually showing title?

app/lib/features/activities/widgets/invitation_card.dart
```diff
@@ -80,17 +100,14 @@ class InvitationCard extends ConsumerWidget {
...
+          inviter(context),
         ],
       ),
     );
   }
 
-  ListTile renderGroupChatTile(BuildContext context, WidgetRef ref) {
-    final roomId = invitation.roomIdStr();
+  ListTile renderGroupChatTile(BuildContext context) {
+    final roomId = widget.invitation.roomIdStr();
```
kumarpalsinh25, line 110: `invitation.room().displayName();` isn't working here?

app/lib/features/activities/widgets/invitation_card.dart
gtalha07: Agree with @kumarpalsinh25 on this. Why having whole widget ```ConsumerStateful``` if ```ref``` is only needed for invalidating in widget methods? Could be simplified with just ```Consumer``` builder?

app/lib/features/activities/widgets/invitation_card.dart
gnunicorn: nope, it's a future

app/lib/features/activities/widgets/invitation_card.dart
gnunicorn: invitations are slightly different. this was the quickest way to make this work... as the invitation changing would reset the widget anyways...

gtalha07 (approved):
Overall I think its good and fixes the issue. Just a minor code improvement thing that I agree with Kumar.

PR #2197, opened 2024-09-24
fix html editor not responding on empty description content

gtalha07:
Fixes: [meta#252](https://github.com/acterglobal/a3-meta/issues/252).

Bug source: https://github.com/AppFlowy-IO/appflowy-editor/issues/549

### DIFF
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -15,6 +15,8 @@ import 'package:flutter_easyloading/flutter_easyloading.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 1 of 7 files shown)


gnunicorn (changes requested):
I'd like to see this in he helper and reducing the parsing to once rather than twice ...

arguably, we could also collapse that helper into just having one function to parse both html or markdown rather than check outside ..

gtalha07:
> I'd like to see this in he helper and reducing the parsing to once rather than twice ...
> 
> arguably, we could also collapse that helper into just having one function to parse both html or markdown rather than check outside ..

Understandable, I have reduced that validity check to be in helper fn ;)

gnunicorn (changes requested):
It seems I wasn't clear in how I wanted this change to be changed and as I wasn't really having the brain to express it, I just put it into the code. You can find it here: https://github.com/acterglobal/a3/pull/2200

PR #2198, opened 2024-09-24
[smallfix] Miscellaneous fixes

bitfriend:
Replace `AsyncValue.data()` with `AsyncValue.guard()`, because `AsyncValue.data()` doesn't contain `try-catch`.


gtalha07 (approved):
Looks good!

PR #2200, opened 2024-09-25
refactor of html editor bug fix

gnunicorn:
Refactor for #2197 .

Just moved it all into the `DocumentHelper.parse` with proper fallback support and remove all instances where it was used differently...

Quick test showed it fixed the issue on pins...


app/lib/common/widgets/edit_html_description_sheet.dart
```diff
@@ -97,13 +97,14 @@ class _EditHtmlDescriptionSheetState
               ActerPrimaryActionButton(
                 onPressed: () {
                   // No need to change
-                  final htmlBodyDescription = textEditorState.intoHtml();
+                  String htmlBodyDescription = textEditorState.intoHtml();
```
gnunicorn, line 100: I honestly don't know what this change is about?!?

Issue #2209, opened 2024-10-01
event list creation can't select space 


gnunicorn:
Events list -> create shows no space in the UI and we can't change it.

https://github.com/user-attachments/assets/13bbcc2d-fd36-4fd4-b416-c8978d060cb2

Same on Pins and alike. Also, if we try to create if no space has been selected we should just open the space selector rather than showing an error...

PR #2217, opened 2024-10-03
Space Selection Fixes & Improvement while creating pin/event

kumarpalsinh25:
fixes, https://github.com/acterglobal/a3/issues/2209

- Fixes issue related to space selection
- If space section is missing then show space selector rather than showing an error

Reference Video:

https://github.com/user-attachments/assets/2f1acecb-b7f0-4166-835b-1f917be8b34e

### DIFF
--- a/app/lib/features/events/pages/create_event_page.dart
+++ b/app/lib/features/events/pages/create_event_page.dart
@@ -4,6 +4,7 @@ import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/utils/utils.dart';
 import 'package:acter/common/widgets/html_editor.dart';
 import 'package:acter/common/widgets/spaces/select_space_form_field.dart';
+import 'package:acter/common/widgets/spaces/space_selector_drawer.dart';
 import 'package:acter/features/events/model/keys.dart';
 import 'package:acter/features/events/utils/events_utils.dart';
 import 'package:acter/features/home/providers/client_providers.dart';
@@ -408,9 +409,10 @@ class CreateEventPageConsumerState extends ConsumerState<CreateEventPage> {
 
   // Create event handler
   Future<void> _handleCreateEvent() async {
-    if (!_formKey.currentState!.validate()) return;
+    String? spaceId = ref.read(selectedSpaceIdProvider);
+    spaceId ??= await selectSpace();
+    if (!mounted) return;
 
-    final spaceId = ref.read(selectedSpaceIdProvider);
     if (spaceId == null) {
       EasyLoading.showError(
         L10n.of(context).pleaseSelectSpace,
@@ -418,6 +420,8 @@ class CreateEventPageConsumerState extends ConsumerState<CreateEventPage> {
       );
       return;
     }
+
+    if (!_formKey.currentState!.validate()) return;
 
     EasyLoading.show(status: L10n.of(context).creatingCalendarEvent);
     try {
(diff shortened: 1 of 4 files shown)


app/lib/router/general_router.dart
```diff
@@ -122,10 +122,12 @@ final generalRoutes = [
     name: Routes.createPin.name,
     path: Routes.createPin.route,
     pageBuilder: (context, state) {
+      final String? spaceId = state.uri.queryParameters['spaceId'];
       return NoTransitionPage(
         key: state.pageKey,
         child: CreatePinPage(
-          initialSelectedSpace: state.uri.queryParameters['spaceId'],
+          initialSelectedSpace:
+              spaceId != null && spaceId.isNotEmpty ? spaceId : null,
```
kumarpalsinh25, line 130: Empty `spaceId` was causing issue in space selector.

app/lib/features/events/pages/create_event_page.dart
```diff
@@ -408,9 +409,10 @@ class CreateEventPageConsumerState extends ConsumerState<CreateEventPage> {
 
   // Create event handler
   Future<void> _handleCreateEvent() async {
-    if (!_formKey.currentState!.validate()) return;
+    String? spaceId = ref.read(selectedSpaceIdProvider);
+    spaceId ??= await selectSpace();
```
gnunicorn, line 413: wait ... what sorcery is this? does this ... it means .. `??=` if null run assign to? Wow .... :exploding_head:

app/lib/features/pins/pages/create_pin_page.dart
```diff
@@ -376,6 +381,17 @@ class _CreatePinConsumerState extends ConsumerState<CreatePinPage> {
     }
   }
 
+  Future<String?> selectSpace() async {
```
gnunicorn, line 384: so considering that this comes in here twice, this should probably be moved into a helper action, should it not?

app/lib/features/events/pages/create_event_page.dart
```diff
@@ -477,4 +481,15 @@ class CreateEventPageConsumerState extends ConsumerState<CreateEventPage> {
       );
     }
   }
+
+  Future<String?> selectSpace() async {
+    final newSelectedSpaceId = await selectSpaceDrawer(
+      context: context,
+      currentSpaceId: ref.read(selectedSpaceIdProvider),
+      canCheck: 'CanPostPin',
```
gnunicorn, line 489: and that helper should have an explicit `canCheck`-param ... as this copy-paste lead to using the wrong key ;) 

```suggestion
      canCheck: 'CanPostEvent',
```

app/lib/router/general_router.dart
gnunicorn: I do wonder if it is more readible to do:
```suggestion
              (spaceId?.isNotEmpty == true) ? spaceId : null,
```

or not .. remain unsure...

app/lib/router/shell_routers/home_shell_router.dart
```diff
@@ -483,10 +483,12 @@ final homeShellRoutes = [
       if (extra != null && extra is CalendarEvent) {
         templateEvent = extra;
       }
+      final String? spaceId = state.uri.queryParameters['spaceId'];
       return NoTransitionPage(
         key: state.pageKey,
         child: CreateEventPage(
-          initialSelectedSpace: state.uri.queryParameters['spaceId'],
+          initialSelectedSpace:
+              spaceId != null && spaceId.isNotEmpty ? spaceId : null,
```
gnunicorn, line 491: same as above.

gnunicorn (approved):
but there is a bug that needs to be fixed. the other things are up for your digression to do..

app/lib/features/events/pages/create_event_page.dart
kumarpalsinh25: Reason is that even after showing space selector, user may close it without selecting any space so in that case spaceId can be null and error message will show over there. I have also shown this use-case in screen recording for reference.

kumarpalsinh25: Check video at 00:34

app/lib/features/events/pages/create_event_page.dart
kumarpalsinh25: Opss! Thanks for noticing.

app/lib/features/events/pages/create_event_page.dart
gnunicorn: all good. this was a comment of surprise of something new I learnt about the dart language!

kumarpalsinh25: Ah I see. You always exited about short syntaxes of programming lang.

PR #2220, opened 2024-10-04
Destroy system tray when app exits

bitfriend:
The following video shows the bug that system tray doesn't clean up when app exits.

https://github.com/user-attachments/assets/30ca56b9-446c-49bb-a795-a0740efa4b7e

Will fix this issue.

### DIFF
--- a/app/lib/config/desktop.dart
+++ b/app/lib/config/desktop.dart
@@ -5,9 +5,9 @@ import 'package:acter/router/router.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
+import 'package:logging/logging.dart';
 import 'package:tray_manager/tray_manager.dart';
 import 'package:window_manager/window_manager.dart';
-import 'package:logging/logging.dart';
 
 final _log = Logger('a3::desktop');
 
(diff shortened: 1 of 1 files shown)


bitfriend:
This video shows that system tray disappears automatically when app exits.

https://github.com/user-attachments/assets/eecc209a-b1c3-4b3d-9d13-a453d723fc2a

This bug was fixed.

app/lib/config/desktop.dart
```diff
@@ -122,6 +125,7 @@ class _DesktopSupportState extends State<DesktopSupport>
   void onTrayMenuItemClick(MenuItem menuItem) async {
     if (menuItem.key == 'exit_app') {
       _log.info('exit app');
+      await trayManager.destroy();
       await windowManager.destroy();
```
gnunicorn, line 129: did you test this on other platforms like Linux and windows as well? Would be unfortunate if a change like this means we can't exit the desktop anymore on any target...

bitfriend: Yeah, you discussed the important thing.
I missed that this PR should be tested on various platforms.
Above videos were created on windows.
All to do is test on linux and macos.
@gnunicorn Can you please build it on them for me?

gnunicorn:
tested and works on mac os and linux

PR #2243, opened 2024-10-10
Acter Search Widget

kumarpalsinh25:
Introducing all new `ActerSearchWidget` to Acter widget library set which makes search bar behaviour and UI-UX consistent across entire Acter App.

### Thing to know about `ActerSearchWidget`:
- Have default leading search icon
- Have default trailing clear icon which will be show/hide based on the search text content
- Have `onChanged` and `onClear` callback method which is most important for manage search data
- Have `initialText` input argument to initialise `ActerSearchWidget` with initial string data
- Have `hintText`, `leading`, `trailing` and `padding` like optional params to override default values.
- Haven it's own `searchTextController` to manage local state data

### Implemented `ActerSearchWidget` in below screens.
- Create Chat
- Select Room Drawer
- Room List (Chat List)
- Public Room Search
- Invite users
- Quick Jump
- Pin List
- Event List
- Task List

### Reference Video:

https://github.com/user-attachments/assets/f407427a-b143-464b-8a11-f778cb3d650a


app/lib/common/providers/common_providers.dart
```diff
@@ -135,3 +135,5 @@ final canRedactProvider = FutureProvider.autoDispose.family<bool, dynamic>(
     }
   }),
 );
+
+final searchValueProvider = StateProvider.autoDispose<String>((ref) => '');
```
kumarpalsinh25, line 139: Added `searchValueProvider` in common providers file as it is common provider used across different use-cases.

app/lib/common/widgets/acter_search_widget.dart
```diff
@@ -1,48 +1,84 @@
 import 'package:atlas_icons/atlas_icons.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
-import 'package:flutter_riverpod/flutter_riverpod.dart';
 
-final searchValueProvider = StateProvider.autoDispose<String>((ref) => '');
+class ActerSearchWidget extends StatefulWidget {
```
kumarpalsinh25, line 5: Optimised `ActerSearchWidget` code to make it generic in nature and can be used across entire app based on all the use-case requirements.

app/lib/common/widgets/acter_search_widget.dart
```diff
@@ -1,48 +1,84 @@
...
   });
 
   @override
-  Widget build(BuildContext context, WidgetRef ref) {
-    return _buildSearchBar(context, ref);
+  State<ActerSearchWidget> createState() => _ActerSearchWidgetState();
+}
+
+class _ActerSearchWidgetState extends State<ActerSearchWidget> {
+  final TextEditingController searchTextController = TextEditingController();
```
gtalha07, line 34: Think we should have it disposed in ```dispose()``` method.

app/lib/features/chat/widgets/rooms_list.dart
```diff
@@ -108,46 +96,22 @@ class _RoomsListWidgetState extends ConsumerState<RoomsListWidget> {
   }
 
   Widget filterBox(BuildContext context) {
-    final hasSearchTerm = ref
-            .watch(roomListFilterProvider.select((value) => value.searchTerm))
-            ?.isNotEmpty ==
-        true;
+    final searchTerm = ref.read(roomListFilterProvider).searchTerm;
```
gtalha07, line 99: This should be ```ref.watch``` right? As its part of ```build()``` method ?

PR #2247, opened 2024-10-11
fix space visibility by fetching correct room for join rules

### LINKED ISSUE #2085: Limited space visibility : Space with access selection not working correctly
https://github.com/user-attachments/assets/abbf43e1-6b0b-4fc9-a48c-01a1071e8d69

gtalha07:
fixes #2085.

 
**Could probably fix [meta#523](https://github.com/acterglobal/a3-meta/issues/523) because not reproducible here. Needs Confirmation.**

Preview:

https://github.com/user-attachments/assets/143257ac-8090-4fd1-a190-5e93342b92a9

### DIFF
--- a/app/lib/features/space/settings/pages/visibility_accessibility_page.dart
+++ b/app/lib/features/space/settings/pages/visibility_accessibility_page.dart
@@ -97,7 +97,8 @@ class _VisibilityAccessibilityPageState
             EasyLoading.showToast(L10n.of(context).visibilityNoPermission);
             return;
           }
-          if (value == RoomVisibility.SpaceVisible) {
+          if (value == RoomVisibility.SpaceVisible &&
+              allowedSpaces.valueOrNull?.isEmpty == true) {
             selectSpace(spaceId);
           } else {
             updateSpaceVisibility(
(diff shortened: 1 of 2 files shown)


app/lib/features/space/settings/pages/visibility_accessibility_page.dart
```diff
@@ -249,7 +249,7 @@ class _VisibilityAccessibilityPageState
       final spaceId = await selectSpaceDrawer(context: context);
       if (spaceId != null) {
         final spaceList =
-            await ref.read(joinRulesAllowedRoomsProvider(spaceId).future);
+            await ref.read(joinRulesAllowedRoomsProvider(roomId).future);
         final isAlreadyAdded = spaceList.any((roomId) => roomId == spaceId);
```
kumarpalsinh25, line 253: What is different between `roomId` and `widget.roomId`?

kumarpalsinh25 (approved):
Nothing much to review from code-wise. Added minor comment to address before merge.

app/lib/features/space/settings/pages/visibility_accessibility_page.dart
gtalha07: the ```roomId``` is the id we selected from drawer while ```widget.roomId``` is the room identifier we've opened settings.

kumarpalsinh25: When I look at entire code file, it looks like both are same. widget.roomId pass to the function argument where roomId is the param.

PR #2248, opened 2024-10-11
add labs flag for next-gen chat switch

gtalha07:
- Adds the labs flag for switching between chat & chat-ng. Nothing is changed in terms of features, and future iteration code is copied.

### ***This is the base starting point for incremental pushes to new generational chat. Features are not meant to be stable and should be carefully considered before switching.

Screenshot 2024-10-11 at 3 04 21 PM

(Mobile) ->

https://github.com/user-attachments/assets/1adecc59-8409-484b-910d-eb1c31a523b3

(Desktop)->

https://github.com/user-attachments/assets/de5f39f0-bb80-41aa-b841-c8ab689fb096

### DIFF
--- a/app/lib/features/chat/pages/room_page.dart
+++ b/app/lib/features/chat/pages/room_page.dart
@@ -7,7 +7,6 @@ import 'package:acter/common/themes/acter_theme.dart';
 import 'package:acter/common/utils/routes.dart';
 import 'package:acter/common/utils/utils.dart';
 import 'package:acter/common/widgets/frost_effect.dart';
-import 'package:acter/features/chat/chat-ng/widgets/chat_room_ng.dart';
 import 'package:acter/features/chat/providers/chat_providers.dart';
 import 'package:acter/features/chat/widgets/avatar_builder.dart';
 import 'package:acter/features/chat/widgets/bubble_builder.dart';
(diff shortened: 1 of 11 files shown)


gtalha07:
> @gtalha07 Can I have screen recording of NG Chat for reference?

This is just cloning ```ChatRoom``` and ```CustomInput``` widget code and adding flag to switch in between. Don't think screen-recording would be any useful. Nothing new added in cloned copy...

app/lib/features/chat/pages/room_page.dart
```diff
@@ -119,13 +121,14 @@ class RoomPage extends ConsumerWidget {
...
   Widget build(BuildContext context, WidgetRef ref) {
+    final isChatNG = ref.watch(isActiveProvider(LabsFeature.chatNG));
     return OrientationBuilder(
       builder: (context, orientation) => Scaffold(
         resizeToAvoidBottomInset: orientation == Orientation.portrait,
         body: Column(
           children: [
             appBar(context, ref),
-            ChatRoom(roomId: roomId),
+            isChatNG ? ChatNGRoom(roomId: roomId) : ChatRoom(roomId: roomId),
```
gnunicorn, line 131: I am not super convinced. I'd rather we don't interfere in the existing features at all. I'd rather have the check being done in the router where if `isNG` we give a different set routes with a different set of widgets...

gnunicorn (approved):

app/lib/features/chat/widgets/chat_layout_builder.dart
```diff
@@ -16,19 +18,26 @@ class ChatLayoutBuilder extends StatelessWidget {
   });
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     final expanded = expandedChild;
     final center = centerChild;
+    final isChatNG = ref.watch(isActiveProvider(LabsFeature.chatNG));
+    final roomRoute = isChatNG ? Routes.chatNGRoom.name : Routes.chatroom.name;
+    print('$isChatNG');
```
gnunicorn: a) print in production
b) this is what I wanted to avoid specifically... The widgets _should not_ have to figure out using different routes, but use the same route and then _the router_ decided were to go to ...

PR #2253, opened 2024-10-14
Acter PinList Widget

kumarpalsinh25:
Introducing all new generic `PinListWidget` to Acter widget library set which makes pin list behaviour and UI-UX consistent across entire Acter App.

### Things to know about `PinListWidget`:
- Optional param name `spaceId` makes this widget to re-used for showing **All Pins** as well as **Space Specific Pins**.
- Optional param name `searchValue` enables this widget to show pins list based on the given search criteria.
- Optional param name `limit` allows you to show only specific pin items in list. 
- You can also show Pin Section Header by passing `showSectionHeader` param to `true`.
- You can also manage scrolling behaviour of the pin list by using `shrinkWrap` param

### Implemented `PinListWidget` in below screens.
- All ping list (When we click on Pin Feature Button from Dashboard)
- Space Overview Pins Section
- All Space specific pins list (When we click on `see all` button from pin sections of space overview)
- [Future use-case] : Quick search screen - Default Pins Section when search value is empty
- [Future use-case] : Quick search screen - Pin list with given search value criteria

Reference Video:

https://github.com/user-attachments/assets/c3d699bc-7966-4ad0-8cfc-4534970cd295


gtalha07 (approved):
Looks good!

PR #2271, opened 2024-10-15
Refactoring news editor

gnunicorn:
Currently news fails to post some times but the errors we get are insufficient as they don't give use enough information nor context, see https://github.com/acterglobal/a3-meta/issues/587 . In order to make it more user and developer friend this, refactored the code to give the user and us specific error messages (rather than `!`-failures without line numbers :( ) and provides the ability for the user to clear all state and start over (as it isn't possible right now to recover from a whatever broken state).

- [x] Removing all `!` operators and replace them with proper throws (as we can't really debug errors right now) 24215b9187426228232036ee29773403c6446921
- [x] Refactor to put each slide type into a sub-function for better legibility and overview 24215b9187426228232036ee29773403c6446921
- [x] A few simplifications on the editor provider for better resilience d2dd61cc9fa4db1526f8040fbef848fcda6bd3a0
- [x] When closing the editor, we also clear the draft, after asking the user bb2da705aac6eea69f05fdcfef30b3dece20a640
- [x] Attempted to create test for this, but the appflowy editor failed in them, so they are disabled but included in bb2da705aac6eea69f05fdcfef30b3dece20a640
- [x]  32aed4d9f12899173e3faca0e37ee68d5519ccc3 adds test coverage support
- [x] tests for closing without info
- [x] test with confirmation dialog and draft behavior fixes


app/lib/common/utils/utils.dart
```diff
@@ -466,3 +466,16 @@ extension Let<T> on T? {
     return value == null ? null : op(value);
   }
 }
+
+/// Helper to allow you to replace `!` with a neat and simple
+/// `.expect('Error Messages')` that will throw with that specific
+/// error message.
+extension Expect<T> on T? {
```
gtalha07, line 473: nice!

app/lib/features/news/pages/add_news_page.dart
```diff
@@ -141,15 +190,12 @@ class AddNewsState extends ConsumerState<AddNewsPage> {
 
   //Action Button
   Widget actionButtonUI(BuildContext context) {
-    return Visibility(
```
gtalha07, line 144: Does this mean it now validates without having singular boost to post???

gtalha07 (approved):
Looks good overall. Please see my comment for address before merge!

app/lib/features/news/pages/add_news_page.dart
gnunicorn: no, this if just has been moved to the outside caller rather than render not animated invisible elements - https://github.com/acterglobal/a3/pull/2271/files/486be7ffe7f856e992d00b5a1f5ab757d9bafbdb#diff-54bc38247b32ff741c73e257f09a65d854d5d6c7f02ae92a24b08f282f86b5b1R99

PR #2273, opened 2024-10-16
All Space list with search option

kumarpalsinh25:
- Optimized all space list code
- Enable search option on all space list screen
- Added all new `SpaceListWidget` which can re-used based on the requirements.

### Things to know about `SpaceListWidget`:
- Optional param name `searchValue` enables this widget to show space list based on the given search criteria.
- Optional param name `limit` allows you to show only specific space items in list.
- You can also show Space Section Header by passing `showSectionHeader` param to `true`.
- You can also manage scrolling behaviour of the space list by using `shrinkWrap` param

### Implemented `SpaceListWidget` in below screens.
- All space list (When we click on` See All` Button from Dashboard My Spaces Header)
- [Future use-case] : Quick search screen - Default Space Section when search value is empty
- [Future use-case] : Quick search screen - Space list with given search value criteria

Reference Video:

https://github.com/user-attachments/assets/e2223af6-1b2c-4fa5-9eaf-cd86e9fcce17


app/lib/features/spaces/widgets/space_list_widget.dart
```diff
@@ -0,0 +1,92 @@
...
+    if (spaceList.isEmpty) return emptyState;
+
+    final count = (limit ?? spaceList.length).clamp(0, spaceList.length);
+    return showSectionHeader
+        ? Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              SectionHeader(
+                title: L10n.of(context).spaces,
+                isShowSeeAllButton: count < spaceList.length,
```
gnunicorn, line 70: as this PR doesn't contain any active version of `showSectionHeader` I am not sure what this does in here, but I _suspect_ this is meant for the future use case of using this as the embedded dashboard widget as well: if that is the case I want to point out that if we showed only the bookmarks, then the spaceList.length might as well be >= count yet we want to have the `isShwoSeeAllButton` visible regardless _if_ there are more spaces _in general_. In case of bookmarks we currently show _all_ bookmarks (as the order _within_ bookmarks isn't deterministic and thus the user can be very annoyed if they bookmarked a spaces but it doesn't show up) and only limit in the non-bookmarked case.

Just to keep in mind how this is meant to work ...

gnunicorn (approved):
Looking good,

some minor code improvements and one remark for the future.

app/lib/features/spaces/widgets/space_list_widget.dart
```diff
@@ -0,0 +1,92 @@
...
+  final int? limit;
+  final bool showSectionHeader;
+  final VoidCallback? onClickSectionHeader;
+  final bool shrinkWrap;
+  final Widget emptyState;
+
+  const SpaceListWidget({
+    super.key,
+    this.limit,
+    this.searchValue,
```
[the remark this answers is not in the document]
kumarpalsinh25: I wanted to make `SpaceListWidget` independent of any params. So `searchValue` shouldn't be mandatory in that way. Yes, in current scenario `searchValue` is given to this widget but this not how generic widget I am thinking off. It can adopt to any use-case.

app/lib/features/spaces/widgets/space_list_widget.dart
kumarpalsinh25: Yes this PR doesn't contain `SectionHeader` work but I already started work on [Next Quick Search PR](https://github.com/acterglobal/a3/pull/2275) on top of this PR where it is require.

Screenshot 2024-10-17 at 11 29 33 AM

I understand the use-case of my bookmarks section of dashboard and that's why haven't touch it at the moment with this code. But I needed section header stuff in Quick search PR as shown in above screenshot.

PR #2274, opened 2024-10-16
Remove bang operator from nullable variables [7]

bitfriend:
This small PR was split from [bigger PR](https://github.com/acterglobal/a3/pull/2125) so that reviewer can see changes easily.


app/lib/features/chat/widgets/chats_list.dart
```diff
@@ -185,17 +185,19 @@ class __AnimatedChatsListState extends State<_AnimatedChatsList> {
...
     }
   }
 
   void _remove(int pos, String data) {
     _currentList.removeAt(pos);
-    if (_listKey.currentState != null) {
-      _listKey.currentState!.removeItem(
+    final curState = _listKey.currentState;
+    if (curState != null) {
+      curState.removeItem(
```
gnunicorn, line 200: The reason you are not using `.let` here is because there is some else? how about extending our `let` helper to take an optional `else:`-function that would be called?

(side note: why is it called `let` if it maps the value to the result of the function in the first place?)

gnunicorn (approved):
Code seems fine, just some minor queries.

app/integration_test/support/chats.dart
```diff
@@ -21,8 +21,8 @@ extension ActerChat on ConvenientTest {
       await select.should(findsNWidgets(counter));
     }
 
-    if (ids?.isNotEmpty == true) {
-      for (final id in ids!) {
+    if (ids != null && ids.isNotEmpty == true) {
```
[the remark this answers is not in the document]
bitfriend: It was intended to remove bang operator from `ids!`.

app/lib/features/chat/chat-ng/widgets/chat_input_ng.dart
```diff
@@ -557,10 +558,9 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
               .width(image.width)
               .height(image.height);
           if (inputState.selectedMessageState == SelectedMessageState.replyTo) {
-            await stream.replyMessage(
-              inputState.selectedMessage!.remoteId!,
-              imageDraft,
-            );
+            final remoteId = inputState.selectedMessage?.remoteId;
+            if (remoteId == null) throw 'remote id of sel msg not available';
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/772cfc5e2a4312afffae1da267c581df99f95f70

app/lib/features/chat/chat-ng/widgets/chat_input_ng.dart
```diff
@@ -569,35 +569,30 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
...
           final videoDraft =
               client.videoDraft(file.path, mimeType).size(file.lengthSync());
-
           if (inputState.selectedMessageState == SelectedMessageState.replyTo) {
-            await stream.replyMessage(
-              inputState.selectedMessage!.remoteId!,
-              videoDraft,
-            );
+            final remoteId = inputState.selectedMessage?.remoteId;
+            if (remoteId == null) throw 'remote id of sel msg not available';
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/772cfc5e2a4312afffae1da267c581df99f95f70

app/lib/features/chat/chat-ng/widgets/chat_input_ng.dart
```diff
@@ -569,35 +569,30 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
...
           final fileDraft =
               client.fileDraft(file.path, mimeType).size(file.lengthSync());
-
           if (inputState.selectedMessageState == SelectedMessageState.replyTo) {
-            await stream.replyMessage(
-              inputState.selectedMessage!.remoteId!,
-              fileDraft,
-            );
+            final remoteId = inputState.selectedMessage?.remoteId;
+            if (remoteId == null) throw 'remote id of sel msg not available';
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/772cfc5e2a4312afffae1da267c581df99f95f70

app/lib/features/chat/chat-ng/widgets/chat_input_ng.dart
```diff
@@ -716,9 +709,13 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
       );
 
       if (inputState.selectedMessageState == SelectedMessageState.replyTo) {
-        await stream.replyMessage(inputState.selectedMessage!.remoteId!, draft);
+        final remoteId = inputState.selectedMessage?.remoteId;
+        if (remoteId == null) throw 'remote id of sel msg not available';
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/772cfc5e2a4312afffae1da267c581df99f95f70

app/lib/features/chat/chat-ng/widgets/chat_input_ng.dart
```diff
@@ -716,9 +709,13 @@ class __ChatInputState extends ConsumerState<_ChatInput> {
...
+        await stream.replyMessage(remoteId, draft);
       } else if (inputState.selectedMessageState == SelectedMessageState.edit) {
-        await stream.editMessage(inputState.selectedMessage!.remoteId!, draft);
+        final remoteId = inputState.selectedMessage?.remoteId;
+        if (remoteId == null) throw 'remote id of sel msg not available';
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/772cfc5e2a4312afffae1da267c581df99f95f70

app/lib/features/chat/widgets/chats_list.dart
bitfriend: Yes, also I was hopping to define `else:` clause in `let` extension.
But I didn't know how to extend it.

bitfriend: `Let` extension was designed by @gtalha07.
I thought he got inspired by `let` statement in rust language.
@gnunicorn If you have better idea in extension name, let's discuss together.

gnunicorn: nah, it is fine for this PR... I just noticed that it is a bit off ... as `map` is what it does ...

@gtalha07 what do you mean with it is reserved? that `.map` is used on lists/iterators? That's the same for rust... I don't see that as being a problem ... but our use of `.let` having side-effects feels quirky too me ... I might propose changing the idioms here a bit ...

bitfriend: I could modify `Let` extension as following.
```dart
extension Let<T> on T? {
  R? let<R>(R? Function(T) op, {void Function()? orElse}) {
    final T? value = this;
    if (value != null) return op(value);
    if (orElse != null) orElse();
    return null;
  }

  // it supports async callback too unlike `extension_nullable`
  Future<R?> letAsync<R>(R? Function(T) op, {void Function()? orElse}) async {
    final T? value = this;
    if (value != null) return op(value);
    if (orElse != null) orElse();
    return null;
  }
}
```

gtalha07: I suppose we could term it as ```option```? Since the type behaviour is derived from it ?? `map` isn't suitable as its dart reserved.

bitfriend: I think there are many occurrences of `.let()` function that should be replaced with `orElse` clauser.
I am going to handle this issue in another PR.

bitfriend: @gnunicorn I think @gtalha07 idea is correct.
Please think about case that we should wrap the original `.map()` by this extension.
If both are `.map()`, which one would be our extension?

gnunicorn: > `map` isn't suitable as its dart reserved.

no it isn't. Or I misunderstand what you mean by that. But if you checked https://github.com/acterglobal/a3/pull/2280 you'll see that using the name `map` isn't a problem at all..

bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2274/commits/710566572d988d450628f56ef7342f6b622b819d

bitfriend: @gnunicorn @gtalha07 Thank you for your earnest cooperation.

PR #2275, opened 2024-10-16
All New Quick search

kumarpalsinh25:
Introducing all new search page. Check below video reference to know more about this PR work.

Mobile:

https://github.com/user-attachments/assets/bc67b4a0-1c24-43fb-a25a-08d835dad7e9

Desktop:

https://github.com/user-attachments/assets/7910f4a8-03f5-4ede-b0b4-47b638438d4a

### Things to know:
- Search page is made with all new Acter Widgets as below which makes search page code pretty short and cleaner.
   - [Acter Search Widget](https://github.com/acterglobal/a3/pull/2243) 
   - [Acter Pin List Widget](https://github.com/acterglobal/a3/pull/2253)
   - [Acter Space List Widget](https://github.com/acterglobal/a3/pull/2273)

- Currently search page only contain spaces and pins but later stage, planning to also add Task and Event over there.
- Quick action button are no more part of search page so they are removed but planning added them in Home Plus Widget right after this PR.


app/lib/common/utils/routes.dart
```diff
@@ -108,7 +108,6 @@ enum Routes {
   settingsSuperInvites('/settings/super_invites'),
   // -- utils
   bugReport('/bug-report'),
-  quickJump('/quick-jump'),
```
kumarpalsinh25, line 111: No need of additional general route as Search is part of shell-routes itself.

app/lib/config/desktop.dart
```diff
@@ -54,6 +54,10 @@ class _DesktopSupportState extends State<DesktopSupport>
     );
     Menu menu = Menu(
       items: [
+        MenuItem(
+          key: 'search',
+          label: 'Search',
+        ),
```
kumarpalsinh25, line 60: I don't know what `desktop.dart` file but I added search page in menu item as there where other Menu where there.

app/lib/config/desktop.dart
```diff
@@ -132,7 +136,10 @@ class _DesktopSupportState extends State<DesktopSupport>
 
     await windowManager.show();
     WidgetsBinding.instance.addPostFrameCallback((Duration duration) async {
-      if (menuItem.key == 'home') {
+      if (menuItem.key == 'search') {
+        _log.info('route search');
+        rootNavKey.currentContext!.pushNamed(Routes.search.name);
+      } else if (menuItem.key == 'home') {
```
kumarpalsinh25, line 142: Same as above.

app/lib/features/search/pages/quick_search_page.dart
```diff
@@ -0,0 +1,82 @@
...
+import 'package:acter/common/widgets/acter_search_widget.dart';
+import 'package:acter/features/pins/widgets/pin_list_widget.dart';
+import 'package:acter/features/search/providers/quick_search_providers.dart';
+import 'package:acter/features/spaces/widgets/space_list_widget.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:go_router/go_router.dart';
+
+class QuickSearchPage extends ConsumerStatefulWidget {
```
kumarpalsinh25, line 11: This is all new search page with minimal and cleaner code.

app/lib/config/desktop.dart
gtalha07: Which design it's following for the desktop now? Could you share video preview?

app/lib/features/search/pages/quick_search_page.dart
```diff
@@ -0,0 +1,82 @@
...
+  const QuickSearchPage({super.key});
+
+  @override
+  ConsumerState<QuickSearchPage> createState() => _QuickSearchPageState();
+}
+
+class _QuickSearchPageState extends ConsumerState<QuickSearchPage> {
+  String get searchValue => ref.watch(quickSearchValueProvider);
```
gtalha07, line 19: So, I like this approach of watching states over ```build()```, IF IT DOESN'T come with side-effects. Can we document this in dev docs recommended practices to follow in all places ?

app/lib/config/desktop.dart
kumarpalsinh25: Yes sure, updated Desktop Video in PR Description.

app/lib/features/search/pages/quick_search_page.dart
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+
+  Widget quickSearchSectionsUI() {
+    return SingleChildScrollView(
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.stretch,
+        children: [
+          SpaceListWidget(
+            limit: 3,
```
gtalha07, line 61: I was thinking this use-case could become more useful by showing recently seen items .i.e. spaces, pins etc? As Element also supports this on quick search view.

gtalha07 (approved):
LGTM!

app/lib/features/search/pages/quick_search_page.dart
kumarpalsinh25: Thinking of the same earlier about recent items and had discussion with @gnunicorn also. We conclude to show list with bookmarked Item first in sort order.

PR #2276, opened 2024-10-16
Routing-based chat-ng checking

gnunicorn:
[following my comment on the original post](https://github.com/acterglobal/a3/pull/2248#discussion_r1802580902) this implements an alternative, less copy-pasted approach for the chat-ng checking by only checking it within the router/upon routing and serving different widgets depending on that.

Specifically, this creates a new roomslist that is based on the existing one but has a different title to indicate you have the feature activated. Which widget to use is decided upon routing. This has the benefit that to the outside both the existing and the ng-version have the same routes and behavior and as they have the same routes.

https://github.com/user-attachments/assets/a23a5da9-26ad-4323-928f-45f5704a59d8

Split into two commits:
- 35785b05585100f857acc14307f27a440e023629 reverts most of the previous PR merge
- 640f8d33a91970636cf535492178e44672401caf implements this approach


app/lib/router/shell_routers/chat_shell_router.dart
```diff
@@ -1,16 +1,35 @@
...
+import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 
+/// define the chat layout builder depending on whether the user has set
+/// the chat-ng feature.
+Widget _chatLayoutBuilder({Widget? centerChild, Widget? expandedChild}) {
+  final isChatNg =
+      rootNavKey.currentContext?.read(isActiveProvider(LabsFeature.chatNG)) ==
+          true;
+  return isChatNg
```
gnunicorn, line 21: this is the actual core of the change: do the `isActivateProvider(LabsFeature.chatNG)` check within the GoRoute-Builder and pick the right Widget accordingly.

gtalha07 (approved):
Thanks for the clarification, seems to misunderstood how we want to handle chat switch on route level. Looks good!

PR #2278, opened 2024-10-16
fix convo card trailing

### LINKED ISSUE #2231: Missing time of latest message in chat rooms list view
Nothing shown anymore on the right side :(

gtalha07:
Fixes #2231. Bug fix include adding fallback ```trailing``` widget renderer as ```trailing``` param isn't passed in any call. That's why the message timestamp was missing from chat list.

### DIFF
--- a/app/lib/features/chat/widgets/convo_card.dart
+++ b/app/lib/features/chat/widgets/convo_card.dart
@@ -84,7 +84,9 @@ class ConvoCard extends ConsumerWidget {
                         overflow: TextOverflow.ellipsis,
                       ),
                 subtitle: buildSubtitle(context, constraints),
-                trailing: constraints.maxWidth < 300 ? null : trailing,
+                trailing: constraints.maxWidth < 300
+                    ? null
+                    : trailing ?? _TrailingWidget(roomId: roomId),
               ),
             ),
           ],


app/lib/features/chat/widgets/convo_card.dart
```diff
@@ -84,7 +84,9 @@ class ConvoCard extends ConsumerWidget {
                         overflow: TextOverflow.ellipsis,
                       ),
                 subtitle: buildSubtitle(context, constraints),
-                trailing: constraints.maxWidth < 300 ? null : trailing,
+                trailing: constraints.maxWidth < 300
```
kumarpalsinh25, line 87: Didn't getting visual idea about issue point and resolution by this code change.

Can you elaborate more about this in PR Description?

gtalha07: Added description.

kumarpalsinh25: what is difference between `trailing`  & `_TrailingWidget(roomId: roomId)` ?

gtalha07: It serves as fallback if ```trailing``` param isn't provided.

PR #2280, opened 2024-10-16
Options Extension Proposal

gnunicorn:
I think we should make `T?.let` into `T?.map` for better consistency. Additionally this adds tests and then expands upon the idea by adding an optional `defaultValue` that is returned instead of the null and adds an `expand` - both with tests of course.

This builds on top of #2279 .


bitfriend (approved):
Looks good to me

bitfriend:
I know you intended this PR to fix [this issue](https://github.com/acterglobal/a3/pull/2274#discussion_r1803250646).
You may implement `orElse` like [this comment](https://github.com/acterglobal/a3/pull/2274#discussion_r1803285315).

gnunicorn:
@bitfriend I've added `orElse` support, including tests.

PR #2288, opened 2024-10-17
Upgrading Build System

gnunicorn:
Upgrades the build system, in particular for Android.

PR #2289, opened 2024-10-17
Synchronize Calendar Event sync

gnunicorn:
This address a the calendar sync labs feature. In particular it ensures that only one sync attempt happens at the same time by debouncing, keeping a lock and run on the latest version provided (again) if we received updates during a run.

The problem was that because we receive the events async and update on the rsvp stream might also happen async we saw the listener triggering the refresh faster than it took to execute, resulting in inconsistency of the managed tracked calendars and events we use for updates and caching. As a result events as well as calendars appeared multiple times on users devices. I was able to reproduce this on the emulator by rsvp'ing on an event very often triggering several updates and thus multiple evens in the device calendars.

With this fix that should be gone now. Which I was also able to confirm with the emulator.

built on top of #2288  (as I was unable to reproduce or test it without that). The only relevant commit is 20fc77405994923f8d94c61ba490111b667a7cb4


app/lib/features/calendar_sync/calendar_sync.dart
```diff
@@ -91,17 +92,54 @@ Future<void> initCalendarSync({bool ignoreRejection = false}) async {
...
+  String calendarId,
+  List<EventAndRsvp> events,
+) {
+  _debounce?.cancel(); // cancel the current debounce;
+  _debounce = Timer(const Duration(seconds: 3), () {
+    _debounce = null;
+    _refreshLoop();
+  });
+  _next = (calendarId, events);
+}
```
gnunicorn, line 122: The actual core to ensure the synchronisity quickly explained:

1. we have a timer (`_debounce`) we reset on every schedule call to issue in 3 seconds time - this allows us to wait for a bunch of things to come in at first before doing the actual update
2. then we save the current values.
3. when the timer triggers it resets the current timer and starts the refreshloop

following below.

app/lib/features/calendar_sync/calendar_sync.dart
```diff
@@ -91,17 +92,54 @@ Future<void> initCalendarSync({bool ignoreRejection = false}) async {
...
+        break;
+      }
+
+      final (calendarId, events) = next;
+      await _refreshCalendar(calendarId, events);
+    }
+  } finally {
+    _running = false;
+  }
+}
```
gnunicorn, line 141: when the loop is triggered it first ensure with the `_running`-flag that it isn't running. if it still is, this run quits as the other loop will take over.

during the loop we take the currently stored params and execute the refreshCalendar with it, once at a time. We reset the globally stored params at the start of the loop, but if an update happens while we are running, the next loop cycle will pull it up - giving us a super simple "latest-only" worker-queue system. If there are no more latest to process it quits and resets the `_running` flag to allow the next loop to start...

kumarpalsinh25 (commented):
Doesn't really getting much idea about code changes. Would love to have review of @gtalha07.

app/lib/features/calendar_sync/calendar_sync.dart
```diff
@@ -97,28 +96,37 @@ Future<void> initCalendarSync({bool ignoreRejection = false}) async {
...
 // makes sure there is only one running at a time
-void _scheduleRefresh(
+@visibleForTesting
+Future<void> scheduleRefresh(
   String calendarId,
   List<EventAndRsvp> events,
 ) {
   _debounce?.cancel(); // cancel the current debounce;
-  _debounce = Timer(const Duration(seconds: 3), () {
+  _completer ??= Completer<void>();
```
gnunicorn, line 118: this completer is reused which allows us to call scheduleRefresh multiple times receiving the same future to complete on the latest of its calls - which is used in the tests to ensure completion is done.

app/test/features/calendar_sync/calendar_sync_test.dart
```diff
@@ -0,0 +1,76 @@
...
+
+      // create in fast succession;
+      final waiter = Future.wait([
+        scheduleRefresh(calendarId, firstTry),
+        scheduleRefresh(calendarId, secondTry),
+        scheduleRefresh(calendarId, thirdTry),
+      ]);
+
+      // advance the timer
+      await tester.pump(const Duration(seconds: 2));
```
gnunicorn, line 57: this is what happened before: we were called multiple times in a very short time frame with slightly different input, mainly on RSVP status and because it was executed simultaneously, the mapping as well as the actual device events got out of sync.


this test does that, then advances below the debounce threshold and confirms nothing happened, then advances way after that threshold and ensures it was only executed _once_ and has the right mapping saved.

gnunicorn:
@kumarpalsinh25 I added a unit test and further comments to explain better what this PR is about as well as ensure we don't have this problem again in the future.

gtalha07 (approved):
Thanks for the effort of putting detailed explanation! This made understanding much easier of the issue and fix. Approved.

PR #2297, opened 2024-10-22
Tracking boost views

gnunicorn:
This PR replaces the existing (and not used anymore) read receipt controller on the rust side with a new mechanism to track boost reads.

See that counter in action:

https://github.com/user-attachments/assets/6d178c55-e357-4c9a-8258-3a9a3ec7b2ba

- [x] remove old receipt controller infrastructure
- [x] integration test for read receipts for boosts
- [x] basic implementation of read receipts
- [x] ensure updates ping signals
- [x] expose API for UI
- [x] implement simple Tracking in the UI
- [ ] Changelog

Due to [the problem lined out below](https://github.com/acterglobal/a3/pull/2297#issuecomment-2430377744), we can't reuse the existing read-receipt system [but I had to implement a custom event and manager for this](https://github.com/acterglobal/a3/pull/2297/commits/e2b47c0748bd31a9f95457ce7b811aaa662afc95).


gnunicorn:
I noticed some odd behavior when trying this in the app itself. Despite all the tests working as expected, when scrolling down the read markers weren't set though the code was sending them. Reading [up on how the spec expects read-receipts to work](https://spec.matrix.org/v1.12/client-server-api/#receipts), I've come to realize that this doesn't quite fit what we are looking: it assumes a linear timeline and thus any receipt is considered an _up to this event_ read-receipt. Thus when the read is sent for the _first_ news we are looking at, all the subsequent sends for other news we seen later (but that happened _before_), the read will not actually be doing anything on the server.

I tried abusing the thread_id but it seems the server is rejecting them if there isn't a clear relation, which we don't have for the primary item.

It appears that I have to redo this entirely differently with completely custom events and models after all, probably similar to the comments & likes. That is a bit sad.

gtalha07 (approved):
LGTM!

bitfriend (approved):
I found you made many changes in both rust side and flutter side.
Great work.

PR #2299, opened 2024-10-22
Auto-updating Event Details via riverpod "utcNowProvider"

### LINKED ISSUE #2256: Upcoming events don't refresh properly
because our riverpod providers don't refresh on time but only on data.

gnunicorn:
Leveraging the riverpod system to have a cached now-utc-time that is updated once a minute and thus refreshes the data to the current time. 

Fixes #2256 .

### DIFF
--- a/app/lib/features/events/providers/event_providers.dart
+++ b/app/lib/features/events/providers/event_providers.dart
@@ -1,6 +1,6 @@
 import 'package:acter/features/bookmarks/providers/bookmarks_provider.dart';
 import 'package:acter/features/bookmarks/types.dart';
-import 'package:acter/features/events/actions/get_event_type.dart';
+import 'package:acter/features/events/providers/event_type_provider.dart';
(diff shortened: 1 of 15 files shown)


app/lib/features/events/providers/event_providers.dart
```diff
@@ -95,7 +99,9 @@ final allPastEventListProvider = FutureProvider.autoDispose
     .family<List<ffi.CalendarEvent>, String?>((ref, spaceId) async {
   final allEventList = await ref.watch(allEventListProvider(spaceId).future);
   List<ffi.CalendarEvent> allPastEventList = allEventList
-      .where((event) => getEventType(event) == EventFilters.past)
+      .where(
+        (event) => ref.watch(eventTypeProvider(event)) == EventFilters.past,
```
bitfriend, line 103: I wonder if overload not occurred when a `ref.watch()` is placed in other `ref.watch()`.

kumarpalsinh25 (approved):
LGTM!

PR #2301, opened 2024-10-23
Boost : Go to latest post

kumarpalsinh25:
Double-Tap on Boost bottom nav button to go at latest Boost.

https://github.com/user-attachments/assets/88c28e94-167d-463b-8a56-a46b31614350

### DIFF
--- a/app/lib/features/home/widgets/custom_selected_icon.dart
+++ b/app/lib/features/home/widgets/custom_selected_icon.dart
@@ -13,8 +13,7 @@ class CustomSelectedIcon extends ConsumerWidget {
     return InkWell(
       onDoubleTap: () {
         if (key == MainNavKeys.updates) {
-          final notifier = ref.read(boostToToPageProvider.notifier);
-          notifier.state = 0;
+          ref.read(boostGoToPageProvider.notifier).goToPage(0);
         }
       },
       child: Container(
--- a/app/lib/features/home/widgets/custom_selected_icon.dart
+++ b/app/lib/features/home/widgets/custom_selected_icon.dart
@@ -1,21 +1,33 @@
+import 'package:acter/features/home/data/keys.dart';
+import 'package:acter/features/news/providers/news_providers.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 
-class CustomSelectedIcon extends StatelessWidget {
+class CustomSelectedIcon extends ConsumerWidget {
   final Widget icon;
+
   const CustomSelectedIcon({super.key, required this.icon});
 
   @override
-  Widget build(BuildContext context) {
-    return Container(
-      height: 30,
-      width: 54,
-      margin: const EdgeInsets.symmetric(horizontal: 8),
-      padding: const EdgeInsets.only(bottom: 2),
-      decoration: BoxDecoration(
-        color: Theme.of(context).colorScheme.primary,
-        borderRadius: BorderRadius.circular(14),
+  Widget build(BuildContext context, WidgetRef ref) {
+    return InkWell(
+      onDoubleTap: () {
+        if (key == MainNavKeys.updates) {
+          final notifier = ref.read(boostToToPageProvider.notifier);
+          notifier.state = 0;
+        }
+      },
+      child: Container(
+        height: 30,
+        width: 54,
+        margin: const EdgeInsets.symmetric(horizontal: 8),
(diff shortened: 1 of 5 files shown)


app/lib/features/home/widgets/custom_selected_icon.dart
```diff
@@ -1,21 +1,32 @@
...
-      margin: const EdgeInsets.symmetric(horizontal: 8),
-      padding: const EdgeInsets.only(bottom: 2),
-      decoration: BoxDecoration(
-        color: Theme.of(context).colorScheme.primary,
-        borderRadius: BorderRadius.circular(14),
+  Widget build(BuildContext context, WidgetRef ref) {
+    return InkWell(
+      onDoubleTap: () {
+        if (key == MainNavKeys.updates) {
+          ref.read(boostGoToPageProvider.notifier).state = 0;
```
kumarpalsinh25, line 16: This code line supposed to trigger `ref.listenManual(boostGoToPageProvider,..` in news_full_view.dart as shown below.
@gnunicorn

app/lib/features/news/widgets/news_full_view.dart
```diff
@@ -25,9 +25,18 @@ class _NewsVerticalViewState extends ConsumerState<NewsFullView> {
   @override
   void initState() {
     super.initState();
-    _pageController = PageController(
-      initialPage: widget.initialPageIndex,
-    );
+    _pageController = PageController(initialPage: widget.initialPageIndex);
+    //LISTEN for boost index
+    ref.listenManual(boostGoToPageProvider, fireImmediately: true,
```
kumarpalsinh25, line 30: `ref.read(boostGoToPageProvider.notifier).state = 0;` supposed to trigger this listenManual Function but it is not.
@gnunicorn

app/lib/features/home/widgets/custom_selected_icon.dart
gnunicorn: but when is it `!= 0`? if you set the state to what it already is, riverpod will not propagate this as an update.

kumarpalsinh25: Ahh, I see. Thank you so much for the help.

gnunicorn, line 16: not a fan of having a _specific_ behavior for a single item in the if based on the key in a generic widget. That's not very flutter idiomatic. How about we pass a custom `onDoubleTap` instead?

app/lib/features/news/widgets/news_full_view.dart
```diff
@@ -25,9 +25,21 @@ class _NewsVerticalViewState extends ConsumerState<NewsFullView> {
...
+        (previous, next) {
+      if (next != null) {
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          _pageController.animateToPage(
+            next,
+            curve: Curves.easeIn,
+            duration: const Duration(milliseconds: 500),
+          );
+          ref.read(boostGoToPageProvider.notifier).state = null;
```
gnunicorn, line 39: okay, that seems like a weird way of triggering actions ... setting to a provider to `0` and resetting it to `0` is not very intuitive for a developer to imagine what this does...

additionally, I am wondering if we can't solve this via either navigation or the `initialPageIndex` that is being used above already ...  Have the doubleTap just set the route again ... which should give it the same initialPageIndex but not as `initState` but instead on the `depdendencyUpdate` ... thus if there is a new `0` (or anything), this should trigger the animation? WDYT?

gnunicorn (changes requested):
I am not convinces this is the proper way to implement this mechanism: it comes across as a strange little hack on all the specific places ...

app/lib/features/home/widgets/custom_selected_icon.dart
kumarpalsinh25: > How about we pass a custom onDoubleTap instead?

hmm.. this look better solution.. Let me change it accordingly.

app/lib/features/news/widgets/news_full_view.dart
kumarpalsinh25: Have the doubleTap just set the route again ... which should give it the same initialPageIndex but not as initState but instead on the depdendencyUpdate ... 

`NewsListPage` is actually not a page but a widget so that will work on doubleTap event?

Is that a good way if I go with static method in `NewsListPage`?

gnunicorn (approved):

PR #2306, opened 2024-10-24
Members list with search

kumarpalsinh25:
fixes, [#2292](https://github.com/acterglobal/a3/issues/2292)

https://github.com/user-attachments/assets/24cbad8c-6833-4172-9345-a2aa44f712bc

### DIFF
--- a/app/lib/common/providers/room_providers.dart
+++ b/app/lib/common/providers/room_providers.dart
@@ -311,9 +311,10 @@ final membersIdWithSearchProvider = FutureProvider.family
   final List<String> foundedMembersId = [];
   for (final memberId in membersIdList) {
     final memberInfo = ref.watch(
-        memberAvatarInfoProvider((userId: memberId, roomId: param.roomId)));
-    final displayName = memberInfo.displayName ?? '';
-    if (displayName.toLowerCase().contains(searchTerm)) {
+      memberAvatarInfoProvider((userId: memberId, roomId: param.roomId)),
+    );
+    if (memberInfo.displayName?.toLowerCase().contains(searchTerm) == true ||
+        memberId.toLowerCase().contains(searchTerm)) {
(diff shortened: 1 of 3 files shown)


app/lib/features/space/pages/members_page.dart
```diff
@@ -72,7 +98,7 @@ class SpaceMembersPage extends ConsumerWidget {
               ),
               itemBuilder: (context, index) => MemberListEntry(
                 memberId: members[index],
-                roomId: spaceIdOrAlias,
+                roomId: widget.spaceIdOrAlias,
               ),
             );
           },
```
gnunicorn, line 104: add a `skipLoadingOnReload: false` to avoid that ugly flickering of the list on typing.

app/lib/common/providers/room_providers.dart
```diff
@@ -295,6 +295,31 @@ final membersIdsProvider =
...
+  final List<String> membersIdList = asDartStringList(members);
+  final searchTerm = param.searchValue.toLowerCase();
+  if (searchTerm.isEmpty) return membersIdList;
+
+  final List<String> foundedMembersId = [];
+  for (final memberId in membersIdList) {
+    final memberInfo = ref.watch(
+        memberAvatarInfoProvider((userId: memberId, roomId: param.roomId)));
+    final displayName = memberInfo.displayName ?? '';
+    if (displayName.toLowerCase().contains(searchTerm)) {
```
gnunicorn, line 316: what about the user-id? I think we want to include that. also, we can shorten this.
```suggestion
    if (memberInfo.displayName?.toLowerCase().contains(searchTerm) == true || memberId.toLowerCase().contains(searchTerm)) {
```

gnunicorn (approved):
looking good. just some minor improvements.

PR #2318, opened 2024-10-25
Fix restarting sync

gnunicorn:
You might have noticed that upon start, the App might show a lot of jigger in the UI of things that have already happened (like old updates to Pins or Tasks). That was due to a bug in the sync mechanism this PR fixes and provides a test to show that it does so. Specifically the bug was caused by:
1. the matrix-rust-sdk not saving the sync token and thus creating a full-sync on every start
2. us restarting the history of all spaces, regardless of whether this first-since-started-sync even said there are updates for us to take care of...

To the reviewers: this is on top of https://github.com/acterglobal/a3/pull/2372 , so please review that first.


.changes/2318-sync-fixes.md
```diff
@@ -0,0 +1 @@
+- Fix to not refresh the entire space history on every app start anymore.
```
gtalha07, line 1: So now, it'd only sync if there's a need for sync ?

gtalha07 (approved):
Looks good!

gtalha07:
Though some tests seem to fail...

.changes/2318-sync-fixes.md
gnunicorn: yeah, that's the idea. so far, we synced the entire history on every start...

gnunicorn:
@bitfriend I was finally able to find the bug introduced by the code and shown in the tests, see https://github.com/acterglobal/a3/pull/2318/commits/1d46b219b3385f3342e6a687a4d1a488f4e006bd

bitfriend (approved):
Looks good to me.
Tried compiling and testing of app in windows and app looks stable.

PR #2324, opened 2024-10-27
Upgrading matrix rust sdk

gnunicorn:
Due to the remaining problems in #2318, I've rebased this PR to only contain the upgrading of the rust SDK and clear out some old stuff.


.github/workflows/check-rust-integration-tests.yml
```diff
@@ -67,12 +67,6 @@ jobs:
       - uses: taiki-e/install-action@cargo-llvm-cov
       - uses: taiki-e/install-action@nextest
 
-      - name: Initialize e2ee mock data
-        run: cargo run -p acter-cli -- mock
-        env:
-          DEFAULT_HOMESERVER_NAME: "localhost"
-          DEFAULT_HOMESERVER_URL: "http://localhost:8118"
-
```
bitfriend, line 75: Is mock data not needed here?

gnunicorn: No, https://github.com/acterglobal/a3/pull/2324/commits/f827014038d941f42a5c56bcbe20a12354ae7dfa replaces all remaining mock-data-based tests with some that just create random local accounts local to the tests. Mock data is not needed anymore after this PR and thus all ability to do that has been removed as well. using the `random_user`-helpers is the recommended way forward.

PR #2331, opened 2024-10-29
Switching from rootNavKey.currentContext! to ProviderContainer

gnunicorn:
This replaces the `rootNavKey.currentContext!` spread throughout our code base with the riverpod provided `ProviderContainer`.

## Why?
The `rootNavKey.currentContext!` requires a context to be present. There are several cases however, when that isn't necessary the case - most notably when we start up or are supposed to handle notifications. As `!` is a failing action, this means those activities can and do fail, as [shown in these logs](https://rageshake.acter.global/api/listing/2024-10-28/225229-WCYFZ7IT/app_2024-10-28_20-29-33.026675261.log.gz) (limited access):

```
[2024-10-28][20:29:36.274547][a3::notifify::matrix][ERROR] Parsing Notification failed: {content_algorithm: m.megolm.v1.aes-sha2, sender_display_name: XXXXX, room_name: YY/Emil/Jo/Ben, sender: @asdaea:acter.global, unread: 4, content_ciphertext: AwgFEpABTgNPP/5gs+zP227dPnd/3qZde5f9GcAhuMc9YzlHk0edAtRzcgABeMv4h/yo3hKpujVbk8aJETUl9k4zuV+FzS0fE+iWD+Dq66KARCVxd9YzKPEcfl7n3OqV0g0x29tcneYoCEDq/GRLBsmEIb3MTTRVWk06lEMHoLHeNqdkCea6xWl32DwBwasdae27gAg8Mj+rNg/3pLVd+5Qyb1SJ9bRC9nREMF8+XYf5cg+ts1HKQZy0FCKUpXkl3w01OMZISVPEUJ, missed_calls: None, prio: high, type: m.room.encrypted, device_id: LBYZFYUHPL, event_id: $pj24lTJda47DtCZ4b9SfEg8u94tE0uHh65rKMHeXudo, content_sender_key: okHn2hBGamjVtMasdafdsnncPHEnVLyn9WrI7teNelo, room_id: !adfasdfasdfasf:acter.global, content_session_id: 1wJKmjgDEuf94xds2Wp95KCTVVU+l1ae5+V7CIm61sc, content_device_id: FECRXDAHMN} Null check operator used on a null value null *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
``` 

We are using `ref.read` in our code base to find out about a huge amount of states and configurations. The main reason we use `rootNavKey.currentContext!` is to get access to `ref.read`. But that isn't reliable, especially outside the widget context.

## What

[Taken from this article](https://codewithandrea.com/articles/riverpod-initialize-listener-app-startup/#conclusion), this PR replaces that `currentContext!.read`-mechanism with a `ProviderContainer` we initialize globally and then hand to the app-widget at startup. This allows us to have a global provider we can `ref.read` with that can be used from anywhere and is independent of any specific context or whether the app is visible at all.

### In Details

 - https://github.com/acterglobal/a3/pull/2331/commits/f8a9844f99bc022512fce6e960565260e9343ea6 switches from rootNavKey to ProviderContainer
 - https://github.com/acterglobal/a3/pull/2331/commits/05c046b862cde9c543452aa38d7e8c015a8b92da is an additional commit to add sentry logging for our notification exceptions, so we get more insights about those going forward.

### DIFF
--- a/app/integration_test/support/util.dart
+++ b/app/integration_test/support/util.dart
@@ -1,14 +1,13 @@
 import 'dart:io';
 
-import 'package:acter/common/extensions/acter_build_context.dart';
 import 'package:acter/common/utils/constants.dart';
 import 'package:acter/common/widgets/spaces/select_space_form_field.dart';
+import 'package:acter/config/setup.dart';
 import 'package:acter/features/home/data/keys.dart';
 import 'package:acter/features/labs/model/labs_features.dart';
 import 'package:acter/features/labs/providers/labs_providers.dart';
 import 'package:acter/features/search/model/keys.dart';
 import 'package:acter/features/settings/widgets/settings_menu.dart';
-import 'package:acter/router/router.dart';
 import 'package:convenient_test_dev/convenient_test_dev.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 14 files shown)


gtalha07 (approved):
Looks good to me. Good approach from moving `!` operator 👏🏻 .

PR #2332, opened 2024-10-29
[Chat-ng]: Appflowy Editor editor integration

gtalha07:
The first set iterative change is to integrate Appflowy Editor integration (with built-in set of mention components block) as a chat-ng input component which will support markdown formatting. This'll also fix the bugs we've seen with input format which we have to accustom with various handlers for ensuring that proper markdown is passed and that won't be needed since the editor would emit correct format and ensure the renderer catches it correctly and display.

As this section of the codebase is pretty complex and time-consuming to follow, the input component isn't entirely migrated. As it accounts numerous issues being appearing when trying to do so. So, having it started from the ground is the best course to take but that also means it's limited in its functionality compared to current state of ```CustomInput```. 

This only contains functional wise:
 - Simple send message state .i.e including trailing rendering of showing either attachment/send button depending on ```editorState``` current state.
 - Emoji handlers with ```editorState```.
 - Attachment handlers.
 - Chat Compose draft state. (Not yet with reply/edit builder)
 
 But also improves UI of the input:
  - Dynamic height adjustment of input field.
  - Proper markdown support for input.
  - Mentions built-in extension for Appflowy. See [html_editor](https://github.com/acterglobal/a3/pull/2332/files) directory for new mentions block support.

## Desktop Demo with Mentions UI:

https://github.com/user-attachments/assets/b75cbe0f-2734-46da-8553-9718adced7fe

## Mobile UI:

https://github.com/user-attachments/assets/d47da144-31a6-4eb1-8508-61ad75ab767d

### DIFF
--- a/app/test/features/chat/custom_input_test.dart
+++ b/app/test/features/chat/custom_input_test.dart
@@ -47,7 +47,8 @@ void main() {
               overrides: [
                 // Same as before
                 canSendMessageProvider.overrideWith(
-                    (ref, roomId) => false), // false means loading
+                  (ref, roomId) => false,
+                ), // false means loading
                 sdkProvider.overrideWith((ref) => MockActerSdk()),
               ],
               child: const InActerContextTestWrapper(


gnunicorn:
@gtalha07 the video doesn't load here. secondly, I tried to start reviewing but my focus is a bit scattered this week and I couldn't get through... so you might want to ping someone else to do it if you want this in earlier...

kumarpalsinh25:
> @gtalha07 the video doesn't load here. secondly, I tried to start reviewing but my focus is a bit scattered this week and I couldn't get through... so you might want to ping someone else to do it if you want this in earlier...

Let me do a review although I don't understand much details about `Applyflowy Editor` and `Mentioned` related work.

kumarpalsinh25 (commented):
I will basic review of this PR. However I don't much understand mentioned and editor related work.

Also can I have video reference for Room/User mention related flow?

app/lib/common/widgets/html_editor/components/room_mention_block.dart
```diff
@@ -0,0 +1,70 @@
+import 'package:acter/common/widgets/html_editor/components/mention_content.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+
+class RoomMentionBlock extends StatefulWidget {
```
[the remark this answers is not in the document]
gtalha07: You're right. This could be reduced on inner layer. 👍🏻

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -293,39 +306,52 @@ class HtmlEditorState extends State<HtmlEditor> {
...
-          shrinkWrap: widget.shrinkWrap,
-          autoFocus: widget.autoFocus,
-          header: widget.header,
-          // local states
-          editorState: editorState,
-          editorScrollController: editorScrollController,
-          editorStyle: mobileEditorStyle(),
-          footer: generateFooter(),
-        ),
+      child: Column(
```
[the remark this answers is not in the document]
gtalha07: It's because of the internal layout of Appflowy widget which is using a lot of scrollable widgets. And making it expandable dynamically is really a complex thing. This pattern is something which I took from the package source code itself (from their example code) and it seems like wrapping it in column solves layout issue.

app/lib/features/chat/widgets/custom_input.dart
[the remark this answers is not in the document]
gtalha07: Yes, I still want to keep it till we migrate it fully to Appflowy. Without ng-flag, it still uses `CustomInput`

app/lib/features/chat_ng/widgets/chat_input.dart
```diff
@@ -0,0 +1,617 @@
...
+import 'package:acter/common/extensions/options.dart';
+
+final _log = Logger('a3::chat::custom_input');
+
+final _allowEdit = StateProvider.family.autoDispose<bool, String>(
+  (ref, roomId) => ref.watch(
+    chatInputProvider
+        .select((state) => state.sendingState == SendingState.preparing),
+  ),
+);
```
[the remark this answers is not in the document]
gtalha07: I guess this could become a common use case and now we have two input widgets using similar provider. I agree that this should be global.

app/lib/features/chat_ng/widgets/chat_input.dart
```diff
@@ -0,0 +1,617 @@
...
+    );
+  }
+
+  Widget _noAccessWidget(BuildContext context) {
+    return FrostEffect(
+      child: Container(
+        padding: const EdgeInsets.symmetric(vertical: 15),
+        decoration: BoxDecoration(color: Theme.of(context).colorScheme.surface),
+        child: Padding(
+          padding: const EdgeInsets.symmetric(horizontal: 10),
```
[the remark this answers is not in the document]
gtalha07: This is copy-pasta  from `CustomInput`, but agreed.

app/lib/features/chat_ng/widgets/chat_input.dart
```diff
@@ -0,0 +1,617 @@
...
+        roomId: roomId,
+        onTyping: onTyping,
+      );
+    }
+    // no permissions to send messages
+    return _noAccessWidget(context);
+  }
+}
+
+class _InputWidget extends ConsumerStatefulWidget {
```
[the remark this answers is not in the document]
gtalha07: I don't really care about developer preferences of choosing way of styling code, as long as it's easy for readability and efficient. So I'm flexible of choosing another approach. So, I can make it more streamlined and consistent.

app/lib/features/chat_ng/widgets/chat_input.dart
```diff
@@ -0,0 +1,617 @@
...
+
+  // handler for expanding editor field height
+  void _updateHeight() {
+    final text = textEditorState.intoMarkdown();
+    final lineCount = '\n'.allMatches(text).length;
+
+    // Calculate new height based on line count
+    // Start with 5% and increase by 4% per line up to 20%
+    setState(() {
+      _cHeight = (0.05 + (lineCount - 1) * 0.04).clamp(0.05, 0.15);
```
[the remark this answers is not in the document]
gtalha07: This is adjusting the height of the editor based on newline input. The values I took are just for the sake of improving UX and aligned with display of cursor and container UI. The container expands upto exceed limit when writing multi-line message after it becomes scrollable input.

gtalha07:
> I will basic review of this PR. However I don't much understand mentioned and editor related work.
> 
> Also can I have video reference for Room/User mention related flow?

Room/User mentions are disabled for now as it seems to need some internal fixes with riverpod and `editorState` to work reliably. I can understand my description made it sound the other way so apologies for that. The subsequent PR will follow more on that details ;)

app/lib/features/chat_ng/widgets/chat_input.dart
```diff
@@ -0,0 +1,617 @@
...
+  static const loadingKey = Key('chat-ng-loading');
+  static const noAccessKey = Key('chat-ng-no-access');
+
+  final String roomId;
+  final void Function(bool)? onTyping;
+
+  const ChatInput({super.key, required this.roomId, this.onTyping});
+
+  Widget _loadingWidget(BuildContext context) {
+    return Skeletonizer(
```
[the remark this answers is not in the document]
gtalha07: The loading state of Skeletonizer actually shows the preview of the actual widget UI, so decoration is needed otherwise it'd go out of match in UI transition. Although we don't need expensive widgets here like ```AnimatedContainer``` etc.

app/lib/common/widgets/html_editor/components/mention_handler.dart
```diff
@@ -0,0 +1,364 @@
...
+    _scrollController.dispose();
+    _focusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Focus(
+      focusNode: _focusNode,
+      onKeyEvent: (node, event) => _handleKeyEvent(node, event, filteredItems),
```
kumarpalsinh25, line 80: Why we are passing `filteredItems` as it is already global variable in class itself?

app/lib/common/widgets/html_editor/components/mention_handler.dart
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    final delta = widget.editorState.getNodeAtPath(selection.start.path)?.delta;
+    if (delta == null) {
+      return false;
+    }
+
+    return delta.isNotEmpty;
+  }
+}
```
kumarpalsinh25, line 313: Ignoring code review of this part as I am not quite understanding logic here.

app/lib/common/widgets/html_editor/components/mention_menu.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  void dismiss() {
+    if (_menuEntry != null) {
+      editorState.service.keyboardService?.enable();
+      editorState.service.scrollService?.enable();
+      keepEditorFocusNotifier.decrease();
+    }
+
+    _menuEntry?.remove();
+    _menuEntry = null;
```
kumarpalsinh25, line 32: Why to put below code outside of null condition check ?
```
   _menuEntry?.remove();
    _menuEntry = null;
```

app/lib/common/widgets/html_editor/components/mention_menu.dart
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  void _onSelectionUpdate() => selectionChangedByMenu = true;
```
kumarpalsinh25, line 35: Why exactly `selectionChangedByMenu` variable will do?
I don't see usage of this here apart from setting it's value from `false` to `true`.

kumarpalsinh25 (changes requested):
It is quite difficulty to judge mentioned related code review without it's visual reference. More over seeing issues in basic code standards during review.

And I see that after first round of PR review, there are additional thing gets added to this PR with below commit which makes PR review process quite difficulty and need to review whole things again carefully to track entire changes.
https://github.com/acterglobal/a3/pull/2332/commits/3c5a6af4d3179d8d2a0fc6e44ddef116b1c22e27

I am not feeling confident to approve this PR, I would like to have once review of @gnunicorn whenever he gets some time.

app/lib/common/widgets/html_editor/components/mention_block.dart
```diff
@@ -0,0 +1,112 @@
...
+        return const SizedBox.shrink();
+    }
+  }
+
+  Widget _mentionContent({
+    required BuildContext context,
+    required EditorState editorState,
+    required String mentionId,
+    String? blockId,
+    required String? displayName,
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_content.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+
+class MentionContentWidget extends StatelessWidget {
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_block.dart
```diff
@@ -0,0 +1,112 @@
...
+    TextStyle? textStyle,
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_content.dart
```diff
@@ -0,0 +1,50 @@
...
+  final EditorState editorState;
+  final Node node;
+  final int index;
+
+  @override
+  Widget build(BuildContext context) {
+    final baseTextStyle = textStyle?.copyWith(
+      color: Theme.of(context).colorScheme.primary,
+      leadingDistribution: TextLeadingDistribution.even,
+    );
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_content.dart
```diff
@@ -0,0 +1,50 @@
...
+
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        if (displayName != null)
+          Text(
+            displayName!,
+            style: baseTextStyle?.copyWith(
+              fontWeight: FontWeight.w500,
+            ),
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_handler.dart
```diff
@@ -0,0 +1,364 @@
...
+
+          return normalizedId.contains(normalizedQuery) ||
+              normalizedName.contains(normalizedQuery);
+        }).toList();
+      }
+    });
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      _focusNode.requestFocus();
+    });
+    startOffset = widget.editorState.selection?.endIndex ?? 0;
```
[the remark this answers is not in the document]
gtalha07: Fixed [def4f25](https://github.com/acterglobal/a3/pull/2332/commits/def4f259431dccdbf9da6e466b0ee2df1437deec)

app/lib/common/widgets/html_editor/components/mention_block.dart
```diff
@@ -0,0 +1,92 @@
+import 'package:acter/common/providers/room_providers.dart';
+import 'package:acter/common/widgets/html_editor/models/mention_block_keys.dart';
+import 'package:acter_avatar/acter_avatar.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class MentionBlock extends ConsumerWidget {
```
gtalha07, line 8: The UI component of the inline mentions in editor (of how it should be displayed when  triggered)

app/lib/common/widgets/html_editor/components/mention_item.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:acter/common/utils/constants.dart';
+import 'package:acter/common/widgets/html_editor/models/mention_type.dart';
+import 'package:acter_avatar/acter_avatar.dart';
+import 'package:flutter/material.dart';
+
+class MentionItem extends StatelessWidget {
```
gtalha07, line 6: Mention list UI item component which gets displayed as part of mention menu list.

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,289 @@
...
+                !HardwareKeyboard.instance.isMetaPressed ||
+            !HardwareKeyboard.instance.isShiftPressed) {
+          widget.editorState.insertTextAtCurrentSelection(event.character!);
+          return KeyEventResult.handled;
+        }
+        return KeyEventResult.ignored;
+    }
+  }
+
+  void _selectItem(String id, String displayName) {
```
gtalha07, line 209: handler when user selects mention item from list and processes how the editor should execute mention block inline

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,289 @@
...
+        itemPosition +
+            kItemHeight -
+            _scrollController.position.viewportDimension,
+        duration: const Duration(milliseconds: 200),
+        curve: Curves.easeOut,
+      );
+    }
+  }
+
+  bool _canDeleteLastCharacter() {
```
gtalha07, line 277: checks whether editor state selection is at mention block.

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -357,6 +388,56 @@ class HtmlEditorState extends State<HtmlEditor> {
                 .bodySmall
                 .expect('bodySmall style not available'),
           ),
+      mobileDragHandleBallSize: const Size(12, 12),
+      textSpanDecorator:
+          widget.roomId != null ? customizeAttributeDecorator : null,
+    );
+  }
+
+  InlineSpan customizeAttributeDecorator(
```
gtalha07, line 397: For customizing the styling of custom component block, here we can extend it beyond mentions (for future reference).

app/lib/common/widgets/html_editor/services/mention_shortcuts.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+
+List<CharacterShortcutEvent> mentionShortcuts(
+  BuildContext context,
+  String roomId,
+) {
+  return [
+    CharacterShortcutEvent(
+      character: '@',
```
gnunicorn, line 12: let's use a global static for these!?!

app/lib/common/widgets/html_editor/components/mention_item.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final isDesktop = desktopPlatforms.contains(Theme.of(context).platform);
+
+    return Container(
+      height: 60,
+      // selection color is only for desktop with keyboard navigation
+      color: (isSelected && isDesktop)
+          ? Theme.of(context).colorScheme.primary
+          : null,
```
gnunicorn, line 34: `isDesktop` is an incorrect check here: pads can also have keyboards and laptops might also have touch devices. what _I think_ you want to detect for here is a hardware keyboard, no?

but then also, how would `isSelected` even come about if not through a hardware-keyboard? I think this check is pointless and should be removed and the rest left to the outer part.


```suggestion
    return Container(
      height: 60,
      color: isSelected ? Theme.of(context).colorScheme.primary : null,
```

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,289 @@
...
+      attributes: {
+        MentionBlockKeys.mention: {
+          MentionBlockKeys.type: widget.mentionType.name,
+          if (widget.mentionType == MentionType.user)
+            MentionBlockKeys.userId: id
+          else
+            MentionBlockKeys.roomId: id,
+          MentionBlockKeys.displayName: displayName,
+        },
+      },
```
gnunicorn, line 248: so ... `attributes`  Attributes = [Map](https://api.flutter.dev/flutter/dart-core/Map-class.html)<[String](https://api.flutter.dev/flutter/dart-core/String-class.html), `dynamic`>, right? so, why not have the "mention"-string-key point to a _typed_ object. all those `MentionBlockKeys`-with strings make it really hard to follow and easy to mess up. I'd rather have a typed:

```suggestion
      attributes: {
        MENTIONS_KEY: widget.mentionType == MentionType.user ? 
            MentionAttributes.user(userId: id, displayName: displayName)
            MentionAttributes.room(roomId: id, displayName: displayName)
      },
```

For a type `MentionAttributes`, along the lines of (pseudocode):

```
class MentionAttributes {
  final String? userId;
  final String? roomId;
  final String? displayName;
  final MentionType type;
  
  const MentionAttributes({this.userId, this.roomId, required this.type, this.displayName});

  factory MentionAtrributes.user(required this.userId, {this.displayName}) {
     type = MentionType.user;
  }
}
```

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,289 @@
...
+      ],
+    );
+    if (isDesktop) {
+      // we only want to listen keyboard events on desktop
+      return Focus(
+        focusNode: _focusNode,
+        onKeyEvent: (node, event) => _handleKeyEvent(node, event, suggestions),
+        child: menuWidget,
+      );
+    }
```
gnunicorn, line 79: We can disregard this flaky desktop check (iPads and Mobiles can have hardware keyboards attached to it) by directly interacting with [`HardwareKeyboardDetector`](https://pub.dev/documentation/flame/latest/events/HardwareKeyboardDetector-class.html). Which we should be doing here rather than doing some flaky check and focus grabbing ...

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,289 @@
...
+  void dispose() {
+    widget.onDismiss();
+    _scrollController.dispose();
+    _focusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final isDesktop = desktopPlatforms.contains(Theme.of(context).platform);
```
gnunicorn, line 56: reminder: this is almost always bad. checking for the specific feature you want is probably the better idea. In this case: check for/directly use the actual hardware keyboard..

app/lib/common/widgets/html_editor/components/mention_block.dart
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  Widget _mentionContent({
+    required BuildContext context,
+    required String mentionId,
+    required String displayName,
+    required WidgetRef ref,
+    required AvatarOptions avatarOptions,
+  }) {
+    final desktopPlatforms = [
```
gnunicorn, line 53: mouse can also be plugged/connected to phones. don't do a check like that.

gnunicorn (commented):
I have not done a full review and I don't pretend that I have understood this whole, but I'd like to see two major suggestions applied before going forward, as it will make it easier to understand what is even meant to be going on:

1. using typed content for the Attributes and everywhere really, rather than a bunch of keyed-strings of keys for a `Map`... that isn't necessary and makes it harder for the reader and the computer to confirm this is correct
2. remove the `isDesktop`-checks. They are almost always wrong and they are for sure here ... if we have a hardware keyboard or a mouse isn't platform specific and pretending it is makes the code more complex than just dealing with that being the case...

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -15,7 +15,7 @@ class ChatMessages extends ConsumerWidget {
 
     return AnimatedList(
       initialItemCount: messages.length,
-      reverse: true,
+      reverse: false,
```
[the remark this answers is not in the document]
gtalha07: Yes, but it's one liner change for showing messages in correct directionality (latest down) so I thought, could've been fixed here...

app/lib/common/widgets/html_editor/components/mention_block.dart
```diff
@@ -0,0 +1,92 @@
...
+    required this.node,
+    required this.index,
+    required this.mention,
+    required this.userRoomId,
+  });
+
+  final Map<String, dynamic> mention;
+  final String userRoomId;
+  final Node node;
+  final int index;
```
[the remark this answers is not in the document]
gtalha07: Hmm, yes. I can change it as I didn't really seem to thought much about it before.

app/lib/common/widgets/html_editor/components/mention_item.dart
gtalha07: I see, good point. I'll change the conditional 👍🏻 !

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -49,3 +53,39 @@ final renderableChatMessagesProvider =
...
   }).toList();
 });
+
+final mentionSuggestionsProvider =
+    StateProvider.family<Map<String, String>?, (String, MentionType)>(
+        (ref, params) {
+  final roomId = params.$1;
+  final mentionType = params.$2;
+  final client = ref.watch(alwaysClientProvider);
+  final userId = client.userId().toString();
```
gnunicorn, line 63: we have a provider for that:
```suggestion
  final userId = ref.watch(ref.watch(myUserIdStrProvider));
```

app/lib/features/chat_ng/widgets/chat_input/chat_editor.dart
```diff
@@ -0,0 +1,306 @@
...
+        // adjust bottom viewport so toolbar doesn't obscure field when visible
+        if (isKeyboardVisible != null && isKeyboardVisible)
+          SizedBox(height: viewInsets + 50),
+      ],
+    );
+  }
+
+  // chat editor UI
+  Widget renderEditorUI(bool emojiPickerVisible) {
+    return FrostEffect(
```
gnunicorn, line 157: what is this, btw? We have that a few times but I never really understood what it is for and why have it? There seems be zero difference in most cases if it just wasn't there... or maybe my :eyes: aren't good enough to catch the difference?

app/lib/features/chat_ng/widgets/chat_input/chat_input.dart
```diff
@@ -0,0 +1,36 @@
...
+      case true:
+        // we have permission, show editor field
+        return ChatEditor(
+          roomId: roomId,
+          onTyping: onTyping,
+        );
+      case false:
+        // no permissions to send messages
+        return const ChatEditorNoAccess();
+    }
```
gnunicorn, line 34: screams for the new switch statement!
```suggestion
    return switch (canSend) {
      true => ChatEditor( // we have permission, show editor field
          roomId: roomId,
          onTyping: onTyping,
        ),
      false => const ChatEditorNoAccess(), // no permissions to send messages
      null => const ChatEditorLoading(), // we're still loading
    };
```

app/lib/common/widgets/html_editor/components/mention_list.dart
```diff
@@ -0,0 +1,281 @@
...
+import 'package:flutter/services.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class MentionList extends ConsumerStatefulWidget {
+  const MentionList({
+    super.key,
+    required this.editorState,
+    required this.roomId,
+    required this.mentionType,
```
gnunicorn, line 18: this widget still feels a bit odd to me. In general, when we have widgets that are generalized but are then supposed to be doing slightly different things depending on _one_ flag passed in (like in this case, pass the mentions type to a different provider, that then just switches again on the same type), it seems like there is a miss for generalization and specialization here.

This isn't necessarily something we have to do before the merge, but it should be done either way.

Let's imagine a `GenericMentionList` of:
```dart
class GenericMentionList extends ConsumerStatefulWidget {

   final EditorState editorState;
   final VoidCallback onDismiss;

   // this is the incoming provider
   final ProviderBase<Map<String, String>> mentionsProvider;

   // these are the missing widgets
   final String headerTitle;
   final String notFoundTitle;

   const GenericMentionList({
      super.key,
      required this.editorState,
      required this.onDismiss
      required this.mentionsProvider,
      required this.headerTitle,
      required this.notFoundTitle,
});
 // ...
}
```

And then we'd have two Widget, each with its specifics:

```dart
class UserMentionList extends ConsumerWidget {
   final EditorState editorState;
   final VoidCallback onDismiss;
   final String roomId;

   const GenericMentionList({
      super.key,
      required this.editorState,
      required this.onDismiss,
      required this.roomId,
});
   @override
   Widget build(BuildContext context, WidgetRef ref) => GenericMentionList(
      editorState: editorState,
      onDismiss: onDismiss,
      // the actual provider
      mentionsProvider: userMentiothat is decided between, where that `MentionType` first comes into nProvider(roomId),
      // the fields
      headerTitle: L10n.of(context).users,
      notFoundTitle: l10n.of(context).noUsersFoundTitle,
    );
}


class RoomMentionList extends ConsumerWidget {
   final EditorState editorState;
   final VoidCallback onDismiss;
   final String roomId;

   const GenericMentionList({
      super.key,
      required this.editorState,
      required this.onDismiss,
      required this.roomId,
});
   @override
   Widget build(BuildContext context, WidgetRef ref) => GenericMentionList(
      editorState: editorState,
      onDismiss: onDismiss,
      // the actual provider
      mentionsProvider: roomMentionProvider(roomId),
      // the fields
      headerTitle: L10n.of(context).chats,
      notFoundTitle: l10n.of(context).noChatsFoundTitle,
   );
}
````


And then, whatever first encounters that `MentionType` would just pick between these two higher level widgets and render that. Where each one has its specific neat and clean stuff together and the inner one doesn't have these nasty switch statement everyone. Also, we get rid of one more provider that has a switch, too...

Comment (gnunicorn) [reply app/lib/common/widgets/html_editor/components/mention_list.dart:18]: after writing it I realised, these don't even need to be consumers, just stateless widgets is enough because we need the L10n.of(Context) there...

File: app/lib/common/widgets/html_editor/components/mention_list.dart [hunk 35]
```diff
@@ -0,0 +1,281 @@
...
+                    ? ref.watch(
+                        memberAvatarInfoProvider(
+                          (roomId: widget.roomId, userId: mentionId),
+                        ),
+                      )
+                    : ref.watch(roomAvatarInfoProvider(mentionId));
+                final options = widget.mentionType == MentionType.user
+                    ? AvatarOptions.DM(avatar, size: 18)
+                    : AvatarOptions(avatar, size: 28);
+
```
gnunicorn, line 115: inline with the switch-removing from above I'd make argue that the generic mention list should have an `AvatarOptions Function(String matchId, WidgetRef ref) avatarBuilder` given as a parameter ... this is pretty easy to provide from the outer thing as well...

Should still be using its own Consumer here though

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -272,10 +280,17 @@ class HtmlEditorState extends State<HtmlEditor> {
...
-          editorScrollController: editorScrollController,
+          editorScrollController:
+              widget.scrollController ?? editorScrollController,
           editorState: editorState,
           editorStyle: desktopEditorStyle(),
           footer: generateFooter(),
+          characterShortcutEvents: [
+            ...standardCharacterShortcutEvents,
+            if (widget.roomId != null)
+              ...mentionShortcuts(context, widget.roomId!),
```
gnunicorn, line 291: we are trying to avoid `!` at most cost... so can we maybe have a local variable `roomId = widget.roomId` and then just do:

```suggestion
            if (roomId != null)
              ...mentionShortcuts(context, widget.roomId),
```

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -293,39 +308,52 @@ class HtmlEditorState extends State<HtmlEditor> {
...
+                // local states
+                editorState: editorState,
+                editorScrollController:
+                    widget.scrollController ?? editorScrollController,
+                editorStyle: mobileEditorStyle(),
+                footer: generateFooter(),
+                characterShortcutEvents: [
+                  ...standardCharacterShortcutEvents,
+                  if (widget.roomId != null)
+                    ...mentionShortcuts(context, widget.roomId!),
```
gnunicorn, line 351: same as above.

app/lib/common/widgets/html_editor/services/mention_shortcuts.dart
```diff
@@ -0,0 +1,64 @@
...
+  BuildContext context,
+  String roomId,
+) {
+  return [
+    CharacterShortcutEvent(
+      character: userMentionChar,
+      handler: (editorState) => _handleMentionTrigger(
+        context: context,
+        editorState: editorState,
+        type: MentionType.user,
```
gnunicorn, line 16: like _here_ is where we'd already decided upon which of the inner listing widgets to use... no need for `MentionType` at all I think ...

gnunicorn (approved):
I have left quite a few comments as I think this can still be structurally be improved a lot. But we can also do that as a next generation and merge this (aside from some smaller things) as-is...

app/lib/common/widgets/html_editor/components/mention_item.dart
```diff
@@ -0,0 +1,49 @@
...
+      color: (isSelected && hasKeyboard)
+          ? Theme.of(context).colorScheme.primary
+          : null,
+      child: ListTile(
+        dense: true,
+        onTap: onTap,
+        contentPadding: const EdgeInsets.symmetric(horizontal: 12),
+        leading: ActerAvatar(options: avatarOptions),
+        title: Text(
+          displayName.isNotEmpty ? displayName : mentionId,
```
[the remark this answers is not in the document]
gtalha07: The suggestion only accounts for ```displayName``` being null but it can never be that as it'll either one of empty or some string. But I discovered it can be nicely put in ```.map()``` extension (if its made nullable which is realistic case) for nullables and it looks much better for ```subtitle```:
```suggestion
          subtitle: displayName.map(
          (name) => Text(name, style: Theme.of(context).textTheme.labelMedium),
          orElse: () => null,
        ),
```

app/lib/features/chat_ng/actions/attachment_upload_action.dart
```diff
@@ -0,0 +1,87 @@
+import 'dart:io';
```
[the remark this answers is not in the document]
gtalha07: Yes this is just moved as it is to ```actions``` folder for code legibility.

app/lib/features/chat_ng/actions/send_message_action.dart
```diff
@@ -0,0 +1,88 @@
+import 'package:acter/common/providers/chat_providers.dart';
```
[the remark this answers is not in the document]
gtalha07: Yep :)

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -49,3 +53,39 @@ final renderableChatMessagesProvider =
...
+    case MentionType.room:
+      final rooms = ref.watch(chatIdsProvider);
+      return rooms.fold<Map<String, String>>({}, (map, roomId) {
+        final displayName = ref.watch(roomDisplayNameProvider(roomId));
+        map[roomId] = displayName.valueOrNull ?? '';
+        return map;
+      });
+  }
+  return null;
+});
```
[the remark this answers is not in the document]
gtalha07: I was a bit hesitant to write too many providers while it can be combined just in one. But that's just my personal preference. If it improves understanding, then I'm happy to split :)

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
gtalha07: Oops, yes. Hard to track too many providers... ^^!

app/lib/features/chat_ng/widgets/chat_input/chat_editor.dart
gtalha07: This is UI wrapper to provide blurry/frost effect to inner child container, I remember this was added to reflect designs of Chat UI.

app/lib/common/widgets/html_editor/components/mention_list.dart
gtalha07: Yeah I can see how this change will improve code cleaning and too many widgets wrapping. Nice thinking!

app/lib/common/widgets/html_editor/components/mention_item.dart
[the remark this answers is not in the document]
gnunicorn: but map also only works on null not on `isEmpty` so this won't work either.

but you could make an `extension` on `String` to do `isEmptyMap` and `isNotEmptyMap`, which are a bit more handy than the inline if ...

PR #2338, opened 2024-10-31
Create generic event list widget and also manage it with Search Page

kumarpalsinh25:
### NOTE: 
REVIEW PR BY COMMITS WHICH EXPLAIN AT THE END OF THE PR DESCRIPTION FOR BETTER UNDERSTANDING.

- - -

Introducing all new generic `EventListWidget` to Acter widget library set which makes event list behaviour and UI-UX consistent across entire Acter App.

### Things to know about `EventListWidget`:
- Optional param name `spaceId` makes this widget to re-used for showing **All Events** as well as **Space Specific Events**.
- Optional param name `searchValue` enables this widget to show events list based on the given search criteria.
- Optional param name `limit` allows you to show only specific event items in list. 
- You can also show Event Section Header by passing `showSectionHeader` param to `true`.
- You can also manage scrolling behaviour of the event list by using `shrinkWrap` param

### Implemented `EventListWidget` in below screens.
- All events list (When we click on Event Feature Button from Dashboard)
- Space Overview Event Section
- All Space specific events list (When we click on `see all` button from event sections of space overview)
- Quick search screen - Default Events Section when search value is empty
- Quick search screen - Event list with given search value criteria

### Implemented `Quick Search Filters` in search page.
- Now you also do filter of search result by different quick filters types (All, Spaces, Pins, Events)

### Reference Video:

https://github.com/user-attachments/assets/047a58ef-bc8e-43bb-bb31-2b6e8cfaaff2

- - -

### PR Commits : Feature Work:
- Remove eventType param from `EventItem` widget as it can be manager by relevant provider value in widget itself.
https://github.com/acterglobal/a3/pull/2338/commits/27e322079f9103ad4046bf236c72e9edbc4748d3

- Create generic `EventListWidget` and use it in `EventListPage`.
https://github.com/acterglobal/a3/pull/2338/commits/94a113d14e1e810f64d2ebe70181b8621123f91f

- Use generic `EventListWidget` in `SpaceDetailsPage` as Event section.
https://github.com/acterglobal/a3/pull/2338/commits/3ea914fadf138f5157e865b85457003810fc3cbf

- Use generic `EventListWidget` in `SearchPage` as Event section.
https://github.com/acterglobal/a3/pull/2338/commits/f68f58e326475957e00efc86dc85126c969d384a

- Remove EventFilter provider and manage it with local state to avoid data conflicts in other screens.
https://github.com/acterglobal/a3/pull/2338/commits/b75ac7c98bd521ce19f093b3d1ce4133e26706a3

- Add `QuickFilters` option in `SearchPage`.
https://github.com/acterglobal/a3/pull/2338/commits/9fe40e5804b8114cd6d2580dd290bea878b74863

- - -

### PR Commits : Widget Testing Work:

- `ActerSearchWidget` testing
https://github.com/acterglobal/a3/pull/2338/commits/f6f57be78f229c84ce952bd02c220e4a03528a4c

- `EventItem` widget testing
https://github.com/acterglobal/a3/pull/2338/commits/988aa3ef2faf0e4d51bab46b1648a7193eb29e16

-  `Rsvp` test in `EventItem` widget testing
https://github.com/acterglobal/a3/pull/2338/commits/d3b8edc2280b4bd14b738ef5ceb89fa04da766fa

- `EventList` widget testing
https://github.com/acterglobal/a3/pull/2338/commits/10d1f8fdd1071776fcd176a971c759beb9cc0189


app/lib/features/events/pages/event_list_page.dart
```diff
@@ -74,42 +76,37 @@ class _EventListPageState extends ConsumerState<EventListPage> {
...
-              );
-            },
-            loading: () => const EventListSkeleton(),
+          child: ValueListenableBuilder(
+            valueListenable: eventFilter,
+            builder: (context, showHeader, child) => EventListWidget(
+              spaceId: widget.spaceId,
+              shrinkWrap: false,
+              searchValue: searchValue,
+              eventFiler: eventFilter.value,
```
gnunicorn, line 104: This isn't really the proper riverpod pattern. Passing around a search-value and filter-params and have local valuelisteners instead of just having the proper riverpod state providers and `ref.watch` on them to bubble the information _up_ not _down_ and have all reactively refresh _from_ the state to the tree.

app/lib/features/events/providers/event_providers.dart
```diff
@@ -131,21 +131,20 @@ enum EventFilters {
...
     .family<List<ffi.CalendarEvent>, EventListSearchParams>(
         (ref, params) async {
   //Declare filtered event list
   List<ffi.CalendarEvent> filteredEventList = [];
 
-  //Filter events based on the selection
-  EventFilters eventFilter = ref.watch(eventFilterProvider);
-  switch (eventFilter) {
+  switch (params.eventFilter) {
     case EventFilters.bookmarked:
```
gnunicorn, line 148: Why did you remove the provider (the proper pattern) and replace it with _several_ input parameters? The way local state is meant to be used is exactly the pattern that was existing before: a provider with a state, that is being updated and then all that is using that provider will update from the bottom up. Family is not meant to pass in local state like a search text or other volatile configuration.

gnunicorn: This is partly my bad, I didn't notice before that we did this change for the search text, but this going the wrong direction now.

gnunicorn (changes requested):
Great to see tests! Lovely!

And now they can show their value, because I don't like the path this is taking regarding providers. "searchtext" and "filtervalue" are really not meant to be family input parameters, but _local_ state provider that something might be manipulating. Replacing those with values we pass down in the rendering tree (again) goes exactly against the concept(s) of provider and separating state in providers. The idea of "everything flows from the providers, which hold the state" is broken by this idea of passing in local state into the family parameters.  Granted, this has already crept in before and I didn't quite notice in a previous review, but we should really not go down this road any further but instead move all that to proper riverpod providers.

And that's where the tests can now be very helpful, because if they are comprehensive enough, this rather big refactor/change would just have the same tests still pass after. Proving that it works correctly regardless of the internal state management solution!

gnunicorn (approved):
replaced the problematic system

PR #2339, opened 2024-10-31
Quick Fixes : Comments

kumarpalsinh25:


app/lib/features/comments/providers/comments_providers.dart
```diff
@@ -42,7 +42,7 @@ final commentsListProvider = FutureProvider.family
     .autoDispose<List<Comment>, CommentsManager>((ref, manager) async {
   final commentList = (await manager.comments()).toList();
   commentList.sort(
-    (a, b) => a.originServerTs().compareTo(b.originServerTs()),
```
gnunicorn, line 45: feels kinda weird that you are reversing the order here ...

app/lib/features/comments/widgets/comment_list_widget.dart
```diff
@@ -42,6 +42,7 @@ class CommentListWidget extends ConsumerWidget {
       shrinkWrap: shrinkWrap,
       itemCount: commentList.length,
       padding: EdgeInsets.zero,
+      reverse: true,
```
gnunicorn, line 45: just to then reverse the order here ...

but I suspect this is due to the scroll-position?

gnunicorn (approved):
lgtm

gnunicorn:
@kumarpalsinh25 please add a changelog entry or the release building will fail when we want to make the release.

app/lib/features/comments/widgets/comment_list_widget.dart
kumarpalsinh25: I found this as better solution then having scroll controller and scroll to last.

Note:
ListView and List Provider are generic so don't think so there will be any issue with this approach.

Happy to change this later if found better solution.

PR #2343, opened 2024-11-01
Create Task as Bottom Sheet, Quick Action and Tests 

gnunicorn:
This replaces the current "inline task"-creation with the existing bottom drawer a29ce458216d91e8b524cc26a76b82705bed16dc . It trims down the bottom drawer to only include the extra fields after being asked to add them and only if they contain content - there are tests to proof it works as expected. Then the "edit" part of the widget is removed as it isn't used anymore anyways and the widgets are moved more appropriate locations. Additionally the widget now allows for no task list to be provided yet but select the space and the task list from within the widget - which required some additional bottom drawer and updates on the existing items.

Which, finally, allows us to replace the `Add Task List`-Quick-Action with the `Add Task` quick action!

Lookie lookie, nom, nom nom!

https://github.com/user-attachments/assets/83ca5a96-7853-465c-ace2-9d6c6efd1e60

Also on desktop. Showing that you can _always_ create a new list from the task list selector, too (if you have the rights):

https://github.com/user-attachments/assets/5a16b6ed-3e62-48c8-9eaa-daae17e4a30a


app/lib/features/tasks/actions/create_task.dart
```diff
@@ -0,0 +1,433 @@
...
+  Widget _widgetCreateButton() {
+    final lang = L10n.of(context);
+    return ElevatedButton(
+      key: CreateTaskWidget.submitBtn,
+      onPressed: addTask,
+      child: Text(lang.addTask),
+    );
+  }
+
+  Future<void> _selectTaskList() async {
```
kumarpalsinh25, line 360: How about putting this in actions folder?

app/lib/features/tasks/actions/create_task.dart
```diff
@@ -0,0 +1,433 @@
...
+      return;
+    }
+
+    final newTaskList = await ref.read(taskListItemProvider(taskListId).future);
+    setState(() {
+      taskList = newTaskList;
+    });
+    }
+
+  Future<void> addTask() async {
```
kumarpalsinh25, line 390: Same for this. May be we can put this also in actions folder.

app/lib/features/tasks/actions/select_tasklist.dart
```diff
@@ -0,0 +1,108 @@
...
+              if (tasklists.isEmpty) {
+                return TaskListsEmptyState(
+                  canAdd: canAdd,
+                  inSearch: false,
+                  spaceId: spaceId,
+                );
+              }
+
+              return SingleChildScrollView(
+                child: ListView.builder(
```
kumarpalsinh25, line 83: Why to overlap `ListView` with `SingleChildScrollView` ?

kumarpalsinh25 (approved):
Very nice UX changes. Looks good to me.

Added few suggestions and thought in which some of things need to answer/correct before merge.

gnunicorn:
Oh, I checked them off in the wrong order, so now all this is merge as-is. Will fix in a follow-up PR, sorry @kumarpalsinh25 !

app/lib/features/tasks/actions/select_tasklist.dart
gnunicorn: I copy-paste that from somewhere. It's not necessary?

app/lib/features/tasks/actions/create_task.dart
```diff
@@ -0,0 +1,433 @@
...
+  bool showDueDate = false;
+
+  TaskList? taskList;
+
+  @override
+  void initState() {
+    super.initState();
+    widget.taskList.map((tl) {
+      WidgetsBinding.instance.addPostFrameCallback((d) {
+        setState(() {
```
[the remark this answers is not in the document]
gnunicorn: > Why this requires `setState` here?

to put in the tasklist given by the widget as the locally selected one? And at initial setup make sure the selected space is the one from that list, too.

Same as below: if the space changed to another one, we have to reset the selected tasklist...

> How about managing taskList with ValueNotifier and ValueListenableBuilder ?.

I don't understand why or how that would improve anything... can you elaborate?

app/lib/features/tasks/actions/create_task.dart
```diff
@@ -0,0 +1,433 @@
...
+class _CreateTaskWidgetConsumerState extends ConsumerState<CreateTaskWidget> {
+  final GlobalKey<FormState> _formKey =
+      GlobalKey<FormState>(debugLabel: 'create task list form');
+  final TextEditingController _taskNameController = TextEditingController();
+  final TextEditingController _taskDescriptionController =
+      TextEditingController();
+  final TextEditingController _taskDueDateController = TextEditingController();
+  DateTime? selectedDate;
+
+  bool showDescriptionField = false;
```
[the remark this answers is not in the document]
gnunicorn: Why? How that improving anything over just using `setState`? This is a stateful widget anyways (due to the controller), and the only "benefit" I can see `ValueNotifier` "provide" is that they do not trigger if the identity didn't change. Which is literally irrelevant for primitives like `bool`s. What are you hoping to achieve by using `ValueNotifier`s that isn't already achieved?

app/lib/features/tasks/actions/create_task.dart
gnunicorn: We are already in the actions folder. I assume you mean make that a separate file in the actions folder!?!

kumarpalsinh25:
> Oh, I checked them off in the wrong order, so now all this is merge as-is. Will fix in a follow-up PR, sorry @kumarpalsinh25 !

No worrier, there were no major issues.

PR #2344, opened 2024-11-03
Tasks overall design improvements

gnunicorn:
On top of #2343 (so be sure to review that first), this improves the UI & UX of TasksList and Task Detail pages:

https://github.com/user-attachments/assets/0e17b577-7129-421c-b1e8-e87a24945182

- 79722f8385997b509bd218867bdfb8cf90bcd2f4 renames the `compactView` (which implies _compatibility_ to `useCompactView` and has minor clean ups around that in the spacechip
- 7785dc78e8febead5fceff8a6bda8410b5c655e8 Improves the TaskList Page
- 99ca5b348b4404b31e548f094a7b45d0779790e0 Improves the Task Page 
- f7e3ebd224e2dc199ef7ea6d42f3352eae88508d finally changes some `Row` with `Wrap` to ensure we don't overflow in our Task Details page

Issue #2227, opened 2024-10-06
Engaging Tasks


gnunicorn:
tasks need to be re-worked and become an engaging feature in the app:

- [ ] Be able to invite people to take over a task
- [ ] We need an actual quick action for tasks (with extra fields)
- [ ] redsign inline-adding (it doesn't work right now)
- [ ] easy overview of tasks and their assignment for larger organizations (?) https://github.com/acterglobal/a3/issues/2033

Further bugs that need to be addressed:

- [ ] https://github.com/acterglobal/a3/issues/1877
- [ ] https://github.com/acterglobal/a3/issues/2047
- [ ] https://github.com/acterglobal/a3/issues/1695
- [ ] https://github.com/acterglobal/a3/issues/1669
- [ ] https://github.com/acterglobal/a3/issues/2195

PR #2345, opened 2024-11-03
Bookmarking for TaskLists & list priotization tests

gnunicorn:
refs #2227 , on top of #2344 so review that one before ;) .

https://github.com/user-attachments/assets/e0196e7f-5efa-490e-8b01-5cec1ddfb2e0

This PR took a test-driven approach:
At first I added a new test on the bookmarks ensuring bookmarked pins are prioritized in the listing 0292f557a6db15e1f66153b5974ce408ee88998c . Once that finished and passed, I added a test to do the same for task lists 782f1eabf98290593c7a6a6520d1da0b402a618a , which, of course failed at that point as it was missing the implementation. As the code was a bit complex and ugly, I then refactored the priotization code to be reusable 4d2f4195907efd7f6c1d2cb0d7438c2ce0cb5ae5 and thanks to the first test was able to see if it still worked fine. While trying to implement the feature in the UI, I noticed another bug: when removing a bookmark the item disappeared from the list. So I added a test 803cd841ccdf682b0e05193e1a1afa87e43aee3a and then a fix 63fbda4985647553a41201d09445002589f5f775 for that as well, before finally committing the new task list side feature 63fbda4985647553a41201d09445002589f5f775 .


kumarpalsinh25 (approved):
Looks good to me.

kumarpalsinh25:
@gnunicorn 
Minor remark : Need to resolve unit test error before merge. :)

app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -40,7 +40,7 @@ final tabsProvider =
     }
 
     if (appSettings.pins().active()) {
-      final pinsList = await ref.watch(allPinListProvider(spaceId).future);
+      final pinsList = await ref.watch(pinsProvider(spaceId).future);
```
kumarpalsinh25, line 43: Why to use `pinsProvider` here which have bookmarks front sorting calculation. Here we need to check just pins list is empty of not. How about directly using `pinListProvider` or separate provider which given only list item counts or check for empty list only.

gnunicorn: Doesn't really make any difference...

I just replaced all uses of `allPinListProvider` with the `pinsProvider` in the merge conflict, that's why. But in practice this will not make a difference here.

kumarpalsinh25: Screenshot 2024-11-08 at 6 08 46 PM

As per above code changes, `pinsProvider` contains `priotizeBookmarked` logic which doesn't require here this code use-case.

Agree that it will not make major impact but why to use provider which have additional logic even though direct list provider is available.

PR #2346, opened 2024-11-03
Refactor referenced managers to remove unnecessary refreshes

gnunicorn:
As of now, we have been passing the comments- and attachment-managers within a future in the tree. This has several disadvantages:

1. when the parent (e.g. Pin, Event, etc) is updated, we are refreshing/replacing that item, as a result, we have to fetch a new manager despite no update there being necessary
2. the future itself is not strongly typed and created a fresh on every refresh of the UI, thus even if the parent object itself didn't change, if anything else caused a rerender of the parent widget (e.g. bookmarking the object), the managers would be newly created and the entire subsection going through a loading -> showing cycle
3. both of the above lead to a bunch of ugly flashes in the UI even if unrelated things are changed ...

This PR changes the behavior by providing a hard-typed wrapper that every parent has to implement. That wrapper is using the parents eventId to ensure their "hash"/identity is understood as not having changed (this is purely for the mentioned effect above) - as we use the ref to update the inner type anyways.

Thus it refactors the entire UI in two main parts: 1. use that new pattern, which even allows you to post in a null to get the loading state; 2. disconnect the loading of the specific subsections from the overall loader, allowing for them to figure out _independently_ whether to show any loading or not.

Notice how this is not flashing on neither the Attachments nor Comments Sections anymore when updates are done on the item itself:

https://github.com/user-attachments/assets/7ccf28ee-1cb7-4646-8f94-98f78cad9a6d

For the reviewer:
 - d8fb07273ed73f0fb901115ff193e997eb8c4861 contains a run-by-fix for using the inline-space-chip rather than the big ugly one
 - 049fd53dcc0ae100e7094de10216138e2675c91b outsources the participants list into a separate provider as the rsvp doesn't refresh the calendar itself anymore and thus the previous "update participants"-code was not working anymore now.


app/lib/features/events/pages/event_details_page.dart
```diff
@@ -62,52 +64,62 @@ class _EventDetailPageConsumerState extends ConsumerState<EventDetailPage> {
...
-        loading: () => const EventDetailsSkeleton(),
+      );
+    }
+    ref.listen(calendarEventProvider(widget.calendarId), (prev, next) async {
+      final ev = next.valueOrNull;
+      if (ev != null) {
+        final participants = asDartStringList(await ev.participants());
+        _log.info('Event Participants => $participants');
+        if (!mounted) return;
+        setState(() {
```
gtalha07, line 90: Are we sure this doesn't have side-implications of throwing errors as we're mutating widget state inside ```build()```? Have it wrapped with ```WidgetInstanceBindings.addPostFrameCallback((){})``` but this could be ```listenManual``` type within ```initState(){}``` ??

app/lib/features/tasks/actions/create_task.dart
```diff
@@ -0,0 +1,433 @@
...
+import 'package:dart_date/dart_date.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_easyloading/flutter_easyloading.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::tasks::create_update_task_item');
+
+Future<void> showCreateTaskBottomSheet(
```
gtalha07, line 21: This is just 'copied-over from somewhere' stuff I assume? Nothing changed in actual ?

app/lib/l10n/app_ur.arb
```diff
@@ -2144,7 +2144,7 @@
   "@audio": {},
   "pinDetails": "پن کی تفصیلات",
   "@pinDetails": {},
-  "inSpaceLabelInline": "میں",
+  "inSpaceLabelInline": ":میں",
   "@inSpaceLabelInline": {},
   "comingSoon": "ابھی تک تعاون یافتہ نہیں، جلد آرہا ہے!",
```
gtalha07, line 2149: Also, would be great if you can correct the exclamation here in `comingSoon`, it should be at the end (from RTL) rather than at the start :)

app/test/features/tasks/tasks_adding_test.dart
```diff
@@ -0,0 +1,352 @@
...
+import 'package:acter/features/tasks/actions/create_task.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mocktail/mocktail.dart';
+
+import '../../helpers/mock_a3sdk.dart';
+import '../../helpers/mock_tasks_providers.dart';
+import '../../helpers/test_util.dart';
+
+void main() {
+  group('Create Task Widget on TaskList', () {
```
gtalha07, line 11: 👏🏻 !

app/lib/l10n/app_ur.arb
gnunicorn: you know you can just do it? either in weblate or within a PR? I am not confident I understood what change you want and I'd rather you fix it ...

app/lib/features/events/pages/event_details_page.dart
gnunicorn: `ref.listen` is meant to be used in `build`. I can wrap the set-state of course.

app/lib/l10n/app_ur.arb
gtalha07: Sure, I can take care of that in weblate.

app/lib/features/events/pages/event_details_page.dart
gnunicorn: @gtalha07 I refactored this as it wasn't properly working anymore with my updates. 049fd53dcc0ae100e7094de10216138e2675c91b contains a separate provider and now all is fine again. We should add tests for these widget soon, as that would have been great to catch earlier :)

PR #2350, opened 2024-11-06
Optimised Pins providers usage

kumarpalsinh25:
As per suggestion from @gnunicorn , this PR improves provider usages in Pins Module.

- Pin module now have it's own `pinListSearchTermProvider` which is independent from other module's search providers
- `PinListWidget` now contains `pinListProvider` as param which makes it more generic in nature with better efficiency.

No changes in Visual side but still sharing video reference to ensure that it is working as expect even after this changes.

https://github.com/user-attachments/assets/98458bb6-5c99-4702-81a2-ddd10a53da8d

### DIFF
--- a/app/lib/features/pins/providers/pins_provider.dart
+++ b/app/lib/features/pins/providers/pins_provider.dart
@@ -5,54 +5,71 @@ import 'package:acter/features/pins/models/pin_edit_state/pin_edit_state.dart';
 import 'package:acter/features/pins/providers/notifiers/create_pin_notifier.dart';
 import 'package:acter/features/pins/providers/notifiers/edit_state_notifier.dart';
 import 'package:acter/features/pins/providers/notifiers/pins_notifiers.dart';
+import 'package:acter/features/search/providers/quick_search_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:riverpod/riverpod.dart';
 
+//Search Value provider for pin list
+final pinListSearchTermProvider = StateProvider<String>((ref) => '');
+
 //SpaceId == null : GET LIST OF ALL PINs
 //SpaceId != null : GET LIST OF SPACE PINs
-final pinListProvider =
+final _pinsProvider =
     AsyncNotifierProvider.family<AsyncPinListNotifier, List<ActerPin>, String?>(
   () => AsyncPinListNotifier(),
 );
 
-//Search any pins
-typedef AllPinsSearchParams = ({String? spaceId, String searchText});
+//All Pins List Provider
+final allPinListProvider =
+    FutureProvider.autoDispose.family<List<ActerPin>, String?>(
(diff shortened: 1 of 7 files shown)


app/lib/features/pins/providers/pins_provider.dart
```diff
@@ -5,54 +5,71 @@ import 'package:acter/features/pins/models/pin_edit_state/pin_edit_state.dart';
...
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:riverpod/riverpod.dart';
 
+//Search Value provider for pin list
+final pinListSearchTermProvider = StateProvider<String>((ref) => '');
+
 //SpaceId == null : GET LIST OF ALL PINs
 //SpaceId != null : GET LIST OF SPACE PINs
-final pinListProvider =
+final _pinsProvider =
```
gtalha07, line 17: Why this has not to be global?

app/lib/features/space/pages/space_details_page.dart
```diff
@@ -264,6 +265,7 @@ class _SpaceDetailsPageState extends ConsumerState<SpaceDetailsPage> {
       TabEntry.overview => AboutSection(spaceId: widget.spaceId),
       TabEntry.news => NewsSection(spaceId: widget.spaceId),
       TabEntry.pins => PinListWidget(
+          pinListProvider: allPinListWithBookmarkFrontProvider(widget.spaceId),
```
gtalha07, line 268: So that means space would only have access to bookmarked pins and not all of them ? Even if we want to see all of them??

gtalha07 (approved):
I am approving this generally but I couldn't understand the purpose of having passed provider down the widgets where they can be accessed globally?

app/lib/features/pins/widgets/pin_list_widget.dart
```diff
@@ -13,6 +12,7 @@ import 'package:logging/logging.dart';
 final _log = Logger('a3::pins::list');
 
 class PinListWidget extends ConsumerWidget {
+  final ProviderBase<AsyncValue<List<ActerPin>>> pinListProvider;
```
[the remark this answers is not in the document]
kumarpalsinh25: Let me explain..

`PinListWidget` is now generic widget and used in multiple area like `SpacePinsSections`, `All Pins Page`,  `Quick Search Pin Sections` etc..
And we have different providers for different use-cases, but ultimately we are giving list of pins..

So `PinListWidget` contains generic providers as param which return `List<ActerPin>`.. Which makes it generic yet independent in terms of data part...

Happy to explain more in call if require further explanation. :)

app/lib/features/space/pages/space_details_page.dart
kumarpalsinh25: `allPinListWithBookmarkFrontProvider`

This is giving all pin list with bookmark pins are in front of link...

Issue #1704, opened 2024-05-01
News: foreground & link color almost never work together


gnunicorn:
Flipping through various background colors, you'll notice that the contrast with either the foreground text or the links in it is always making at least one of them illegible:

https://github.com/acterglobal/a3/assets/40496/cb0c4401-5831-4150-be7b-7a0724daae49

gnunicorn:
improved but not yet fixed.

I think we need to take a much more radical approach here and either define high-contrast foreground-colors for the set of background colors -- including a link-blue -- or define some sort of shaded-box that the content is in so that it will always have enough of a contrast to the then darker background.

Issue #2358, opened 2024-11-10
Swiping Boost Slides difficult on desktop 


gnunicorn:
Reported in https://github.com/acterglobal/a3-meta/issues/626 :
> I can't swipe to see other slides on the boosts when I am on desktop version. 

I think this is due to the system using "vertical scroll" to "swipe" on desktop and that simply isn't always available for nice. So maybe some touch-swipe and/or click on the indicators would be useful?!?

kumarpalsinh25:
@gnunicorn @jo-acter 

Swipe works fine on desktop as well. Check below video for reference.
https://github.com/user-attachments/assets/b3fcb89e-1bc4-4c5e-9809-ae7dd6896cfa

Default Behaviour is as below.

1) TrackPad : 
- Swipe Up-Down for vertical scrolling
- Swipe Left-Right for horizontal scrolling

2) Mouse :
- Scroll wheel Up-Down for vertical scrolling
- Scroll wheel Up-Down with Shift Key pressed for horizontal scrolling

I will check for additional alternate options for till that time you can go with above behaviour and let me know in case of any difficulties.

gnunicorn:
> I will check for additional alternate options for till that time you can go with above behaviour and let me know in case of any difficulties.

being able to hold and drag with the mouse would be useful regardless, I think ... shift-scroll isn't really that commonly used by regular people, I believe...

kumarpalsinh25:
> being able to hold and drag with the mouse would be useful regardless, I think ... shift-scroll isn't really that commonly used by regular people, I believe...

Yeh true and agree with you, I will look for better approach but just post here current behaviour for knowledge purpose. :)

Issue #2356, opened 2024-11-10
 new pin icon doesnt refresh on space overview


gnunicorn:
first reported in https://github.com/acterglobal/a3-meta/issues/623 

1. Create pin and change icon.
2. Go to space overview and see if it matches

Issue #2351, opened 2024-11-06
Small space changes


emilvincentz:
- Change title on space overview from "Suggested chats" -> "Chats"
- Reverse the order on spaces and chats (so that chats shows first)

emilvincentz:
@kumarpalsinh25 if you have a little time window, i would appriciget if you could have a look at this one

kumarpalsinh25:
Yes sure @emilvincentz

PR #2361, opened 2024-11-11
General Fixes

kumarpalsinh25:
1) Fixes, [New pin icon doesnt refresh on space overview](https://github.com/acterglobal/a3/issues/2356)
Commit: https://github.com/acterglobal/a3/pull/2361/commits/f7a4d0ab13bb291b68f9b60d9f301a131a33dc63

https://github.com/user-attachments/assets/86ee0e51-f6fb-4893-8320-c969394af17d

2) Fixes, [Keyboard overlay issue when do Add Task from Quick Action Buttons](https://github.com/acterglobal/a3/issues/2362)
Commit: https://github.com/acterglobal/a3/pull/2361/commits/5ea4214ad8ae435c1573dbc82be0e155e8662f11

https://github.com/user-attachments/assets/b8a06290-1c01-43ac-9dc0-653c5674058b

3) Fixes, [News: foreground & link color almost never work together](https://github.com/acterglobal/a3/issues/1704)
Commit: https://github.com/acterglobal/a3/pull/2361/commits/77f051dc156ebcb594e90dceb5569eec19c5df9c

https://github.com/user-attachments/assets/9e725509-9595-475d-87c5-1cf7a6d1592b

4) Fixes, [Small space changes](https://github.com/acterglobal/a3/issues/2351)
Commit: https://github.com/acterglobal/a3/pull/2361/commits/0cfca1709c8c2a2aec6c351efce3efa8aa4beaf1

https://github.com/user-attachments/assets/7e1abbde-6c13-4114-8614-bd8fd501ade9

5) Fixes, [Swiping Boost Slides difficult on desktop](https://github.com/acterglobal/a3/issues/2358)
Commit: https://github.com/acterglobal/a3/pull/2361/commits/becf82f89bffac147b20807956dc313a741a43f3

https://github.com/user-attachments/assets/4ab24beb-5731-479d-97f8-24eedfd54175

6)Fixes,  [Event list order is not based on bookmark priority](https://github.com/acterglobal/a3/issues/2366)
Commit : https://github.com/acterglobal/a3/pull/2361/commits/ac272436c5aad506e4370893b0d1caebc32f25bd

https://github.com/user-attachments/assets/c01372ea-e7e9-4895-992a-5efc7753ca23


gnunicorn:
I don't see how you think https://github.com/acterglobal/a3/commit/77f051dc156ebcb594e90dceb5569eec19c5df9c would fix https://github.com/acterglobal/a3/issues/1704 the contrast is still way to low and bad. The word "link" is barely legible.

I think we need a different approach there.

app/lib/features/pins/actions/pin_update_actions.dart
```diff
@@ -115,6 +115,7 @@ Future<void> updatePinIcon(
     await updateBuilder.send();
     EasyLoading.dismiss();
     ref.invalidate(pinProvider);
+    ref.invalidate(pinListProvider);
```
gnunicorn, line 118: this only fixes the case where _we_ do the update. if the change comes from outside - another user - this will not trigger.

the proper fix for that entire situation is to switch the `pinList` to just return the `eventId`s and have each item be observed internally. But that is a bigger change not necessary now. We should write that up somewhere though.

app/lib/features/tasks/actions/create_task.dart
```diff
@@ -24,9 +24,12 @@ Future<void> showCreateTaskBottomSheet(
     showDragHandle: true,
     useSafeArea: true,
     isScrollControlled: true,
-    builder: (context) => CreateTaskWidget(
-      taskList: taskList,
-      taskName: taskName,
+    builder: (context) => Padding(
+      padding: MediaQuery.of(context).viewInsets,
```
gnunicorn, line 28: curios ... why is that not needed in other cases? Is it because the object then pulls the auto-focus? But others should be doing that, too, no? I'd like to understand the cause here ...

Or should we wrap that in a safe-area?

app/lib/features/events/providers/event_providers.dart
```diff
@@ -186,28 +187,52 @@ final eventListSearchedProvider = FutureProvider.autoDispose
...
+    priotizeBookmarkedEvents,
   );
 });
 
 final eventListSearchedAndFilterProvider = FutureProvider.autoDispose
     .family<List<ffi.CalendarEvent>, String?>((ref, spaceId) async {
   //Declare filtered event list
-  final filteredEventList =
+
+  final bookmarkedEvents =
```
gnunicorn, line 208: why this? Why `await` on _all_ of these now if only some are needed in any case? I understand the other might be a bit harder to parse but this is executes unnecessary code (and `ref.watch`es for unnecessary changes).

I'd suggest reverting to the old swtich-statement.

gnunicorn (approved):
Generally fine, just two minor remarks.

app/lib/features/events/providers/event_providers.dart
kumarpalsinh25: Screenshot 2024-11-13 at 4 26 18 PM

app/lib/features/pins/actions/pin_update_actions.dart
kumarpalsinh25: Adding TODO for now.

Screenshot 2024-11-13 at 4 29 11 PM

app/lib/features/tasks/actions/create_task.dart
kumarpalsinh25: Don't know exact root cause. I also faced similar kind of issue where I developer createEditTitle common widget where it was working fine on some area and misbehave at some places.

Above code just fixed the issue earlier also so applied it here.

kumarpalsinh25: I can also confirm that wrapping with `safe-area` doesn't fixed this.

app/lib/features/events/providers/event_providers.dart
kumarpalsinh25: Also old switch statements doesn't sorting list by ongoing first then upcoming and at the last past event when switch case-all is selected.

kumarpalsinh25:
> I don't see how you think [77f051d](https://github.com/acterglobal/a3/commit/77f051dc156ebcb594e90dceb5569eec19c5df9c) would fix #1704 the contrast is still way to low and bad. The word "link" is barely legible.
> 
> I think we need a different approach there.

For now, I removed color which causing issue in visibility of link colors. Open for any better solution / suggestions.

app/lib/features/events/providers/event_providers.dart
gnunicorn: I can't follow. The error message you are showing me is for the new code ... what am I supposed to see there?

app/lib/features/tasks/actions/create_task.dart
gnunicorn: Hm... that is unsatisfying ... would be good to figure out the root cause at some point...

app/lib/features/events/providers/event_providers.dart
kumarpalsinh25: As mentioned in above comment, your older code doesn't sorting list with ongoing follow by upcoming and past when ALL switch case selected.

app/lib/features/tasks/actions/create_task.dart
kumarpalsinh25: Hmm true, let me check if I can figure out something

PR #2372, opened 2024-11-15
Upgrading Deps again

gnunicorn:
Upgrading to latest matrix-rust-sdk, flutter dependencies.


Cargo.toml
```diff
@@ -11,13 +11,16 @@ default-members = ["native/acter"]
 git = "https://github.com/matrix-org/matrix-rust-sdk"
 # path = "../matrix-rust-sdk/crates/matrix-sdk"
 default-features = false
-features = ["experimental-sliding-sync"]
+features = [
+  "experimental-sliding-sync",
+  "rustls-tls",
```
gnunicorn, line 16: we are adding `matrix-sdk` in the acter-core now and that means this flag is needed for clippy to succeed in all cases :( .

PR #2373, opened 2024-11-15
New Room Preview facilities

gnunicorn:
You will now get the room preview shown (if the server supports it), with the ability to join the space, if you have the permissions to, 

## When you click on an Space / Chat Invite

https://github.com/user-attachments/assets/dc588364-a9c5-461d-bf94-1aee8687467d

## When you try to access an object via Ref you don't have access to yet

https://github.com/user-attachments/assets/7e764fc2-553f-4ee6-93a6-fb8c8ac45848

## When you scan a QR code to an object you aren't in the space of yet

https://github.com/user-attachments/assets/29c5d5be-d8bd-48f4-91ce-ef82859ccd53

For additional convienience, we are also showing exactly what will be seen/shown/shared over the QR code to the other user, too:

image

_To the reviewer_:
Sorry, for again coming with such a large pull request. But in certain parts of the app (especially with really old code, like this joining rooms one), it's like the famous "pulling one thread and the entire thing unravels" ... and cascading changes are needed.

In particular these are (not the order they were committed):
 - The actual RoomPreview rust, ffi and dart support, obviously ;) 
 - linking the room preview to room invites: 0b5f89a41296b2a00e8c4433216cbe3e176571ef
 - a bigger refactor of ref-item to allow for opening the room preview and have people join and be forwarded from there (as shown above) in 753abc4285bc732614301810bd079e451dc9b97d
 - A minor update in the error handler to give us the option to show better, more specific access information if the server provides it for us: which is the case if we aren't allowed to see the room preview. This is all the `textBuilder`-changes on those `ActerError`-Widget in db8e099ff337c3d58f8778b5b8806913e752821b
 - Two big changes were done around joining rooms:
   - the via-servers are now provided as _a list_ on the dart side as well (as it should be) and we provide our main server in the ref-details for external sharing for proper support of that - which also brings some new rust API to generate lists of strings
   - a refactor of the `joinRoom`-action to remove the forwarding-callback and instead use the future-syntax as we do in other actions as well, in f2036f98c522fa61e87344f41e312ee6a408c8ae  
   - which also brings the - hopefully final - fix for showing a 404 after forwarding to a joined space in 26f921460fdc51e5f7dfec922670f12d6c2815bb


app/lib/common/toolkit/errors/util.dart
```diff
@@ -1,12 +1,20 @@
+typedef ErrorTextBuilder = String Function(Object error, ErrorCode code);
+
 enum ErrorCode {
   notFound,
+  forbidden,
```
gnunicorn, line 5: Extended the Error types to be more specific and more precise and _allow us_ to show a different error depending on the underlying code we were given by the server: most notably if we were forbidden access.

app/lib/features/preview/widgets/room_preview.dart
```diff
@@ -0,0 +1,221 @@
...
+    BuildContext context,
+    WidgetRef ref,
+    Object error,
+    StackTrace? stack,
+  ) =>
+      ActerInlineErrorButton(
+        error: error,
+        stack: stack,
+        textBuilder: (error, errCode) => switch (errCode) {
+          ErrorCode.forbidden =>
```
gnunicorn, line 87: Showing different errors depending on the error indicated by the server.

app/lib/features/room/actions/join_room.dart
```diff
@@ -14,54 +16,68 @@ Future<T> _ensureLoadedWithinTime<T>(
...
     final roomId = newRoom.roomIdStr();
-    final isSpace = newRoom.isSpace();
-    // ensure we re-evaluate the room data on our end. This is necessary
-    // if we knew of the room prior (e.g. we had left it), but hadn’t joined
-    // this should properly re-evaluate all possible readers
-    ref.invalidate(maybeRoomProvider(roomId));
+    final isSpace = await _ensureLoadedWithinTime(() async {
+      final room = await ref.refresh(maybeRoomProvider(roomId).future);
+      return room?.isJoined() == true ? room!.isSpace() : null;
+    });
```
gnunicorn, line 67: this now properly checks for the room being available _and_ accessible to the UI before continuing.

app/lib/features/room/widgets/room_chip.dart
```diff
@@ -0,0 +1,147 @@
...
+        ),
+      ),
+      label: Text(displayName, overflow: TextOverflow.ellipsis),
+      onPressed: () async {
+        await navigateToRoomOrAskToJoin(context, ref, widget.roomId);
+      },
+    );
+  }
+
+  Widget buildPreview(BuildContext context) {
```
gnunicorn, line 123: new feature to show the preview in the chat for a room we don't have access to

app/lib/features/share/action/share_space_object_action.dart
```diff
@@ -133,10 +135,25 @@ class ShareSpaceObjectActionUI extends ConsumerWidget {
   Widget externalShareOptionsUI(BuildContext context) {
     return ExternalShareOptions(
       qrContent: internalLink,
+      qrCodeHeader: qrCodeHeader(context),
       shareContentBuilder: shareContentBuilder,
     );
   }
 
+  Widget? qrCodeHeader(BuildContext context) {
```
gnunicorn, line 143: Minor feature (and useful for debugging): show the _to be shared_ information within the QRcode above the QRcode as well.

app/test/features/deep_linking/parse_acter_uri_test.dart
gnunicorn, line 1: tests updated for additional `via`-support that we always include now

app/lib/features/deep_linking/widgets/reference_details_item.dart
```diff
@@ -0,0 +1,80 @@
...
+          Padding(
+            padding: EdgeInsets.symmetric(vertical: 8),
+            child: Text(
+              L10n.of(context).needToBeMemberOf,
+              style: Theme.of(context).textTheme.bodyMedium,
+            ),
+          ),
+        ],
+      ),
+    );
```
kumarpalsinh25, line 78: Probably we can design better design for this. (Later part..)

Also can be generalised this code as I see this used couple of times..

app/lib/features/deep_linking/widgets/reference_details_item.dart
```diff
@@ -0,0 +1,80 @@
...
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class ReferenceDetailsItem extends ConsumerWidget {
+  final RefDetails refDetails;
+  final bool tapEnabled;
+  final EdgeInsets? margin;
+
+  const ReferenceDetailsItem({
```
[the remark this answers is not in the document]
kumarpalsinh25: What is the need of `ReferenceDetailsItem` as we have `ItemPreviewCard`. Can't we combined those in single widget?

app/lib/features/preview/widgets/room_preview.dart
```diff
@@ -0,0 +1,221 @@
...
+        onRetryTap: () {
+          ref.invalidate(maybeRoomProvider(roomId));
+          ref.invalidate(roomPreviewProvider(query));
+        },
+      );
+
+  Widget roomHeader(RoomPreview preview) => Consumer(
+        builder: (context, ref, child) => ListTile(
+          leading: ActerAvatar(
+            options: AvatarOptions(ref.watch(roomPreviewAvatarInfo(query))),
```
kumarpalsinh25, line 100: Don't understand this `query` thing.

app/lib/features/room/actions/join_room.dart
```diff
@@ -14,54 +16,68 @@ Future<T> _ensureLoadedWithinTime<T>(
...
-    final res = await callback();
-    if (res != null) {
-      return res;
+    try {
+      final res = await callback();
+      if (res != null) {
+        return res;
+      }
+    } catch (e) {
+      if (throwError) rethrow;
```
kumarpalsinh25, line 30: Don't understand need of this change.

kumarpalsinh25 (approved):
Nice addition!!

Added few comments few none of that blocking for merge. PR changes looks good to me.

app/lib/common/toolkit/errors/util.dart
```diff
@@ -1,12 +1,20 @@
...
 
   static ErrorCode guessFromError(Object error) {
     final errorStr = error.toString();
     // yay, string-based error guessing!
     if (errorStr.contains('not found')) {
       return ErrorCode.notFound;
+    } else if (errorStr.contains('[400 / UNKNOWN]')) {
+      return ErrorCode.unknown;
+    } else if (errorStr.contains('[403 / M_FORBIDDEN]')) {
+      return ErrorCode.forbidden;
```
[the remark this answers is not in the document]
gnunicorn: yes, this is exact on purpose.

app/lib/common/extensions/record_helpers.dart
```diff
@@ -0,0 +1,29 @@
...
+  @override
+  bool operator ==(Object other) {
+    if (other is AllHashed<T>) {
+      return items == other.items;
+    } else if (other is List<T>) {
+      return items == other;
+    }
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
gnunicorn: two list of items, even if they are all the same, will come out as "different" if they are different instances. This is used to ensure that https://github.com/acterglobal/a3/pull/2373/files#diff-b2a403334550e3f97cc376cf1228271526121ca2ecf70cf62e6c50ba41e69289R14 caches the input value and not keeps triggering for every render.

app/lib/features/deep_linking/types.dart
```diff
@@ -83,4 +83,12 @@ class UriParseResult {
...
     this.preview = const ObjectPreview(),
   });
+
+  ObjectType? finalType() {
+    ObjectRef? cur = objectPath;
+    while (cur?.child != null) {
+      cur = cur?.child; // find the deepest object
+    }
+    return cur?.objectType;
+  }
```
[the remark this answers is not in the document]
gnunicorn: it traverses the object path of the reference type to give the most internal objects type.

the way this is constructed is that the actor URI allows unlimited deep linking. Think:

`/tl/asdf/t/339390/c/2323` -> this is a reference to the comment `2323` of task `33930` of task list `asdf`. The way this is represented in the parse uri result is that every object with an object path can have an additional `child` which (going left to right) gives the next inner type. So this would be `TaskList[id=asdf, child=Task[id=339390. child=Comment[id=2323]]]`.  In order to find out what the _actual_ final object is (in this case "comment"), you need to traverse that tree. Which is what this code does: look for the most inner item that doesn't have any child (in the `while`) and then return its objectType.

app/lib/features/deep_linking/widgets/reference_details_item.dart
gnunicorn: it is already generalized some part. But I didn't get this much further without making things a lot more specific again ... Happy to go further though.

app/lib/features/preview/widgets/room_preview.dart
gnunicorn: query needs roomId and via-server listing, which is combined as `query` record.

app/lib/features/room/actions/join_room.dart
gnunicorn: this code was meant to repeatedly try to call the function until a timeout occured. But that wasn't the case if there was an exception raised in the inner code, it would just fail immediately instead of retrying. This fixes the retry behaviour to what we really wanted to have in the first place.

PR #2384, opened 2024-11-24
Fixing Workshop issues: Linking any Space & Chat, unlinking and tasks creation

gnunicorn:
While running the workshop the other day, we noticed some smaller bugs when going through the flow. This PR fixes them:

1. Fixes unlinking of chats wasn't working
2. You get prompted about activating the Tasks feature like with all other features, too

https://github.com/user-attachments/assets/29af3d4b-d0ab-41b3-89d9-ada2f35e631f

3. You can now link any Chat or any space as a "sub" one and we cleverly figure out whether that is "recommended" or "sub". This finishes the clean up of "recommended" vs "sub" and removes a bunch of artificial distinction between them for the UI, too

https://github.com/user-attachments/assets/fdbb346f-30d7-4404-96d6-7603a5978477


app/lib/features/link_room/actions/unlink_child_room.dart
```diff
@@ -20,7 +20,7 @@ Future<void> unlinkChildRoom(
   await space.removeChildRoom(roomId, reason);
   //Fetch selected room data and add given parentSpaceId as parent
   final room = await ref.read(maybeRoomProvider(roomId).future);
-  if (room != null && room.isSpace()) {
+  if (room != null) {
```
gnunicorn, line 23: this is what prevented the previous code from properly unlinking chats.

app/lib/features/link_room/pages/link_room_page.dart
```diff
@@ -0,0 +1,205 @@
+import 'package:acter/common/extensions/options.dart';
```
gnunicorn, line 1: the core structure of these has remained the same, but they appears a new files because I moved the content into its own feature and split it up into several distinct widgets that I can also easier test on their own.

app/lib/features/link_room/pages/link_room_page.dart
```diff
@@ -0,0 +1,205 @@
...
+  });
+
+  @override
+  ConsumerState<LinkRoomPage> createState() => _LinkRoomPageConsumerState();
+}
+
+class _LinkRoomPageConsumerState extends ConsumerState<LinkRoomPage> {
+  final TextEditingController searchTextEditingController =
+      TextEditingController();
+
```
gnunicorn, line 37: removed a bunch of stateful ness here that could be done with providers instead. Later realized that we can even collapse most of these providers.

app/lib/features/link_room/types.dart
```diff
@@ -0,0 +1,5 @@
+// ChildRoomType configures the sub child type of the `Spaces`
+enum ChildRoomType {
+  chat,
+  space,
```
gnunicorn, line 4: removed the "recommneded" type as we consider all of them the same, it is really just a question of which listing provider we are using.

app/lib/features/link_room/widgets/link_room_trailing.dart
```diff
@@ -0,0 +1,177 @@
...
+
+    //Make subspace
+    if (canLink) {
+      //Fetch selected room data and add given parentSpaceId as parent
+      await room.addParentRoom(parentId, true);
+      if (!context.mounted) return;
+      await checkJoinRule(context, ref, room, parentId);
+    } else {
+      EasyLoading.showSuccess(lang.roomLinkedButNotUpgraded);
+    }
```
gnunicorn, line 152: This now accepts both the non upgradable and the uprgadable cases and only shows the dialog if you can upgrade. If you can't uprgade you'll get a success messages with a note instead.

app/lib/features/space/widgets/space_sections/space_actions_section.dart
```diff
@@ -37,118 +39,205 @@ class SpaceActionsSection extends ConsumerWidget {
...
+          membership,
+          settings,
         ),
-      if (canAddTask || canChangeSetting)
-        simpleActionButton(
-          context: context,
-          iconData: Atlas.list,
-          title: lang.addTask,
-          onPressed: () async {
-            if (!canAddEvent && canChangeSetting) {
```
gnunicorn, line 126: _this_ was the actual bug. `canAddEvent` is the wrong check here. However, while looking at it, I also noticed that we are only checking the permission but not if the feature is activated, so I did refactor the entire file to check for the feature-activation more specifically.

app/test/features/link_room/link_room_page_test.dart
```diff
@@ -0,0 +1,147 @@
+import 'package:acter/common/providers/chat_providers.dart';
```
gnunicorn, line 1: yay tests \o/ !

gtalha07 (approved):
Looks good!

app/lib/features/link_room/widgets/link_room_trailing.dart
```diff
@@ -0,0 +1,177 @@
...
+                  style: OutlinedButton.styleFrom(
+                    side: BorderSide(color: colorScheme.success),
+                  ),
+                  child: Text(lang.link),
+                )
+              : null,
+    );
+  }
+
+  Future<void> checkJoinRule(
```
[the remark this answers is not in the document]
gnunicorn: I was thinking about it but consider this very specific to this flow at the moment and rather not have this be reusable for others yet ...

PR #2387, opened 2024-11-26
Content-hashed external URI parsing support

gnunicorn:
This PR brings support for providing non matrix-to https links with content hashing support for additional security. These work analogous to the previous `acter:`-style-links but with that entire path put into the `#`-fragment (query stays the same, but the order might change) and replacing the path with a final hash of the entire URI in this style _without_ that hash as the final parameter). This allows any server to provide that feature under any domain + path as long as the last path-item is the hash itself.

Example: `acter:roomid/room:acter.global/e/someEvent?via=acter.global&via=example.org` becomes `https://app.acter.global/p/$HASH?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent` (where `$HASH = SHA1.of(https://app.acter.global/p/?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent`)

This ensures that we can a) still parse most information from the URI itself without querying the server and b) ensure that the data wasn't messed with in transit or by any third party compared to the data that the server would be showing under that hash.

To the reviewer:
1. I renamed the files and functions for better lookup
2. I have refactored the code to be reusable for both tests of the `acter:` and new https-links, that's why the test file looks different
3. I decided to update the super-invite-uris to use the same pattern (rather than a host-url-one) to make it easier to reuse the code (which required updating our current qr-code generator)
4. links with a hash are checked for containing a user-id - if not the result is thrown in a particular error (allowing the UI to still use it, but make clear that this isn't the usual way it is meant to be)

### DIFF
--- a/app/lib/features/deep_linking/parse_acter_uri.dart
+++ b/app/lib/features/deep_linking/parse_acter_uri.dart
@@ -20,7 +20,7 @@ class ObjectNotSupported extends UriParseError {
 class ParsingFailed extends UriParseError {}
 class IncorrectHashError extends ParsingFailed {}
 class MissingUserError extends ParsingFailed {
-  final UriParseError result;
+  final UriParseResult result;
   MissingUserError({required this.result});
 }
 
@@ -59,7 +59,11 @@ UriParseResult _parseHttpsUri(Uri uri) {
 
   // put the query as the path
   final extractableUri = strippedUri.replace(path: strippedUri.fragment, fragment: null);
-  return _parseActerUri(extractableUri);
+  final result = _parseActerUri(extractableUri);
+  if (result.preview.userId == null || result.preview.userId?.isEmpty == true) {
+    throw MissingUserError(result: result);
+  }
+  return result;
 }
 
 UriParseResult _parseSuperInvite(Uri uri) {
--- a/app/lib/features/deep_linking/parse_acter_uri.dart
+++ b/app/lib/features/deep_linking/parse_acter_uri.dart
@@ -19,15 +19,19 @@ class ObjectNotSupported extends UriParseError {
 
 class ParsingFailed extends UriParseError {}
 class IncorrectHashError extends ParsingFailed {}
+class MissingUserError extends ParsingFailed {
+  final UriParseError result;
+  MissingUserError({required this.result});
+}
 
-UriParseResult parseUri(Uri uri) => switch (uri.scheme) {
-      'acter' => _parseUri(uri),
+UriParseResult parseActerUri(Uri uri) => switch (uri.scheme) {
+      'acter' => _parseActerUri(uri),
       'matrix' => _parseMatrixUri(uri),
       'https' || 'http' => _parseHttpsUri(uri),
       _ => throw SchemeNotSupported(scheme: uri.scheme),
     };
 
-UriParseResult _parseUri(Uri uri) {
+UriParseResult _parseActerUri(Uri uri) {
   final path = uri.pathSegments.first;
   return switch (path) {
     'o' => _parseActerEvent(uri),
@@ -55,7 +59,7 @@ UriParseResult _parseHttpsUri(Uri uri) {
 
   // put the query as the path
   final extractableUri = strippedUri.replace(path: strippedUri.fragment, fragment: null);
-  return _parseUri(extractableUri);
+  return _parseActerUri(extractableUri);
 }
 
 UriParseResult _parseSuperInvite(Uri uri) {
--- a/app/test/features/deep_linking/parse_acter_uri_test.dart
+++ b/app/test/features/deep_linking/parse_acter_uri_test.dart
@@ -185,10 +185,11 @@ void newSpecLinksTests(UriMaker makeUri) {
   });
 }
 
-UriMaker makeUriMakerForPublicPrefix(String uriPrefix) {
+UriMaker makeUriMakerForPublicPrefix(String uriPrefix, {String? userId = 'test:example.org'}) {
   Uri makeDomainLink(String path, String? query) {
-    final hash = sha1.convert(utf8.encode('$uriPrefix?$query#$path')).toString();
-    return Uri.parse('$uriPrefix/$hash?$query#$path');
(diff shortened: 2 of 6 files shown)


app/lib/features/deep_linking/parse_acter_uri.dart
```diff
@@ -32,14 +40,41 @@ UriParseResult _parseActerUri(Uri uri) {
   };
 }
 
+UriParseResult _parseHttpsUri(Uri uri) {
```
gnunicorn, line 43: this is the core of the new scheme

app/lib/features/deep_linking/parse_acter_uri.dart
```diff
@@ -32,14 +40,41 @@ UriParseResult _parseActerUri(Uri uri) {
...
+  if (uri.host == 'matrix.to') {
+    return _parseMatrixHttpsUri(uri);
```
gnunicorn, line 45: matrix.to is a special case with a special format for legacy links

app/lib/features/deep_linking/parse_acter_uri.dart
```diff
@@ -32,14 +40,41 @@ UriParseResult _parseActerUri(Uri uri) {
...
+    throw IncorrectHashError();
+  }
+
+  final pathWithoutHash = uri.path.substring(0, uri.path.length - hash.length -1);
+  final strippedUri = uri.replace(path: pathWithoutHash);
+  final hashableUri = strippedUri.toString();
+  final calculatedHash = sha1.convert(utf8.encode(hashableUri)).toString();
+  if (calculatedHash != hash) {
+    throw IncorrectHashError();
+  }
```
gnunicorn, line 58: 1. look for the hash
2. remove the hash
3. compare the hash

app/lib/features/deep_linking/parse_acter_uri.dart
```diff
@@ -32,14 +40,41 @@ UriParseResult _parseActerUri(Uri uri) {
...
+
+  // put the query as the path
+  final extractableUri = strippedUri.replace(path: strippedUri.fragment, fragment: null);
+  final result = _parseActerUri(extractableUri);
```
gnunicorn, line 62: extract the fragment as the path to just use the existing parsing code

app/lib/features/deep_linking/parse_acter_uri.dart
```diff
@@ -32,14 +40,41 @@ UriParseResult _parseActerUri(Uri uri) {
...
+  if (result.preview.userId == null || result.preview.userId?.isEmpty == true) {
+    throw MissingUserError(result: result);
```
gnunicorn, line 64: fail _with_ the result if no userId was found.

app/test/features/deep_linking/parse_acter_uri_test.dart
```diff
@@ -0,0 +1,295 @@
...
+import 'package:acter/features/deep_linking/types.dart';
+import 'package:acter/features/deep_linking/parse_acter_uri.dart';
+import 'package:crypto/crypto.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+typedef UriMaker = Uri Function(String path, String? query);
+
+// re-usable test cases
+
+void acterObjectLinksTests(UriMaker makeUri) {
```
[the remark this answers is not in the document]
kumarpalsinh25: 👏

app/test/features/deep_linking/parse_acter_uri_test.dart
```diff
@@ -0,0 +1,295 @@
...
+    acterObjectPreviewTests(makeUriMakerForPublicPrefix('https://app.acter.global/p/')),
+  );
+
+  group('Testing broken https://app.acter.global/ ', () {
+    test('faulty hash fails', () async {
+      expect(() => parseActerUri(
+          Uri.parse('http://acter.global/p/faultyHash?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent'),
+        ),
+        throwsA(TypeMatcher<IncorrectHashError>()),
+      );
```
kumarpalsinh25, line 242: Getting bit difficulty to grasp entire logic.

app/lib/features/deep_linking/parse_acter_uri.dart
kumarpalsinh25: - - -
Example: `acter:roomid/room:acter.global/e/someEvent?via=acter.global&via=example.org` becomes `https://app.acter.global/p/$HASH?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent` (where `$HASH = SHA1.of(https://app.acter.global/p/?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent`)
- - -
Not quite understand link structure and logic. Sorry but looks complex to me to understand.

kumarpalsinh25 (commented):
Deep-link structure looks complex to me, not sure if I am not getting it well or it is actually complex itself.

app/lib/features/deep_linking/parse_acter_uri.dart
gnunicorn: Well, not sure what to say. it is _complex_ by nature, as we try to have a link contain all the information of itself in a hash. It is just a question of how to do the pattern. Not sure what else to say here... other than:

```dart
final url = 'acter:roomid/room:acter.global/e/someEvent?via=acter.global&via=example.org'
final scheme = url.scheme; // = 'acter'
final path = url.path; // = 'roomid/room:acter.global/e/someEvent'
final query = url.query; // = 'via=acter.global&via=example.org'

final targetUriPrefix = 'https://app.acter.global/p/';

// which is then constructed a new as
final hash = SHA1.of('$targetUriPrefix?$query#$path'); // = '0c51512e60831746f59b479bc7371693b15f0ad0'

and then becomes:
final finalUri = `$targetUriPrefix$hash?$query#$path` // = 'https://app.acter.global/p/0c51512e60831746f59b479bc7371693b15f0ad0?via=acter.global&via=example.org#roomid/room:acter.global/e/someEvent'

```

PR #2388, opened 2024-11-27
[Chat-NG]: Room messages UI implementation

gtalha07:
- [x] provider for handling when to show avatar based on messages.
- [x] message bubble widget for message content rendering.
- [x] cover basic content types rendering .i.e. text, image, file etc.
- [x] tests coverage: includes unit test for ```shouldShowAvatarProvider``` and copied image message test from older chat version to chat-ng version with updated objects.

Simple previews of messages list with UI:

## Mobile:

https://github.com/user-attachments/assets/806f993f-a2dd-4282-b61a-b0729644bb14

## Desktop

https://github.com/user-attachments/assets/16d7ab08-96ab-45ae-b089-9de5e365ec08


app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -49,3 +53,30 @@ final renderableChatMessagesProvider =
     return _supportedTypes.contains(msg.eventItem()?.eventType());
   }).toList();
 });
+
+// Provider to check if we should show avatar by comparing with the next message
+final shouldShowAvatarProvider = Provider.family<bool, RoomMsgId>(
```
gnunicorn, line 58: Nice idea of just moving the logic into a provider. Do you think we could have some tests for it? That'd be awesome, especially since it is pretty simple now but then could increase in complexity and we could always have test showing it works as wanted!

gtalha07: Yes, they will be included in this PR :)

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+    RoomEventItem item,
+    bool isUser,
+    bool showAvatar,
+  ) {
+    final eventType = item.eventType();
+    switch (eventType) {
+      case 'm.room.message':
+        return _buildMsgEventItem(roomId, messageId, item, isUser, showAvatar);
+      case 'm.room.redaction':
+        return ChatBubble(
```
gtalha07, line 110: This is the new UI wrapper implemented for message events inner content. Rest of the implementation is same and copied from old chat widgets.

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -10,6 +10,32 @@ import 'package:logging/logging.dart';
...
+  'm.room.join.rules',
+  'm.room.name',
+  'm.room.pinned_events',
+  'm.room.power_levels',
+  'm.room.server_acl',
+  'm.room.third_party_invite',
+  'm.room.tombstone',
+  'm.room.topic',
+  'm.space.child',
+  'm.space.parent',
```
gnunicorn, line 37: I'd rather we don't add them here until we have a proper support. Like showing a "has changed the topic to XYZ" ... Let's not do the same mistake of just doing brush-of-a-stroke not-really-thought-through rendering please ... rather keep them off screen until we have decent support...

app/lib/features/chat_ng/widgets/events/image_message_event.dart
```diff
@@ -0,0 +1,170 @@
...
+import 'package:acter/features/chat/models/media_chat_state/media_chat_state.dart';
+import 'package:acter/features/chat/providers/chat_providers.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' show MsgContent;
+import 'package:flutter/material.dart';
+import 'package:flutter_chat_ui/flutter_chat_ui.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:phosphor_flutter/phosphor_flutter.dart';
+
+class ImageMessageEvent extends ConsumerWidget {
```
gtalha07, line 15: The inner implementation is same except that we now access ```MsgContent``` and identifiers directly instead of old ```types.Message```

app/lib/features/chat_ng/widgets/events/file_message_event.dart
```diff
@@ -0,0 +1,103 @@
...
+import 'package:acter/features/chat/providers/chat_providers.dart';
+import 'package:acter/features/files/actions/file_share.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' show MsgContent;
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_chat_ui/flutter_chat_ui.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:path/path.dart' as p;
+
+class FileMessageEvent extends ConsumerWidget {
```
gtalha07, line 12: Same as explained in ```ImageMessageEvent``` comment.

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -55,6 +79,33 @@ final renderableChatMessagesProvider =
...
+    final currentIndex = messages.indexOf(eventId);
+
+    // Always show avatar for the first message (last in the list)
+    if (currentIndex == messages.length - 1) return true;
+
+    // Get current and next message
+    final currentMsg = ref.watch(chatRoomMessageProvider(roomMsgId));
+    final nextMsg = ref.watch(
+      chatRoomMessageProvider((roomId, messages[currentIndex + 1])),
+    );
```
gnunicorn, line 97: I wonder if we shouldn't just make that its own provider, that would have the benefit of making caching nicer _and_ mocking easier if we just did a `nextMessageProvider(eventId) -> Message?` ...

app/lib/features/chat_ng/widgets/events/video_message_event.dart
```diff
@@ -0,0 +1,172 @@
...
+import 'package:acter/common/widgets/video_dialog.dart';
+import 'package:acter/features/chat/models/media_chat_state/media_chat_state.dart';
+import 'package:acter/features/chat/providers/chat_providers.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart' show MsgContent;
+import 'package:flutter/material.dart';
+import 'package:flutter_chat_ui/flutter_chat_ui.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class VideoMessageEvent extends ConsumerWidget {
```
gtalha07, line 13: Same as explained in ```ImageMessageEvent``` file comment

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'package:flutter/material.dart';
+import 'package:acter/common/extensions/options.dart';
+import 'package:flutter_gen/gen_l10n/l10n.dart';
+
+// Chat Bubble UI
+class ChatBubble extends StatelessWidget {
+  final int? messageWidth;
+  // inner content
+  final Widget child;
+  final bool isUser;
```
gnunicorn, line 11: I don't understand what these mean. Could you please add inline documentation what `isUser` and `showAvatar` are meant to convey?

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -0,0 +1,70 @@
...
+  final bool showAvatar;
+  final bool wasEdited;
+
+  const ChatBubble({
+    super.key,
+    required this.child,
+    this.messageWidth,
+    this.isUser = false,
+    this.showAvatar = true,
```
gnunicorn, line 20: Form the code `showAvatar` seems to only decide whether what the bottom corner looks like. But that isn't really what `showAvatar` implies as semantic-information. For e.g. in a 1-on-1 DM there is a good reason to not show either avatar (so `showAvatar = false`), while we still want to show the bottom corner differently some times ... we should rename that field to more appropriately describe what this is about.

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -0,0 +1,70 @@
...
+              color: isUser
+                  ? theme.colorScheme.primary
+                  : theme.colorScheme.surface,
+              borderRadius: BorderRadius.only(
+                topLeft: const Radius.circular(16),
+                topRight: const Radius.circular(16),
+                bottomLeft: Radius.circular(isUser || !showAvatar ? 16 : 4),
+                bottomRight: Radius.circular(isUser ? 4 : 16),
+              ),
+            ),
```
gnunicorn, line 49: so ... `isUser` is about which design we use? How about instead have that be a factory?

````dart
class ChatBubble extends StatelessWidget {
  final int? messageWidth;
  final BoxDecoration? decoration;
  //
  // ...
  //
  factory ChatBubble.mine({this.messageWidth, ...} : (boxDecoration: BoxDecoration(
              color: theme.colorScheme.primary,
              borderRadius: BorderRadius(
                topLeft: const Radius.circular(16),
                topRight: const Radius.circular(16),
                bottomLeft: Radius.circular(4),
                bottomRight: Radius.circular(4),
              ),
);

```

I prefer that pattern over passing boolean that are then used a lot within the rendered to switch between values of the same fields ...

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -1,27 +1,92 @@
...
-        roomId: roomId,
-        eventId: messages[index],
-        animation: animation,
-      ),
+  ConsumerState<ChatMessages> createState() => _ChatMessagesConsumerState();
+}
+
+class _ChatMessagesConsumerState extends ConsumerState<ChatMessages> {
+  final ScrollController _scrollController = ScrollController();
+  bool isLoading = false;
```
gnunicorn, line 16: don't keep local state that isn't necessary.
```suggestion
  bool get isLoading() => ref.watch(
      chatStateProvider(widget.roomId).select((v) => v.loading.isLoading),
    );
```

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -1,27 +1,92 @@
...
+    if (_scrollController.position.pixels >=
+        _scrollController.position.maxScrollExtent) {
+      setState(() => isLoading = true);
+
+      // Get the notifier to load more messages
+      final notifier = ref.read(chatStateProvider(widget.roomId).notifier);
+      await notifier.loadMore();
+
+      setState(() => isLoading = false);
+    }
```
gnunicorn, line 40:
```suggestion

    // Check if we're near the top of the list
    if (_scrollController.position.pixels >=
        _scrollController.position.maxScrollExtent) {
        if (isLoading) return; // FIXME: check for error

      // Get the notifier to load more messages
      ref.read(chatStateProvider(widget.roomId).notifier).loadMore();
    }
```

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -1,27 +1,92 @@
...
+            controller: _scrollController,
+            reverse: true,
+            itemBuilder: (_, index, animation) => Padding(
+              padding: const EdgeInsets.symmetric(vertical: 12),
+              child: ChatEvent(
+                roomId: widget.roomId,
+                eventId: messages[messages.length - 1 - index],
+              ),
+            ),
+          ),
```
gnunicorn, line 87: so we reverse to then in the itemBuilder reverse again? Are we sure this isn't messing with the `animatedListKey`'s idea of the index? What if we deleted or added entries? I feel a bit uneasy about this ... Can I have some tests that show this is working fine?

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+import 'package:acter/features/chat_ng/widgets/events/text_message_event.dart';
+import 'package:acter/features/chat_ng/widgets/events/video_message_event.dart';
+import 'package:acter_avatar/acter_avatar.dart';
+import 'package:flutter/material.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show RoomMessage, RoomVirtualItem, RoomEventItem;
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat::widgets::room_message');
```
gnunicorn, line 20:
```suggestion
final _log = Logger('a3::chat_ng::widgets::room_message');
```

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+    required this.eventId,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final msg = ref.watch(chatRoomMessageProvider((roomId, eventId)));
+
+    if (msg == null) {
+      _log.severe('Msg not found $roomId $eventId');
+      return const SizedBox.shrink();
```
gnunicorn, line 38: this is such a severe inconsistency problem in our code, I'd argue that we should be showing an error message to the user as well.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  Widget _buildEventItem(
+    String messageId,
+    RoomEventItem item,
+    bool isUser,
+    bool showAvatar,
+  ) {
+    final eventType = item.eventType();
+    switch (eventType) {
```
gnunicorn, line 106: screams for new switch-statement

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+      case 'm.room.topic':
+      case 'm.space.child':
+      case 'm.space.parent':
+        return StateUpdateEvent(item: item);
+      default:
+        return _buildUnsupportedMessage(eventType);
+    }
+  }
+
+  Widget _buildMsgEventItem(
```
gnunicorn, line 149: This will get complex soon. I'd suggest we already moved that into its own widget.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+    final msgType = item.msgType();
+    final content = item.msgContent();
+
+    // shouldn't happen but in case return empty
+    if (msgType == null || content == null) return const SizedBox.shrink();
+
+    switch (msgType) {
+      case 'm.emote':
+      case 'm.text':
+        return TextMessageEvent(
```
gnunicorn, line 165: Sorry, but I am a sucker for the new switch-statement ;)

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -0,0 +1,199 @@
...
+    bool showAvatar,
+  ) {
+    final msgType = item.msgType();
+    final content = item.msgContent();
+
+    // shouldn't happen but in case return empty
+    if (msgType == null || content == null) return const SizedBox.shrink();
+
+    switch (msgType) {
+      case 'm.emote':
```
gnunicorn, line 163: shouldn't we make is a separate widget as well now? I'd argue that an emote and sticker shouldn't have the bubble for example ...

app/lib/features/chat_ng/widgets/events/chat_event_widget.dart
gnunicorn, line 1: bye, bye :wave: !

gnunicorn (approved):
Just some minor structural things

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
gtalha07: Yeah, good idea. This can be degraded for ease in testing 👍🏻 !

PR #2393, opened 2024-11-29
Detect debug mode in platform code of desktop

bitfriend:
Flutter provides `kDebugMode` constant to detect debug mode in dart level.
But we don't know how to detect debug mode in cpp level.
We should be able to detect debug mode in cpp level for windows/linux/macos etc.

### DIFF
--- a/app/linux/main.cc
+++ b/app/linux/main.cc
@@ -1,6 +1,12 @@
 #include "my_application.h"
 
 int main(int argc, char** argv) {
+#ifndef NDEBUG
+  printf("This app works in debug mode\n");
+#else
+  printf("This app works in release mode\n");
+#endif
+
   g_autoptr(MyApplication) app = my_application_new();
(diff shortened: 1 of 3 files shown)


bitfriend:
During app build, flutter calls cmake with flutter build mode (one of debug/release/profile).
cmake converts `CMAKE_BUILD_TYPE` into `/D _DEBUG` when calling `CL.exe`.
We can use the primitive flag `_DEBUG`, because flutter doesn't provide debug flag in cpp level.

bitfriend:
Linux provides `NDEBUG` instead of `_DEBUG`, that has opposite meaning.
https://stackoverflow.com/questions/2290509/debug-vs-ndebug

bitfriend:
I also implemented macos code, but couldn't test it.
@gnunicorn Can you please test that code on your machine?

bitfriend:
This is the screenshot of linux app on my side.
image

PR #2394, opened 2024-12-02
External Linking facilities

gnunicorn:
Uses the new [link share generator service](https://github.com/acterglobal/synapse-super-invites/pull/8) (deployed on m-1 already) to generate external links. Also provides the facilities to generate the internal links.

Implemented for:
 - Pin
 - TaskList
 - CalendarEvent

Also exposes a new `ref_details()`-generator feature on each of them (which is used internally), which can be used as a system-wide reference object to attach to other objects (e.g. boosts) directly. Which can also be used as a new attachment type on all attachment-able objects (see generated api). 

Brings end-2-end-tests and includes the super invites branch in the tests already


native/test/src/tests/news.rs
```diff
@@ -862,3 +862,66 @@ async fn multi_news_read_receipt_test() -> Result<()> {
 
     Ok(())
 }
+
+// #[tokio::test]
```
gtalha07, line 866: This doesn't work yet ? Then we might have it removed ?

native/test/src/tests/thumbnail.rs
```diff
@@ -518,7 +518,7 @@ async fn image_attachment_can_support_thumbnail() -> Result<()> {
     assert_eq!(attachment.event_id(), attachment_id);
     assert_eq!(attachment.type_str(), "image");
 
-    let msg_content = attachment.msg_content();
+    let msg_content = attachment.msg_content().unwrap();
```
gtalha07, line 521: is this safe? Better give ```.expect()``` here for tests easier debugging?

gtalha07 (approved):
Looks good to me. Just some minor queries at your disposal :)

native/test/src/tests/thumbnail.rs
gnunicorn: in tests we get full tracebacks including line-numbers, `unwrap` is fine here.

PR #2396, opened 2024-12-04
fix flutter unit tests (chat input)

gtalha07:


app/lib/features/chat/providers/chat_providers.dart
```diff
@@ -280,9 +280,9 @@ final allowSendInputProvider = StateProvider.family.autoDispose<bool, String>(
 );
 
 // whether user has enough permissions to send message in room
-final canSendMessageProvider = FutureProvider.family<bool, String>(
+final canSendMessageProvider = FutureProvider.family<bool?, String>(
   (ref, roomId) async {
     final membership = ref.watch(roomMembershipProvider(roomId));
-    return membership.valueOrNull?.canString('CanSendChatMessages') == true;
```
gtalha07, line 286: This was the one source of tests failing as it seemed our input states logic is dependent on nullable for loading instance. So I had to revert it... (came from Feedback review change)

app/lib/features/chat/widgets/custom_input.dart
```diff
@@ -68,25 +69,29 @@ class CustomChatInput extends ConsumerWidget {
...
-              size: 14,
-              color: Theme.of(context).unselectedWidgetColor,
-            ),
-            const SizedBox(width: 4),
-            Text(
-              key: noAccessKey,
-              L10n.of(context).chatMissingPermissionsToSend,
-              style: TextStyle(
-                color: Theme.of(context).unselectedWidgetColor,
+        child: Padding(
```
gtalha07, line 74: Also it seemed necessary to move padding inner tree here to avoid overflow...

bitfriend (approved):
Looks good to me

app/lib/features/chat/widgets/custom_input.dart
gtalha07: I've seen cases where visually the effects look different comparably. Don't know how it's so but it is what it is...

PR #2397, opened 2024-12-04
Enforce singleton in release only

gnunicorn:
superseeds #2393  .

This allows Windows and Linux developers to run the dev-version _next to_ an installed Acter with its own secret-store identity and not being blocked by their single-window problem.

The core problem on linux is the ``SingleMainWindow=true` flag in the `.desktop`-file. If acter is installed and an  app starts with that same identity, the desktop will switch to the first. This patch now overwrites the identity with a `-dev`-suffixed one to get around the issue, as long as this is a debug build. Thus the desktop environment understoods them as different apps and allows them to be run simultaneously.

### DIFF
--- a/app/windows/runner/main.cpp
+++ b/app/windows/runner/main.cpp
@@ -40,9 +40,15 @@ int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
   FlutterWindow window(project);
   Win32Window::Point origin(10, 10);
   Win32Window::Size size(1280, 720);
+#ifdef _DEBUG
+  if (!window.CreateAndShow(L"Acter-dev", origin, size)) {
+    return EXIT_FAILURE;
+  }
+#else
   if (!window.CreateAndShow(L"Acter", origin, size)) {
     return EXIT_FAILURE;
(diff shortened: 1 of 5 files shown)


app/windows/runner/main.cpp
```diff
@@ -8,8 +8,11 @@
 
 int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
                       _In_ wchar_t *command_line, _In_ int show_command) {
-
+#ifdef _DEBUG
+  HWND hwnd = ::FindWindow(L"FLUTTER_RUNNER_WIN32_WINDOW", L"Acter-dev");
```
bitfriend, line 12: In windows, which stage in compilation will deal with `Acter-dev`?

gnunicorn: To make sure we use a different window system and don't block the installed version.

bitfriend: Then windows and linux will detect release/debug flag using the different logic?

gnunicorn: this is the windows one, yes.

PR #2398, opened 2024-12-04
App availability check for external share options

kumarpalsinh25:


app/lib/common/providers/app_install_check_provider.dart
```diff
@@ -0,0 +1,31 @@
...
+      ExternalApps.signal => await checkAppAvailability(
+          Platform.isAndroid ? 'org.thoughtcrime.securesms' : 'sgnl://',
+        ),
+    };
+  }, // this means we are running
+);
+
+Future<bool> checkAppAvailability(String package) async {
+  try {
+    final app = await AppCheck().checkAvailability(package);
```
gnunicorn, line 26: Why not use [`isAppInstalled(package)`](https://pub.dev/documentation/appcheck/latest/appcheck/AppCheck/isAppInstalled.html)?

app/lib/common/providers/app_install_check_provider.dart
```diff
@@ -0,0 +1,31 @@
...
+    return app != null;
+  } catch (e) {
```
gnunicorn, line 28: logging the error? What kind of error are we ever expecting here?

app/lib/common/widgets/share/widgets/external_share_options.dart
```diff
@@ -23,9 +25,16 @@ class ExternalShareOptions extends StatelessWidget {
...
+  Widget build(BuildContext context, WidgetRef ref) {
     final lang = L10n.of(context);
 
+    final isWhatsAppInstalled =
+        ref.watch(isAppInstalledProvider(ExternalApps.whatsApp)).valueOrNull;
+    final isTelegramInstalled =
+        ref.watch(isAppInstalledProvider(ExternalApps.telegram)).valueOrNull;
+    final isSignalInstalled =
+        ref.watch(isAppInstalledProvider(ExternalApps.signal)).valueOrNull;
+
```
gnunicorn, line 37: I'd move that `== true` check up here... just makes more sense to me ...

gnunicorn (approved):

app/lib/common/providers/app_install_check_provider.dart
kumarpalsinh25: Ahh, good notice! Let me use it.

app/lib/common/providers/app_install_check_provider.dart
kumarpalsinh25: If app is not install then `checkAvailability` function will give `PlatformException` so managed in this way.

Screenshot 2024-12-05 at 11 33 36 AM

But anyway, as per your suggestion, let me use `isAppInstalled` which is basically doing the same.
Screenshot 2024-12-05 at 11 34 58 AM

app/lib/common/widgets/share/widgets/external_share_options.dart
kumarpalsinh25: Sure, let me change it accordingly.

PR #2410, opened 2024-12-05
Caching of provider information for fewer UI updates

gnunicorn:
when not necessary.

Starting out at trying to get the `bookmark`ing of a space to not make that page refresh all its items anymore. Brings several improvements:

 - [x] a new EnumList type that allows comparing lists of enums - which on regular `List<Enum>` will never be `==` even if they are the same items inside
 - [x] a new caching provider wrapper that listens to the inner provider and if upon reloading/refreshing it comes up with the same (as in `==`-equal) result is happening, _doesn't_ update the inner value, thus breaking the update-chain accurately. 
 - [ ] a small break off the an unrelated display-name loading in the tabs-providing UI so that it won't refresh unless it has to


gnunicorn:
Abandoning approach as no real improvement can be shown.

PR #2411, opened 2024-12-07
Wrap receiver in broadcast so that receiver can be called multiple times

bitfriend:
Will replace `Mutext<Option<Receiver<...>>>` with `BroadcastStream<Receiver<...>>`.

### DIFF
--- a/app/lib/features/cross_signing/widgets/cross_signing.dart
+++ b/app/lib/features/cross_signing/widgets/cross_signing.dart
@@ -21,9 +21,7 @@ final _log = Logger('a3::cross_signing::widget');
 // it just pops up stage dialogs for verification
 @immutable
 class CrossSigning extends ConsumerStatefulWidget {
-  final Client client;
-
-  const CrossSigning({super.key, required this.client});
+  const CrossSigning({super.key});
 
   @override
   ConsumerState<ConsumerStatefulWidget> createState() => CrossSigningState();
@@ -114,7 +112,7 @@ class CrossSigningState extends ConsumerState<CrossSigning> {
       builder: (BuildContext context) => RequestCreatedView(
         onCancel: (BuildContext context) async {
           // cancel verification request launched by this device
-          await event.cancelVerificationRequest(widget.client);
+          await event.cancelVerificationRequest();
         },
       ),
       isDismissible: false,
@@ -140,13 +138,11 @@ class CrossSigningState extends ConsumerState<CrossSigning> {
       builder: (BuildContext context) => RequestReadyView(
         isVerifier: isVerifier,
         onCancel: (BuildContext context) async {
-          // occurs request.cancelled
-          await event.cancelVerificationRequest(widget.client);
+          await event.cancelVerificationRequest(); // occurs request.cancelled
         },
         onAccept: (BuildContext context) async {
           // start sas verification from this device
-          // occurs request.transitioned
-          await event.startSasVerification(widget.client);
+          await event.startSasVerification(); // occurs request.transitioned
         },
       ),
       isDismissible: false,
@@ -232,11 +228,11 @@ class CrossSigningState extends ConsumerState<CrossSigning> {
         sender: event.sender(),
         onCancel: (BuildContext context) async {
           // cancel verification request from other device
-          await event.cancelVerificationRequest(widget.client);
+          await event.cancelVerificationRequest();
         },
         onAccept: (BuildContext context) async {
           // accept verification request from other device
-          await event.acceptVerificationRequest(widget.client);
+          await event.acceptVerificationRequest();
         },
       ),
       isDismissible: false,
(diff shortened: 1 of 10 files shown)


native/acter/src/api/verification.rs
```diff
@@ -23,46 +20,37 @@ use matrix_sdk_base::ruma::{
...
-    controller: VerificationController,
     event_type: String,
     flow_id: String,
-    sender: OwnedUserId,
+    sender: String,
     /// event content
     content: HashMap<String, String>,
     /// emoji array
     emojis: Vec<VerificationEmoji>,
 }
```
bitfriend, line 46: `impl Stream<Item = VerificationEvent>` needs for `VerificationEvent` to derive `Default` trait.
I removed `client` and `controller` fields from `VerificationEvent`, because they didn't derive `Default`.

gnunicorn: this seems wrong. why does it need to be `Default`?

bitfriend: It needs `Default` trait.
image

app/lib/features/cross_signing/widgets/cross_signing.dart
```diff
@@ -19,8 +19,11 @@ final _log = Logger('a3::cross_signing::widget');
 
 // this widget has no elements
 // it just pops up stage dialogs for verification
+@immutable
 class CrossSigning extends ConsumerStatefulWidget {
-  const CrossSigning({super.key});
+  final Client client;
+
+  const CrossSigning({super.key, required this.client});
```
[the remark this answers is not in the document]
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2411/commits/2e2689406697bbf8c3a46f50fe2bafc7a7d4789b

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2411/commits/2e2689406697bbf8c3a46f50fe2bafc7a7d4789b

native/acter/src/api/verification.rs
```diff
@@ -1077,11 +1127,9 @@ impl SessionManager {
...
 
 impl Client {
-    pub fn verification_event_rx(&self) -> Option<Receiver<VerificationEvent>> {
-        match self.verification_controller.event_rx.try_lock() {
-            Ok(mut r) => r.take(),
-            Err(e) => None,
-        }
+    pub fn verification_event_rx(&self) -> impl Stream<Item = OptionVerificationEvent> {
+        BroadcastStream::new(self.verification_controller.event_rx.resubscribe())
+            .map(|o| OptionVerificationEvent::new(o.ok()))
```
gnunicorn, line 1132: Do I understand correctly that this entire new type and the wrapping and all is due to that event sometimes being `None`? Wouldn't `.filter_map` do enough for us here?

```suggestion
    pub fn verification_event_rx(&self) -> impl Stream<Item = VerificationEvent> {
        BroadcastStream::new(self.verification_controller.event_rx.resubscribe())
            .filter_map(|o| o.ok())
```

gnunicorn (changes requested):
I think this can be simplified further.

native/acter/src/api/verification.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2411/commits/8ff8191aad386d987fffd4a938a9d636fe9efbcc

native/acter/src/api/verification.rs
```diff
@@ -1077,11 +1113,11 @@ impl SessionManager {
 }
 
 impl Client {
-    pub fn verification_event_rx(&self) -> Option<Receiver<VerificationEvent>> {
-        match self.verification_controller.event_rx.try_lock() {
-            Ok(mut r) => r.take(),
-            Err(e) => None,
-        }
+    // this return value should be unpinned when wait_for_verification_event ends, because it calls pin_mut
+    // this return value should be wrapped in Box::pin, to make unpin possible
```
gnunicorn, line 1117:
```suggestion
```

not necessary doc comments

gnunicorn (approved):

native/acter/src/api/verification.rs
bitfriend: Frankly, I don't want to forget this logic.
If I leave this comment, I can remember that later.

native/acter/src/api/device.rs
```diff
@@ -98,18 +98,16 @@ impl DeviceController {
...
 
 impl Client {
-    pub fn device_event_rx(&self) -> Option<Receiver<DeviceEvent>> {
-        match self.device_controller.event_rx.try_lock() {
-            Ok(mut r) => r.take(),
-            Err(e) => None,
-        }
+    pub fn device_event_rx(&self) -> impl Stream<Item = DeviceEvent> {
+        BroadcastStream::new(self.device_controller.event_rx.resubscribe())
+            .map(|o| o.unwrap_or_default())
```
gnunicorn, line 110: this should _not_ create a default even when there is none, this is wrong. Please rewrite this to also use `filter_map`

gnunicorn (approved):
Just one more minor thing I missed earlier.

native/acter/src/api/device.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2411/commits/65f1e301d3f52063f81cebb6fa60450eade7ebcc

PR #2412, opened 2024-12-08
Simplify custom event handler into single callback

gnunicorn:
All so much simpler this way. Test's are passing \o/ .


native/core/src/events.rs
```diff
@@ -207,11 +232,238 @@ impl<'de> serde::Deserialize<'de> for AnyActerEvent {
...
+                    tasks::TaskEventContent::TYPE,
+                    tasks::TaskUpdateEventContent::TYPE,
+                    tasks::TaskSelfAssignEventContent::TYPE,
+                    tasks::TaskSelfUnassignEventContent::TYPE,
+                    comments::CommentEventContent::TYPE,
+                    comments::CommentUpdateEventContent::TYPE,
+                    attachments::AttachmentEventContent::TYPE,
+                    attachments::AttachmentUpdateEventContent::TYPE,
+                    rsvp::RsvpEventContent::TYPE,
+                    reaction::ReactionEventContent::TYPE,
```
bitfriend, line 459: `read_receipt::ReadReceiptEventContent::TYPE` seems to be missed.

native/core/src/events.rs
```diff
@@ -207,11 +232,238 @@ impl<'de> serde::Deserialize<'de> for AnyActerEvent {
                             attachments::AttachmentEventContent::TYPE,
                             attachments::AttachmentUpdateEventContent::TYPE,
                             rsvp::RsvpEventContent::TYPE,
-                            ReactionEventContent::TYPE,
+                            reaction::ReactionEventContent::TYPE,
```
bitfriend, line 235: `read_receipt::ReadReceiptEventContent::TYPE` seems to be missed.

native/core/src/events.rs
gnunicorn:
```suggestion
                    read_receipt::ReadReceiptEventContent::TYPE,
                    reaction::ReactionEventContent::TYPE,
```

Issue #2423, opened 2024-12-10
keyboard covers space selection tab


emilvincentz:

emilvincentz:
![Image](https://github.com/user-attachments/assets/b6a961b2-168e-48fc-8b78-5b21052993de)

emilvincentz:
@kumarpalsinh25 a little ui fix:)

Issue #2422, opened 2024-12-09
fixes and improvements


emilvincentz:
- [ ] "create space" button no longer shows on home (even though user has less than 5 spaces)
- [ ] People dont understand that you can click on "My spaces" (we need to make it more "clickable" or add "see all")
- [ ] Add "Add Space" and "Add Chat" to quick actions

emilvincentz:
@kumarpalsinh25 are you able to take a look a these ones? 
Users have a hard time finding the way to create new spaces, e.g. below

![Image](https://github.com/user-attachments/assets/165300db-e6e0-45ad-863f-f90de4c65617)

PR #2424, opened 2024-12-11
General Fixes

kumarpalsinh25:
1) Issue: Fixes, https://github.com/acterglobal/a3/issues/2422
Commit: https://github.com/acterglobal/a3/pull/2424/commits/41429773f5700a676bc44e4020eaebb337c92abc

Reference:

https://github.com/user-attachments/assets/daed7742-48a7-46ec-93f9-aa8dba4a1e95

2) Issue: Fixes, https://github.com/acterglobal/a3/issues/2423
Commit: https://github.com/acterglobal/a3/pull/2424/commits/6caad8af34378a4d0ea282f169f32ee4303eb164

Reference:

https://github.com/user-attachments/assets/c17019c5-96c5-4bdf-b230-8c2431a1bdca

### DIFF
--- a/app/lib/common/widgets/spaces/space_selector_drawer.dart
+++ b/app/lib/common/widgets/spaces/space_selector_drawer.dart
@@ -16,13 +16,17 @@ Future<String?> selectSpaceDrawer({
     enableDrag: true,
     context: context,
     isDismissible: true,
-    builder: (context) => SelectRoomDrawer(
-      key: key,
-      canCheck: canCheck,
-      currentSpaceId: currentSpaceId,
-      title: title ?? Text(L10n.of(context).selectSpace),
-      keyPrefix: 'select-space',
-      roomType: RoomType.space,
+    isScrollControlled: true,
+    builder: (context) => Container(
+      constraints: BoxConstraints(maxHeight: 700),
+      child: SelectRoomDrawer(
+        key: key,
+        canCheck: canCheck,
+        currentSpaceId: currentSpaceId,
+        title: title ?? Text(L10n.of(context).selectSpace),
+        keyPrefix: 'select-space',
+        roomType: RoomType.space,
+      ),
     ),
   );
   if (selected == null) {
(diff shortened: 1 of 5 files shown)


app/lib/common/widgets/spaces/space_selector_drawer.dart
```diff
@@ -16,13 +16,17 @@ Future<String?> selectSpaceDrawer({
...
-    builder: (context) => SelectRoomDrawer(
-      key: key,
-      canCheck: canCheck,
-      currentSpaceId: currentSpaceId,
-      title: title ?? Text(L10n.of(context).selectSpace),
-      keyPrefix: 'select-space',
-      roomType: RoomType.space,
+    isScrollControlled: true,
+    builder: (context) => Container(
+      constraints: BoxConstraints(maxHeight: 700),
```
gtalha07, line 21: Shouldn't we use ```MediaQuery``` here to determine fixed size ? The pixels might calculate differently based on device ratio...

app/lib/features/main/widgets/bottom_navigation_widget.dart
```diff
@@ -41,7 +41,7 @@ class BottomNavigationWidget extends ConsumerWidget {
                 bottomNavNar(ref),
                 AnimatedContainer(
                   duration: const Duration(milliseconds: 300),
-                  height: showQuickActions ? 180 : 0,
+                  height: showQuickActions ? 250 : 0,
```
gtalha07, line 44: Also having this hardcoded means it might not be as responsive as we'd like it to be...

gtalha07 (approved):
I have some doubts about how widget sizing would come in effect based on device ratio but overall looks good!

app/lib/features/main/widgets/bottom_navigation_widget.dart
kumarpalsinh25: This is known thing to manage slide-up and slide-down animation. And checked this in Android, iOS and Desktop platforms with this height value.

PR #2425, opened 2024-12-11
Space Object : Internal sharing expansion

kumarpalsinh25:
Space objects can be share internally with other objects.

For example,
 - You can share Pin with Other Pin, Event or TaskList.
- You can share Event with Other Event, Pin or TaskList.
- You can share TaskList with Other TaskList,Pin or Event.

Space access Management: 
There is no restriction of within space boundary during object sharing. You can share space object to any other space's object.

For example,
- User-1 have Space-A which contain Pin-A object and Space-B which contains Pin-B.
- User-2 have access to Space-B only.
- User-1 shares Pin-A with Pin-B.
- Both users can see Pin-A as Attachment Ref in Pin-B's details screen. However User1 can also go to Pin-A details screen and access all the details but User-2 can only see the static details as he doesn't have access to Space-A.

Video Reference:

https://github.com/user-attachments/assets/ff0311b8-3cc4-4f59-abcd-83c67937b558


app/lib/common/widgets/share/action/share_space_object_action.dart
```diff
@@ -167,4 +219,25 @@ class ShareSpaceObjectActionUI extends StatelessWidget {
...
+  }) async {
+    final objectId = objectDetails.objectId;
+    switch (objectDetails.objectType) {
+      case ObjectType.pin:
+        final sourcePin = await ref.watch(pinProvider(objectId).future);
+        return await sourcePin.refDetails();
+      case ObjectType.calendarEvent:
+        final sourcePin =
+            await ref.watch(calendarEventProvider(objectId).future);
+        return await sourcePin.refDetails();
```
gnunicorn, line 235: not a pin.
```suggestion
        final source =
            await ref.watch(calendarEventProvider(objectId).future);
        return await source.refDetails();
```

app/lib/common/widgets/share/action/share_space_object_action.dart
```diff
@@ -167,4 +219,25 @@ class ShareSpaceObjectActionUI extends StatelessWidget {
...
+      case ObjectType.taskList:
+        final sourcePin = await ref.watch(taskListProvider(objectId).future);
+        return await sourcePin.refDetails();
```
gnunicorn, line 238:
```suggestion
        final source = await ref.watch(taskListProvider(objectId).future);
        return await source.refDetails();
```

not a pin.

these are clearly copy paste, another indicator that this API probably doesn't need this match...

app/lib/features/attachments/widgets/reference_attachment_item.dart
```diff
@@ -25,74 +35,178 @@ class ReferenceAttachmentItem extends ConsumerWidget {
...
       ),
     );
   }
 
-  Widget title(BuildContext context) {
-    final title = refDetails.title();
-    return Column(
-      crossAxisAlignment: CrossAxisAlignment.start,
+  Widget refAttachmentTitle(WidgetRef ref) {
+    final defaultTitle = Text(refDetails.title().toString());
```
gnunicorn, line 55: can still technically be `null` and should be dealt with... 
```suggestion
    final defaultTitle = Text(refDetails.title() ?? L10n.of(context).unknown);
```

 (or something)

app/lib/features/attachments/widgets/reference_attachment_item.dart
```diff
@@ -25,74 +35,178 @@ class ReferenceAttachmentItem extends ConsumerWidget {
...
+    }
+  }
+
+  void onTapRefAttachment(
+    BuildContext context,
+    WidgetRef ref,
+  ) {
+    final lang = L10n.of(context);
+    final refObjectType = refDetails.typeStr();
+    final refObjectId = refDetails.targetIdStr();
```
gnunicorn, line 127: having several functions all call the exact same details feels wrong and in particular that they then have matches on them...

I'd rather say, let's make separate Widgets for each type (`pin`, `calendar-event`, etc) and have one global match that selects the right widget and passes the ref-details over..

makes it also a lot neater and easier to extend than extending several functions with their internal match ...

gnunicorn (approved):
Improvements possible :)

app/lib/common/widgets/share/action/share_space_object_action.dart
```diff
@@ -167,4 +219,25 @@ class ShareSpaceObjectActionUI extends StatelessWidget {
       _ => null,
     };
   }
+
+  Future<RefDetails?> getRefDetails({
+    required WidgetRef ref,
+    required SpaceObjectDetails objectDetails,
+  }) async {
+    final objectId = objectDetails.objectId;
+    switch (objectDetails.objectType) {
```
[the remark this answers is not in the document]
kumarpalsinh25: No, this require here.
This is for getting `RedDetails` of source object whereas `onTapPin` is for destination object.

`objectDetails.objectType == ObjectType.calendarEvent` mean you are in Calendar Details screen and this code will get you `RefDetails` of calendar object.

Whereas `onTapPin` define destination Pin Object in which you want to put Calendar source object.

I hope this will clarify the above code. Still happy to elaborate more if needed.

app/lib/common/widgets/share/action/share_space_object_action.dart
kumarpalsinh25: `another indicator that this API probably doesn't need this match..`

Sorry not getting you on this, can you explain more on this?

app/lib/features/attachments/widgets/reference_attachment_item.dart
kumarpalsinh25: Yes agree with you and also felt the same during this code development. Tried couple of way but then reverted it to minimised this PR Complexity. 

Surely will do this in next follow-up PR.

app/lib/common/widgets/share/action/share_space_object_action.dart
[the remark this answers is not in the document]
gnunicorn: I understand what the API does / how it works _right_ now. I am arguing that this should be changed because it is wrong and leads to unnecessarily complicated code like this.

Simply put. Why is `onTapPin: Function(ObjectDetails { ObjectType.pin, ActerPin })` and not simply `onTapPin : Function(ActerPin)`? This seems like an unnecessary attempt of generalizing something that is _always_ specific...

And that more specific API would then allow to simplify this code to : 
```
onTapPin: (pin) async => await attachRefDetailToPin(
            context: context,
            ref: ref,
            refDetails: await pin.refDetails())
```

I'd rather do these smaller refactors that show an earlier approach can be improved massively sooner rather than later. As with every use the API gets more entrenched in the way it is even if that means more complicated code around it. and changing it later becomes harder.

If you still don't understand what I mean, let me know and I push a commit refactoring it.

app/lib/common/widgets/share/action/share_space_object_action.dart
gnunicorn: this is in reference to the open discussion above.

PR #2426, opened 2024-12-11
chat-ng messages support types ```m.emote``` and ```m.location```

gtalha07:
- Brings improvements in events widget code implementation .i.e.  meta information will grow larger so having it combined in map just like in old code instead of passing boolean singletons.
- Add support for `m.location` and `m.emote` type render. 
- Emoji detection function ```isOnlyEmojis``` is updated to include extended modern emojis support and also with unit tests!

## PREVIEW:

Screenshot 2024-12-11 at 18 19 26


gnunicorn:
@gtalha07 I don't understand the screenshot. what am I seeing there or should I notice? Please elaborate

gtalha07:
> @gtalha07 I don't understand the screenshot. what am I seeing there or should I notice? Please elaborate

It is showcasing how emoji's only text are rendered now in different formats of inputs (along with tests supporting use cases). The method which is used to calculate ```isOnlyEmojis``` was broken and didn't seem to enlarge emojis which is fixed and also extending regex to support more emojis.

app/lib/features/chat/utils.dart
```diff
@@ -405,3 +405,35 @@ Future<void> onMessageLinkTap(
...
+      emoji.input.substring(emoji.start, emoji.end),
+      '',
+    );
+  }
+
+  // remove all whitespace
+  remainingText = remainingText.replaceAll(RegExp(r'\s'), '');
+
+  // return true if nothing else left
+  return remainingText.isEmpty;
```
gnunicorn, line 438: this is quite an excessive way of figuring this out... why not have a regexp containing a space in the group and make sure it matches the entire string?

I know this isn't technically code you wrote, but it seems something obvious to improve...

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -106,12 +108,15 @@ class ChatBubble extends StatelessWidget {
...
-                  .copyWith(fontSize: 12),
+          Align(
+            alignment: Alignment(0.9, 0.0),
+            child: Visibility(
+              visible: wasEdited,
+              child: Text(
+                L10n.of(context).edited,
+                style: chatTheme.emptyChatPlaceholderTextStyle
+                    .copyWith(fontSize: 12),
+              ),
```
gnunicorn, line 119: Unnecessary wrapping in `Visibility` ...

```suggestion
          if (wasEdited)
          Align(
            alignment: Alignment(0.9, 0.0),
            child: Text(
              L10n.of(context).edited,
              style: chatTheme.emptyChatPlaceholderTextStyle
                  .copyWith(fontSize: 12),
```

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -94,20 +96,29 @@ class ChatEventItem extends StatelessWidget {
     String messageId,
     RoomEventItem item,
     bool isUser,
-    bool showAvatar,
+    bool nextMessageGroup,
```
gnunicorn, line 99: what is this flag supposed to indicate? it doesn't make sense to me from its name...

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -94,20 +96,29 @@ class ChatEventItem extends StatelessWidget {
...
+      'm.emote' ||
+      'm.notice' ||
+      'm.server_notice' ||
+      'm.text' =>
+        TextMessageEvent(
           roomId: roomId,
           content: content,
-          isUser: isUser,
-          nextMessageGroup: showAvatar,
+          metadata: metadata,
```
gnunicorn, line 121: why not just have them as proper parameters? `Map<String, dynamic>` are hazaradous and easy to get wrong in later version. I'd rather not have `dynamic` passed around if we can avoid it.

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -124,6 +135,7 @@ class ChatEventItem extends StatelessWidget {
           messageId: messageId,
           content: content,
         ),
+      'm.location' => LocationMessageEvent(content: content),
```
gnunicorn, line 138: I think I mentioned before that I prefer to _not_ include support for items if the support is bad. This doesn't seem like a good way to support this. I'd rather skip supporting `m.location` and show a "not supported" message than a badly supported one...

app/lib/features/chat_ng/widgets/events/location_message_event.dart
```diff
@@ -0,0 +1,22 @@
...
+
+class LocationMessageEvent extends StatelessWidget {
+  final MsgContent content;
+  const LocationMessageEvent({super.key, required this.content});
+
+  @override
+  Widget build(BuildContext context) {
+    final body = content.body();
+    final geoUri = content.geoUri();
+    return Container(
```
gnunicorn, line 12: this seems pointless. As said above, let's not support this type.

app/lib/features/chat_ng/widgets/events/text_message_event.dart
```diff
@@ -8,23 +11,47 @@ import 'package:flutter_matrix_html/text_parser.dart';
 class TextMessageEvent extends StatelessWidget {
   final String roomId;
   final MsgContent content;
-  final bool isUser;
-  final bool nextMessageGroup;
-  final bool wasEdited;
+  final Map<String, dynamic> metadata;
```
gnunicorn, line 14: why swap nicely typed specific fields that our compiler can check for us with a dynamic map of undefined stuff that then goes bad at runtime? I'd rather like to see _more_ typing and less...

app/lib/features/chat_ng/widgets/events/text_message_event.dart
```diff
@@ -8,23 +11,47 @@ import 'package:flutter_matrix_html/text_parser.dart';
...
+    final textTheme = Theme.of(context).textTheme;
+    final chatTheme = Theme.of(context).chatTheme;
+    final colorScheme = Theme.of(context).colorScheme;
     final body = content.formattedBody() ?? content.body();
 
+    String? msgType = metadata['msgType'];
+    bool isUser = metadata['isUser'];
+    bool isNotice = (msgType == 'm.notice' || msgType == 'm.server_notice');
+    bool nextMessageGroup = metadata['nextMessageGroup'];
+    bool wasEdited = metadata['wasEdited'];
```
gnunicorn, line 34: as stated above. let's just make that proper typed params to put in. there are several here that if removed from the sender would lead to _runtime_ errors...

app/test/features/chat_ng/messages/chat_message_test.dart
```diff
@@ -105,4 +106,45 @@ void main() {
       });
     });
   });
+
+  group('emoji only detection unit tests', () {
```
gnunicorn, line 110: yay tests!

gnunicorn (approved):
I don't want to badly support types, let's not add this `m.location` support. And let's make that map for a dynamic proper types please.


Last but not least - up to you to change it - the isOnlyEmoji function clearly can be improved upon.

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
gtalha07: This is a check whether the next message comes as group from same user. The flag term is taken from ```flutter_chat_ui``` => [isNextMessageInGroup](https://pub.dev/documentation/flutter_chat_ui/latest/flutter_chat_ui/Message/bubbleBuilder.html)

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
gtalha07: Yeah I think we can have separate model for that which can contain these information. Because this parameters count will get very large as it grows so I did it as clean way. But it's not type-safe! 👍🏻

app/lib/features/chat_ng/widgets/events/text_message_event.dart
gtalha07: Yeah as stated above, will incorporate in a class model.

PR #2430, opened 2024-12-12
Chat-NG reply messages preview

gtalha07:
build on top of #2426 as it brings core structural improvements to messages UI, this includes reply messages preview along with proper automatic provider updates on ```repliedTo``` message ids whenever valid reference gets available. Also supports nice error UI handling.

### showcases the __replied to content__ UI and notice how the original event gets available on server fetch when paginating:

https://github.com/user-attachments/assets/e3782aa2-59a6-473c-bf42-f62566715383


app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat::reply_notifier');
+
+class ReplyMessageNotifier
+    extends AutoDisposeFamilyAsyncNotifier<ReplyMsgState, ReplyMsgInfo> {
+  Future<void> retryLoad() async {
+    state = const AsyncValue.loading();
+    state = await AsyncValue.guard(() => build(arg));
```
gnunicorn, line 13: guard crates a loading around the inner fn ...
```suggestion
    state = AsyncValue.guard(() => build(arg));
```

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
```diff
@@ -0,0 +1,66 @@
...
+      chatStateProvider(arg.roomId).select((value) => value.messageList),
+    );
+
+    // First try to find message in current state
+    final messageId = messages.firstWhere(
+      (eventId) => eventId == arg.originalId,
+      orElse: () => '',
+    );
+
+    if (messageId.isNotEmpty) {
```
gnunicorn, line 29: Please use types to indicate state, not the content if it ;)
```suggestion
    final String? messageId = messages.firstWhere(
      (eventId) => eventId == arg.originalId,
      orElse: () => null,
    );

    if (messageId != null) {
```

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
```diff
@@ -0,0 +1,66 @@
...
+      RoomMsgId msgInfo = (arg.roomId, arg.originalId);
+      final message = ref.watch(chatRoomMessageProvider(msgInfo));
```
gnunicorn, line 31: I don't understand why we do anything of the above if we then just go do this ref.watch ... what is the stuff above useful for that this ref.watch wouldn't cover? Can't we move whatever is needed into the ref.watch or an adjacent provider instead?

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
```diff
@@ -0,0 +1,66 @@
...
+      final convo = await ref.watch(chatProvider(arg.roomId).future);
+      if (convo == null) {
+        _log.severe(
+          'Failed to load room ${arg.roomId})',
+          null,
+          null,
+        );
+        return ReplyMsgError(arg.originalId, arg.messageId, null, null);
+      }
+      final timeline = convo.timelineStream();
```
gnunicorn, line 54: don't we have a `timelineProvider`? We should, because this should be cached and not create a new timeline object for every reply we have to load...

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -1,40 +1,40 @@
...
+  final MessageMetadata metadata;
   final BoxDecoration decoration;
   final CrossAxisAlignment bubbleAlignment;
 
   // default private constructor
   const ChatBubble._inner({
     super.key,
     required this.child,
-    required this.wasEdited,
+    required this.metadata,
```
gnunicorn, line 19: I feel like the chat-bubble widget is getting too much intelligence and builds up into another mega widget that does a lot of things depending on several inputs...

I think rather than have it pack in a lot information that is unpacked again, I'd rather it to have a `final List<Widgets>? children` and have the outer renderer be _smart_ about what should go in ...

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -30,36 +30,42 @@ class ChatEventItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final eventType = item.eventType();
+    final metadata = MessageMetadata(
```
gnunicorn, line 33: As said above, I think this place should decide _directly_ which children to render, rather than have that intelligence moved into the chat-bubble.

app/lib/features/chat_ng/widgets/events/reply_original_event.dart
```diff
@@ -0,0 +1,168 @@
...
+    RoomEventItem item,
+    bool isUser,
+  ) {
+    final msgType = item.msgType();
+    final content = item.msgContent();
+    final metadata = MessageMetadata(
+      roomId: roomId,
+      messageId: messageId,
+      isUser: isUser,
+      isReply: true,
```
gnunicorn, line 132: this is such an important flag which indicating _how_ the redering is supposed to happen, yet it is bundled into a `metadata`-object that is about the messages. The messages indicated here is _not_ a reply, it is a message to be replied to - and that flag only matters for the rendering, this isn't _metadata_ but _rendering information_. We should really differentiate between the concerns. This _flag_ should really be a `TextMessagesEvent.noBubble`-factory I think...

app/lib/features/chat_ng/widgets/reply_preview.dart
```diff
@@ -0,0 +1,72 @@
...
+        metadata.repliedTo.expect('should always contain replied id');
+    final ReplyMsgInfo replyInfo =
+        (roomId: roomId, messageId: messageId, originalId: originalId);
+
+    final roomMsgState = ref.watch(replyToMsgProvider(replyInfo)).valueOrNull;
+
+    return switch (roomMsgState) {
+      ReplyMsgLoading() => replyBuilder(
+          context,
+          Center(child: CircularProgressIndicator()),
```
gnunicorn, line 31: please let's not use any circular progress indicators going forward. They make for a messy and constantly switching UI. Let's use skeletons!

app/lib/features/chat_ng/widgets/reply_preview.dart
```diff
@@ -0,0 +1,72 @@
+import 'package:acter/features/chat_ng/models/message_metadata.dart';
+import 'package:acter/features/chat_ng/models/reply_message_state.dart';
+import 'package:acter/features/chat_ng/providers/chat_room_messages_provider.dart';
+import 'package:acter/features/chat_ng/widgets/events/reply_original_event.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:acter/common/extensions/options.dart';
+
+// Reply State UI widget
+class ReplyPreview extends ConsumerWidget {
```
gnunicorn, line 10: But this isn't a ReplyPreview, this is a `RepliedToPreview`, isn't it?

gnunicorn (changes requested):
UI/UX wise looks good but I think we can do better code wise. In particular around two aspects of the way this is being implemented: a) better separation of concerns - it's not metadata if it is purely about the way the same object is rendered differently depending on context (I marked that situation)
b) I think the _bubble_ should not become too intelligent, but be a dump widget that other higher level objects use to render their widgets (or not). For reusability we can have common code (e.g. a Reply-To-Widget-Builder) in a (abstract) widget that others mix in or can inherit from. I think that is better, cleaner separation and leads to bigger reusability than tweaking a "metadata" object you pass around to maybe show a bubble or not...

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
gtalha07: I can conclude based on your below comment that this isn't needed at all. My thinking was to have that message fetched already if it's available in current state but since we providers are cached, this doesn't seem to have any noticeable difference in rendering speed, even if removed.

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
gtalha07: yes, only ```timelineStreamProvider``` is sufficient enough.

app/lib/features/chat_ng/providers/notifiers/reply_messages_notifier.dart
gtalha07: Yes, we have too many providers to know which objects are actually cached and forgot. Thanks for pointing that out!

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -17,7 +19,7 @@ const _supportedTypes = [
   'm.room.encrypted',
 ];
 
-typedef RoomMsgId = (String roomId, String uniqueId);
+typedef RoomMsgId = ({String roomId, String uniqueId});
```
gnunicorn, line 22: :+1:

app/lib/features/chat_ng/widgets/events/reply_original_event.dart
```diff
@@ -0,0 +1,160 @@
...
+    required this.msg,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    // reply is allowed for only EventItem not VirtualItem
+    // user should be able to get original event as RoomMessage
+    RoomEventItem replyEventItem =
+        msg.eventItem().expect('room msg should have event item');
+    final messageId = msg.uniqueId();
```
gnunicorn, line 32: if so, why not directly ask for both of these as the input param rather than only `msg`?

app/lib/features/chat_ng/widgets/events/reply_original_event.dart
```diff
@@ -0,0 +1,160 @@
...
+    required this.messageId,
+    required this.item,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final myId = ref.watch(myUserIdStrProvider);
+    final isUser = myId == item.sender();
+    final eventType = item.eventType();
+    return switch (eventType) {
```
gnunicorn, line 105: :pray:

app/lib/features/chat_ng/widgets/reply_preview.dart
gnunicorn, line 1: filename also needs updating...

app/lib/features/chat_ng/widgets/reply_preview.dart
```diff
@@ -0,0 +1,76 @@
...
+
+    final roomMsgState = ref.watch(replyToMsgProvider(replyInfo)).valueOrNull;
+    final avatarInfo = AvatarInfo(uniqueId: '#');
+
+    return switch (roomMsgState) {
+      ReplyMsgLoading() => replyBuilder(
+          context,
+          Skeletonizer(
+            child: ListTile(
+              leading: ActerAvatar(options: AvatarOptions.DM(avatarInfo)),
```
gnunicorn, line 33: only needed for the loading case, no reason to generate it otherwise.
```suggestion
    return switch (roomMsgState) {
      ReplyMsgLoading() => replyBuilder(
          context,
          Skeletonizer(
            child: ListTile(
              leading: ActerAvatar(options: AvatarOptions.DM(AvatarInfo(uniqueId: '#'))),
```

app/lib/features/chat_ng/widgets/reply_preview.dart
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  Widget replyErrorUI(
+    BuildContext context,
+    WidgetRef ref,
+    RoomMsgId replyInfo,
+  ) {
+    final originalId = replyInfo.uniqueId;
+    return Text(
+      'Failed to load original message id: $originalId',
```
gnunicorn, line 55: l10n please

gnunicorn (approved):
nice!

Issue #2149, opened 2024-09-04
[tracking] Verification NG 


gnunicorn:
tracking everything that is important for the next iteration of the verification system:

- https://github.com/acterglobal/a3/issues/2148
- https://github.com/acterglobal/a3/issues/1240

PR #2445, opened 2024-12-19
Switch verification from event stream-based to timeout-based

bitfriend:
Will implement [next generation of verification](https://github.com/acterglobal/a3/issues/2149).


bitfriend:
Will keep `SessionVerificationController` and `VerificationController` in parallel.
When `SessionVerificationController` is ready, will remove `VerificationController`.

Issue #2453, opened 2024-12-19
Users can't find Chat Invitations


gnunicorn:
The Activity's Stream isn't clear enough in its purpose for most users (on on boarding). While we try to improve that by building out the [Activities Stream](https://github.com/acterglobal/a3/issues/2224) we might need to consider a shorter term solution. There are a few things proposed:

- Make the activities-dot on the Icon on the bottom (and sidebar) at lot stronger in size and/or color to draw more attention
- Add a block for invitations on the top of the chat area as well
- Add a block about invitations on the main dashboard
- Add "you have important activities"-block on the main dashboard's top.

gnunicorn:
> Add a block for invitations on the top of the chat area as well

While this sounds to be simpliest and easiest fix (in line with what the users want), I want us to be carefully consider the consequences and potential draw backs of this:
1. Do we show it invites for Chats _and_ Spaces? If we don't do it for Spaces, won't they still have the same problem? If we do it for spaces, and they accept the Invite from there then Nothing changes on that screen, causing confusion.
2. We draw the attention away from the activities center and even more attention to the busy chat area - the entire point of the notifications center was to move the other interaction points _out of there_
3. Due to 2, we'd just delay the learning of what the activities center is about for the user, creating a potential gap in terms of them understanding its importance.

The Dashboard solutions have _some_ of the same draw backs but aren't as much of a problem case, as they don't tie into idea the user already has, but thus they aren't addressing the problem that well either.

gnunicorn:
My current thinking is to do a combination of the ideas but all of them _nudging_ the user towards the notification center. What I am thinking of is:

1. a stronger indicator on the icon when there are invitations
2. an Invitations sub-screen
3. a box on the top of the dashboard for the Activities center, giving a summary and an easy call to action. Like `You have 3 Invitations pending` or `There are 2 unauthorized Sessions` (though we'd focus on the invitations at first). On the first, we'd directly open the invitations sub screen and let the user deal with it from there
4. We'd have a similar small messages on the top of the chat rooms listing if there are open invitations, clearly indicating to find them in the Activities Center. Like `[ActiviesIcon] You have 2 pending Invitations` and upon click directly open said sub-page. Allow the user to dismiss this message for the session / until the counter changes.

gnunicorn:
considered done for now with #2455 .

PR #2455, opened 2024-12-19
Make Invitations more obvious and cleaner

gnunicorn:
Fixing #2453 by:

1. Marking "open invitations" as an urgent-style badge
2. brighten up the red color dot on the Activities icon to make it more clear
3. Creating a Invites Overview Page and Overview Widget with clean and nice layout
4. use that Widget in the Dashboard to highlight open Invitations, link directly to the sub-page
6. Improve the UI/UX flow of the Activities Screen & Invite Card, specifically:
  - Use Section Headers
  - Put Invites on top
  - Drop Slivers, use regular list
  - Use Summary Widget if more than 1 invitation
  - Have invites load the local Avatar
  - Replace the Decline Button with an InlineText to make the flow nicer for the eye

See it in Action:

https://github.com/user-attachments/assets/eadce463-83af-4994-ac6d-63028bdb4254

_Notes to the Reviewer_:
The Features has been moved from the Activities Feature into its own Features, but the main widgets haven't changed much, nor have the providers and notifiers - they only moved. The Main changes on the Activity Page itself are about removing the Slivers and replacing the order of sections.

_Notes to Product_:
This should vastly improve visibility of invitations without compromising going into the hairy details of expecting them to show up in chat screen and its drawbacks. If people still don't find them we can consider adding the same pending widget to the chat room list screen, but I'd like to avoid that if possible. Let's try it with what we have here and see what people say.


kumarpalsinh25 (approved):
Very nice implementation! 👏

PR #2463, opened 2024-12-26
Refactor the dependency versions

bitfriend:
Use the dependency version that other packages are already using


gnunicorn:
we agreed to not do it this way.

PR #2440, opened 2024-12-17
RefDetail enhancement on Space Objects

kumarpalsinh25:
THIS PR GETS BIT LARGER THEN EXPECTED BUT I WILL TRY TO EXPLAIN ALL THE DETAIL IN PR DESCRIPTION AND COMMENTS TO MAKE IT EASIER FOR REVIEWERS. ENTIRE PR WORK IS WELL SEPARATED BY COMMITS BUT IT REQUIRES BIT OF IN-DEPTH DETAIL WHICH I TRIED TO COVER AS MUCH AS POSSIBLE IN PR DESCRIPTION.

- - -
### Boost Actions Enhancement: Make it beyond selected space boundry.
- Currently boost actions are only belongs to selected space in which we are going to post Boost but it is now enhanced and user can post any space action items to space boost.

**1) Changes related to Add Boost:**
- First I have removed space selection restriction while adding actions on the boost by below commit.
https://github.com/acterglobal/a3/pull/2440/commits/86ccfd88b20a23f5d61ac5de88960bc97a7e31e5
- Then I have removed custom `NewsReferencesModel` class  and use default `RefDetails` class in relevant code of Boosts by below commit.
https://github.com/acterglobal/a3/pull/2440/commits/e9cc2a0ed409d75a8ebe7096af80f22d500b4284
- Reference Video :

https://github.com/user-attachments/assets/2cb7edc8-3d92-457e-bb1a-d1be21ccb112

**2) Changes related to View Boost:**
- Now on view part, Boost Actions (Space Objects) on which we have access (Space we already joined) are visible as expected.
- But Boost Actions (Space Objects), on which we do not have any access (Space we haven't joined) are not visible directly so for that I created common `ReferenceDetailsItem` class which shows static data based on the `refDetails`.
- To handle it in more better way, I have used `ReferenceDetailsItem` in `PinListItemWidget`, `EventItem` and in `TaskListItemCard` based on the required conditions.
- `NewsSlideActions` for boost post is now becomes more clear and can directly used `PinListItemWidget`, `EventItem` and in `TaskListItemCard` classes without working about anything else.
- All the work is implemented in below commit.
https://github.com/acterglobal/a3/pull/2440/commits/92c37c15f5ec58e073f139b65b75feec4488308c
- Reference Video:

https://github.com/user-attachments/assets/048aae2c-3ff3-4988-865e-d6975e3ccd47

- - -

### Attachment List Improvements:
[...]

PR #2464, opened 2024-12-27
Add reference

kumarpalsinh25:
As continuation of [PR - RefDetail enhancement on Space Objects](https://github.com/acterglobal/a3/pull/2440), This PR allow you to directly add reference in similar way we are adding attachments.

https://github.com/user-attachments/assets/ac5313f3-6a99-4103-b5fd-d2f8c1886bec

### DIFF
--- a/app/lib/features/attachments/widgets/attachment_section.dart
+++ b/app/lib/features/attachments/widgets/attachment_section.dart
@@ -1,5 +1,6 @@
 import 'package:acter/common/models/types.dart';
 import 'package:acter/common/toolkit/buttons/inline_text_button.dart';
+import 'package:acter/common/widgets/share/action/share_space_object_action.dart';
 import 'package:acter/features/attachments/actions/handle_selected_attachments.dart';
 import 'package:acter/features/attachments/actions/select_attachment.dart';
 import 'package:acter/features/attachments/providers/attachment_providers.dart';
@@ -22,9 +23,11 @@ class AttachmentSectionWidget extends ConsumerWidget {
   static const addAttachmentBtnKey = Key('attachments-add-btn');
   static const confirmRedactKey = Key('attachments-confirm-redact');
   final AttachmentsManagerProvider? manager;
+  final SpaceObjectDetails spaceObjectDetails;
 
   const AttachmentSectionWidget({
     super.key,
+    required this.spaceObjectDetails,
     this.manager,
   });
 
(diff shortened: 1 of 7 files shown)


app/lib/features/tasks/pages/task_list_details_page.dart
```diff
@@ -162,7 +162,7 @@ class _TaskListPageState extends ConsumerState<TaskListDetailPage> {
   Widget _buildTaskListInner(TaskList? taskListData) {
     return SingleChildScrollView(
       child: Padding(
-        padding: const EdgeInsets.symmetric(horizontal: 20.0),
+        padding: const EdgeInsets.symmetric(horizontal: 8.0),
```
gnunicorn, line 165: what is this about? I didn't see this in the promo video...?

app/lib/features/attachments/widgets/attachment_section.dart
```diff
@@ -187,40 +219,20 @@ class FoundAttachmentSectionWidget extends ConsumerWidget {
...
-              );
-            },
+        Expanded(
+          child: Text(
+            title,
+            style: Theme.of(context).textTheme.titleSmall,
           ),
-          child: Text(lang.add),
         ),
+        IconButton(onPressed: onTapAdd, icon: Icon(Icons.add)),
```
gnunicorn, line 235: shouldn't the ability to this be conditioned by `canEdit`? Because if they can't edit the item, we don't want to show the button...

gnunicorn (approved):
Generally fine, but two requests.

also wondering if we could have tests for this ...

app/lib/features/tasks/pages/task_list_details_page.dart
kumarpalsinh25: Ahh, I was play with TaskList Details padding-spacing improvements. Here with [this commit](https://github.com/acterglobal/a3/pull/2464/commits/de1c7fa4d7f9148702d8dede7acb7b3da647ebc2) I completed improvements. Also sharing screenshot for reference.

TaskList

app/lib/features/attachments/widgets/attachment_section.dart
kumarpalsinh25: Ya true, let me add it. Good notice!

kumarpalsinh25:
> Generally fine, but two requests.

- Addressed given feedback and added comments over there.

> also wondering if we could have tests for this ...

For attachment section?

gnunicorn:
well, generally have tests, @kumarpalsinh25 - e.g. that canEdit is observed etc ...

kumarpalsinh25:
> well, generally have tests, @kumarpalsinh25 - e.g. that canEdit is observed etc ...

Facing some difficulty here, would like to discuss about this on our next call.

Issue #2137, opened 2024-09-03
[Tracking] Chat-NG


gnunicorn:
Collecting issues to keep in mind when starting on the next generation of the chat interface - in no particular order.

- https://github.com/acterglobal/a3-meta/issues/463
- https://github.com/acterglobal/a3/issues/1907
- https://github.com/acterglobal/a3-meta/issues/491 -> Probably due to state error failures that are hidden from view
- https://github.com/acterglobal/a3-meta/issues/459
- https://github.com/acterglobal/a3-meta/issues/458 -> Probably due to mistake in list management (too many replaceAt and a race condition)
- https://github.com/acterglobal/a3/issues/1614 -> most comes from rust, but there is a bunch of balony we do in the architecture that delays it further as well.

emilvincentz:
Really great to see the ng coming together!!
Here is a couple of bugs observed.
1. When enter chat on mobile keyboard is always open (and not closing on scroll)
2. Grey box is shown when using text formmatting or pasting in url link 
3. Flicker on desktop when navigatin beetween chats

https://github.com/user-attachments/assets/94faa627-2d00-4d0a-95b9-1434e13f9ff0
![Image](https://github.com/user-attachments/assets/19015576-333a-46bc-a46d-08e3fbf57bbd)
![Image](https://github.com/user-attachments/assets/41bc436a-13b5-4fe7-b6cc-0094b008a486)

emilvincentz:
@gtalha07 some feedback from testing latest release (today)

1. Doesnt scroll to the bottom of the newest message
2. Text format bar sometimes doesnt show or is shown below menu (see image)
3. Not able to drag and select (when wanting to highlight/select text in input field)
4. Tesxt size seems rather small
5. When activating text format’er, all text dissapears
6. How much are we able to edit the text editor?  E.g. Dont think we need the background colour change
7. Keyboard does not close upon scrolling
8. Keyboard automatically activaed upon entering chat (should not)
9. Not able to react to messages 
10 Line spacing should be the same when posting, as when writing (currently its almost same spacing although i made new paragraph, see image)


![Image](https://github.com/user-attachments/assets/72e6c42d-9bbc-4809-aeed-efe6d417947b)
![Image](https://github.com/user-attachments/assets/13623052-3bba-4a40-8d30-214e4f9ba68e)
![Image](https://github.com/user-attachments/assets/70d5ffde-d59c-4a49-bdf1-ed9745cd712a)
[...]

PR #2498, opened 2025-01-15
Chat-NG Reactions

### LINKED ISSUE #2204: chat react-emoji UX
reported by @emilvincentz :

- Make chat reactions bigger (can be hard to press the right emojii
    image
image
    
- Consider making chat text and emojii reaction indicators slighly bigger
    
image

gtalha07:
# Description:
- Brings out fresh new emoji selector row UX along with implementing reactions view on messages in Chat-NG.
- Most of the backend logic is same, the only new changes are on the emoji-row UX side (frost/blurred effect) when long pressing the message. I have pointed that out in change-set. 
- With these new changes, also fixes #2204 as a part of tracking issue #2137.
- Moved `Edited` indicator in the message bubble body.

### Video showcasing reactions view (on long-press message):
https://github.com/user-attachments/assets/b5182df2-cce9-4b06-9125-09a406be9e86

### Emoji Selector view on messages:
Screenshot 2025-01-16 at 18 30 52

Screenshot 2025-01-16 at 18 33 24

### Reaction UI on messages:
Screenshot 2025-01-16 at 18 34 03

Screenshot 2025-01-16 at 18 35 25

### Detailed reaction view (on long press reaction UI) Updated:

Screenshot 2025-01-17 at 18 22 37


app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -113,19 +119,21 @@ class ChatBubble extends StatelessWidget {
                     const SizedBox(height: 10),
                   ],
                   child,
+                  if (isEdited) ...[
```
gtalha07, line 122: moved ```Edited``` indicator inside bubble for better UX.

app/lib/features/chat_ng/widgets/reactions/reaction_chips_widget.dart
```diff
@@ -0,0 +1,106 @@
+import 'package:acter/common/extensions/acter_build_context.dart';
+import 'package:acter/common/themes/app_theme.dart';
+import 'package:acter/features/chat_ng/providers/chat_room_messages_provider.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
+    show ReactionRecord;
+import 'package:flutter/material.dart';
+
+class ReactionChipsWidget extends StatelessWidget {
```
gtalha07, line 8: This part is logically same and pasted from legacy but more linear and structured.

app/lib/features/chat_ng/widgets/reactions/reaction_selector_row.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:acter/common/widgets/emoji_picker_widget.dart';
+import 'package:acter/features/chat/providers/chat_providers.dart';
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat::reactions_selector_row');
+
+class ReactionSelectorRow extends ConsumerWidget {
```
gtalha07, line 33: This is same as ```EmojiRow``` widget logic-wise but have been more degraded for simplicity and structured.

gtalha07:
> Just from the UI I notice the following:
> 
> 1. the user-counter is flickering when you click an emoji in the window. Do we know why? That seems like something we want to fix (if not this in this PR then at least record it so we do it later).
> 2. why not show display name & avatar for the users in the emoji overview dialog?

1. It was from memberIdsProvider ```.when```:
```
 membersLoader.when(
              data: (members) => Text(
                lang.membersCount(members.length),
                style: Theme.of(context).textTheme.bodySmall,
              ),
              skipLoadingOnReload: true,  --> previous was false
              error: (e, s) {
                _log.severe('Failed to load active members', e, s);
                return Text(lang.errorLoadingMembersCount(e));
              },
              loading: () => Skeletonizer(
                child: Text(lang.membersCount(100)),
              ),
            ),
```

2. Emoji overview dialog inner UI is unchanged and been copied as it is from before. I can probably put that here now since we have more refined avatar providers.

app/lib/features/chat_ng/widgets/reactions/reactions_list.dart
```diff
@@ -0,0 +1,59 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat::reactions_list');
+
+class ReactionsList extends ConsumerWidget {
+  final String roomId;
+  final String messageId;
+  final RoomEventItem item;
+  final bool isNextMessageInGroup;
```
kumarpalsinh25, line 17: What is `isNextMessageInGroup` and where it is used in this class?

app/lib/features/chat_ng/widgets/reactions/reaction_selector.dart
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) 2024, Acter Global, (c) 2022 Simform Solutions
```
kumarpalsinh25, line 2: `2022 Simform Solutions` ?

app/lib/features/chat_ng/widgets/reactions/reaction_selector.dart
```diff
@@ -0,0 +1,153 @@
...
+    String uniqueId,
+    String emoji,
+  ) async {
+    try {
+      final stream = await ref.read(timelineStreamProvider(roomId).future);
+      await stream.toggleReaction(uniqueId, emoji);
+    } catch (e, s) {
+      _log.severe('Reaction toggle failed', e, s);
+    }
+  }
```
kumarpalsinh25, line 133: How about putting this in action folder as this also used in other class.

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,204 @@
...
+          emojis: userReactions[userId] ?? [],
+        );
+      },
+      separatorBuilder: (context, index) => const SizedBox(height: 12),
+    );
+  }
+}
+
+class ReactionUserItem extends StatelessWidget {
+  final String roomId;
```
kumarpalsinh25, line 179: What is the need of `roomId` here?

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,204 @@
...
+    );
+  }
+
+  ({
+    List<String> allUsers,
+    Map<String, List<String>> userReactions,
+    Map<String, List<String>> reactionUsers,
+  }) _processReactionData() {
+    final userReactions = <String, List<String>>{};
+    final reactionUsers = <String, List<String>>{};
```
kumarpalsinh25, line 120: `userReactions` and `reactionUsers` wording looks confusing.

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,204 @@
...
+        (a, b) => (userReactions[b]?.length ?? 0)
+            .compareTo(userReactions[a]?.length ?? 0),
+      );
+
+    return (
+      allUsers: allUsers,
+      userReactions: userReactions,
+      reactionUsers: reactionUsers,
+    );
+  }
```
kumarpalsinh25, line 144: This way of `_processReactionData` looks confusing to me.

kumarpalsinh25 (commented):
Taking bit of time to review and understand the things. Added few comments related to suggestion/queries.

Here are some more UI Feedback.

1) Reaction on PDF file:  Reaction on this file look off.
Screenshot 2025-01-17 at 4 50 23 PM



2) Reaction Details : I don't see which tab is selected. Also as @gnunicorn mentioned it will be good if we show User avatar and Display name here. And overall UI-UX can be improve for this dialog.
Screenshot 2025-01-17 at 4 51 41 PM





3) Reaction Chip Design : Chip design looks different for send and receiver messages. Also ReactionList overlaying/position and overall design can be improve by taking references of well-known social media app.
Screenshot 2025-01-17 at 4 51 31 PM

app/lib/features/chat_ng/actions/reaction_selection_action.dart
```diff
@@ -0,0 +1,120 @@
...
+
+import 'package:acter/features/chat_ng/widgets/reactions/reaction_selector.dart';
+import 'package:flutter/material.dart';
+
+// reaction selector action on chat message
+void reactionSelectionAction({
+  required BuildContext context,
+  required Offset position,
+  required Widget messageWidget,
+  required bool isUser,
```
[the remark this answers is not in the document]
gnunicorn: I think this is meant to be `isAuthor` or `isMe` - probably should be renamed as such.

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -35,11 +37,56 @@ class ChatEventItem extends StatelessWidget {
 
     return switch (eventType) {
       // handle message inner types separately
-      'm.room.message' => buildMsgEventItem(
-          context,
-          roomId,
-          messageId,
-          item,
+      'm.room.message' => Column(
```
gnunicorn, line 40: I think this is too much of a tree inside a `switch`, probably better to keep it in a function or better even as a separate widget.

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,221 @@
...
+
+  void _initializeTabs() {
+    final total = widget.reactions.fold<int>(
+      0,
+      (sum, reaction) => sum + reaction.$2.length,
+    );
+
+    _tabs = [
+      Tab(child: Chip(label: Text('All $total'))),
```
gnunicorn, line 41: translations support please

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,221 @@
...
+      ...widget.reactions.map(
+        (reaction) => Tab(
+          child: Chip(
+            avatar: Text(reaction.$1, style: EmojiConfig.emojiTextStyle),
+            label: Text('${reaction.$2.length}'),
```
gnunicorn, line 46: isn't this faster?
```suggestion
            label: Text(reaction.$2.length.toString()),
```

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
gnunicorn: I think this is meant as a bi-directional mapping for the main to show the list sorted by users and their emojis, and then per emoji-tap to just show the list of users.

the naming isn't great and it could use a doc comment to explain it, but I think this isn't actually new code but just existing code copied over...

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,221 @@
...
+                userReactions: reactionData.userReactions,
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  ({
```
gnunicorn, line 114: comment below said ... this is a great case for splitting the function into a utility function instead and have proper tests for it to make sure it works and doesn't break in the future.

While "copying over", I think we _do_ want to improve the code quality and in particular the code test coverage in chat NG as well.

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,221 @@
...
+        borderRadius: BorderRadius.all(Radius.circular(12)),
+      ),
+      indicatorSize: TabBarIndicatorSize.tab,
+      dividerColor: Colors.transparent,
+      tabs: _tabs,
+    );
+  }
+
+  Widget _buildTabBarView() {
+    final reactionData = _processReactionData();
```
gnunicorn, line 88: I don't understand why we rebuild this every time, even though we only rebuild the tabs on init... we should probably have them in sync and done at the same time.

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
```diff
@@ -0,0 +1,221 @@
...
+class _ReactionDetailsSheetState extends ConsumerState<ReactionDetailsSheet>
+    with TickerProviderStateMixin {
+  late TabController _tabController;
+  late List<Tab> _tabs;
+
+  @override
+  void initState() {
+    super.initState();
+    _initializeTabs();
+  }
```
gnunicorn, line 32: I understand that the tabs controller needs a stateful widget and the crap, _but_ then we also need to rebuild that list upon didWidgetChange, no? Otherwise we have new emojis coming in and show them in the render, but don't have the reactions in the tabs ...?!?

app/lib/features/chat_ng/widgets/reactions/reaction_selector.dart
gnunicorn: agreed. what is this about, where does it come from?

app/lib/features/chat_ng/widgets/reactions/reaction_chips_widget.dart
gnunicorn, line 1: You nicely split the reactions up into their own subfolder ... but I wonder: are these _chat_ specific? We do (later) want users to be able to react to all types of things (pins, comments, tasks and tasklists), doesn't it make sense to put these out into their own _features_ folder already?

app/lib/features/chat_ng/widgets/reactions/reaction_selector.dart
gnunicorn: agreed, should be in the actions folder.

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -139,3 +141,18 @@ final repliedToMsgProvider = AsyncNotifierProvider.autoDispose
     .family<RepliedToMessageNotifier, RepliedToMsgState, RoomMsgId>(() {
   return RepliedToMessageNotifier();
 });
+
+final messageReactionsProvider = StateProvider.autoDispose
```
gnunicorn, line 145: I wonder if that should be per _item_ object or rather be bound to the roomId+eventId combination ... not sure yet.

gnunicorn (approved):
Generally looks fine to me, but there's a few things to be addressed.

app/lib/features/chat_ng/actions/reaction_selection_action.dart
```diff
@@ -0,0 +1,120 @@
...
+  required String roomId,
+  required String messageId,
+}) {
+  final RenderBox box = context.findRenderObject() as RenderBox;
+  final Offset position = box.localToGlobal(Offset.zero);
```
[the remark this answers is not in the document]
gtalha07: Yeah I realized it's not necessary. 👍🏻!

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -35,11 +37,56 @@ class ChatEventItem extends StatelessWidget {
...
+                );
+              },
+              child: Hero(
+                tag: messageId,
+                child: buildMsgEventItem(
+                  context,
+                  roomId,
+                  messageId,
+                  item,
+                ),
```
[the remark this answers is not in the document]
gtalha07: on upper side you mean out of ```switch``` statement? But this is based on a case when we have message event, otherwise we would be creating widget for all other event cases regardless which seems unnecessary and not needed.

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -35,11 +37,56 @@ class ChatEventItem extends StatelessWidget {
...
+              ),
+            ),
+            // reactions of message
+            Container(
+              padding: EdgeInsets.only(
+                right: isUser ? 12 : 0,
+                left: isUser ? 0 : 12,
```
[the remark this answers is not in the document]
gtalha07: will change it to suggested by ben

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -35,11 +37,56 @@ class ChatEventItem extends StatelessWidget {
...
+              ),
+              child: FractionalTranslation(
```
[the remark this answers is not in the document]
gtalha07: this gives the reaction UI widget on message a 'stacked effect'. This could be done with more known ```Stack``` widget also but it seems to have problems with gestures.

app/lib/features/chat_ng/widgets/reactions/reactions_list.dart
gtalha07: Sorry, this isn't needed here and might be unintentionally placed while copy-pasting! Thanks for notifying. 👍🏻

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
gtalha07: This is a [record pattern](https://dart.dev/language/records) introduced in dart 3.0 for more functional and concise programming. While it's still new and might not be grasping enough coming from traditional ways, I think it's much better than enclosing attributes class-like and saves boilerplate ;)

app/lib/features/chat_ng/widgets/reactions/reaction_detail_sheet.dart
gtalha07: yeah, logic-wise it's same but I agree with better naming and explanation.

app/lib/features/chat_ng/widgets/reactions/reaction_chips_widget.dart
gtalha07: > You nicely split the reactions up into their own subfolder ... but I wonder: are these _chat_ specific? We do (later) want users to be able to react to all types of things (pins, comments, tasks and tasklists), doesn't it make sense to put these out into their own _features_ folder already?

So, reaction toggling is operated through ```timelineStreamProvider``` which is from ```Convo``` object. Do we have support of reaction API on other room objects? If so, I can maybe change this action a bit and separate it neatly as singleton.

gnunicorn: keep it for the future, @gtalha07 . leave it as is.

Issue #1977, opened 2024-07-24
froze on last session on macos


emilvincentz:
Version:          1.24.7231 (1721760909)

gnunicorn:
it appears we do get freeze bug reports in sentry, too. I think this is the relevant one: https://acter.sentry.io/issues/3169691/?project=4507612337012816&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&stream_index=0

gnunicorn:
@emilvincentz can you describe the conditions and what you were doing? I can see in the logs that the app lost focus for like 2 seconds ("was backgrounded") but not much else. Did you switch to another app and come back and then it froze? did you copy-paste anything over? 

Asking because there seem to be various flutter-caused freeze issues and I am trying to track down if it is any of them...

emilvincentz:
hmm cant remember exactly, but feel it was like going from one tab to another. from chat to home or from chat to search.
but im not 100% sure

emilvincentz:
@gnunicorn just froze again. This time it happened when I went from one chat and clicked to go into another chat

PR #2508, opened 2025-01-20
Reduce `NoClientException`-Problems

gnunicorn:
### The user problem:

we have seen several reports of the app "freezing" or indicating to the user that the client was logged out, when it fact, it was not.

### The diagnosis:
These were most often seen on Apple devices. Looking through the logs we find the curious case of
```
[2025-01-19][18:53:02.077422][acter::api::auth][INFO] Successfully logged in user @XXXX:acter.global, device Some("YYYYYYY") with token.
[2025-01-19][18:53:02.312033][a3::calendar_sync][INFO] Previous key found ASSASS-2212-898-B3A1-DF113B2A71A6 null null null
[2025-01-19][18:53:02.312498][a3::calendar_sync][INFO] Existing calendar found 6D0008DD-2212-4176-B3A1-DF113B2A71A6 null null null
[2025-01-19][18:53:02.312808][a3::calendar_sync][INFO] ignoring state change without value null null null
[2025-01-19][18:53:02.580610][a3::home::client_notifier][ERROR] platform dispatch error Instance of 'NoClientException' null #1      pid: 13446, tid: 6109229056, name io.flutter.1.ui (unparsed:0:0)
```
Yes, that's right, we see a `NoClientException` despite just four lines above it states that the user successfully logged in. How can that be?

Well, we have been wrapping the async `acterSdk.instance`-call (which trigger that first line of logging) inside a sync provider that will raise said exception if no client is found. This can happen if the providers actual value is looked up _quicker_ than the async code actually setting the internal value. Classic race condition because we are not properly `await`ing the code that issues that first log entry.

As you might now, several parts of the application _assume_ there to be a client, and if they are sent to a screen that causes a `NoClientException` to occur, they behave differently: sometimes a proper error is shown (if that is wrapped in another provider), sometimes you are redirected to the login screen (if you happen to come across the auth guard while the data is not yet available) and often times, when the provider is used directly, we have an exception right in the render thread, which leads to flutter just rendering a box of nothing - which is easily understood as a "app freeze" (despite it not actually, technically, being a freeze).

### The fix 

The fix is simple but substantial: we need to make the `clientProvider` and it's parent the `alwaysClientProvider` async and await that initialization phase. Simple enough. _Unfortunately_ a lot of other providers directly or indirectly link to that provider and thus also need updating. And thus a cascade of refactoring was necessary.

#### Specific

This happens in a few specific parts that I commented in the code. But here is the gist:
 1. make the Client provider async and have it await for the initialization before handing over a client
 2. Make all those `StateNotifier`s that rely on it regular `Notifier`s or `AsyncNotifiers` that `listen` to the client being updated internally rather than assuming a state notifier that is already available (as this is through `ref.listen` make sure they are set to `fireImmediately: true` or you won't see the first draw)
 3. Follow the cascade and replace all other instances of Providers accordingly, simplify cases were possible
 4. Some minor drive-by updates on top:
   - removed support for `isGuest` as we don't actually use that anywhere anyhow
   - have the main app read these providers to preload that data: dca2436c75ea1d454b344ee5ecd5573a61e9f47a
   - some minor refactoring in the notification forward handler related to the same problem: ba180d0316edc1fe958a299bf3c10940bf18a057

Fixes https://github.com/acterglobal/a3-meta/issues/714, https://github.com/acterglobal/a3-meta/issues/715 https://github.com/acterglobal/a3-meta/issues/411 https://github.com/acterglobal/a3/issues/1977

### DIFF
--- a/app/lib/common/providers/common_providers.dart
+++ b/app/lib/common/providers/common_providers.dart
@@ -13,7 +13,7 @@ final _log = Logger('a3::common::common_providers');
 
 final genericUpdatesStream =
     StreamProvider.family<int, String>((ref, key) async* {
-  final client = ref.watch(alwaysClientProvider);
+  final client = await ref.watch(alwaysClientProvider.future);
   int counter = 0; // to ensure the value updates
 
   // ignore: unused_local_variable
@@ -24,14 +24,14 @@ final genericUpdatesStream =
 });
 
 final myUserIdStrProvider = Provider(
-  (ref) => ref.watch(
-    alwaysClientProvider.select((client) => client.userId().toString()),
-  ),
+  (ref) =>
+      ref.watch(alwaysClientProvider).valueOrNull?.userId().toString() ??
+      '@acter:acter.global',
 );
 
-final accountProvider = Provider(
+final accountProvider = FutureProvider(
   (ref) => ref.watch(
-    alwaysClientProvider.select((client) => client.account()),
+    alwaysClientProvider.selectAsync((client) => client.account()),
   ),
 );
 
@@ -39,12 +39,9 @@ final hasFirstSyncedProvider = Provider(
   (ref) => ref.watch(syncStateProvider.select((v) => !v.initialSync)),
 );
 
-final isGuestProvider =
-    Provider((ref) => ref.watch(alwaysClientProvider).isGuest());
-
-final deviceIdProvider = Provider(
+final deviceIdProvider = FutureProvider(
   (ref) => ref.watch(
-    alwaysClientProvider.select(
+    alwaysClientProvider.selectAsync(
       (v) => v.deviceId().toString(),
     ),
   ),
@@ -52,8 +49,7 @@ final deviceIdProvider = Provider(
 
 /// Gives [AvatarInfo] object for user account. Stays up-to-date internally.
 final accountAvatarInfoProvider = StateProvider.autoDispose<AvatarInfo>((ref) {
-  final account = ref.watch(accountProvider);
-  final userId = account.userId().toString();
+  final userId = ref.watch(myUserIdStrProvider);
 
   final displayName = ref.watch(accountDisplayNameProvider).valueOrNull;
   final avatar = ref.watch(_accountAvatarProvider).valueOrNull;
@@ -74,14 +70,14 @@ final accountAvatarInfoProvider = StateProvider.autoDispose<AvatarInfo>((ref) {
 /// Caching the name of each Room
 final accountDisplayNameProvider =
     FutureProvider.autoDispose<String?>((ref) async {
-  final account = ref.watch(accountProvider);
+  final account = await ref.watch(accountProvider.future);
   return (await account.displayName()).text();
 });
 
 final _accountAvatarProvider =
     FutureProvider.autoDispose<MemoryImage?>((ref) async {
   final sdk = await ref.watch(sdkProvider.future);
-  final account = ref.watch(accountProvider);
+  final account = await ref.watch(accountProvider.future);
   final thumbSize = sdk.api.newThumbSize(48, 48);
   final avatar = await account.avatar(thumbSize);
   // Only call data() once as it will consume the value and any subsequent
@@ -100,7 +96,7 @@ class EmailAddresses {
 }
 
 final emailAddressesProvider = FutureProvider((ref) async {
-  final account = ref.watch(accountProvider);
+  final account = await ref.watch(accountProvider.future);
   // ensure we are updated if the upgrade comes down the wire.
   ref.watch(genericUpdatesStream('global.acter.dev.three_pid'));
   final confirmed = asDartStringList(await account.confirmedEmailAddresses());
--- a/app/lib/common/providers/notifiers/chat_notifiers.dart
+++ b/app/lib/common/providers/notifiers/chat_notifiers.dart
@@ -19,7 +19,7 @@ class AsyncConvoNotifier extends FamilyAsyncNotifier<Convo?, String> {
   @override
   FutureOr<Convo?> build(String arg) async {
     final roomId = arg;
-    final client = ref.watch(alwaysClientProvider);
+    final client = await ref.watch(alwaysClientProvider.future);
     _listener = client.subscribeStream(roomId); // keep it resident in memory
     _poller = _listener.listen(
       (data) async {
@@ -50,7 +50,7 @@ class AsyncLatestMsgNotifier extends FamilyAsyncNotifier<RoomMessage?, String> {
   @override
   FutureOr<RoomMessage?> build(String arg) async {
     final roomId = arg;
-    final client = ref.watch(alwaysClientProvider);
+    final client = await ref.watch(alwaysClientProvider.future);
     _listener = client.subscribeStream('$roomId::latest_message');
     _poller = _listener.listen(
       (data) async {
@@ -70,20 +70,15 @@ class AsyncLatestMsgNotifier extends FamilyAsyncNotifier<RoomMessage?, String> {
   }
 }
 
-class ChatRoomsListNotifier extends StateNotifier<List<Convo>> {
-  final Client client;
+class ChatRoomsListNotifier extends Notifier<List<Convo>> {
   late Stream<ConvoDiff> _listener;
-  late StreamSubscription<ConvoDiff> _poller;
+  StreamSubscription<ConvoDiff>? _poller;
+  late ProviderSubscription _providerSubscription;
 
-  ChatRoomsListNotifier({
-    required Ref ref,
-    required this.client,
-  }) : super(List<Convo>.empty(growable: false)) {
-    _init(ref);
-  }
-
-  void _init(Ref ref) {
+  void _reset(Client client) {
+    state = List<Convo>.empty(growable: false);
     _listener = client.convosStream(); // keep it resident in memory
+    _poller?.cancel();
     _poller = _listener.listen(
       _handleDiff,
       onError: (e, s) {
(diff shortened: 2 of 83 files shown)


app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -204,6 +199,24 @@ class ChatRoomsListNotifier extends StateNotifier<List<Convo>> {
...
+        final client = newVal.valueOrNull;
+        if (client == null) {
+          // we don't care for not having a proper client yet
+          return;
+        }
+        _reset(client);
+      },
+      fireImmediately: true,
+    );
+    ref.onDispose(() => _providerSubscription.close());
```
gnunicorn, line 217: main pattern used to `listen` to the client now from within the other `Notifiers`. Logic otherwise is mainly the same, just some more better cleaning up procedures added.

app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -70,20 +70,15 @@ class AsyncLatestMsgNotifier extends FamilyAsyncNotifier<RoomMessage?, String> {
   }
 }
 
-class ChatRoomsListNotifier extends StateNotifier<List<Convo>> {
-  final Client client;
+class ChatRoomsListNotifier extends Notifier<List<Convo>> {
```
gnunicorn, line 73: Notifier already has `ref` given and without `Client` passed in I mad most `StateNotifiers` to simple `Notifiers` instead.

app/lib/common/providers/notifiers/client_pref_notifier.dart
```diff
@@ -7,40 +7,36 @@
...
-    prefs = await sharedPrefs();
-    final newValue = (prefs.get(prefKey) as T? ?? defaultValue);
-    state = newValue;
-  }
+  String get prefKey => '$deviceId-$prefKeySuffix';
+  T defaultValue;
 
   /// Updates the value asynchronously.
-  Future<void> update(T value) async {
+  Future<void> set(T value) async {
```
gnunicorn, line 39: rename due to `update` already being defined by `Notifier` itself with a different signature.

app/lib/common/providers/notifiers/sync_notifier.dart
```diff
@@ -3,33 +3,47 @@ import 'dart:math';
...
+        }
+        // on release we have a really weird behavior, where, if we schedule
+        // any async call in rust too early, they just pend forever. this
+        // hack unfortunately means we have two wait a bit but that means
+        // we get past the threshold where it is okay to schedule...
+        client = newClient;
+        Future.delayed(
+          const Duration(milliseconds: 1500),
+          () => _restartSync(),
+        );
```
gnunicorn, line 41: it stands to be seen whether this comment still applies now and thus the delay is necessary or whether that means we can now remove that as all this happens "later" anyways.

app/lib/common/providers/notifiers/sync_notifier.dart
```diff
@@ -48,46 +62,44 @@ class SyncNotifier extends StateNotifier<SyncState> {
...
     });
-    ref.onDispose(() => _syncPoller.cancel());
+    ref.onDispose(() => _syncPoller?.cancel());
 
     _errorListener = syncState.syncErrorRx(); // keep it resident in memory
-    _errorPoller = _errorListener.listen((msg) {
+    _errorPoller = _errorListener?.listen((msg) {
       Sentry.captureMessage('Sync failure: $msg', level: SentryLevel.error);
-      if (mounted) {
-        if (msg == 'SoftLogout' || msg == 'Unauthorized') {
```
gnunicorn, line 70: `mounted` isn't necessary anymore as the internal `ref` directly put on the widget is dispose correctly otherwise and so are we.

app/lib/features/chat/widgets/mention_profile_builder.dart
```diff
@@ -24,8 +24,7 @@ class MentionProfileBuilder extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final client = ref.watch(alwaysClientProvider);
-    final userId = client.userId().toString();
```
gnunicorn, line 28: multiple cases, where we could simplify this to just using the proper provider

app/lib/features/cross_signing/widgets/cross_signing.dart
```diff
@@ -218,11 +218,12 @@ class CrossSigningState extends ConsumerState<CrossSigning> {
     });
 
     // start request event loop
-    final client = ref.read(alwaysClientProvider);
+    final client = await ref.read(alwaysClientProvider.future);
     client.installRequestEventHandler(fId);
 
     // open verification.request dialog
     showModalBottomSheet(
+      // ignore: use_build_context_synchronously
```
gnunicorn, line 226: this entire code isn't really being used now, so I opted for just ignoring the lint in this case.

app/lib/features/events/pages/event_details_page.dart
```diff
@@ -449,10 +449,11 @@ class _EventDetailPageConsumerState extends ConsumerState<EventDetailPage> {
         openShareSpaceObjectDialog(
           context: context,
           refDetails: refDetails,
-          internalLink: internalLink, shareContentBuilder: () async {
-          Navigator.pop(context);
-          return await refDetails.generateExternalLink();
-        },
+          internalLink: internalLink,
+          shareContentBuilder: () async {
```
gnunicorn, line 453: just format fixing that was automatically done.

app/lib/features/home/pages/dashboard.dart
```diff
@@ -38,15 +34,15 @@ class Dashboard extends ConsumerWidget {
               ? FloatingActionButtonLocation.miniEndDocked
               : FloatingActionButtonLocation.miniEndFloat,
           floatingActionButtonAnimator: FloatingActionButtonAnimator.scaling,
-          appBar: _buildDashboardAppBar(context, client),
+          appBar: _buildDashboardAppBar(context),
```
gnunicorn, line 37: all unnecessary if we don't have `isGuest` support at this point. nicely removed.

app/lib/features/main/app_shell.dart
```diff
@@ -60,16 +64,27 @@ class AppShellState extends ConsumerState<AppShell> {
...
+    _initProviders();
 
     // these want to be sure to execute in order
     await _initNotifications();
     // calendar sync
     await _initCalendarSync();
   }
 
-  Future<void> initShake() async {
+  Future<void> _initProviders() async {
```
gnunicorn, line 78: new preloaders

app/lib/router/router.dart
```diff
@@ -70,14 +70,27 @@ Future<String?> forwardRedirect(
...
           .expect('query params should contain device id');
       client = await acterSdk.getClientWithDeviceId(deviceId, true);
-      // ignore: use_build_context_synchronously
-      final ref = ProviderScope.containerOf(context);
-      // ensure we have selected the right client
-      ref.invalidate(clientProvider);
     } catch (e, s) {
-      _log.severe('Client not found', e, s);
-      return null;
+      _log.severe('Specified Client not found', e, s);
```
gnunicorn, line 74: this update on the forwarding code now falls back to an existing client if the deviceId failed to materialize and only if that isn't possible, forwards to the intro/login screens.

Before this would return `null` here some times leading to an empty screen being rendered.

app/lib/router/router.dart
```diff
@@ -91,7 +104,7 @@ Future<String?> forwardRedirect(
       // we haven’t joined yet or have been kicked
       // either way, we are to be shown the thing on the activities page
       return state.namedLocation(
-        Routes.activities.name,
+        Routes.myOpenInvitations.name,
```
gnunicorn, line 107: invitations have moved, this update was missed.

gnunicorn (commented):
Explainers

kumarpalsinh25 (approved):
Very good finding of Freezing issue!!

PR changes looks good overall. Added some minor comment for better clarity few things.

Good to go for merge.

kumarpalsinh25:
@gnunicorn 
Few lint errors and Widget testing failures need correction before merge. :)

app/lib/common/providers/common_providers.dart
```diff
@@ -24,36 +24,32 @@ final genericUpdatesStream =
 });
 
 final myUserIdStrProvider = Provider(
-  (ref) => ref.watch(
-    alwaysClientProvider.select((client) => client.userId().toString()),
-  ),
+  (ref) =>
+      ref.watch(alwaysClientProvider).valueOrNull?.userId().toString() ??
+      '@acter:acter.global',
```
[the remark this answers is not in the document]
gtalha07: What does default value determine on the UI level ? Normally it shouldn't equate to some error widget or a simple refresh till actual client gets available ?

gtalha07 (approved):
Looks good overall!

app/lib/common/providers/notifiers/chat_notifiers.dart
gnunicorn: only _once_ a client is found, do we actually call the inner function that's what this code does. at build there might not be one yet (as that provider is async).

app/lib/common/providers/common_providers.dart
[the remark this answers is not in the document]
gnunicorn: in practice, this will never be the case as all pages calling this are protected by having a client. BUT this is directly used by rendering functions and wrapping that into an async-case just for type system being happy felt like overreach. So I gave it a default value that satisfies the type system ... again, never actually happens in practice.

PR #2512, opened 2025-01-22
fix flutter unit tests

gtalha07:


app/test/helpers/mock_chat_providers.dart
```diff
@@ -169,15 +169,20 @@ class MockRoomListFilterNotifier extends StateNotifier<RoomListFilterState>
         );
 }
 
-class MockPersistentPrefNotifier extends StateNotifier<FilterSelection>
+class MockPersistentPrefNotifier extends Notifier<FilterSelection>
```
gtalha07, line 172: moved to newer riverpod API as ```StateNotifier``` (legacy API) seems to cause issue.

PR #2515, opened 2025-01-22
[Rust] Proper typed for internal indizes and subscription keys

gnunicorn:
Working on providing support for the event history for the activity screens, I came across and decided to resolve a long standing issue in our internal system: rather than being properly typed the executor and storage just use "known" strings to communicate and store data. It has been bugging me for a while, as we opted for it just as an initial flexible system to learn what was really needed, but we have grown to the point where we know and can do a proper typing system here.

And that is exactly what this does. It replaces the string-based subscription and index identification system with properly typed indizes and typed methods instead. 

To the reviewer:
This starts out with a bunch of new enums defined here in `native/core/src/referencing.rs` (which is where I recommend you start reviewing) and this then bubbles up through the entire infrastructure and all the APIs, from the models, which have been extended over these types up to the final API layer and the consequences on the dart side for that. As most of the stuff is covered well enough by the type system and the unit tests, I recommend glossing over most parts and focus on logic and consistency of objects and related types (e.g. the `indizes`-methods of the various models or the typed subscription function in the dart providers). As per the comment above, this _does_ already introduce some history-related indizes we didn't have before (for rooms and objects), which is why this all started, so I left those in ;) .


app/lib/common/providers/notifiers/chat_notifiers.dart
```diff
@@ -51,7 +52,7 @@ class AsyncLatestMsgNotifier extends FamilyAsyncNotifier<RoomMessage?, String> {
   FutureOr<RoomMessage?> build(String arg) async {
     final roomId = arg;
     final client = await ref.watch(alwaysClientProvider.future);
-    _listener = client.subscribeStream('$roomId::latest_message');
+    _listener = client.subscribeRoomParamStream(roomId, 'latest_message');
```
gnunicorn, line 55: there is still a few magical keywords for these functions that are then understood via strums EnumString, that I'd like to also switch into typed enums on the dart-side, but I opted for doing that in a follow up as this PR is already quite large as it is...

native/acter/api.rsh
```diff
@@ -2342,9 +2342,6 @@ object Space {
     /// the Tasks lists of this Space
     fn task_lists() -> Future<Result<Vec<TaskList>>>;
 
-    /// the Tasks list of this Space
-    fn task_list(key: string) -> Future<Result<TaskList>>;
```
gtalha07, line 2346: any reason for having it removed ???

app/lib/common/providers/notifiers/chat_notifiers.dart
gtalha07: Yeah, it's good to have also these string be typed as enums also. But overall good refactor on making these more safe!

gtalha07 (approved):
Looks good overall. Though I understand the changes and all the reflections around it, would still like @bitfriend to review once for any suggestions, particularly around ```ExecuteReference```.

Though I generally had it approved but please address the removal of api method which I have commented already :)

native/acter/api.rsh
gnunicorn: a) wasn't used and isn't meant to be used anymore - objects should not be taken directly from the spaces, only left the others as tests currently use them. but ideally, we'd remove them as well

b) it was easier to remove the function that wasn't used than include it in the refactor ;) .

PR #2520, opened 2025-01-23
[Rust] Observable Sorted Indizes for Stored Models

gnunicorn:
After switching to a [typed key referencing system in the rust store](https://github.com/acterglobal/a3/pull/2515), this ism the next stop towards allowing for retrieving the room and model histories: Sorted, observable indizes.

The way this works is by providing a custom Index type for the given index type, with two implementations we currently have: the default one just adds items to the front (which is kinda what we had, just that it used to be at the back) upon seeing it, and now this also brings a _Ranked_ one, where items with a higher rank are added to the beginning - which we can use to construct a simple _latest first history_. Thus this ranking system is used for history-indizes -- and now it becomes evident why we should have that as types rather than strings :D .

While they are already observable as is, actually using that via the API is left for a future PR to keep things tidy and simple. 

Tips for reviewing: The first commit only adds the new indizes with its unit tests, while the second does all the work to integrate it nicely into the store and add unit tests for that as well.


native/core/src/store/index.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Returns the current list of values in order of when they were added
+    pub fn values(&self) -> Vec<&OwnedEventId> {
+        match self {
+            StoreIndex::Lifo(lifo_index) => lifo_index.values(),
+            StoreIndex::Ranked(ranked_index) => ranked_index.values(),
+        }
+    }
+
+    // pub fn update_stream(&self) -> impl Stream<Item = VectorDiff<OwnedEventId>> {
```
gnunicorn, line 214: this is the future API that _should_ work, but the rust type system currently doesn't allow us to do this way. I kept it here for reference, but it will be changed in the future version soon.

native/core/src/models/rsvp.rs
```diff
@@ -142,7 +142,7 @@ impl ActerModel for Rsvp {
 
         let manager = {
             let model = store.get(&belongs_to).await?;
-            if !model.capabilities().contains(&Capability::Commentable) {
```
gtalha07, line 145: Does this mean we were wrongfully checking before ? As ```Commentable``` and ```RSVPable``` has different semantics?

gtalha07 (approved):
LGTM!

native/core/src/models/rsvp.rs
gnunicorn: yes, we did. but as only events are rsvp-able and they are also commentable it didn't really matter in practice.

Issue #2507, opened 2025-01-20
Events: ListView scrolling is weird / tends to block scrolling


gtalha07:
Seen at least in iOS and macOS: 

https://github.com/user-attachments/assets/7341878c-ba85-4f37-9635-a7fb30de6756

PR #2528, opened 2025-01-28
Solved weird Listview scrolling.

anisha-e10:
Fixes, #2507 

Scroll behavior before changes,
 
https://github.com/user-attachments/assets/f8b7b6e9-dea0-4dc7-8538-1ce2d52af213

Scroll behavior after fixes,

https://github.com/user-attachments/assets/1a5fd5d7-e27e-4044-9c15-cb822ef0258c

### DIFF
--- a/.changes/2528-listview-scroll-behavior.md
+++ b/.changes/2528-listview-scroll-behavior.md
@@ -1,0 +1,1 @@
+-[Fix] Solved weird behavior of listview scroll issue events page.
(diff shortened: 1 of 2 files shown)


gnunicorn (approved):
Great work, @anisha-e10 !

Can you elaborate what that parameter does, what the problem before was and how that value of that parameter fixes it? Just to educate me and the others in the team about what's going on and learn to spot and fix it in the future...

anisha-e10:
In Flutter, the cacheExtent property of the  Listview widget determines how much content (in terms of pixels) before and after the currently visible portion of the list should be preloaded into memory.

By default, the Listview only keeps the visible items in memory to save on performance and memory usage. However, in some cases, you may want to load more items to provide a smoother scrolling experience, especially in cases where you have lists with a large number of items or if you want to ensure that items are ready to display immediately when they come into view.

PR #2531, opened 2025-01-29
Events sorting improvements.

anisha-e10:
Fixes [#707](https://github.com/acterglobal/a3-meta/issues/707)

This PR addresses the following points:

1. Event Sorting:

 Events are now sorted with priority:
 Happening Now events will appear first, followed by Upcoming events and then, Past Events will be listed at the end.

 Note: Previously, bookmarked events were prioritized and displayed first, but this is now updated according to the above 
 sorting order.

2. Improved Event List Representation:

 The Events list now categorizes events into distinct sections:
 Happening Now, Upcoming, and Past Events,  providing a clearer and more organized view.

 Reference Video:

 https://github.com/user-attachments/assets/17b95c9c-45c7-485e-bff0-b193a35c2bf2

 Additional Notes:

 Due to the changes in event categorization, the error page event test cases were previously failing. These test cases have now 
 been Fixed.

### DIFF
--- a/.changes/2531-events-sorting-improvement.md
+++ b/.changes/2531-events-sorting-improvement.md
@@ -1,1 +1,1 @@
-- [Improvement] : Better events filter representation added for Happening Now, Upcoming and Past events.
+- [Improvement] : Better management of event list sorting mechanism and also improve visualisation
--- a/.changes/2531-events-sorting-improvement.md
+++ b/.changes/2531-events-sorting-improvement.md
@@ -1,0 +1,1 @@
+- [Improvement] : Better events filter representation added for Happening Now, Upcoming and Past events.
--- a/.changes/2531-events-sorting-improvement.md
+++ b/.changes/2531-events-sorting-improvement.md
@@ -1,1 +1,2 @@
-- [Improvement] : Better management of event list sorting mechanism and also improve visualisation
+- [Improvement] : Reorganized event list to show "Happening Now" first, followed by "Upcoming" and "Past" events - replacing the previous bookmark-based sorting.
+- [Improvement] : Events list now also categorizes events into distinct sections for better visibility.
--- a/app/lib/features/events/pages/event_list_page.dart
+++ b/app/lib/features/events/pages/event_list_page.dart
@@ -5,6 +5,7 @@ import 'package:acter/common/widgets/space_name_widget.dart';
 import 'package:acter/features/events/providers/event_providers.dart';
 import 'package:acter/features/events/widgets/event_list_empty_state.dart';
 import 'package:acter/features/events/widgets/event_list_widget.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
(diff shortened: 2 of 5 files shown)


gtalha07:
> Fixes [#707](https://github.com/acterglobal/a3-meta/issues/707)
> 
> This PR addresses the following points:
> 
> 1. Event Sorting:
> 
> Events are now sorted with priority: Happening Now events will appear first, followed by Upcoming events and then, Past Events will be listed at the end.
> 
> Note: Previously, bookmarked events were prioritized and displayed first, but this is now updated according to the above sorting order.
> 
> 2. Improved Event List Representation:
> 
> The Events list now categorizes events into distinct sections: Happening Now, Upcoming, and Past Events, providing a clearer and more organized view.


Sorry, but this defeats the purpose of bookmarking. I didn't glanced at the code but reading over the notes which is substantially not the behavior we would want to implement. Priority-wise bookmarking should still stay at top regardless. So the order would become bookmarked -> happening now -> upcoming -> past. I have to stop code review till I hear thoughts from @gnunicorn @kumarpalsinh25!

kumarpalsinh25:
> I have to stop code review till I hear thoughts from @gnunicorn @kumarpalsinh25!

Giving priority in Event case doesn't makes sense. Consider if you have bookmarked 3-4 event and those are now status with Past events still they will show in top and we aren't able to see events which are actually live or upcoming in time.

Also had discussion with @gnunicorn regarding the same and got mutual confirmation of him on this. :)

gnunicorn:
@gtalha07 as you can easily _filter_ on top in events to find the bookmarks, we opted for not putting them at the beginning of the list. For other objects putting the bookmarked at the front makes sense, as they don't have any "natural" order. But events have a natural order and re-ordering them according to bookmarking is highly confusing. That's why I proposed to change it to this.

It remains to be seen why you'd bookmark an event in the first place. My only reasonable idea is that you use it as a template to quickly "copy as new" from or because there is some other information on the event itself that you want to be able to quickly look up. But in any case, you have the handy top filters, so it seems reasonable to break that bookmarked-first-pattern in this case.

gtalha07:
> @gtalha07 as you can easily _filter_ on top in events to find the bookmarks, we opted for not putting them at the beginning of the list. For other objects putting the bookmarked at the front makes sense, as they don't have any "natural" order. But events have a natural order and re-ordering them according to bookmarking is highly confusing. That's why I proposed to change it to this.
> 
> It remains to be seen why you'd bookmark an event in the first place. My only reasonable idea is that you use it as a template to quickly "copy as new" from or because there is some other information on the event itself that you want to be able to quickly look up. But in any case, you have the handy top filters, so it seems reasonable to break that bookmarked-first-pattern in this case.

Okay the intention seems plausible, since we have filters for them to render and this only existing at the all events list section! 👍🏻

.changes/2531-events-sorting-improvement.md
```diff
@@ -0,0 +1 @@
+- [Improvement] : Better management of event list sorting mechanism and also improve visualisation
```
gtalha07, line 1: Think we can put it in more user-friendly and direct tone. Example:

```
[Improvement] Reorganized event list to show "Happening Now" first, followed by "Upcoming" and "Past" events - replacing the previous bookmark-based sorting.
[Improvement] Events list now also categorizes events into distinct sections for better visibility.

```

gtalha07 (approved):
Approved, just some minor feedback to address!

PR #2517, opened 2025-01-23
Object notifications : Unit testing enhancements

kumarpalsinh25:
Ref, https://github.com/acterglobal/a3/issues/2501

This PR contains comprehensive Unit Tests for Acter Notifications and Dart side implementation as below.

Screenshot 2025-01-26 at 10 15 41 PM

Screenshot 2025-01-26 at 10 15 30 PM

Screenshot 2025-01-26 at 10 15 08 PM

Screenshot 2025-01-26 at 10 14 44 PM

Screenshot 2025-01-26 at 10 16 40 PM

Screenshot 2025-01-26 at 10 16 55 PM

Screenshot 2025-01-26 at 10 17 08 PM

Screenshot 2025-01-26 at 10 17 21 PM

Screenshot 2025-01-26 at 10 17 31 PM

Screenshot 2025-01-26 at 10 17 50 PM

Screenshot 2025-01-27 at 11 21 46 AM

PR #2539, opened 2025-01-30
Implement rust notifications for creation, title and description change of major objects

gnunicorn:
✅ comes with tests.

Implements the basic starter notifications of "creation" and change of title and description for the main types Pin, CalendarEvent & TaskList. Tests are mostly copy-paste but ensure that all these work fine and according to spec modeled in #2517 .

This is a first set, split out into a separate PR to make reviewing easier. Further types will be added on top. Also contains a few tests for redaction, attachments and references that are disabled for now as these are not yet implemented.


gtalha07 (approved):
Looks good!

PR #2552, opened 2025-02-04
Make alwaysClientProvider never fail

gnunicorn:
This PR is the final step in making the `alwaysClientProvider` truly async and infallible. It came about that we might still have a minor race in the registration code, where the `alwaysClientProvider` could still be in the `NoClientException`-failed-mode when the client setting hasn't fully trickled through the `async` loops of the system at the time we try to use that to redeem an invite. Leading to, depending on the situation, a freeze on registration or at the very least the invite not being redeemed.

This PR fixes that behavior by replacing the earlier `FutureProvider` with a more actively listening `Notifier` that will be stay in `loading` (and return the same inner `Future`) until there is a `Client` available. Removing the need for `NoClientException` altogether, as all other code will just `await` forever in that case now.

It also comes with tests showing that this new behavior is working es expected, also when clients are switched and reset to ensure we didn't introduce further issues in that logic.

### DIFF
--- a/.changes/2552-always-client.md
+++ b/.changes/2552-always-client.md
@@ -1,0 +1,1 @@
+- Fix an internal race conditions that lead to some systems not properly redeeming invitations upon registration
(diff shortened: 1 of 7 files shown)


kumarpalsinh25 (approved):
Looks good to me!

Issue #2483, opened 2025-01-08
Blank page on clicking space on Edit Invite Code side-drawer


kokodavid:
Steps to reproduce
Invite Code Page - side drawer (Edit Invite Code) - Click Space/Chat

Outcome
What did you expect?|
Go to Space profile or chat detail

What happened instead?
Blank page displayed

The video below shows the behavior.
https://github.com/user-attachments/assets/35eadcc6-74ac-407f-b1ab-82e0efc35248

Happening on: Windows, MacOS, iOS

PR #2566, opened 2025-02-07
Super Invite Refactor

kumarpalsinh25:
Complete refactoring of Super Invites module UI-UX. before starting code review, go through video reference for better idea.

### Video Reference:

https://github.com/user-attachments/assets/b85d128f-fd33-4a19-96c1-455a29d8c72a

### Key Highlights:
- Improvement on General UI-UX of Create invite and Invite List Page
- If custom invite code is not given then create auto-generated code based on room display name. https://github.com/acterglobal/a3/pull/2566/commits/321938674c447df2a1d6f14cee5c471a523cdbf5
- General code structure improvements
- Remove side drawer and replace it with page view for create invite

### Fixes:
- https://github.com/acterglobal/a3/issues/2535
- https://github.com/acterglobal/a3/issues/2483

### Notes:
- While removal of old classes, there were few errors in `integration_test` so I replace it with new classes name however I haven't check integrate_test as we aren't considering it as of now.
- This PR is not fully covered by test, but made it ready for view to not wait for test. However I have added test for list widget and planning to add few more test before merging. :)


app/lib/common/utils/utils.dart
```diff
@@ -149,6 +149,29 @@ String formatBytes(int bytes) {
   }
 }
 
+String generateInviteCodeName(String? roomName) {
```
kumarpalsinh25, line 152: Created common function to generate room name specific random code.

gnunicorn: but shouldn't this be in the `feature/super_invites` then?

app/lib/features/super_invites/pages/create_super_invite_page.dart
```diff
@@ -0,0 +1,375 @@
...
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.stretch,
+          children: [
+            isEdit && widget.token != null
+                ? InviteListItem(
+                    inviteToken: widget.token!,
+                    cardMargin:
+                        EdgeInsets.symmetric(horizontal: 0, vertical: 12),
+                  )
+                : Form(
```
gnunicorn, line 107: May I suggest to split this into a more readable format, which also removes the `!` (which we want to avoid whenever possible)

```suggestion
    final token = widget.token;
    return SingleChildScrollView(
      child: Padding(
        padding: const EdgeInsets.symmetric(horizontal: 16),
        child: Column(
          crossAxisAlignment: CrossAxisAlignment.stretch,
          children: [
            if (isEdit && token != null)
               InviteListItem(
                    inviteToken: token,
                    cardMargin:
                        EdgeInsets.symmetric(horizontal: 0, vertical: 12),
                  )
            else
                Form(
```

gnunicorn (approved):

PR #2574, opened 2025-02-10
Expose new account data subscriptions and use them

gnunicorn:
fixes that bookmarking didn't update, due to our improved internal listening system. Now works again:

https://github.com/user-attachments/assets/59d6dd29-cf75-45ce-bc62-b4dfeb8cad39


kumarpalsinh25 (commented):
Nothing much to review on Dart side. :)

gtalha07 (approved):
looks good!

PR #2577, opened 2025-02-11
Flutter upgrade 3.27

anisha-e10:
### 1.  The `withOpacity`  is deprecated:
Commit : https://github.com/acterglobal/a3/pull/2577/commits/ab9c1ed815ecb55a0d82699a8204a65f72d05f3a

- With new flutter version 3.27.4 `withOpacity` is deprecated so based on that  changed it to `withValues`.
- Reference  :  **https://docs.flutter.dev/release/breaking-changes/wide-gamut-framework#opacity**

### 2.  The `color.value`  is deprecated:
Commit : https://github.com/acterglobal/a3/pull/2577/commits/a4232c6d9efc0136c57b0a4385be2c1d44230aa5

- With new flutter version 3.27.4 the `color.value` is deprecated so based on that changed it to `color.toInt()`.
- The `color.toInt()`  is an extension function that created from Color class as there were no default alternative option I found.
- Reference : https://github.com/flutter/flutter/issues/160184

### 3. Removed `default` clause from switch case:
commit: https://github.com/acterglobal/a3/pull/2577/commits/db5a83116feb4ba876790828d009ae1c7a6b402d

- The default clause is removed from the switch statement because the previous cases already cover all possible enum values, meaning there is no need for a fallback.

### 4. Update flutter version in `pubspec.yaml` file:
commit : https://github.com/acterglobal/a3/pull/2577/commits/e24342698106c2359e11967de51cc7e0043c8638

- Update the flutter version from `flutter: ^3.24.0` to `flutter:^3.27.4` and also update sdk from `sdk: >=3.5.0` to` sdk: >=3.6.2`.
- After version updated there is an issue of `flutter_math_fork` plugin so, that  need to update from `flutter_math_fork: ^0.7.2` to `flutter_math_fork: ^0.7.3`.

### 5. Update flutter version in github workflow files:
commit : https://github.com/acterglobal/a3/pull/2577/commits/ff80a6d07a40d1a69d3881897cc040d7a971b54c

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -204,7 +204,7 @@ jobs:
       - uses: subosito/flutter-action@v2
         name: Set up flutter
         with:
-          flutter-version: '3.24'
+          flutter-version: '3.27.4'
           channel: 'stable'
 
       - name: Flutter info
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -148,3 +148,14 @@ String formatBytes(int bytes) {
     return '${size.toStringAsFixed(2)} ${suffixes[i]}';
   }
 }
+
+extension ColorUtils on Color {
+  int toInt() {
+    final alpha = (a * 255).toInt();
+    final red = (r * 255).toInt();
+    final green = (g * 255).toInt();
+    final blue = (b * 255).toInt();
+    // Combine the components into a single int using bit shifting
+    return (alpha << 24) | (red << 16) | (green << 8) | blue;
+  }
+}
--- a/app/lib/features/space/actions/set_acter_feature.dart
+++ b/app/lib/features/space/actions/set_acter_feature.dart
@@ -40,8 +40,6 @@ extension SpaceFeatureActivator on ActerAppSettings {
         updated.active(newVal);
         builder.tasks(updated.build());
         break;
-      default:
-        throw 'Unsupported SpaceFeature Type';
     }
     return builder;
   }
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup flutter
         uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.24'
+          flutter-version: '3.27.4'
           channel: 'stable'
 
       - name: Install supported toolchain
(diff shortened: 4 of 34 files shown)


.github/workflows/build-app.yml
```diff
@@ -204,7 +204,7 @@ jobs:
       - uses: subosito/flutter-action@v2
         name: Set up flutter
         with:
-          flutter-version: '3.24'
+          flutter-version: '3.27.4'
```
gnunicorn, line 207: Well done!

app/lib/common/utils/utils.dart
```diff
@@ -148,3 +148,14 @@ String formatBytes(int bytes) {
     return '${size.toStringAsFixed(2)} ${suffixes[i]}';
   }
 }
+
+extension ColorUtils on Color {
```
gnunicorn, line 152: can we have a minimal unittest for this? Just adding a bunch of colors and ensuring they come out with the correct value should be fine...

.github/workflows/build-app.yml
gtalha07, line 207: I think earlier we set this up due to some dependencies breaking up on CI, but its generally good to have it run always against `latest`. @gnunicorn do we want to keep it that way or just having it specified ?

app/lib/features/space/actions/set_acter_feature.dart
```diff
@@ -40,8 +40,6 @@ extension SpaceFeatureActivator on ActerAppSettings {
         updated.active(newVal);
         builder.tasks(updated.build());
         break;
-      default:
```
gnunicorn, line 43: why does this go away?

gnunicorn (approved):
nicely done. Just one missing in a package according to the ci tests

gtalha07 (approved):
LGTM!

kumarpalsinh25 (approved):
Looks good overall.

One minor thing I need to confirm with @gnunicorn before going for merge.

I see lint error on `acter_flutter_sdk.dart` related to `color.value` deprecation.
https://github.com/acterglobal/a3/actions/runs/13257086546/job/37005701468?pr=2577

Instead of implementing `toInt()` with `extension` of `color` class, can't we follow new color code syntax on `acter_flutter_sdk.dart`?

.github/workflows/build-app.yml
gnunicorn: this is an attempt to upgrade, if CI agrees, we want to upgrade and CI seems to agree.

gnunicorn:
@kumarpalsinh25 the helper that is used isn't even ffi-specific. we should probably move that out there.

that said, I am not sure we want to move to a different rgb color system on the backend just yet... at least not in this PR.

PR #2580, opened 2025-02-11
Activites types

gnunicorn:
Starting out building up the rust types for activities for the activities screen. Started with Membership based events. End-to-End-tested and API exposed.

Split this off at this point to make it easier for this and other to get reviewed after.


native/acter/src/api/activities.rs
```diff
@@ -0,0 +1,125 @@
...
+use tokio_stream::wrappers::BroadcastStream;
+
+use super::{Client, RUNTIME};
+
+#[derive(Clone, Debug)]
+pub struct MembershipChange(CoreMembershipChange);
+
+impl MembershipChange {
+    pub fn user_id_str(&self) -> String {
+        self.0.user_id.to_string()
```
gtalha07, line 20: is this some kind of indexing ? Think we could use ```inner``` or some better access name ?

native/core/src/activities/status.rs
```diff
@@ -0,0 +1 @@
+
```
gtalha07, line 1: Empty file. Is that intentionally placed ?

gtalha07 (approved):
Looks good, some minor comments but not blockers for merge ;)

native/acter/src/api/activities.rs
gnunicorn: yes, if your struct has unnamed fields (think `struct MytType(String, u32)`) then you access them with the counters. As we only have one inner type, there is no reason to have named fields ...

PR #2585, opened 2025-02-12
Chat as boost action

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2567

In the "Boost" section, I have added a "Share Chat" action, similar to other actions. This allows the user to select a chat for the boost action, and the selected chat will be applied to the selected slide.

There are two scenarios  in a share chat:

**1. User is a member of the chat:** 
if the logged-in user is a member of the chat, they can view the details of that chat.

**2. User is not a member of the chat:**
 If the logged-in user is not a member of the chat, they will see the chat name, but they won’t be able to view it's details.

This behavior is handled similarly to how other actions are managed, ensuring consistent user experience.

Reference video :

https://github.com/user-attachments/assets/036aaf00-9824-4d2d-bfc9-51a947f98496

### Reference image(User is a member of the chat) :

image

### Reference image(User is not a member of the chat) :

image

### In this video, we’ll walk through what happens when a user clicks on a chat they aren’t yet a part of  and a user click on a chat they are part of : 

https://github.com/user-attachments/assets/b73a0640-13de-413c-829d-05947ca8927e


app/lib/features/news/providers/news_post_editor_providers.dart
```diff
@@ -104,6 +106,18 @@ class NewsStateNotifier extends StateNotifier<NewsPostState> {
     state = state.copyWith(currentNewsSlide: selectedNewsSlide);
   }
 
+  Future<void> selectChatToShare(BuildContext context) async {
+    final selectedChatId = await selectChatDrawer(context: context);
+    RefDetails? refDetails;
+    if (selectedChatId != null) {
+      final selectedChat = await ref.read(chatProvider(selectedChatId).future);
+      refDetails = await selectedChat!.refDetails();
```
gtalha07, line 114: possible null check before (!) to avoid it:
```suggestion
      if(selectedChat == null) return;
      refDetails = await selectedChat.refDetails();
```

app/lib/features/news/widgets/news_post_editor/selected_action_button.dart
```diff
@@ -30,13 +30,13 @@ class SelectedActionButton extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     if (refDetails == null) return SizedBox();
     final refObjectType = refDetails!.typeStr();
```
gtalha07, line 32: we have checked possible ```null``` case for ```refDetails``` already above, its good to use ```?``` to avoid UI exception :)

app/lib/features/news/widgets/news_post_editor/selected_action_button.dart
```diff
@@ -30,13 +30,13 @@ class SelectedActionButton extends ConsumerWidget {
...
-
     return switch (refObjectType) {
       'pin' => pinActionButton(context, ref, refDetails!),
```
gtalha07, line 34: same for this switch cases, refrain from using it.

Issue #2224, opened 2024-10-06
Activities Stream (Notifications & Screen)


gnunicorn:
The fourth screen is still very limited in its usage. This is a tracking issue about what we do there. Currently in a discussion form.

emilvincentz:
Content:
1.Activity stream quickly get noise and cluttered
Would:
- Cluster alle likes to a boost or story (instead if showing every single one)
- Remove alle spaces member changes (joined, kicked, etc) this info isn't really important and just clutter signal vs. noise
- Dont show my own engagement.  no need to see that I joined x event or liked x (I know this)

2. Add activity when objects are created
- Right now I dont see "Ben created a event", I only see "Ben is going to an event"

3. Order
I understand the idea about having it being clustered by space.
However I often find myself missing some info, because I scroll down to view latest "unread ones" and then I stop scrolling until I get to the ones from today, however there might be an activity that I havent read, which is "newer" than the above ones, but is shown below because its part of another space. 
Would:
- consider not having it clustered by space (I see the idea, but its not very often that I collapse the space) also it doesnt provide that much value that its clustered by space, I would rather that its linear order and maybe just a small icon or name, indicating which space the activity belong too. 

4. not sure which activities are unread
Itts not always clear which ones I have read and which I havent.

5. Spacing. 
- There is a lot if spacing and empty space. it feel it can be done more condensed and in that way have a quicker overview rather meeting to scroll that much

gnunicorn:
I largely agree with @emilvincentz 's comments, but I'd like to emphasize a few things in particular (via repeating them)

_Reminder that this feature is inspired by Github's History and the [Basecamp Latest Activity feature](https://basecamp.com/#whats-actually-been-happening):

**Signal vs. Noise isn't**  is dealt with very well in this current version in three main ways:
1. there are many minuscule events of low importance, like likes and views -- with very little information -- but they are given the same size and importance as other events that actually have relevant content, like comments. These should probably be grouped (maybe per object per day?) but also just shouldn't get that much attention (compared to more important things happening).
2.  the cards are huge and with the grouping all this takes up soo much space, that even on my large phone (with rather tiny font settings), I can see barely three events at once , even without grouping by space, it fits barely 6 items. This is a lot of scrolling to find anything, not even talking about:
![Image](https://github.com/user-attachments/assets/1a82d61f-0a7c-4cdb-81e7-ac2533294c10)
[...]

PR #2593, opened 2025-02-13
Room and Object  Activities

gnunicorn:
On top #2224  this:

1. 367e844587d5d68f5c44bd9ec7851cf2713f6379 adds `roomName` as a new status type, `roomCreate` added but isn't working in e2e test
2. 4dec1ffcd35d090fe7c887328c88e081861eaf78 refactors the notification into a common object activity system

I added some comments for the latter to understand the changes better.


native/acter/api.rsh
```diff
@@ -2199,8 +2199,27 @@ object MembershipChange {
     fn reason() -> Option<string>;
 }
 
+object ActivityObject {
```
gnunicorn, line 2202: previously `NotificaitonItemParent`, now wrapped into an `ActiviyObject`

native/acter/src/api/activities.rs
```diff
@@ -67,24 +84,26 @@ impl Activities {
...
                         .take(limit as usize)
-                        .collect(),
+                        .collect()
+                        .await,
                 )
             })
             .await?
     }
 
+    pub async fn iter(&self) -> anyhow::Result<impl Stream<Item = CoreActivity> + '_> {
```
gnunicorn, line 99: makes it more handy for testing to find the specific item in question

native/acter/src/api/push/notification_item.rs
```diff
@@ -157,138 +159,6 @@ impl NotificationRoom {
             .await?
     }
 }
-#[derive(Clone, Debug)]
```
gnunicorn, line 160: this has all moved to the new `Activity` area, which has stayed mostly the same internally though

native/acter/src/api/push/notification_item.rs
```diff
@@ -303,154 +173,24 @@ pub enum NotificationItemInner {
...
-        event_id: OwnedEventId,
-    },
-    // catch-all for other object changes
-    OtherChanges {
-        parent_obj: Option<NotificationItemParent>,
-        parent_id: OwnedEventId,
-        room_id: OwnedRoomId,
-        event_id: OwnedEventId,
-    },
+    Activity(Activity),
```
gnunicorn, line 176: Anything we can understand as an object activity is now wrapped in here, coming from the common internal API based on the actual inner object

native/acter/src/api/push/notification_item.rs
```diff
@@ -721,55 +258,34 @@ impl NotificationItemInner {
...
-                | NewsContent::Text(msg_content) => Some(MsgContent::from(msg_content)),
-                NewsContent::Fallback(FallbackNewsContent::Video(msg_content))
-                | NewsContent::Video(msg_content) => Some(MsgContent::from(msg_content)),
-                NewsContent::Fallback(FallbackNewsContent::Audio(msg_content))
-                | NewsContent::Audio(msg_content) => Some(MsgContent::from(msg_content)),
-                NewsContent::Fallback(FallbackNewsContent::File(msg_content))
-                | NewsContent::File(msg_content) => Some(MsgContent::from(msg_content)),
-                NewsContent::Fallback(FallbackNewsContent::Location(msg_content))
-                | NewsContent::Location(msg_content) => Some(MsgContent::from(msg_content)),
+            NotificationItemInner::Activity(activity) => match activity.content() {
```
gnunicorn, line 261: as the core types are properly typed, we have to wrap the inner content into the local ffi-api of `MsgContent` here

native/acter/src/api/push/notification_item.rs
```diff
@@ -900,12 +422,14 @@ impl NotificationItem {
         // acter specific items:
         if let RawNotificationEvent::Timeline(raw_tl) = &inner.raw_event {
             if let Ok(event) = raw_tl.deserialize_as::<SyncAnyActerEvent>() {
-                return NotificationItem::for_acter_object(
-                    client,
-                    builder,
-                    event.into_full_any_acter_event(room_id),
-                )
-                .await;
+                if !matches!(
```
gnunicorn, line 425: this actual if fixes a bug I only realized when looking at the code: as the `SyncAnyActerEvent` also catches regular matrix messages now, in the latest version of the code we _never_ go ahead to the `chat`-variant below. This match check actually ensures we skip those cases so we can do the chat message as below.

native/acter/src/api/push/notification_item.rs
```diff
@@ -926,552 +450,116 @@ impl NotificationItem {
 
         Ok(builder.build()?)
     }
+}
 
-    async fn for_acter_object(
+async fn convert_acter_model(client: Client, event: AnyActerEvent) -> Result<Activity> {
```
gnunicorn, line 455: separate function to make the match below easier without early quitting.

native/acter/src/api/push/notification_item.rs
```diff
@@ -926,552 +450,116 @@ impl NotificationItem {
...
+    Ok(Activity::for_acter_model(client.store(), AnyActerModel::try_from(event)?).await?)
```
gnunicorn, line 456: Thanks to the new implementation, we can now have the model from the event and then directly turn that into an activity right here.

native/acter/src/api/push/notification_item.rs
```diff
@@ -926,552 +450,116 @@ impl NotificationItem {
...
-                let parent_obj = client
-                    .store()
-                    .get(&e.content.task.event_id)
-                    .await
-                    .map_err(|error| {
-                        tracing::error!(?error, "Error loading parent of comment");
+        let mut builder = self;
+
+        // a few special cases we want to deal with
+        let builder = match activity.content() {
```
gnunicorn, line 476: Notifications have a peculiar pattern and we need to wrap some of those inner types into the specific patterns -- most notably to have the right `title` set on the notification itself. This is what this match here does.

There are e2e tests for all of these.

native/acter/src/api/push/notification_item.rs
```diff
@@ -926,552 +450,116 @@ impl NotificationItem {
...
+                ..
+            } => builder.title(new_due_date.format("%Y-%m-%d").to_string()),
+            ActivityContent::TaskDueDateChange { new_due_date, .. } => {
+                builder.title("removed due date".to_owned())
             }
+            ActivityContent::TaskAdd { task, .. } => builder.title(task.title().clone()),
+            ActivityContent::DescriptionChange {
+                object,
+                content: Some(content),
+            } => builder.msg_content(MsgContent::from(content)),
```
gnunicorn, line 557: description doesn't mean adding the title, but adding the content.

native/core/src/activities.rs
```diff
@@ -1,55 +1,624 @@
...
+pub mod object;
 pub mod status;
 
 #[derive(Clone, Debug)]
-pub enum Activity {
+pub enum ActivityContent {
     MembershipChange(MembershipChange),
+    RoomCreate(RoomCreateEventContent),
+    RoomName(String),
+    Boost {
```
gnunicorn, line 30: this and the following are essentially moved over from the `NotificationItemInner` but cleaned up (as we can be sure we have some activity object, and we have the other data in the meta info instead)

native/core/src/activities.rs
```diff
@@ -1,55 +1,624 @@
...
+            }
+            ActivityContent::MembershipChange(_)
+            | ActivityContent::RoomCreate(_)
+            | ActivityContent::RoomName(_) => todo!(),
+        }
+    }
+}
+
+impl Activity {
+    pub async fn for_acter_model(store: &Store, mdl: AnyActerModel) -> Result<Self, crate::Error> {
```
gnunicorn, line 261: we are now converting the activity right from the model, but the logic is basically the same as before and most of this is still covered by the same and additional tests.

native/core/src/activities/object.rs
```diff
@@ -0,0 +1,145 @@
+use matrix_sdk::ruma::OwnedEventId;
+use urlencoding::encode;
+
+use crate::models::{ActerModel, AnyActerModel};
+
+#[derive(Clone, Debug)]
+pub enum ActivityObject {
```
gnunicorn, line 7: Formerly known as `NotificationItemParent`

native/core/src/activities/object.rs
```diff
@@ -0,0 +1,145 @@
...
+        title: String,
+    },
+    Task {
+        tl_id: OwnedEventId,
+        object_id: OwnedEventId,
+        title: String,
+    },
+    Unknown {
+        object_id: OwnedEventId,
+    },
```
gnunicorn, line 30: Just with an additional case, where we might not have been able to load the item locally, so we just pass out the parent id

native/core/src/models/calendar/event.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct CalendarEvent {
-    inner: CalendarEventEventContent,
+    pub(crate) inner: CalendarEventEventContent,
```
gnunicorn, line 24: these are now necessary for the activity parsing from model code

native/test/src/tests/activities/calendar.rs
```diff
@@ -0,0 +1,61 @@
...
+    assert_eq!(user.calendar_events().await?.len(), 1);
+
+    let spaces = user.spaces().await?;
+    assert_eq!(spaces.len(), 1);
+
+    let main_space = spaces.first().unwrap();
+    assert_eq!(main_space.calendar_events().await?.len(), 1);
+
+    let activity = get_latest_activity(&user, main_space.room_id().to_string(), "creation").await?;
+    assert_eq!(activity.type_str(), "creation");
```
gnunicorn, line 56: see, it also works for activities now \o/

native/test/src/tests/activities/status.rs
```diff
@@ -28,6 +31,75 @@ async fn _setup_accounts(
...
+    // })
+    // .await?;
+    // assert_eq!(created.type_str(), "created");
+
+    // let ActivityContent::MembershipChange(r) = activity.content() else {
+    //     bail!("not a membership event");}
+    // ;
+    // assert!(matches!(r.change, MembershipChangeType::Invited));
+    // assert_eq!(r.as_str(), "invited");
+    // assert_eq!(r.user_id, to_invite_user_name);
```
gnunicorn, line 98: For reasons not clear to me, I do not receive the `RoomCreate` events, but I wanted to leave the tests code here for future exercise and fixing of the test.

gnunicorn (commented):
Just some comments for clarification

bitfriend (approved):
Looks good to me

PR #2594, opened 2025-02-14
Super Invite Code as Boost action

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2588

In the "Boost" section, I have added a "Share Invitation Code" action, similar to the other actions in the system. This allows the user to select a super invite token for the boost action, and the selected token will be applied to the corresponding slide.

Reference Video : 

https://github.com/user-attachments/assets/3b7c7a3c-4d24-4789-8bad-c7d1229cf0e5

Reference Video : 

Note: If the user owns the super invite code, they can view or edit it. However, if the user does not own the code, they can only copy the token for use.

https://github.com/user-attachments/assets/3b6dac28-f163-4ba5-89d6-f05a96febaea

https://github.com/user-attachments/assets/1cd17247-5e0a-4f9a-ab48-baa7e055e518


app/lib/features/news/widgets/news_item_slide/news_slide_actions.dart
```diff
@@ -115,6 +123,48 @@ class NewsSlideActions extends ConsumerWidget {
...
+              if (!context.mounted) return;
+              context.pushNamed(
+                Routes.createSuperInvite.name,
+                extra: token,
+              );
+            } catch(e) {
+              await Clipboard.setData(ClipboardData(text: title));
+              EasyLoading.showToast(lang.messageCopiedToClipboard);
+            }
+          },
```
anisha-e10, line 149: Managed by try catch because if the superInviteTokenProvider throws an error
```
final superInviteTokenProvider = FutureProvider.autoDispose
    .family<SuperInviteToken, String>((ref, tokenCode) async {
  final tokens = await ref.watch(superInvitesTokensProvider.future);
  for (final token in tokens) {
    if (token.token() == tokenCode) return token;
  }
  throw 'SuperInvite $tokenCode not found';
});
```
The user is able to copy the invite code

bitfriend:
Pushed [the above commit](https://github.com/acterglobal/a3/pull/2594/commits/402a2b1fa7e2afbed369420fa6595c1bca83ca94) in ubuntu because the latest matrix-sdk has [an issue in windows](https://github.com/acterglobal/a3/pull/2613).

app/lib/features/news/providers/news_post_editor_providers.dart
```diff
@@ -132,7 +134,18 @@ class NewsStateNotifier extends StateNotifier<NewsPostState> {
...
     );
+  }
 
+  Future<void> selectInvitationCodeToShare(BuildContext context) async {
+    final superInviteCode = await selectSuperInviteListDrawer(context: context);
+    RefDetails? refDetails;
+    if (superInviteCode != null) {
+      final selectedInviteCode = await ref.watch(superInviteTokenProvider(superInviteCode.token()).future);
+      refDetails = selectedInviteCode.refDetails();
+    }
```
kumarpalsinh25, line 145: `final selectedInviteCode = await ref.watch(superInviteTokenProvider(superInviteCode.token()).future);`

What is the need of this? 
As you are getting `SuperInviteToken` object directly from `selectSuperInviteListDrawer`

```suggestion
    final superInviteToken = await selectSuperInviteListDrawer(context: context);
    RefDetails? refDetails;
    if (superInviteToken != null) {
      refDetails = superInviteToken.refDetails();
    }
```

app/lib/features/super_invites/widgets/invite_list_item.dart
```diff
@@ -36,12 +38,16 @@ class InviteListItem extends StatelessWidget {
             style: Theme.of(context).textTheme.labelMedium,
           ),
           onTap: () {
-            context.pushNamed(
-              Routes.createSuperInvite.name,
-              extra: inviteToken,
-            );
+            if (onSelectInviteCode != null && inviteToken != null) {
+              onSelectInviteCode!(inviteToken!);
```
kumarpalsinh25, line 42:
```suggestion
            if (onSelectInviteCode != null) {
              onSelectInviteCode!(inviteToken);
```

final Function(SuperInviteToken)? onSelectInviteCode returns nullable data no need to check for null condition here.

kumarpalsinh25 (approved):
Overall flutter side changes looks good.

Added few comments regarding code structure and some null condition checking which need to fix prior to code merge.

app/lib/features/news/widgets/news_item_slide/news_slide_actions.dart
```diff
@@ -115,6 +123,48 @@ class NewsSlideActions extends ConsumerWidget {
...
   }
 
+  Widget renderInvitationCodeActionButton(
+    BuildContext context,
+    WidgetRef ref,
+    RefDetails referenceDetails,
+  ) {
+    final lang = L10n.of(context);
+    final title = referenceDetails.title();
+    if (title != null) {
```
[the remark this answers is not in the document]
anisha-e10: ok, got it

app/lib/features/super_invites/widgets/invite_list_item.dart
anisha-e10: Actually `SuperInviteToken` is not nullable callback function is nullable, and the `invitetoken != null` condition is already handled in the onTap method, so it will not cause any issues. I need to put this condition because I have to pass `invitetoken` for super invite code item to  `onClick` method and also it's helpful in test cases.

PR #2612, opened 2025-02-17
Enable system-wide object autosubscribe

gnunicorn:
With the now default-on "autosubscribe" feature, we subscribe the user to all objects they created or interacted with to receive notifications for (unless they "unring" the bell on the object).

This means, that next to the bookmark, all objects now have the :bell: button indicating whether the user is subscribed to updates from this object and allowing them to toggle that.

Upon creation, edit, comment and attachment, we auto-subscribe the user if they have set the notification setting for that (which is on by default). See it in action:

## Autosubscribe upon creation:

https://github.com/user-attachments/assets/a925ebcd-5d16-4f04-b46a-c952bd16a10b

## Autosubscribe for pins:

https://github.com/user-attachments/assets/43a8c54a-376f-4761-b692-f7bfa1faacfa

## Autosubscribe for rsvp:

https://github.com/user-attachments/assets/15712dfa-5c08-4963-bfcc-750a4e809d55

## Autosubscribe on edit:

https://github.com/user-attachments/assets/4519eeb9-7fa1-4fec-8142-786612eb9ba4

## And for task creation:

_subscribes to the task as well as its list_:

https://github.com/user-attachments/assets/ff5f932c-5c4b-4363-8097-79923f114a23

## on comments too

https://github.com/user-attachments/assets/f09e6553-7512-4b54-866b-f59e14d9363a


app/lib/features/notifications/providers/notifiers/notification_settings_notifier.dart
```diff
@@ -10,21 +10,26 @@ final _log = Logger('a3::common::notification_settings_notifier');
...
     extends AsyncNotifier<NotificationSettings> {
-  late Stream<bool> _listener;
-  late StreamSubscription<bool> _poller;
+  // ignore: unused_field
+  Stream<bool>? _listener;
+  StreamSubscription<bool>? _poller;
 
   @override
   Future<NotificationSettings> build() async {
+    ref.onDispose(() => _poller?.cancel());
```
gnunicorn, line 19: this PR revealed a bug in this notifier which lead to it only updating once. The reason was that we need to re-subscribe after each time when we create a new settings object as the old one gets dropped. the new code does that: triggering reset will always subscribe to the new one and add a listener to call reset again.

native/acter/src/api/attachments.rs
```diff
@@ -388,6 +387,10 @@ impl AttachmentsManager {
         self.room.room_id().to_string()
     }
 
+    pub fn object_id_str(&self) -> String {
```
gnunicorn, line 390: the managers for comments and attachments didn't give us a way to subscribe to the parent object, so I added these helpers from the rust side.

app/lib/features/notifications/actions/autosubscribe.dart
```diff
@@ -21,11 +19,6 @@ Future<bool> autosubscribe({
   required L10n lang,
   SubscriptionSubType? subType,
 }) async {
```
gnunicorn, line 21: not a labs anymore

kumarpalsinh25 (approved):
Get set ready for Acter Buzz!! 🔔

Changes look good to me.

PR #2635, opened 2025-02-19
Fix bug that app exit on system tray doesn't work in windows

bitfriend:
I found `Exit app` menu of system tray doesn't work in windows.
Will fix it here.


bitfriend:
[This callback](https://github.com/acterglobal/a3/blob/5f148f16d84a07adcc899f592a7614add7f566fe/app/lib/config/desktop.dart#L202) doesn't work in windows.
[This comment](https://github.com/leanflutter/tray_manager/issues/53#issuecomment-2538661428) was helpful to me.

bitfriend:
Now user can exit app via system tray in windows.

PR #2642, opened 2025-02-21
Fix issue that sync callback is started twice

bitfriend:
I found that sync callback is started twice.
This affects the initial loading time and probably may occur app crash due to too frequent mutex locks.


bitfriend:
The following screenshot shows that sync callback is started twice.
image

bitfriend:
This screenshot shows that `alwaysClientProvider` callback is called twice.
image

bitfriend:
`alwaysClientProvider` is the wrapper of `clientProvider`.
So that callback seems to be called twice.

bitfriend:
When `SyncNotifier` listens to `clientProvider` instead of `alwaysClientProvider`, sync callback is started only once.
image

bitfriend:
I recall `alwaysClientProvider` was used here to avoid that guest client (not-logged-in) starts sync callback.

bitfriend:
This issue will be fixed by https://github.com/acterglobal/a3/pull/2651
Then will close this PR.

PR #2651, opened 2025-02-25
Ensure always client and sync runs only onces

gnunicorn:
The assumption of `alwaysClientProvider` is that it is only triggered _once_ per client. However, upon startup, under specific circumstances, we actually set it twice for the very first client created. This trickles down and leads to a double sync, which we believe is a problem for some users and leads to freeze and white UI screens.

This was due to a problem in the way the provider tried to early-quit and handle the completer future in those cases, leading to it being set once via the callback, but then also through the completing future. There are several tests added in e1234fdd5fc5e771507708659d9466c8ac8d8f30 , most notably the last one is able to show the problem by proofing it is being set twice. df5959613f209a92c53f4292426919d2dd6a841e fixes the problem by _always_ returning the future and _never_ setting the state if there is still a future pending but only completing that. Only sets the state locally if there is no more completer from `build` pending.

An additional debug sessions showed that https://github.com/acterglobal/a3/blob/df5959613f209a92c53f4292426919d2dd6a841e/app/lib/common/providers/notifiers/sync_notifier.dart#L38 which was called twice before the fix, is now only called once, as expected.

fixes https://github.com/acterglobal/a3-meta/issues/807, https://github.com/acterglobal/a3-meta/issues/804, https://github.com/acterglobal/a3-meta/issues/801, https://github.com/acterglobal/a3-meta/issues/800, https://github.com/acterglobal/a3-meta/issues/799, https://github.com/acterglobal/a3-meta/issues/798, https://github.com/acterglobal/a3-meta/issues/794

### DIFF
--- a/app/lib/features/home/providers/notifiers/always_client_notifier.dart
+++ b/app/lib/features/home/providers/notifiers/always_client_notifier.dart
@@ -9,13 +9,12 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
   late ProviderSubscription subscription;
   @override
   FutureOr<Client> build() async {
-    subscription = ref.listen<AsyncValue<Client?>>(clientProvider, _updated);
-    final client = await ref.read(clientProvider.future);
-    if (client != null) {
-      return client;
-    }
-
     final cpl = completer ??= Completer<Client>();
+    subscription = ref.listen<AsyncValue<Client?>>(
+      clientProvider,
+      _updated,
+      fireImmediately: true,
+    );
     return cpl.future;
   }
 
@@ -23,12 +22,16 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
     // set up the refresh
     final newClient = newV.valueOrNull;
     if (newClient != null) {
-      state = AsyncData(newClient);
-      // and inform any pending completers
-      completer?.complete(newClient);
-      completer = null;
-    } else {
-      state = AsyncValue.loading();
+      final cmpl = completer;
+      if (cmpl != null) {
+        // we need to complete the previous future
+        // which internally sets the new state
+        cmpl.complete(newClient);
+        completer = null;
+      } else {
+        // we need to update to the new value
+        state = AsyncData(newClient);
+      }
     }
   }
 }
--- a/app/lib/features/home/providers/notifiers/always_client_notifier.dart
+++ b/app/lib/features/home/providers/notifiers/always_client_notifier.dart
@@ -32,6 +32,10 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
         // we need to update to the new value
         state = AsyncData(newClient);
       }
+    } else if (old?.valueOrNull != null) {
+      // we had a value previously, but now we don't, go back to endlessly
+      // pending to stop all dependents in their track.
+      state = AsyncValue.loading();
     }
   }
 }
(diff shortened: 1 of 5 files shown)


app/lib/features/home/providers/notifiers/always_client_notifier.dart
```diff
@@ -9,26 +9,29 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
...
   void _updated(AsyncValue<Client?>? old, AsyncValue<Client?> newV) {
     // set up the refresh
     final newClient = newV.valueOrNull;
     if (newClient != null) {
-      state = AsyncData(newClient);
-      // and inform any pending completers
-      completer?.complete(newClient);
-      completer = null;
-    } else {
-      state = AsyncValue.loading();
```
gnunicorn, line 31: only finish the completer _or_ set the state now. Before, due to us doing both, the same client was set twice.

app/test/common/providers/client_providers_test.dart
```diff
@@ -90,5 +90,48 @@ void main() {
...
+          // no client
+          clientProvider.overrideWith(() => provider),
+        ],
+      );
+      expect(container.read(alwaysClientProvider).isLoading, true);
+      int called = 0;
+      container.listen(alwaysClientProvider, (p, a) => called += 1);
+      provider.setClient(cl);
+      await tester.pumpAndSettle();
+      expect(called, 1);
```
gnunicorn, line 134: without this fix, this fails with `1 != 2`

app/lib/features/home/providers/notifiers/always_client_notifier.dart
```diff
@@ -9,26 +9,29 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
...
-      return client;
-    }
-
     final cpl = completer ??= Completer<Client>();
+    subscription = ref.listen<AsyncValue<Client?>>(
+      clientProvider,
+      _updated,
+      fireImmediately: true,
+    );
     return cpl.future;
```
bitfriend, line 18: I think destruction code is needed.
```dart
ref.onDispose(() => subscription.close());
```

app/lib/features/home/providers/notifiers/always_client_notifier.dart
```diff
@@ -9,26 +9,29 @@ class AlwaysClientNotifier extends AsyncNotifier<Client> {
...
+      if (cmpl != null) {
+        // we need to complete the previous future
+        // which internally sets the new state
+        cmpl.complete(newClient);
+        completer = null;
+      } else {
+        // we need to update to the new value
+        state = AsyncData(newClient);
+      }
     }
```
bitfriend, line 35: If `newClient` is `null`, we would do nothing here?

gnunicorn: yep

app/lib/features/home/providers/notifiers/always_client_notifier.dart
gnunicorn: not really. it will be dropped and the drop will cause a close.

gtalha07 (approved):
Good to see tests added around it which verifies the problem being fixed. I looked vaguely on fix due to limited knowledge but explicitly approving !

PR #2652, opened 2025-02-26
Chat Message URL Previews Rust Support

gnunicorn:
This PR brings support for sharing `UrlPreview`s and `RefDetails` (via those previews) attached to any text message in chat.

This is done via providing a few new API's, one must understand in tandem:

1. The `MsgContent` of a text chat message (any kind of markdown, html or plain) can now have one or many [`UrlPreview`s](https://github.com/acterglobal/a3/pull/2652/files#diff-12c0480bc493ffbb94ce49b75db168b6aec7504019bcbdea587c1192ffe429c2R116-R129) ([via this API](https://github.com/acterglobal/a3/pull/2652/files#diff-12c0480bc493ffbb94ce49b75db168b6aec7504019bcbdea587c1192ffe429c2R1110)), which can and should be used to render those previews (rather than fetching them on-demand)
2. those `url()` values can be `RefDetails`-Url, so you can and should parse them via the existing system and can construct `RefDetails` from that URL and the preview data provided
3. `Client` has a new [`url_preview` function](https://github.com/acterglobal/a3/pull/2652/files#diff-12c0480bc493ffbb94ce49b75db168b6aec7504019bcbdea587c1192ffe429c2R3297), that given a url will load the information and give a [`LocalUrlPreview`-object](https://github.com/acterglobal/a3/pull/2652/files#diff-12c0480bc493ffbb94ce49b75db168b6aec7504019bcbdea587c1192ffe429c2R131-R146) to preview data
4. the `MsgDraft` for chat messages has a new API that accepts any `LocalUrlPreview` and `RefDetails` as input types to be added to the chat message during the construction process.

_Note_: While the `LocalUrlPreview` and `UrlPreview` have API's for image previews, image previews are actually yet supported as a missing exposed type in `ruma` doesn't allow us to extract the information until the rust sdk has update to it.


Cargo.toml
```diff
@@ -46,6 +46,7 @@ default-features = false
 # version = "0.12.1"
 git = "https://github.com/ruma/ruma"
 rev = "7755c7cbc580f8d8aea30d78cc1a6850b1a6fd39"
+features = ['unstable-msc4095']
```
gnunicorn, line 49: for `url_preview`s in the rust types

native/acter/src/api/common.rs
```diff
@@ -326,11 +331,46 @@ impl TryFrom<&AttachmentContent> for MsgContent {
...
+    pub fn description(&self) -> Option<String> {
+        self.0.description.clone()
+    }
+
+    pub fn has_image(&self) -> bool {
+        false // not yet supported
+              // !self.0.image.is_none()
+    }
+    pub fn image_source(&self) -> Option<MediaSource> {
+        None // not yet support
```
gnunicorn, line 358: should work but we are missing the right types from upstream. PR has already been merged.

native/test/src/tests/url_previews.rs
```diff
@@ -0,0 +1,179 @@
+use anyhow::{bail, Result};
```
gnunicorn, line 1: url preview tests showing both types work end-to-end

gnunicorn (commented):
comments for the reviewer

bitfriend (approved):
Nice job.
Looks good to me.

PR #2654, opened 2025-02-26
Fix bug that could not find replied_to event outside of page

bitfriend:
We are facing this error log.
`[2025-02-26][13:42:37.387603][a3::chat::room_notifier][ERROR] Failing to load reference $eZQvskOrFiumOJhfHLtbNCjLH7vi8qAGV0037npDcOg (from $fsm-GcMno7rnqMOFmgpyZKEZsUGToPKKDDGSLNo3LWk) Event not found null`
Because original message in `replied_to` feature cannot be fetched.
Will fix it here.

Fixes [#meta-740](https://github.com/acterglobal/a3-meta/issues/740)


bitfriend:
The key point of this PR is `timeline.fetch_details_for_event().await`.
This fn must be called earlier, before `if let TimelineDetails::Ready(replied_to) = in_reply_to.event.clone() {}` can be `true`.

bitfriend:
I could fix the error `Event not found` via this PR.

gnunicorn (changes requested):
We agreed to not make significant changes to the existing chat but instead work on chat-ng. this instead changes the existing chat significantly and thus won't be merged.

PR #2658, opened 2025-02-27
Better color contrast on Update Slides

kumarpalsinh25:
Fixes, https://github.com/acterglobal/a3/issues/1704

Added separate colors for Background, Foreground and Link which make better contrast between each other.

https://github.com/user-attachments/assets/772de679-319f-4bbc-b9cf-5971e20490df


app/lib/features/news/model/news_post_color_data.dart
```diff
@@ -1,12 +1,52 @@
...
+  PostColorScheme(
+    backgroundColor: const Color.fromARGB(255, 38, 48, 104),
+    foregroundColor: Colors.white,
+    linkColor: Colors.amberAccent,
+  ),
+  PostColorScheme(
+    backgroundColor: const Color.fromARGB(255, 10, 10, 10),
+    foregroundColor: Colors.white,
+    linkColor: Colors.amberAccent,
+  ),
```
gnunicorn, line 51: So ... I am confused. Why have a new `linkColor` and `foregroundColor` added and passed over the wire and types, if they are always the exact same? Not seeing the point of doing that at all...

Also, due to the foreground and link color being bright all these background colors are now rather dark and boring :( . I don't find this a convincing set of colors we want to work with.

app/lib/features/news/model/news_post_color_data.dart
```diff
@@ -1,12 +1,52 @@
...
+    required this.linkColor,
+  });
+}
+
+//https://www.google.com
+final List<PostColorScheme> postColorSchemes = [
+  PostColorScheme(
+    backgroundColor: const Color.fromARGB(255, 107, 65, 50),
+    foregroundColor: Colors.white,
+    linkColor: Colors.amberAccent,
```
gnunicorn, line 20: I assume you picked `amberAccess` because it has a strong contrast to the background colors? However, I have two issue with using that color:
1. _traditionally_ links are some shade of blue, not using blue makes it harder for people to recognise them, especially since we don't ever use that same color/pattern anywhere else. I'd really like to stick with blue for links
2. accessiblity guidelines recommend a contrast ratio of 3:1 between the link and the text color (so it easy to recognise especially for impaired people), but [amberAccent to white has a meager 1.39:1 ratio](https://webaim.org/resources/contrastchecker/?fcolor=FFD740&bcolor=FFFFFF).

If we intend to fix the colors for contrast we should not introduce a new contrast problem but fix it wholesale IMHO.

gnunicorn (approved):
Code-wise looks fine, but UX wise this leaves things open to be desired

kumarpalsinh25:
@gnunicorn 
I agree with all your feedback related selection of colors and usage. 

Actually after implantation of Background , Foreground and LinkColors dynamically, I faced one issue related to state update in ApplyFlowy Editor. Editor doesn't gets updated with new fg or link color. Due to which I go with approach.

But it is true that we are resolving one issue of color contract and raising another color realted issue here. Let me try again with some of the way to manage ApplyFlowwy editor state with new colors.

kumarpalsinh25:
@gnunicorn 
There was one issue I was facing due to which random bg, fg and link color were misbehaving. So due to that I went for fixed colors.

But now I face figure out that issue and updating new Post Color New scheme accordingly.

kumarpalsinh25:
@gnunicorn 
I have updated video reference in PR description based on updated colors contrast.

Note: 
- I felt difficulty in finding best contrast of colors if we stick with Blue color for link so feel free to update in this PR if you find better combination of the same.

app/lib/features/news/model/news_post_color_data.dart
```diff
@@ -1,12 +1,61 @@
...
+final List<PostColorScheme> postColorSchemes = [
+  PostColorScheme(
+    backgroundColor: Colors.pinkAccent, // Purple 200
+    foregroundColor: Colors.white, // Deep Purple 900
+    linkColor: Colors.lightBlueAccent.shade700, // Blue 800
+  ),
+  PostColorScheme(
+    backgroundColor: Colors.green, // Purple 200
+    foregroundColor: Colors.white, // Deep Purple 900
+    linkColor: Colors.lightBlueAccent.shade100, // Blue 800
```
gnunicorn, line 24:
```suggestion
    linkColor: Colors.lightBlue.shade900
```

[Color Contrast Ratio of 2.7](https://colors.artyclick.com/contrast-color-finder/?color1=#4caf50&color2=#01579b)

gnunicorn (approved):
Yes, blue is traditional but also very hard to get a good contrast on against anything but white. I've added two changes where the contrast is way too low with stronger ones. Used https://colors.artyclick.com/contrast-color-finder/ to pick them.

PR #2666, opened 2025-03-03
Read Markers

gnunicorn:
1. Switch to the proper API for sending read markers, as it seems to work now (removes old code)
2. Ensure we only show and use read markers when the lab is actually activated
3. Implement fresh and improved code for read marker sending in chat and chat-ng upon scroll to end.


gtalha07 (approved):
Looks good

PR #2674, opened 2025-03-04
[Chat-NG]: fix desktop auto-focus set to field initially and after sending message

### LINKED ISSUE #2670: chat-ng: auto focus field on desktop
on desktop (when we don't have a soft keyboard) entering the chat screen, as well as after pressing `enter` to send, I'd assume the text field is autofocused. Unfortunately, it is not and I have to click with the mouse every time after having sent a message ...

gtalha07:
## Screenshot verification:

https://github.com/user-attachments/assets/fc565b6b-7714-44fd-b7d3-594fabe78a00

### DIFF
--- a/app/lib/common/widgets/html_editor/html_editor.dart
+++ b/app/lib/common/widgets/html_editor/html_editor.dart
@@ -102,7 +102,6 @@ extension ActerEditorStateHelpers on EditorState {
   void clear() async {
     if (!document.isEmpty) {
       final transaction = this.transaction;
-      final selection = this.selection;
 
       // Delete all existing nodes
       int nodeIndex = 0;
@@ -115,11 +114,14 @@ extension ActerEditorStateHelpers on EditorState {
 
       transaction.insertNode([0], paragraphNode(text: ''));
 
-      updateSelectionWithReason(
-        selection,
-        reason: SelectionUpdateReason.transaction,
-      );
       apply(transaction);
+
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        updateSelectionWithReason(
+          Selection.single(path: [0], startOffset: 0, endOffset: 0),
+          reason: SelectionUpdateReason.uiEvent,
+        );
+      });
     }
   }
 }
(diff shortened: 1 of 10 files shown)


app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -370,7 +372,7 @@ class HtmlEditorState extends State<HtmlEditor> {
           // widget pass through
           editable: widget.editable,
           shrinkWrap: widget.shrinkWrap,
-          autoFocus: widget.autoFocus,
+          autoFocus: true,
```
gtalha07, line 375: keeping auto-focus  `true` for accessibility on desktop.

gnunicorn: but shouldn't this be this then?
```suggestion
          autoFocus: isDesktop,
```

I don't think we want to open this on mobile every time as well...

gnunicorn (approved):
concern about mobile support

app/lib/common/widgets/html_editor/html_editor.dart
gtalha07: > but shouldn't this be this then?
> 
> I don't think we want to open this on mobile every time as well...

Yes, we don't and we don't have to put ```isDesktop``` check here as it only belongs in `desktopEditor()` method:
```
Widget desktopEditor(String? roomId) {
    return FloatingToolbar(
      items: [
        paragraphItem,
        ...headingItems,
        ...markdownFormatItems,
        quoteItem,
        bulletedListItem,
        numberedListItem,
        linkItem,
        buildHighlightColorItem(),
      ],
      textDirection: _textDirection,
      editorState: editorState,
      editorScrollController: editorScrollController,
      style: FloatingToolbarStyle(
        backgroundColor: Theme.of(context).colorScheme.surface,
        toolbarActiveColor: Theme.of(context).colorScheme.tertiary,
      ),
      child: Directionality(
        textDirection: _textDirection,
        child: AppFlowyEditor(
          // widget pass through
          editable: widget.editable,
          shrinkWrap: widget.shrinkWrap,
          autoFocus: true,
          header: widget.header,
          // local states
          editorScrollController:
              widget.scrollController ?? editorScrollController,
          editorState: editorState,
          editorStyle: desktopEditorStyle(),
          footer: generateFooter(),
          blockComponentBuilders: _buildBlockComponentBuilders(),
          characterShortcutEvents: [
            ...standardCharacterShortcutEvents,
            if (roomId != null) ...mentionShortcuts(context, roomId),
          ],
          commandShortcutEvents: [...standardCommandShortcutEvents],
          disableAutoScroll: true,
        ),
      ),
    );
  }
  ```

PR #2677, opened 2025-03-05
add name and subtype for attachment activities

### LINKED ISSUE #2676: Activity Object : Add param from where I can get details about content
- Like if someone added attachment then I need which type of attachment user added.

gnunicorn:
we were missing ways to display that information in the activity stream.

fixes #2676


bitfriend (approved):
Looks good to me

PR #2687, opened 2025-03-07
Apply new non-trailing-comma formatting

gnunicorn:
With 3.29 the default `dart_style`-formatting is to use a more condensed style with _some_ trialing commas, but not always. Which directly contradicts the `require_trailing_comma` lint we currently have for that makes `dart --fix` and the dart analyze server fight over the correct style.

see https://github.com/dart-lang/dart_style/issues/1652 for the discussion about this.

This PR thus removes that lint in4ec86f91810308b4e1a2351195d641ad648a488f and applies the style across the entire repository in eea4bbb8fd1a315d8e0890e5f596b556bf2dbc11 . No other changes are being made.

Please raise your voice about whether we want to go down that road, @kumarpalsinh25 , @gtalha07 .


kumarpalsinh25:
@gnunicorn 
I really like the changes that this PR brings. In-fact I personally alway try to remove additional `comma` where things can be fit with one line.

So I am good with PR changes.

gtalha07 (commented):
On the same page as of @kumarpalsinh25. Sometimes it leads to unnecessary indentation and depending upon line break I've set on IDE, it seems fine.

PR #2691, opened 2025-03-07
Appflowy with Codeblocks

gnunicorn:
on top of #2685 this upgrades the Appflowy Editor and includes the AppFlowyEditor-Plugins for code-blocks. This only works if we drop the `flutter_matrix_html`-package though due to some incompatible dependencies...

thus, I want to finish the main PR first and ensure it works as expected and only once that is out of labs, continue on this. Just leaving it here as the reference point.


gnunicorn:
superseeded.

PR #2697, opened 2025-03-09
Improving User Task Assignment UI & UX

### LINKED ISSUE #1877: Task Assignment Issues
1) When there are two persons in single task assignments and I remove myself then it is showing no assignments instead of just removing myself and showing other user's assignment.

https://github.com/acterglobal/a3/assets/51526480/08a4487e-5cee-4312-95ed-5ba6bf891fb8

2) Sometimes assignment myself removes other user's assignments

gnunicorn:
This PR overhauls the user assignment UI and UX to improve legibility and context. In particular this PR:

1. shows a user pill instead of just a user name
2. fixes #1877 by showing all users, even if we are not assigned
3. highlight if the user is us, by giving it a primary background, as well as a border in the listing
4. replace the action button with a bottom drawer menu (for future extensions)
5. direct action on my own user pill to remove the assignment (and a `x` next to the name to indicated that behavior)
6. improved UI and wording around the process but also in the internal code structure

See it in action:

https://github.com/user-attachments/assets/b9dfc77d-22c2-49bf-ae46-213ead1820c8


app/lib/features/tasks/pages/task_item_detail_page.dart
```diff
@@ -396,77 +399,137 @@ class TaskItemDetailPage extends ConsumerWidget {
...
+                child: Text(
+                  lang.notAssigned,
+                  style: textTheme.bodyMedium?.copyWith(
+                    decoration: TextDecoration.underline,
+                    fontStyle: FontStyle.italic,
+                  ),
+                ),
+              ),
       subtitle:
-          task.isAssignedToMe()
```
gnunicorn, line 425: this was the bug of #1877 .

app/lib/common/toolkit/buttons/user_chip.dart
```diff
@@ -1,50 +1,110 @@
...
 
 class UserChip extends ConsumerWidget {
   final String roomId;
   final String memberId;
+  final TextStyle? style;
+  final Widget? Function(
+    BuildContext context, {
+    required bool isMe,
+    required double fontSize,
+  })?
```
kumarpalsinh25, line 17: Why to have this type of complex callback methods?
- `isMe` can get fetch with `myUserIdStrProvider` and `memberId` at the calling area without having any dependency of `UserChip`
- What is the need of this `fontSize` param?

app/lib/common/toolkit/buttons/user_chip.dart
```diff
@@ -1,50 +1,110 @@
...
+  trailingBuilder;
+  final Function(
+    BuildContext context, {
+    required bool isMe,
+    required VoidCallback defaultOnTap,
+  })?
```
kumarpalsinh25, line 23: Same here, things can't be manage without callback functions?

kumarpalsinh25 (approved):
Nice improvement.

Overall code looks good apart from one minor feedback regarding `callback` function.

I personally would like to avoid `callback` functions whenever possible to reduce complexity of code understanding.

app/lib/common/toolkit/buttons/user_chip.dart
gnunicorn: I wouldn't consider these `callbacks` but rather `builders` .

And it _probably_ can be managed in a different way, sure. But if you look at the pattern we use this chip, it just means _reproducing_ the complexity of the checks the user-chip is already doing outside again. Doing this via a builder is _actually_ simpler code.

app/lib/common/toolkit/buttons/user_chip.dart
gnunicorn: same as above: yes, it _could_ be done without a `builder` function, but actually makes things _more_ complex rather than simpler ...

PR #2711, opened 2025-03-12
[Chat-NG]: add reply swipe 

gtalha07:
### PREVIEW:

https://github.com/user-attachments/assets/b1ed4e2c-6c31-4b5a-9d9d-f244005bdb7c

### DIFF
--- a/app/lib/features/chat_ng/widgets/events/message_event_item.dart
+++ b/app/lib/features/chat_ng/widgets/events/message_event_item.dart
@@ -1,6 +1,7 @@
 import 'package:acter/common/providers/room_providers.dart';
 import 'package:acter/features/chat/utils.dart';
 import 'package:acter/features/chat_ng/dialogs/message_actions.dart';
+import 'package:acter/features/chat_ng/providers/chat_room_messages_provider.dart';
 import 'package:acter/features/chat_ng/widgets/chat_bubble.dart';
 import 'package:acter/features/chat_ng/widgets/events/file_message_event.dart';
 import 'package:acter/features/chat_ng/widgets/events/image_message_event.dart';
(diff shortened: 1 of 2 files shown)


app/lib/features/chat_ng/widgets/events/message_event_item.dart
```diff
@@ -36,17 +38,25 @@ class MessageEventItem extends ConsumerWidget {
...
-      crossAxisAlignment:
-          isMe ? CrossAxisAlignment.end : CrossAxisAlignment.start,
-      mainAxisSize: MainAxisSize.min,
-      children: [
-        _buildMessageUI(context, ref, roomId, messageId, item, isMe),
-        _buildReactionsList(roomId, messageId, item, isMe),
-      ],
+    return SwipeTo(
+      onRightSwipe: !isMe ? (_) => _handleReplySwipe(ref, item) : null,
+      onLeftSwipe: isMe ? (_) => _handleReplySwipe(ref, item) : null,
```
gnunicorn, line 43: why have different directional swipe for the same action depending on whose message that is? this feels odd. I'd expect that whatever the message, if I swipe one direction I get one action and the other direction, I get another action. In that case, edit only works for my messages ...
```suggestion
      onRightSwipe: (_) => _handleReplySwipe(ref, item)
      // onLeftSwipe: isMe ? (_) => _handleEditSwipe(ref, item) : null,
```

gnunicorn (approved):
UX query

app/lib/features/chat_ng/widgets/events/message_event_item.dart
gtalha07: Oh yes, thanks for correction. I confirmed the same UI behaviour on whatsapp, signal use-cases.

PR #2719, opened 2025-03-14
Fix comment notifications

gnunicorn:
On story-type and if the parent has no known title. With tests and a minor refactor of the utils into extension types.

See 19102c8dcfb8e55079fc5dc36a3cfbe8004285ab which has the tests failing and 814deea46b43a9ddd0d607c30095850c2a054cd2 which fixes those tests.

### DIFF
--- a/packages/acter_notifify/lib/processing/utils.dart
+++ b/packages/acter_notifify/lib/processing/utils.dart
@@ -1,31 +1,34 @@
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 
-String parentPart(ActivityObject parent) {
-  final emoji = parent.emoji();
-  final title = switch (parent.typeStr()) {
-    'news' => "boost",
-    _ => parent.title(),
-  };
-  return "$emoji $title";
+extension ActerNotififyActivityObjectExtension on ActivityObject {
+  String parentPart() {
+    final emoji = this.emoji();
+    final title = switch (typeStr()) {
+      'news' => 'boost',
+      'story' => 'story',
+      _ => this.title(),
+    };
+    return "$emoji $title";
+  }
+
+  String getObjectCentricTitlePart(
+    String suffix,
+  ) {
+    final parentInfo = parentPart();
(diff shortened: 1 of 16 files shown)


packages/acter_notifify/lib/processing/utils.dart
```diff
@@ -1,31 +1,34 @@
...
-  String suffix,
-) {
-  final sender = item.sender();
-  final username = sender.displayName() ?? sender.userId();
+extension ActerNotififyActivityObjectExtension on ActivityObject {
+  String parentPart() {
+    final emoji = this.emoji();
+    final title = switch (typeStr()) {
+      'news' => 'boost',
+      'story' => 'story',
```
gnunicorn, line 8: fix for stories.

packages/acter_notifify/lib/processing/utils.dart
```diff
@@ -1,31 +1,34 @@
...
 
-  String titlePart = '$username $suffix';
-  if (emoji != null) titlePart = '$emoji $titlePart';
-  return titlePart;
+  String? getObjectCentricTitlePart(
+    String suffix,
+  ) {
+    final parentInfo = parentPart();
+    return parentInfo != null ? '$parentInfo $suffix' : null;
+  }
```
gnunicorn, line 19: and with `parentPart` nullable we can have both of these just also return null if we have no title and make the actual code go back to the fallback cases.

packages/acter_notifify/lib/processing/utils.dart
```diff
@@ -1,31 +1,34 @@
...
-}
-
-String getUserCentricTitlePart(
-  NotificationItem item,
-  String? emoji,
-  String suffix,
-) {
-  final sender = item.sender();
-  final username = sender.displayName() ?? sender.userId();
+extension ActerNotififyActivityObjectExtension on ActivityObject {
```
kumarpalsinh25, line 3: Nice work with `extension` approach.

kumarpalsinh25 (approved):
Looks good to me.

PR #2720, opened 2025-03-14
task invitations notifications

gnunicorn:
on top of #2719 and #2698 this adds support to the acter notifify plugin for showing object notifications (on tasks).

In particular this adds another `mentionsYou`-flag that allows us to figure out, if we are the target, in which case we put that into the title. Otherwise, we make it an object-notification with the object in the title (if given).

Tests for all of it, of course.


.changes/2719-fix-comments-notifcations.md
```diff
@@ -0,0 +1 @@
+- Fix notifications format for comments on stories
```
kumarpalsinh25, line 1: Probably this is from last PR work. We also need to add changes logs for task invitation.

packages/acter_notifify/lib/model/push_styles.dart
```diff
@@ -21,6 +21,7 @@ enum PushStyles {
   taskAccept('🤝'),
   taskDecline('✖️'),
   taskDueDateChange('🕒'),
+  objectInvitation('📨'),
```
kumarpalsinh25, line 24: Is it general object invitation or task invitation only?

packages/acter_notifify/test/invitations_notification_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      final sender = MockNotificationSender(name: "Bernd Maur");
+      when(() => item.sender()).thenReturn(sender);
+      when(() => item.mentionsYou()).thenReturn(true);
+
+      // Act: process processing and get tile and body
+      final (title, body) = genTitleAndBody(item);
+
+      // Assert: Check if tile and body are as expected
+      expect(title, '📨 Bernd Maur invited you');
+      expect(body, "☑️ Website Redesign");
```
kumarpalsinh25, line 64: @gnunicorn 
Don't you think structure should be like below...

### 1) When mentioned You:
```
Title : "📨 Bernd Maur invited you"
Body : "On ☑️ Website Redesign"
```

### 2) Invite someone else:
```
Title : "📨 Michael May invited @id:acter.global"
Body : "On ☑️ Website Redesign"
```

### 3) Invite more than 1 person:
```
Title : "📨 Michael May invited 3 people"
Body : "On ☑️ Website Redesign"
```

kumarpalsinh25 (approved):
Apart from one notification structural suggestion which we are currently following, other changes looks good to me.

packages/acter_notifify/test/invitations_notification_test.dart
gnunicorn: well, the proper proposition here would be `To` though ;) .

kumarpalsinh25: Yupp, `To` will be wording :)

packages/acter_notifify/lib/model/push_styles.dart
gnunicorn: theoretically it is more general. but it only works on tasks at the moment. not sure - product-wise - where and how to extend it.

packages/acter_notifify/test/invitations_notification_test.dart
gnunicorn: Invites are currently one person. And I think if that isn't you, we want to keep this as an object-first-style. So I'd rather keep the different styles. I've added the `To` for invited-me though in 5ac16bbf00 .

PR #2724, opened 2025-03-17
Fix live event

gnunicorn:
Lotsa additional tests for event providers, to find the bug causing "live events" to not properly update. And a few tests showing that this is the case. And the actual fix: watching the now provider not just reading it.

Fixes https://github.com/acterglobal/a3-meta/issues/821 and https://github.com/acterglobal/a3-meta/issues/779 .


app/lib/features/events/providers/event_type_provider.dart
```diff
@@ -8,7 +8,7 @@ final eventTypeProvider = Provider.autoDispose
     .family<EventFilters, ffi.CalendarEvent>((ref, event) {
       DateTime eventStartDateTime = toDartDatetime(event.utcStart());
       DateTime eventEndDateTime = toDartDatetime(event.utcEnd());
-      DateTime currentDateTime = ref.read(utcNowProvider);
+      DateTime currentDateTime = ref.watch(utcNowProvider);
```
kumarpalsinh25, line 11: Ah, good catch!

kumarpalsinh25 (approved):

PR #2728, opened 2025-03-18
Chat-ng scrolling fixes

gtalha07:
# PR Description:

- Fixes [meta-797](https://github.com/acterglobal/a3-meta/issues/797). 
- Added scroll to bottom functionality for easier navigation to latest message when scrolling through previous messages.
- Smooth scrolling improvements when doing back-pagination.

### Mobile:

https://github.com/user-attachments/assets/d06a4952-6602-4963-ad85-a1ba42e141db

### Desktop:

https://github.com/user-attachments/assets/02c0b3c9-02e9-42ac-9c73-18c8addc28b2

### DIFF
--- a/app/lib/features/chat_ng/providers/notifiers/chat_room_messages_notifier.dart
+++ b/app/lib/features/chat_ng/providers/notifiers/chat_room_messages_notifier.dart
@@ -84,6 +84,7 @@ ChatRoomState handleDiff(
   RoomMessageDiff diff, // the diff to apply
 ) {
   final action = diff.action();
+  // the diff is applied in reverse order as the animated list is provided to UI in reverse .i.e. for improved scrolling
   switch (action) {
     case 'Append':
       List<RoomMessage> incoming =
@@ -96,14 +97,20 @@ ChatRoomState handleDiff(
       for (final m in incoming) {
         final uniqueId = m.uniqueId();
         messages[uniqueId] = m;
-        messageList.add(uniqueId);
+        messageList.insert(0, uniqueId);
       }
       final endLen = messageList.length;
-      animatedList?.insertAllItems(startLen, endLen - startLen);
+
+      animatedList?.insertAllItems(0, endLen - startLen);
+
       return state.copyWith(messageList: messageList, messages: messages);
+
     case 'Set': // used to update UnableToDecrypt message
       RoomMessage m = diff.value().expect('set diff must contain value');
       final index = diff.index().expect('set diff must contain index');
+
+      final reversedIndex =
+          state.messageList.isEmpty ? 0 : state.messageList.length - 1 - index;
 
       final uniqueId = m.uniqueId();
       if (state.messageList.isEmpty) {
@@ -111,21 +118,29 @@ ChatRoomState handleDiff(
         return state.copyWith(messageList: [uniqueId], messages: {uniqueId: m});
       }
       final messageList = state.messageList.toList();
-      final removedItem = messageList.removeAt(index);
-      messageList.insert(index, uniqueId);
+      final removedItem = messageList.removeAt(reversedIndex);
+      messageList.insert(reversedIndex, uniqueId);
 
       final messages = Map.fromEntries(
         state.messages.entries.where((entry) => entry.key != removedItem),
       );
       messages[uniqueId] = m;
+
       return state.copyWith(messageList: messageList, messages: messages);
     case 'Insert':
       RoomMessage m = diff.value().expect('insert diff must contain value');
       final index = diff.index().expect('insert diff must contain index');
-      return state.copyWithNewMessageAt(index, m, animatedList);
+
+      final reversedIndex =
+          state.messageList.isEmpty ? 0 : state.messageList.length - index;
+
+      return state.copyWithNewMessageAt(reversedIndex, m, animatedList);
     case 'Remove':
       int index = diff.index().expect('remove diff must contain index');
-      return state.copyWithRemovedMessageAt(index, animatedList);
+
+      final reversedIndex =
+          state.messageList.isEmpty ? 0 : state.messageList.length - 1 - index;
+      return state.copyWithRemovedMessageAt(reversedIndex, animatedList);
     case 'PushBack':
       RoomMessage m = diff.value().expect('push back diff must contain value');
 
@@ -134,24 +149,27 @@ ChatRoomState handleDiff(
         animatedList?.insertItem(0);
         return state.copyWith(messageList: [uniqueId], messages: {uniqueId: m});
       }
+
+      return state.copyWithNewMessageAt(0, m, animatedList);
+    case 'PushFront':
+      RoomMessage m = diff.value().expect('push front diff must contain value');
+
       return state.copyWithNewMessageAt(
         state.messageList.length,
         m,
         animatedList,
       );
-    case 'PushFront':
-      RoomMessage m = diff.value().expect('push front diff must contain value');
-      return state.copyWithNewMessageAt(0, m, animatedList);
     case 'PopBack':
       if (state.messageList.isEmpty) {
         return state;
       }
+
+      return state.copyWithRemovedMessageAt(0, animatedList);
+    case 'PopFront':
       return state.copyWithRemovedMessageAt(
         state.messageList.length - 1,
         animatedList,
       );
-    case 'PopFront':
-      return state.copyWithRemovedMessageAt(0, animatedList);
     case 'Clear':
       if (state.messageList.isNotEmpty && animatedList != null) {
         animatedList.removeAllItems((b, a) => const SizedBox.shrink());
@@ -160,16 +178,18 @@ ChatRoomState handleDiff(
     case 'Reset':
       List<RoomMessage> incoming =
           diff.values().expect('reset diff must contain values').toList();
+
       final (messageList, messages) = incoming.fold(
         (List<String>.empty(growable: true), <String, RoomMessage>{}),
         (val, m) {
           final (list, map) = val;
           final uniqueId = m.uniqueId();
-          list.add(uniqueId);
+          list.insert(0, uniqueId);
           map[uniqueId] = m;
           return (list, map);
         },
       );
+
       if (animatedList != null) {
         animatedList.removeAllItems((b, a) => const SizedBox.shrink());
         animatedList.insertAllItems(0, messageList.length);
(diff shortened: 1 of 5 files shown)


app/lib/features/chat_ng/providers/notifiers/chat_room_messages_notifier.dart
```diff
@@ -84,6 +84,7 @@ ChatRoomState handleDiff(
   RoomMessageDiff diff, // the diff to apply
 ) {
   final action = diff.action();
+  // the diff is applied in reverse order as the animated list is provided to UI in reverse .i.e. for improved scrolling
```
gtalha07, line 87: Tried various ways to fix scrolling without reversing the list but unfortunately, couldn't find a way to have optimal scroll offsets for various requirements we have for the chat .i.e. encountering offset bugs like not scrolling over all the bottom and how the pagination loading affects the UI refresh before. 

It seems like keeping `reverse:true` is only thing fixing the offset issue, which is also a normal behavior of the chat messages.

gtalha07: The change-set only addresses the updates over indexing, as we have reversed the list.

kumarpalsinh25 (commented):
Not sharing any explicit approval here even though changes addressing the exact issue. Because I am not sure whether approach of implementation is right or wrong.

Instead of managing reverse indexing and adding data from 0 indexes, why just just sorting list in reverse direction?

Because for each work which is relates to chat list, we need taken reverse indexing in consideration otherwise if we forgot then it may misbehave.

@gnunicorn
Would love to here you feedback here.

app/lib/features/chat_ng/providers/notifiers/chat_room_messages_notifier.dart
```diff
@@ -181,36 +201,32 @@ ChatRoomState handleDiff(
...
+
         final messages = Map.fromEntries(
-          state.messages.entries.where((entry) => !after.contains(entry.key)),
+          state.messages.entries.where(
+            (entry) => !removedMessages.contains(entry.key),
+          ),
         );
-        return state.copyWith(messageList: before, messages: messages);
+
+        return state.copyWith(messageList: keptMessages, messages: messages);
```
gnunicorn, line 229: this section not being covered by tests concerns me.

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -36,11 +38,9 @@ class _ChatMessagesConsumerState extends ConsumerState<ChatMessages> {
     ref.listenManual(
       chatMessagesStateProvider(
         widget.roomId,
-      ).select((value) => value.messageList.length),
-      (_, __) {
-        if (_scrollController.hasClients &&
-            _scrollController.position.pixels >
-                _scrollController.position.maxScrollExtent - 150) {
+      ).select((value) => value.messageList),
```
gnunicorn, line 41: I know this doesn't sound easy, but I think especially now that we are growing its complexity, this widget _needs tests_ for all these complex problems. all these complex problems that are being tried to address are really hard to recreate in real life and even harder to follow in any review. this needs tests.

gnunicorn (changes requested):
This needs tests. One important thing why we wanted to do chat-ng is to also have a _more robust and well tested_ section and we are adding complexity without the appropriate _large_ number of tests ensuring this works as we want it to work. It is hard to understand the impact of this by reviewing ...

I can also totally understand @kumarpalsinh25 concern about swapping the order of the diffing algorithm. But I also see that just using `reversed` isn't going to cut it as it also implicitly manages the AnimatedList-State (for fewer rerenders). That said, while other things are covered - and I feel more tests would be even better, I am concerned that if we missed anything in there, it will be hard to debug and track down - the AnimatedList changes not being covered by tests is a problem.

In general: tests, all this should be covered by tests, also higher level widget tests that make sure the providers up to the UI update properly under various different states and circumstances . For e.g. what is totally unclear to me right now: if I scroll up and a new message comes in, does it scroll back down? How would the user know there is a new message if not? Tests!

gtalha07:
> Not sharing any explicit approval here even though changes addressing the exact issue. Because I am not sure whether approach of implementation is right or wrong.
> 
> Instead of managing reverse indexing and adding data from 0 indexes, why just just sorting list in reverse direction?
> 
> Because for each work which is relates to chat list, we need taken reverse indexing in consideration otherwise if we forgot then it may misbehave.
> 
> @gnunicorn Would love to hear your feedback here.

Just sorting the list in reverse doesn't help here really, all of this mechanism is handled with diff applier which would break as it directly manipulates animated list ordering and would have to take care of indexing too.

gtalha07:
> This needs tests. One important thing why we wanted to do chat-ng is to also have a _more robust and well tested_ section and we are adding complexity without the appropriate _large_ number of tests ensuring this works as we want it to work. It is hard to understand the impact of this by reviewing ...
> 
> I can also totally understand @kumarpalsinh25 concern about swapping the order of the diffing algorithm. But I also see that just using `reversed` isn't going to cut it as it also implicitly manages the AnimatedList-State (for fewer rerenders). That said, while other things are covered - and I feel more tests would be even better, I am concerned that if we missed anything in there, it will be hard to debug and track down - the AnimatedList changes not being covered by tests is a problem.
> 
> In general: tests, all this should be covered by tests, also higher level widget tests that make sure the providers up to the UI update properly under various different states and circumstances . For e.g. what is totally unclear to me right now: if I scroll up and a new message comes in, does it scroll back down? How would the user know there is a new message if not? Tests!

Yeah, I totally understand the insecurity here. I was avoiding to take this approach by trying out various possibilities without reversal but it seemed futile.  The only thing that fixes and improves it is with reverse UI list ordering and indexing as implication of it. I'd add more tests for all code coverage here.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void main() {
+  group('ChatMessages Widget Tests', () {
+    final testRoomId = 'test-room-id';
+
+    setUp(() {
+      registerFallbackValue(const ChatRoomState());
+      registerFallbackValue(true); // Register fallback for boolean parameters
+      registerFallbackValue(false); // Register fallback for boolean parameters
```
gnunicorn, line 56: I assume this comes from some AI generation code, but are these really _both_ necessary?

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+            testRoomId,
+          ).overrideWith((ref) => GlobalKey<AnimatedListState>()),
+        ],
+        child: ChatMessages(roomId: testRoomId),
+      );
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+    });
+
+    testWidgets('hides loading indicator when not loading', (tester) async {
```
gnunicorn, line 147: this is not what the tests checks for. It checks whether the loading animation isn't show _if it is started that way_. but I would like it to test what the test says instead: that it is shown in loading and when _changed_ switches to not showing. Maybe we can update the tests to check for that?

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+          ).overrideWith((ref) => animatedListKey),
+        ],
+        child: ChatMessages(roomId: testRoomId),
+      );
+
+      // Update with messages to trigger auto-scroll
+      final updatedState = initialState.copyWith(
+        messageList: List.generate(5, (index) => 'msg$index'),
+      );
+      notifier.state = updatedState;
```
gnunicorn, line 204: does this actually work? I thought it was necessary to do actually modify the notifier via the `ref` and not just standalone. And after it is not tested in here that the change actually did anything on the object, no check whether more items are rendered. I suspect the test passes because this change doesn't actually happen. Can we have additional checks that the update actually took place.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+      final updatedState = initialState.copyWith(
+        messageList: [...initialState.messageList, 'newMessage'],
+      );
+      notifier.state = updatedState;
+
+      await tester.pump();
+      await tester.pump(const Duration(milliseconds: 300));
+
+      // verify the list was updated
+      expect(animatedListKey.currentState?.widget.initialItemCount, equals(4));
```
gnunicorn, line 301: this doesn't actually check for what the comment says. it is checking only whether the incoming parameter was updated, not whether the actual list items updated. I am not sure we can actually check for the items of the state but this tests creates a false sense of security that will make debugging harder as it would show a false negative about the actual state of the system.

In such a case I'd rather have the diffing test extended with a mocked animatedList that is checking for the right methods to be called. In tandem they can be fairly accurate then.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+
+      await (finalState as dynamic).scrollToEnd();
+
+      await tester.pump();
+      //wait for the animation to complete
+      await tester.pumpAndSettle();
+
+      expect(scrollController?.position.pixels, equals(0.0));
+      // button should be hidden
+      expect((finalState as dynamic).showScrollToBottom, isFalse);
```
gnunicorn, line 357: are these `as dynamic` in here really necessary? It seems to me like we know the proper types here and call proper functions on them. I assume this is AI generated code? But we should clean that up regardless. That's a lot of `dynamic` on _known_ types.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+    });
+
+    testWidgets('preserves correct message order when displayed in reverse', (
```
gnunicorn, line 360: does this matter to us (at all)? this feels like a pointless test of whether the builtin `reverse` of dart lists work. Let's drop this test.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
```diff
@@ -0,0 +1,406 @@
...
+      expect(animatedList.initialItemCount, equals(5));
+
+      // Check the order of messages in the state
+      expect(
+        mockNotifier.state.messageList,
+        equals(['msg0', 'msg1', 'msg2', 'msg3', 'msg4']),
+      );
+    });
+  });
+}
```
gnunicorn, line 406: so close! the chat room message notifier in codecov just complains about one line never being met, and that is removed items. Maybe we can add a tests for removing items as well?

gnunicorn (approved):
Thanks for adding tests. But some of them need some fixes/cleaning up.

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
gtalha07: Yes, I've modified the test case to do actual update on notifier and added additional checks ! 👍🏻

app/test/features/chat_ng/widgets/chat_messages_list_test.dart
gtalha07: changed the test scenario which now includes switches loading and not loading state while checking indicator state synchronously! 👍🏻

PR #2732, opened 2025-03-19
Replace wording Boost with Updates

kumarpalsinh25:


gnunicorn (approved):
quick actions (and space actions?) also need an update

kumarpalsinh25:
> quick actions (and space actions?) also need an update

Screenshot 2025-03-19 at 5 07 47 PM

Space action already have separate buttons as there are different conditional thins associated with it.

Will add story button in Quick actions.

PR #2735, opened 2025-03-19
Onboarding : Encryption backup screen with Auto Enable Backup

kumarpalsinh25:
Enable ENCRYPTION BACKUP automatically for new users on sign-up and show screen to take back-up of the same.

Screenshot Reference:
Back-up screen

Video Reference:

https://github.com/user-attachments/assets/39a98074-0358-41a9-ba9a-5db801f72480

Note:
- At the moment, I have only added option to copy key and share key as action buttons, but in later PRs we can add options for support to specific Apps (if available) to have it stored in, like Keepass(X), 1Password, Bitwarden, ProtonPass, LastPass ...
- We aren't storing key anywhere in local encryption store, just enabling backup automatically and showing option to take back-up externally.


app/lib/features/onboarding/pages/encrption_backup_page.dart
```diff
@@ -0,0 +1,187 @@
+import 'package:acter/common/utils/routes.dart';
```
gnunicorn, line 1: how about some tests?

app/lib/features/onboarding/pages/encrption_backup_page.dart
```diff
@@ -0,0 +1,187 @@
...
+    super.initState();
+    enableBackup(context);
+  }
+
+  void enableBackup(BuildContext context) async {
+    try {
+      final manager = await ref.read(backupManagerProvider.future);
+      encryptionKey.value = await manager.enable();
+    } catch (e) {
+      encryptionKey.value = e.toString();
```
gnunicorn, line 35: eh... probably not good idea to store the string of the error as the value. Tests would show that in case of an error you might want to allow the user to retry, not let them assume things are fine with a completely broken key..

app/lib/features/onboarding/pages/encrption_backup_page.dart
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  @override
+  void dispose() {
+    isShowNextButton.dispose();
```
gnunicorn, line 41: why is this needed?

app/lib/features/onboarding/pages/encrption_backup_page.dart
```diff
@@ -0,0 +1,187 @@
...
+    return Container(
+      padding: const EdgeInsets.symmetric(vertical: 16, horizontal: 24),
+      decoration: BoxDecoration(
+        color: Theme.of(context).colorScheme.surfaceContainerLow,
+        borderRadius: BorderRadius.circular(12),
+      ),
+      child: ValueListenableBuilder<String>(
+        valueListenable: encryptionKey,
+        builder: (context, key, _) {
+          if (key.isEmpty) {
```
gnunicorn, line 104: I know you like your value notifiers, but this is an exemplary case for while a future-provider/notifier system  - or at an `AsyncValue` at least - is more advanced and fitting for the use case: the error case is handled completely wrong...

gnunicorn (changes requested):
error cases are mishandled and we should have tests for this.

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: Sure, let me add it.

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: Let me handle it in better way.

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: I was testing something with `ValueNotifier` that's added it but let me remote it.

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: I am generally using `ValueNotifier` when things are limit to specific class or screen only. But that also true that `Providers/Notifier` are more advance. Let me do needful here.

kumarpalsinh25:
@gnunicorn 
As [Implement encryption backup key storage](https://github.com/acterglobal/a3/pull/2742) is merged, let me implement those changes here in this along with the given PR Feedback points.

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: Error state reference.

Screenshot 2025-03-21 at 12 10 45 PM

app/lib/features/onboarding/pages/encrption_backup_page.dart
```diff
@@ -92,39 +65,55 @@ class _EncryptionBackupPageState extends ConsumerState<EncryptionBackupPage> {
...
-            ).textTheme.bodyMedium?.copyWith(fontSize: 16, letterSpacing: 1.2),
-            textAlign: TextAlign.center,
-          );
-        },
-      ),
+      child: SelectableText(encKey, style: style, textAlign: TextAlign.center),
+    );
+  }
+
+  Widget _buildEncryptionKeyError(BuildContext context, String error) {
```
gnunicorn, line 101: we have an acter error button widget with a retry option... why don't we use this here and allow the user to report and/or retry in case of an error?

also ... if an error occurred, the user will never be able to move forward, will they? they will be stuck here forever ... that doesn't fly...

app/lib/features/backups/providers/backup_manager_provider.dart
```diff
@@ -6,3 +6,23 @@ final backupManagerProvider = FutureProvider(
     alwaysClientProvider.selectAsync((client) => client.backupManager()),
   ),
 );
+
+final storedEncKeyProvider = FutureProvider((ref) async {
```
gnunicorn, line 10: not sure this is the best way of organizing these tbh - feels like this should be a full notifier with an internal state rather than separate providers, but I won't insist on that here.

app/lib/features/backups/providers/backup_manager_provider.dart
```diff
@@ -6,3 +6,23 @@ final backupManagerProvider = FutureProvider(
...
+  final manager = await ref.watch(backupManagerProvider.future);
+  return manager.storedEncKey();
+});
+
+final enableEncrptionBackUpProvider = FutureProvider((ref) async {
+  final manager = await ref.watch(backupManagerProvider.future);
+  return manager.enable();
```
gnunicorn, line 17: doesn't this also fail if the state is already enabled? I know this isn't technically possible on registration, but if this is a generic provider we should cover those cases and problems ...

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: > we have an acter error button widget with a retry option... why don't we use this here and allow the user to report and/or retry in case of an error?

Ok let me use that widget here.

> also ... if an error occurred, the user will never be able to move forward, will they? they will be stuck here forever ... that doesn't fly...
Provided skip button so user will not get stuck in case of error.

app/lib/features/backups/providers/backup_manager_provider.dart
kumarpalsinh25: I am not getting on it and I am not using it at the moment in this PR so let me remove this provider for now.

app/lib/features/backups/providers/backup_manager_provider.dart
kumarpalsinh25: Then this provider will throw error which will be display to user.

If I add condition like `isEnable` then what to do in else case?

gnunicorn: the error should be shown to the user, sure. but they should be able to retry or skip :) .

app/lib/features/onboarding/pages/encrption_backup_page.dart
kumarpalsinh25: Screenshot 2025-03-21 at 7 46 31 PM


Probably this error dialog widget won't work in case of this screen as it is part of onboarding and there is no back machanism.

gnunicorn: for this error it would be sufficient, but assume the problem was teh server wasn't availalbe. I'd argue in that case `next` should be `retry`, don't you think?

kumarpalsinh25: What to do on click of `retry` ?

kumarpalsinh25: Not every time error need Retry option, some times error can be just an informations without requiring additional call of actions.

kumarpalsinh25:
@gnunicorn
Added retry button as requested.

https://github.com/user-attachments/assets/19ccf351-2602-4258-bc16-28fe7a166a8e

gnunicorn (approved):

PR #2742, opened 2025-03-20
Implement encryption backup key storage

gnunicorn:
This PR changes the behavior of the backup manager to store the last used encryption key in the encrypted local storage for latter retrieval. It is overwritten when the backup is reset locally and will only return if the storage backup in not disabled. Tests cover all major uses.

### DIFF
--- a/native/acter/api.rsh
+++ b/native/acter/api.rsh
@@ -3660,4 +3660,13 @@ object BackupManager {
     /// Open the existing secret store using the given key and import the keys
     fn recover(secret: string) -> Future<Result<bool>>;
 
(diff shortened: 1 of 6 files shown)


gnunicorn:
@kumarpalsinh25  after this is merged you can easily retrieve the current key from teh backup manager `storedEncKey` (and `storedEncKeyWhen` for the recorded time) and show a helper in the activity screen if one is found allowing the user to back it up else where and when done call `destroyStoredEncKey` to get rid of it.

bitfriend (approved):
Looks good to me

PR #2745, opened 2025-03-21
My Invitation Model system

gnunicorn:
This rewrites the room invitations API's to provide a unified entry point for room and object invitations. It reactivates the integration tests for the room invitations under the new layout, updates the flutter API to use them and eventually ensures that the object invitation will become accessible through the same API's for showing them in the future.


bitfriend (approved):
Nice job.

PR #2750, opened 2025-03-25
Improvements : Space Creation Flow

kumarpalsinh25:
Improvements in the Space Creation flow by giving different configuration options at the time of the space creation.

### Reference Sketch:
Screenshot 2025-03-28 at 1 09 37 PM

So with the new flow in space creation, you can..
- Create space quickly by providing basic details like space name and description.
- You can also configure below things additional along with the space creation.
   - Default Chat 
   - Parent Space Selection
   - Space Visibility
   - (NEW) - Space Feature Activation with permissions configurations.
   
### Reference Video :
  
https://github.com/user-attachments/assets/279c6376-843c-4106-8882-33c0654a60b5


app/lib/features/spaces/widgets/create_space/space_features_widget.dart
```diff
@@ -164,21 +191,29 @@ class _SpaceFeaturesWidgetState extends ConsumerState<SpaceFeaturesWidget> {
               ),
           child: Text(
             permissionItem.permissionLevel.name.toUpperCase(),
```
gnunicorn, line 193: no l10n for this?

gnunicorn (approved):
Code-wise fine, But where are the default set of settings declared? I think we want to review that because all-off is not the usual set we want to start with.

app/lib/features/spaces/widgets/create_space/space_features_widget.dart
kumarpalsinh25: Yes sure, let me do the needful here.

kumarpalsinh25:
> Code-wise fine, But where are the default set of settings declared? I think we want to review that because all-off is not the usual set we want to start with.

From `featureActivationStateProvider`, I can easily set the default values for Activations and also permissions.

Screenshot 2025-04-01 at 3 40 55 PM

Can you guide me on what default activation and permissions I need to set?

app/lib/features/spaces/widgets/create_space/space_features_widget.dart
kumarpalsinh25: Resolved by https://github.com/acterglobal/a3/pull/2750/commits/92cfcb4e73dd8c6eb555452f80d572241351dfe0

app/lib/features/spaces/providers/space_creation_providers.dart
```diff
@@ -0,0 +1,33 @@
...
+// create default chat provider
+final createDefaultChatProvider = StateProvider<bool>((ref) => false);
+
+// create default chat provider
+final showSpaceCreationConfigurationProvider = StateProvider<bool>(
+  (ref) => false,
+);
+
+// Feature activation state provider
+final featureActivationStateProvider = StateProvider<
```
gnunicorn, line 21: how is this being reset after submission?

gnunicorn:
> @kumarpalsinh25 Can you guide me on what default activation and permissions I need to set?

I wanted to directly do it on the code, but could really find it. Maybe a dedicated `generateRecommendedSettings()` fn would be useful? Anyways the permissions should be: all on, with boosts set to moderators and all else to `Everyone`.

kumarpalsinh25:
> I wanted to directly do it on the code, but could really find it. Maybe a dedicated generateRecommendedSettings() fn would be useful? Anyways the permissions should be: all on, with boosts set to moderators and all else to Everyone.

Yes agree on your thought. With below commit, I have created separate class for managing default feature activation and permission and it is easy find it.
https://github.com/acterglobal/a3/pull/2750/commits/0311f66dac25d9aa848351c3bf1a38722c8614bb

Also enable all the feature ON by default and sets permissions as suggested.

PR #2746, opened 2025-03-24
Space Membership Changes added in activity.

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2669

PR #2741, opened 2025-03-20
Space Detail Changes added in activity

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2668

Followed this UI: 
https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=3482-9&p=f

✅ THIS PR COMES WITH WIDGET TESTING OF SPACE ITEM WIDGETS IN ACTIVITY.

This PR includes below space details changes in activities,

- Space avatar change
- Space title change
- Space description change

### Reference Image : 

Screenshot 2025-04-21 at 5 30 38 PM

PR #2740, opened 2025-03-20
Space Objects changes added in activities.

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2667

Followed this UI:
https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=3482-9&p=f

✅ THIS PR COMES WITH WIDGET TESTING OF OBJECTS IN ACTIVITY.

This PR includes below space details changes in activities,

- Object creation
- Object titleChange
- Object descriptionChange
- Object otherChanges

### Reference video :

https://github.com/user-attachments/assets/12476221-959d-4ae8-a10a-bd1fc064ccb2

PR #2751, opened 2025-03-25
Split m.room.member event handler into ProfileChange into MembershipChange

bitfriend:
Will split `m.room.member` into `ProfileChange` and `MembershipChange` like `matrix_sdk_ui`.


bitfriend:
This PR is intended for https://github.com/acterglobal/a3/pull/2740, https://github.com/acterglobal/a3/pull/2741, https://github.com/acterglobal/a3/pull/2746.

native/acter/api.rsh
```diff
@@ -1115,6 +1115,43 @@ object MsgContent {
 
     /// the list of url previews 
     fn url_previews() -> Vec<UrlPreview>;
+
+    /// available when user changed his profile
+    fn profile_change() -> Option<ProfileChange>;
+
+    /// available when a user changed another user
+    fn membership_change() -> Option<MembershipChange>;
+}
```
gnunicorn, line 1124: I'd rather not plug these into `MsgContent`, they aren't . How about we have them on the activity / notifications & RoomEvent directly - these aren't "messages", so "message content" is not the place  I'd look for that at.

native/acter/src/api/common.rs
```diff
@@ -188,6 +193,56 @@ pub enum MsgContent {
...
+                .displayname_change()
+                .and_then(|c| Change::new(c.new.clone(), c.old.clone())),
+            avatar_url_change: value
+                .avatar_url_change()
+                .and_then(|c| Change::new(c.new.clone(), c.old.clone())),
+        }
+    }
+}
+
+impl From<&RoomMembershipChange> for MsgContent {
```
gnunicorn, line 221: assuming nothing much has changed here.

native/acter/src/api/message.rs
```diff
@@ -194,115 +195,23 @@ impl RoomEventItem {
...
-                    }
-                    Some(MembershipChange::KnockDenied) => {
-                        me.msg_type(Some("KnockDenied".to_string()));
-                        "denied knock".to_string()
-                    }
-                    Some(MembershipChange::NotImplemented) => {
-                        me.msg_type(Some("NotImplemented".to_string()));
-                        "not implemented change".to_string()
+                me.event_type("membershipChange".to_string()); // some of m.room.member
+                let msg_content = match m.content() {
```
gnunicorn, line 199: as said above, I don't think this should be a msg_content.

native/core/src/models/status.rs
```diff
@@ -60,20 +71,60 @@ impl TryFrom<AnyStateEvent> for RoomStatus {
...
+                            old_val: c.old.map(ToOwned::to_owned),
+                        }),
+                        avatar_url_change: avatar_url_change.map(|c| Change {
+                            new_val: c.new.map(ToOwned::to_owned),
+                            old_val: c.old.map(ToOwned::to_owned),
+                        }),
+                    }
+                } else {
+                    let change = match membership_change {
+                        MChange::None => "None",
```
gnunicorn, line 99: this if else, match should be one big match instead IMHO

native/core/src/models/status/membership.rs
```diff
@@ -1,147 +1,131 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ProfileChange {
+    user_id: OwnedUserId,
+    display_name_change: Option<Change<Option<String>>>,
+    avatar_url_change: Option<Change<Option<OwnedMxcUri>>>,
+}
 
-    /// User knocked.
-    Knocked,
+impl ProfileChange {
```
gnunicorn, line 31: very few of the newly added lines in this file are covered by tests. Please extend the tests to include these lines.

gnunicorn (changes requested):
I have a request regarding the API and additional tests on the rust side. Once they are applied happy to review the dart side as well, but kinda pointless before hand ;) .

native/acter/api.rsh
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/2751/commits/a085729d8603818c361be06ab1978f05f1a0fff0.
Renamed `msg_content()` API to `message()` in order to highlight it is room message and distinguish it from room state events.
So changed some flutter files for api change.

native/acter/src/api/common.rs
bitfriend: Moved it into `/native/core/src/models/status/membership.rs`, because activity is implemented in `core` module and we have to support space state event for activity.

native/core/src/models/status.rs
bitfriend: Appended room state event handler for core object via https://github.com/acterglobal/a3/pull/2751/commits/c007e3dc00a0dde7eeac7035b6e7a99d9ed51be1.

bitfriend: I could extend `ActerSupportedRoomStatusEvents` and `ActivityContent` easily by using `OtherState`.

app/lib/common/utils/room_state.dart
```diff
@@ -0,0 +1,1637 @@
+import 'package:acter/common/extensions/options.dart';
+import 'package:acter/l10n/generated/l10n.dart';
+import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
+
+String? getStateOnPolicyRuleRoom(
```
gnunicorn, line 5: Tried to find out what this used for, as it looks like UI code more than anything else and see that it is being shown in both the existing chat and the chat-ng. But the way this approached isn't the way we want to continue to do stuff in chat-ng: rather than compile some weird arbitrary metadata and strings from the input, in chat-ng we hand the actual model to an ever more fine-grained widget that knows how to render _this specific_ scenario.

I will elaborate more about this on the chat-ng side of things.

This file in general should probably be gone, as it mainly serves as a place for reusability of chat and chat-ng, while the point of chat-ng _was to rewrite in a new pattern_, which means _reusing_ with legacy chat should basically not happen _ever_ but instead we copy over and modify. This file in its entirely is making this approach harder.

app/lib/features/chat_ng/widgets/events/chat_event_item.dart
```diff
@@ -69,19 +69,18 @@ class ChatEventItem extends StatelessWidget {
...
-      'm.room.member' || 'ProfileChange' => MemberUpdateEvent(
+      'membershipChange' || 'profileChange' => MemberUpdateEvent(
         isMe: isMe,
         roomId: roomId,
         item: item,
       ),
       'm.policy.rule.room' ||
       'm.policy.rule.server' ||
       'm.policy.rule.user' ||
-      'm.poll.start' ||
```
gnunicorn, line 80: what happened to the poll support?

native/acter/src/api/timeline/message.rs
```diff
@@ -0,0 +1,986 @@
+use acter_core::models::status::{
```
gnunicorn, line 1: more tests needed. please look at this file in review mode to see which parts a not covered.

native/core/src/models/status/room_state.rs
```diff
@@ -0,0 +1,1513 @@
...
+    RoomJoinRules(RoomJoinRulesContent),
+    RoomName(RoomNameContent),
+    RoomPinnedEvents(RoomPinnedEventsContent),
+    RoomPowerLevels(RoomPowerLevelsContent),
+    RoomServerAcl(RoomServerAclContent),
+    RoomThirdPartyInvite(RoomThirdPartyInviteContent),
+    RoomTombstone(RoomTombstoneContent),
+    RoomTopic(RoomTopicContent),
+    SpaceChild(SpaceChildContent),
+    SpaceParent(SpaceParentContent),
```
gnunicorn, line 58: Lacks Acter specific changes for `ActerAppsSettings`.

native/acter/src/api/timeline/message.rs
```diff
@@ -0,0 +1,986 @@
...
+    pub fn space_parent(&self) -> Option<SpaceParentContent> {
+        if let Some(TimelineEventContent::OtherState(OtherState::SpaceParent(c))) = &self.content {
+            Some(c.clone())
+        } else {
+            None
+        }
+    }
+}
+
+impl TimelineEventItemBuilder {
```
gnunicorn, line 656: tests....

native/core/src/activities.rs
```diff
@@ -123,9 +132,34 @@ impl Activity {
 
     pub fn type_str(&self) -> String {
         match &self.inner {
-            ActivityContent::MembershipChange(c) => c.as_str(),
-            ActivityContent::RoomCreate(_) => "roomCreate",
-            ActivityContent::RoomName(_) => "roomName",
+            ActivityContent::ProfileChange { .. } => "profileChange",
+            ActivityContent::MembershipChange { .. } => "membershipChange",
```
gnunicorn, line 136: I don't like this change. The more specific `left` type has been wrapped into a more opaque "membershipChange" type that the UI then first has to figure out again that this was actually someone leaving rather than just updating their display name. This is inconvenient for the UI to build around. In general we took the approach to be as specific as possible as highly in the layers as possible, so this should return `left` and then can hold internally the `membershipChange`-object that is handed out for details, but it shouldn't be required to go that route.

native/core/src/activities.rs
```diff
@@ -123,9 +132,34 @@ impl Activity {
...
+            ActivityContent::RoomState(OtherState::RoomPowerLevels(_)) => "roomPowerLevels",
+            ActivityContent::RoomState(OtherState::RoomServerAcl(_)) => "roomServerAcl",
+            ActivityContent::RoomState(OtherState::RoomThirdPartyInvite(_)) => {
+                "roomThirdPartyInvite"
+            }
+            ActivityContent::RoomState(OtherState::RoomTombstone(_)) => "roomTombstone",
+            ActivityContent::RoomState(OtherState::RoomTopic(_)) => "roomTopic",
+            ActivityContent::RoomState(OtherState::SpaceChild(_)) => "spaceChild",
+            ActivityContent::RoomState(OtherState::SpaceParent(_)) => "spaceParent",
+
```
gnunicorn, line 162: as said above, we decided for a highly-specific-naming approach while most of these aren't that specific but rather descriptive. Has a roomName been set or unset? has the topic been changed or removed?

native/core/src/models/status/membership.rs
```diff
@@ -1,147 +1,162 @@
...
+    pub(crate) change: Option<String>,
+}
 
-    /// User rejected the invite.
-    InvitationRejected,
+impl From<&RoomMembershipChange> for MembershipChange {
+    fn from(value: &RoomMembershipChange) -> Self {
+        MembershipChange {
+            user_id: value.user_id().to_owned(),
+            change: value.change().map(|c| {
```
gnunicorn, line 37: why remove the neatly organized and typed enum and replace it with  less easy to enforce `String` here? what's the benefit?

native/core/src/models/status/room_state.rs
```diff
@@ -0,0 +1,1513 @@
...
+        child::{PossiblyRedactedSpaceChildEventContent, SpaceChildEventContent},
+        parent::{PossiblyRedactedSpaceParentEventContent, SpaceParentEventContent},
+    },
+};
+use serde::{Deserialize, Serialize};
+
+use crate::util::do_vecs_match;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum OtherState {
```
gnunicorn, line 38: I don't understand the point of this enum. Next to the bad name (it should probably be `RoomState` not just `OtherState`), it seems to be only used where we already have an enum of different states as an additional wrapper that is the unwrapped.

native/acter/api.rsh
```diff
@@ -1022,8 +1022,80 @@ object RoomEventItem {
...
+    fn message() -> Option<MsgContent>;
+
+    /// covers m.sticker
+    fn sticker() -> Option<Sticker>;
+
+    /// covers some of m.room.member
+    fn membership_change() -> Option<MembershipChange>;
+
+    /// covers some of m.room.member
+    fn profile_change() -> Option<ProfileChange>;
```
gnunicorn, line 1035:
```suggestion
    fn user_profile_change() -> Option<ProfileChange>;
```

native/acter/api.rsh
```diff
@@ -1022,8 +1022,80 @@ object RoomEventItem {
...
+    fn room_tombstone() -> Option<RoomTombstoneContent>;
+
+    /// covers room state change
+    fn room_topic() -> Option<RoomTopicContent>;
+
+    /// covers room state change
+    fn space_child() -> Option<SpaceChildContent>;
+
+    /// covers room state change
+    fn space_parent() -> Option<SpaceParentContent>;
```
gnunicorn, line 1095: can't mark all here but from a API usage perspective, I'd rather have the functions of 1037-1095 on a separate `RoomStateEvent`-wrapper object which could also hand out for `Activity` then.

gnunicorn (changes requested):
I already complained that this is _way_ to large for review. Several 10th of thousands of lines changed are unacceptable to put of review. These _must_ be split up into separate PRs of specific parts. Piling on existing PRs is never a good idea, instead please make PRs that build on top of each other. Which also allows for finding things developing into the wrong direction early and prevent work going to waste because of a faulty approach and the subsequent rewrite that needs to happen.


I have to request that this PR is split up. There are several good parts in here:
1. the refactor of the rust types to provider a better more appropriately named pattern :+1: 
2. support for state events (though some API changes are requested)


But there are also many things that changing to make them acceptable will require several further large-scale reviews, I am not willing to make:
1. the entire `room_state.dart`-pattern of reuse in legacy chat and chat-ng
2. thousand+ line files without tests
3. specificity of the added types, rather than adding one by one adding them in a big batch means they weren't properly thought through.
4. minor changes to the way this API is exposed.

bitfriend:
This PR was replaced with another smaller PRs.

PR #2755, opened 2025-03-26
Implement new Permission defaults and API

gnunicorn:
We want stories to be default-on on spaces we create. this adds a new API to do that right at initialization, including power level settings support. Has tests!

### DIFF
--- a/native/core/src/spaces.rs
+++ b/native/core/src/spaces.rs
@@ -1,5 +1,8 @@
 use derive_builder::Builder;
-use matrix_sdk::{room::Room, ruma::events::room::join_rules::JoinRule};
+use matrix_sdk::{
+    room::Room,
+    ruma::{events::room::join_rules::JoinRule, Int},
+};
 use matrix_sdk_base::{
     ruma::{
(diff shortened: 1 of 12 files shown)


native/core/src/spaces.rs
```diff
@@ -223,7 +240,7 @@ impl CoreClient {
                 // local uri
                 let path = PathBuf::from(avatar_uri);
                 let guess = mime_guess::from_path(path.clone());
-                let content_type = guess.first().expect("don’t know mime type");
+                let content_type = guess.first().expect("don't know mime type");
```
bitfriend, line 243: Frankly, I used prime instead of single quote, because single quote is default of flutter side and double quote is default of rust side.

bitfriend (approved):
Looks good to me

native/core/src/spaces.rs
gnunicorn: this was just the editor auto-setting it. If we want to ensure a specific style we should put that into an editorconfig or something.

PR #2774, opened 2025-04-01
Fixes : Invitation list not update after accepting invite

kumarpalsinh25:
Fixes, https://github.com/acterglobal/a3-meta/issues/894

I have did missing `ref.invalidate(invitationListProvider);` on Accept Invite success and Error cases and which is fixes the Invitation list updation issues.

Reference Video:

https://github.com/user-attachments/assets/9a1eece1-b48d-4dbb-bafa-02eb2944d5da

Side-note:
- However randomly I notice one issue that even after successfully accepting invite, list is not updating and showing invitation and when we click on it then it is showing below error.
`Failed to accept invite: Unable to join a room we are not invited`

Screenshot 2025-04-01 at 2 17 32 PM

- At the moment everything seems to be working fine after `ref.invalidate(invitationListProvider);` changes and not able to see any at the moment.

### DIFF
--- a/app/lib/features/invitations/widgets/invitation_item_widget.dart
+++ b/app/lib/features/invitations/widgets/invitation_item_widget.dart
@@ -164,6 +164,7 @@ class _InvitationWidgetState extends ConsumerState<InvitationItemWidget> {
     final client = await ref.read(alwaysClientProvider.future);
     try {
       await widget.invitation.accept();
+      ref.invalidate(invitationListProvider);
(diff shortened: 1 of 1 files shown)


gnunicorn (approved):
the PR is fine but won't fix the problem. which is in the underlying API on how the list of invitations is compiled and send to the UI. that is just an old code that hasn't been fixes in a while. I had it already on my menu somewhere but didn't finish the job...

PR #2777, opened 2025-04-01
Rename room message to timeline item

bitfriend:
Will rename `RoomMessage` to `TimelineItem`, `RoomEventItem` to `TimelineEventItem`, `RoomVirtualItem` to `TimelineVirtualItem`.
We are going to append sticker handler or room state event handler in another PRs.
We don't want that sticker handler or room state event handler is included to `MsgContent`.
We are going to keep them on the equal level to room message.

This PR was split from https://github.com/acterglobal/a3/pull/2751.


gnunicorn (approved):
no logic changes to review. rename is fine.

Issue #2772, opened 2025-04-01
Add push notification permission.


anisha-e10:
![Image](https://github.com/user-attachments/assets/77e5eb77-cd00-4ed0-8b86-de943ae15f0e)

Figma reference : 

https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=3613-11394&p=f&t=wjqLSY7DavUQ9WEw-0

PR #2778, opened 2025-04-02
Push Notification Permission startup wizzard added with UI

anisha-e10:
Implemented,https://github.com/acterglobal/a3/issues/2772

- **This PR introduces a new step in the onboarding flow** : A Push Notification Permission startup wizard. The wizard presents a prompt to users, asking for their permission to enable push notifications. 

### Reference video for login flow (Don't allow permission):

https://github.com/user-attachments/assets/8e011233-838c-4f49-8450-ab66f4436df7

### Reference video for login flow  (Allow permission):

https://github.com/user-attachments/assets/018c3258-87db-483b-bf7b-e0870faf0f4d

### Reference video for notification setting flow  (Allow permission):

https://github.com/user-attachments/assets/de1f32ab-2efe-4edb-80c0-97b7e80d9b59

### Reference video for notification setting flow  (Don't Allow permission):

https://github.com/user-attachments/assets/dbdc6949-f51d-4f65-8ff4-42db668a90b2


app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -74,13 +78,58 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+      onPressed: () => _onSkipPressed(context),
       child: Text(
         L10n.of(context).done,
         style: Theme.of(context).textTheme.bodyMedium,
       ),
     );
   }
+
+  Future<void> _onSkipPressed(BuildContext context) async {
+    if (Platform.isAndroid) {
```
gnunicorn, line 92: as pointed out above, this shouldn't happen in this specific page but instead be moved to a single instance that deals with all the missing states upon login or startup and compiles the list of pages that need to be dealt with.

As the overall requests shows why this is a better approach. This needs to move ...

app/test/features/notifications/notification_permission_page_test.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:acter/common/toolkit/buttons/primary_action_button.dart';
+import 'package:acter/features/notifications/pages/notification_permission_page.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:phosphor_flutter/phosphor_flutter.dart';
+import '../../helpers/test_util.dart';
+
+void main() {
+  testWidgets('renders correctly with all UI elements', (tester) async {
```
gnunicorn, line 9: :+1:

gnunicorn (approved):
The order is incorrect though. It should be asking for the permissions first, and then go to the analytics screen.

Which brings up the point to move the _order_ and _which_ screens a user still has to complete into a single instance elsewhere then bind each page together like this. So that these checks if a page should be shown are not bound to the goNamed-call of the previous page but dealt with close to the actual page.

app/lib/features/notifications/pages/notification_permission_page.dart
```diff
@@ -0,0 +1,244 @@
...
+    );
+  }
+
+  // AppBar for the Notification Permission Page
+  AppBar _buildAppBar(BuildContext context) {
+    return AppBar(
+      actions: [
+        IconButton(
+          icon: Icon(Icons.close),
+          onPressed: () => context.goNamed(Routes.main.name),
```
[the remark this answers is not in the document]
kumarpalsinh25: Yes agree, Notification permission screen should be independent of any routing so that I can call it from anywhere based on the requirement.

For Example, calling it before we open notification screen.
Screenshot 2025-04-02 at 7 46 49 PM

As notification screen only make sense when you have granted notification permission.

And in current case, if user have by chanced disabled notification from settings then we aren't not guide or providing enough information from our app that permission is not granted.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -74,13 +78,58 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          // Handle successful login, navigate as needed
+          context.goNamed(Routes.main.name);
+        }
+      }
+    } else {
+      if (context.mounted) {
+        // Handle successful login, navigate as needed
+        context.goNamed(Routes.main.name);
+      }
+    }
```
kumarpalsinh25, line 110: What about iOS devices?

kumarpalsinh25: And this entire code (Which checks platform specific things for notifications) should be at some generic place like in `utils/device_perimissions/notification.dart` so that we can use common function to trigger notification permission.

kumarpalsinh25 (changes requested):
As we are working on Push Notification screen, I would like also cover different use-cases around that.

### 1) ASK Permission when needed instead of Directly asking at app start:

https://github.com/user-attachments/assets/769b6d40-0cd9-423e-b763-caeeddf2f9c8

- Currently we are asking for Notification permission right after app start which we should avoid.
- Notification permission only requires after user logged in or sign-up.
- So ask for it after login/signup instead of right after app start.
- As many of the user will decline it as they are new to app and they want to provide any permission without seeing actual need of that.
- And we aren't showing any indication of Permission is not grated in anywhere in our app, and at the end user is not receiving push notification.


2) Notification Settings screens :

https://github.com/user-attachments/assets/833b658d-c8d9-468e-bd6b-6fc49e67b04c

- We have notification settings screen to manage multiple things related push services.
- But we aren't providing any indication if notification permission is not granted from the system. 
- Without notification permissions, none of the settings make sense.
- So I would like either so indication that  Notification is not granted or open notification screen only if permission granted and else show this new request permission screen.

anisha-e10:
@kumarpalsinh25

I have updated the flow of notification permissions in the onboarding process as well as in the settings screen. Please review my PR, as I have re-requested your review.

I have also updated the video in the description for the PR.

app/lib/features/auth/pages/login_page.dart
```diff
@@ -209,8 +210,20 @@ class _LoginPageState extends ConsumerState<LoginPage> {
...
+      if (!mounted) return;
+
+      // Request notification permission
+      await handleNotificationPermission(context);
+
+      // Check if context is still valid after async operation
+      if (!mounted) return;
+
+      // Navigate to the stored route
+      navigator.goNamed(routeToNavigate);
```
kumarpalsinh25, line 226:
```suggestion
      // Check if context is still valid
      if (!mounted) return;

      // Request notification permission
      await handleNotificationPermission(context);

      // Check if context is still valid after async operation
      if (!mounted) return;
      
      // Navigate to the analyticsOptIn route
      navigator.goNamed(Routes.analyticsOptIn.name);
```

kumarpalsinh25 (approved):
Looks much better now! Nice work!

PR #2780, opened 2025-04-02
A better way to generate space tabs

### LINKED ISSUE #2695: Space Overview doesn't load when offline
We changed the infrastructure to _enforce_ that we fetch _and wait for_ the remote room hierarchy when opening a space overview, regardless of whether we need it. Leading to the screen becoming unusable in offline scenarios. See https://github.com/acterglobal/a3-meta/issues/833 .

It also fails with a message about the url not being able to load, but still doesn't give access to the screen. Instead the menu must be refactored to allow for the spaces- and chats-sections to not having reported their status yet (or report a temporary status) but allow accessing and interacting with the space overview page already.

gnunicorn:
Fixes #2695 and #2375 .

This takes a fresh approach to loading the space overview screen: load without errors.

By replacing the existing mechanism to generate the space features bar (with tests) and some other providers, this is able to always show stuff instead of 404ing or failing if things couldn't be loaded. It makes errors a bit more opaque though :( .

But most importantly, it means faster loading and easier access of a space page, while offline:

https://github.com/user-attachments/assets/17930e10-d7ae-475b-a141-d6ddecaa7e19


app/lib/features/space/widgets/space_sections/spaces_loading_error_section.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:acter/common/providers/room_providers.dart';
```
gnunicorn, line 1: shown in case of error or still loading

app/lib/common/toolkit/skeletons/room_card_skeleton.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:flutter/material.dart';
```
gnunicorn, line 1: simple minimal room skeleton for loading

app/test/features/spaces/error_pages_test.dart
gnunicorn, line 1: not relevant anymore

app/test/helpers/test_util.dart
```diff
@@ -49,6 +50,7 @@ extension ActerProviderTesting on WidgetTester {
     List<Override>? overrides,
     MockNavigator? navigatorOverride,
     GoRouter? goRouter,
+    String? screenshotPath,
```
gnunicorn, line 53: new additional screenshot support for easier debugging :)

app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -3,6 +3,7 @@ import 'package:acter/common/providers/space_providers.dart';
 import 'package:acter/features/events/providers/event_providers.dart';
```
gnunicorn, line 3: this is the heart of it all:

Instead of having this await for all the internal states and thus fails if any of inner ones fails - in particular the space relations - this is now a list that is automatically updated with the proper tabs once they are loaded. This lowers the time-to-first-load and removes the need to have a full-space error message on failures.

gnunicorn (commented):
Some review comments.

app/lib/features/space/pages/space_details_page.dart
```diff
@@ -126,50 +122,30 @@ class _SpaceDetailsPageState extends ConsumerState<SpaceDetailsPage> {
...
-      data: (tabs) {
-        return ScrollableListTabScroller(
-          headerKey: SpaceDetailsPage.headerKey,
-          itemCount: tabs.length,
-          itemPositionsListener: itemPositionsListener,
-
-          //Space Details Header UI
-          headerContainerBuilder:
-              (context, menuBarWidget) => spaceHeaderUI(menuBarWidget),
+    final tabs = ref.watch(tabsProvider(widget.spaceId));
```
kumarpalsinh25, line 125: I see in below codes that spaces loading is manage with tabs enum like `TabEntry.spacesLoading` and latter on `space_navbar_provider.dart` there is condition for empty tabs checking.

Why not showing loader state directly here with empty tabs condition? It there any specific need of using it in `TabEntry` enum?

app/lib/features/space/providers/space_navbar_provider.dart
```diff
@@ -16,100 +17,126 @@ enum TabEntry {
...
-  bool canLinkSpaces =
-      membership.valueOrNull?.canString('CanLinkSpaces') == true;
+    if (tabs.isEmpty) {
+      final currentState = ref.watch(spaceRelationsProvider(spaceId));
+      if (currentState.isLoading || currentState.error != null) {
+        // show while loading or on error
+        tabs.add(TabEntry.spacesLoading);
+      }
+    }
+    return tabs;
```
kumarpalsinh25, line 118: I would suggest to use loading state this in space overview page itself instead of tab entry level code.

kumarpalsinh25 (changes requested):
Core changes looks good which improves and manage SpaceOverview Page data loading time.

However I am not happy with the way `spacesLoading` is managed with `TabEntry.spacesLoading`. spacesLoading isn't any tab and space be managed at the page level instead of tabs entry.

app/lib/features/space/pages/space_details_page.dart
```diff
@@ -249,6 +225,7 @@ class _SpaceDetailsPageState extends ConsumerState<SpaceDetailsPage> {
       TabEntry.suggestedSpaces => lang.suggestedSpaces,
       TabEntry.chats => lang.chats,
       TabEntry.spaces => lang.spaces,
+      TabEntry.spacesLoading => lang.spaces,
```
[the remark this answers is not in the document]
gnunicorn: No, it is also use to show a loading section below.

app/lib/features/pins/providers/pins_provider.dart
```diff
@@ -14,7 +14,15 @@ final pinListSearchTermProvider = StateProvider<String>((ref) => '');
...
-final pinListProvider =
+final pinListProvider = FutureProvider.family<List<ActerPin>, String?>((
+  ref,
+  spaceId,
+) async {
+  final pinList = await ref.watch(_pinListProvider(spaceId).future);
+  return pinList; // this is easier for us to mock
+});
+
+final _pinListProvider =
```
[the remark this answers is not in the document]
gnunicorn: As the comment says, this made mocking quite a bit easier ;) .

app/lib/features/space/pages/space_details_page.dart
gnunicorn: Because we have four potential sections while we load, in most cases all will disappear. But especially for lacking Internet, there will be 4 sections of loading animations the user has to scroll over. For no reason other than us being too lazy to make it better.

I first open for only showing when there was an actual error, to inform the user and allow them to retry. But for the cases of being offline that means you don't see any tab entering the UI, even if you know it has subspaces - which is perceived as broken.

So I opted instead for having the `Spaces` entry show up (once) in the menu and show a loading and failure in those cases, I forming the user without blocking things. Arguably we could show it as `Spaces & Chats` in the UI though to make clear what's happening.

app/lib/features/space/providers/space_navbar_provider.dart
gnunicorn: We don't want to block rendering of the space overview because we have trouble reaching the server - that's the entire point of this PR.

Instead if this is still loading or errors we show a specific other section meanwhile - not blocking the rendering of the other tabs.

gnunicorn:
@kumarpalsinh25 I've elaborate more on the approach to make it easier to understand why it has an additional tab. As you didn't have any other addressable feedback, I've rerequested review. If you still don't want this approach, I am happy to try another approach if you suggest one - keeping in mind the requirements of not blocking the tabs from loading and not make the user unnecessarily scroll over pointless loading blocks that then just disappear.

kumarpalsinh25 (approved):
As mentioned earlier, core changes looks good and improving space overview page data management with much better way.

Regarding structural things of Space loading with Tab Entry, I will share thoughts on call so that there will be no back and forth on communication.

And as this addressing critical space loading and error related issue, I would like to go for approval so that we can it on Thursday release :)

PR #2781, opened 2025-04-02
Refactor timeline-related objects

bitfriend:
This PR was split from https://github.com/acterglobal/a3/pull/2751.

- Will remove unnecessary fields in `TimelineItem`, except needed fields.
- Will integrate timeline-related objects into single directory.

### DIFF
--- a/native/acter/src/api/convo.rs
+++ b/native/acter/src/api/convo.rs
@@ -172,7 +172,7 @@ impl Convo {
                 };
                 let room_id = latest_msg_room.room_id();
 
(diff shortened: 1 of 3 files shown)


native/acter/src/api/timeline/msg_content.rs
```diff
@@ -0,0 +1,452 @@
+use acter_core::events::{
```
gnunicorn, line 1: skipping as I assume this is just a code-move. If there is logic changes to review, please point them out and I will review them.

gnunicorn (approved):
looking good :+1:

Issue #2763, opened 2025-03-28
Desktop Chat-NG: sending of empty messages


gnunicorn:
see https://github.com/acterglobal/a3-meta/issues/785

Just open chat ng in desktop, press enter to send an empty message.

PR #2793, opened 2025-04-04
[Chat-NG fix] - chat editor content validation

gtalha07:
## Description:

Fix for #2763. Adds a validation for proper html structure checking whether its content is valid or exists with proper tests supporting the case .i.e. Appflowy html encoders still bounds to return some structure with empty text. So that needs to taken care of.

### Preview:

- Video where I showcased, trying different scenarios of empty content format but send button state doesn't appear unless, contains valid content.

https://github.com/user-attachments/assets/aaf0a46c-403b-46de-87ea-57a112bbe18e

### DIFF
--- a/app/test/common/utils/utils_test.dart
+++ b/app/test/common/utils/utils_test.dart
@@ -1,0 +1,92 @@
+import 'package:acter/common/utils/utils.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('Editor Content Validation', () {
+    test('returns false for empty plain text', () {
+      expect(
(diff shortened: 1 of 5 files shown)


app/test/common/utils/utils_test.dart
```diff
@@ -0,0 +1,92 @@
+import 'package:acter/common/utils/utils.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('Editor Content Validation', () {
```
kumarpalsinh25, line 5: Very Nice!

kumarpalsinh25 (approved):
Changes looks good to me.

Issue #2797, opened 2025-04-07
Telemetry & Analytics startup wizzard added with UI


anisha-e10:
![Image](https://github.com/user-attachments/assets/3cd94ddb-8c35-4744-afe6-eeafd2dde048)

Figma reference :

https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=3613-11394&p=f&t=wjqLSY7DavUQ9WEw-0

PR #2801, opened 2025-04-07
Telemetry & Analytics startup wizzard added with UI 

anisha-e10:
Implemented, https://github.com/acterglobal/a3/issues/2797

This PR introduces a "Contribute to Acter" screen, which encourages users to support the development of the app by allowing optional telemetry and crash report sharing. The screen features:

- A clear explanation of how users can help by contributing data

Toggle switches for:

1. Crash Reporting
2. Basic Telemetry
3. In-App Analytics
4. Research

- A "Toggle All" switch to quickly enable/disable all options.

### Reference Image : 

image

### Reference Video ( Startup of  app after logged in) :

https://github.com/user-attachments/assets/377a85b7-4cca-4d6f-8dfc-ecbe1d5d58ed

### Reference Video ( Added in app settings) :

https://github.com/user-attachments/assets/53967329-f21a-4d31-b7eb-9d0ce9bd4e6c


gnunicorn:
there shouldn't be any skip, as turning all of them off and pressing `continue` is the same action.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+              final newValue = !allEnabled;
+              await setCanReportToSentry(newValue);
+              await setMatomoAnalyticsEnabled(newValue);
+              await _setAnalyticsPreference(basicTelemetryPref, newValue);
+              await _setAnalyticsPreference(researchPref, newValue);
+              ref.invalidate(allowSentryReportingProvider);
+              ref.invalidate(allowMatomoAnalyticsProvider);
+            }, context);
+          },
+        ),
```
gnunicorn, line 141: can we please unwrap that future builder into a ref.watch as well? It is impossible to follow what is to be happening when here and what is meant to update and stuff ... we have riverpod for a reason ;)

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+        _buildCrashAnalyticsCard(
+          lang.sendCrashReportsTitle,
+          lang.sendCrashReportsInfo,
+          allowReportSending,
+          (value) async {
+            await setCanReportToSentry(value);
```
gnunicorn, line 147: so ... I see these actually change the variables on click already. But what does it then mean if we clicked `skip`? It would still be applied, right? That shouldn't be the case, I'd argue. But then I'm also against `skip` in the first place, so I suppose I am fine with direct toggles...

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+    String title,
+    String subtitle,
+    String prefKey,
+    BuildContext context,
+    TextTheme textTheme,
+    WidgetRef ref,
+  ) {
+    return FutureBuilder<bool>(
+      future: _getAnalyticsPreference(prefKey),
+      builder: (context, snapshot) {
```
gnunicorn, line 241: same as above, let's please use a custom riverpod provider. Generally `FutureBuilder` isn't something that went well for us ever...

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          ),
+        );
+      },
+    );
+  }
+
+  Future<bool> _getAnalyticsPreference(String prefKey) async {
+    final prefs = await sharedPrefs();
+    return prefs.getBool(prefKey) ?? isNightly;
+  }
```
gnunicorn, line 273: so close to just already being its own provider .... wait, there is one... just look up how `isCalendarSyncActiveProvider` is done ... also just a boolean and has a proper notifier and doesn't need you to `invalidate` things all the time.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          textTheme,
+          ref,
+        ),
+        _buildCrashAnalyticsCard(
+          lang.appAnalytics,
+          lang.appAnalyticsInfo,
+          matomoAnalyticsEnabled,
+          (value) async {
+            await setMatomoAnalyticsEnabled(value);
+            ref.invalidate(allowMatomoAnalyticsProvider);
```
gnunicorn, line 167: using `ref.invalidate` is a strong indicator that the way this provider works should be re-evaluated. Because if this was a `ref.read(provider.notifier).set(newValue)` instead, it wouldn't need a `ref.invalidate` ...

app/lib/main.dart
```diff
@@ -86,6 +91,15 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
       // and prevent reporting otherwise.
       options.beforeSend = sentryBeforeSend;
     }, appRunner: () => runApp(wrappedApp));
+  } else if (withMatomo) {
```
gnunicorn, line 94: so this _only_ works if sentry isn't enabled? That is not correct...

app/lib/main.dart
```diff
@@ -86,6 +91,15 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
...
+    await MatomoTracker.instance.initialize(
+      siteId: '',
+      url: '',
+    );
+
+    // Set Matomo tracking based on user preference
+    final analyticsEnabled = await getMatomoAnalyticsEnabled();
+    MatomoTracker.instance.setOptOut(optOut: !analyticsEnabled);
```
gnunicorn, line 102: generally, I don't think this is where we want to be doing this. But rather in providers. Considering that we start with analytics being off - thus optOut - , but if you switched it on, we aren't putting it into `optIn`, it will only be doing this _after the next restart_. Which is a lot later than what I would be expecting.

gnunicorn (changes requested):
I think this needs some further refactoring and restructuring for clarity:

1. using proper providers rather than futurebuilders
3. think through the opt-in and opt-out flows for matomo proper

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+    BuildContext context,
+  ) {
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.end,
+      children: [
+        Text(text),
+        Transform.scale(
+          scale: 0.6,
+          child: Switch(value: value, onChanged: (_) => onToggle()),
+        ),
```
kumarpalsinh25, line 198:
```
Transform.scale(
          scale: 0.6,
          child: Switch(value: value, onChanged: (_) => onToggle()),
```

I see above code is used repeatedly, why not put in common widget function with `onChanged(value)` callback?

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
-      key: AnalyticsOptInPage.skipBtn,
-      onPressed: () => context.goNamed(Routes.main.name),
-      child: Text(
-        L10n.of(context).done,
-        style: Theme.of(context).textTheme.bodyMedium,
+  // Reusable widget for switch toggle option without card
+  Widget _buildSwitchOption(
+    String text,
+    bool value,
+    Future<void> Function() onToggle,
```
kumarpalsinh25, line 188: `Future<void> Function() onToggle`

Why it is without new value in function arg?

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+        textAlign: TextAlign.center,
+      ),
+    );
+  }
+
+  // Refactored crash analytics section with toggle functionality
+  Widget _buildCrashAnalytics(
+    BuildContext context,
+    WidgetRef ref,
+    TextTheme textTheme,
```
kumarpalsinh25, line 108: No need to pass `TextTheme textTheme` to deeper layer if we are passing `BuildContext context`.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
kumarpalsinh25, line 167: Why to have separate function code for each cards? 

Screenshot 2025-04-09 at 9 02 21 AM

If we are passing title, subtile and getting callback of switch then what else is difference?

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+  /// Handles the tap action for the "More Details" link.
+  Future<void> _handleMoreDetailsTap(BuildContext context) async {
+    final url = Uri.parse(
+      'https://acter.global/faq/in-app-tracking-analytics/',
+    );
+    try {
+      await launchUrl(url, mode: LaunchMode.externalApplication);
+    } catch (e) {
+      _log.severe('Could not launch URL: $e');
+    }
```
kumarpalsinh25, line 311: How about using `openLink` function from `lib/common/open_link.dart` as this is our general action for click of any link.

kumarpalsinh25 (changes requested):
Apart from the feedback from @gnunicorn , I have added few more things which need to address before proceed ahead.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          child: Text(
+            lang.wizzardContinue,
+            style: Theme.of(context).textTheme.bodyMedium,
+          ),
+        ),
+        const SizedBox(height: 15),
+        OutlinedButton(
+          onPressed: () => context.goNamed(Routes.main.name),
+          child: Text(lang.skip),
+        ),
```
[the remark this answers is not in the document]
anisha-e10: ok, will remove it

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+  }
+
+  // Action button to proceed
+  Widget _buildActionButton(BuildContext context, L10n lang) {
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        ActerPrimaryActionButton(
+          key: AnalyticsOptInPage.continueBtn,
+          onPressed: () => context.goNamed(Routes.main.name),
```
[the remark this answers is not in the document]
anisha-e10: In the analytics page, the continue button was previously using `context.goNamed` to navigate, and I understand your concern. I didn't change this behavior initially because it was already integrated in the flow, but now will change it like the notification and calendar.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,61 +34,280 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          child: Text(lang.skip),
+        ),
+      ],
+    );
+  }
+
+  /// Handles the tap action for the "More Details" link.
+  Future<void> _handleMoreDetailsTap(BuildContext context) async {
+    final url = Uri.parse(
+      'https://acter.global/faq/in-app-tracking-analytics/',
```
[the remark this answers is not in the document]
anisha-e10: sure will add it in env

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: Absolutely, I agree! I’ll refactor the code to use ref.watch instead of FutureBuilder.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: ok got it . will  remove the "Skip" button and handle the actions using toggles only.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: ok, will check and apply.

app/lib/main.dart
anisha-e10: ok, got your point the functionality should work regardless of whether Sentry is enabled or not.

app/lib/main.dart
```diff
@@ -86,6 +91,15 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
       // and prevent reporting otherwise.
       options.beforeSend = sentryBeforeSend;
     }, appRunner: () => runApp(wrappedApp));
+  } else if (withMatomo) {
+    await MatomoTracker.instance.initialize(
```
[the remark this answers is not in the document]
anisha-e10: let me check it and add it

app/lib/main.dart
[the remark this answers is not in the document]
anisha-e10: Yeah, I think we’ll need to wait for this. Right now, it’s just integrated with empty` url and siteId` but nothing is being executed yet it will just uses the preference value.

app/lib/main.dart
```diff
@@ -86,6 +91,15 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
...
+      siteId: '',
+      url: '',
```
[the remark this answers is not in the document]
anisha-e10: okay for visitors  specific identity i will try to use the Visitor property with `setVisitorUserId()` with our userId.

app/lib/main.dart
anisha-e10: It sounds like you’re concerned about the delay in transitioning from the "optOut" to "optIn" state i will handle this within providers, it seems like we could apply the change dynamically right after the switch is made, without waiting for a restart.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: I will extract the repeated code into a common widget with an onChanged(value) callback to make the code cleaner,

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: Thank you for the suggestion! I'm not familiar with that function, but I will use it.

app/lib/main.dart
gnunicorn: I am not concerned about the delay transition. I am talking about when this is happening in terms of startup flow. What happens at the moment with this code is:
1. startup of app
2. disabling of matomo due to optout
3. registration happens
4. toggle is set to on
5. nothing else

Which means matomo is not informed about the opt-in, until a new restart of the app. It will stay optout until then.

I am suggesting to switch the opt-in/opt-out into a provider that handles the matomo side of things (I think we are doing something similar for sentry) and toggle that directly, rather than doing it only once at startup.

anisha-e10: ok, got your point the flow will be based on toggle,

1. if the toggle is on then enable matomo with optIn
2. if the toggle is off then disable matomo with optOn

All things will managed with provider for matomo.

.changes/2801-telemetry-analytics-UI.md
```diff
@@ -0,0 +1 @@
+- [Implemented] : - Added user telemetry consent screen for community contributions and it includes the crash reports, basic telemetry, analytics and research.
```
kumarpalsinh25, line 1:
```suggestion
- [Implemented] : Added user telemetry consent screen for community contributions which includes the crash reports, basic telemetry, analytics and research.
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
-              _buildCrashAnalytics(context, ref),
-              const SizedBox(height: 40),
-              _buildSkipActionButton(context),
-              const Spacer(),
-            ],
+          child: SingleChildScrollView(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: [
+                _buildIcon(context),
```
kumarpalsinh25, line 33:
```suggestion
                _buildCloseIconButton(context),
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+              ],
+            ),
           ),
         ),
       ),
     );
   }
 
-  Widget _buildCrashAnalytics(BuildContext context, WidgetRef ref) {
+  Widget _buildIcon(BuildContext context) {
```
kumarpalsinh25, line 53:
```suggestion
  Widget _buildCloseIconButton(BuildContext context) {
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+      alignment: Alignment.topRight,
+      child: IconButton(
+        onPressed: () => Navigator.pop(context),
+        icon: const Icon(Icons.close),
+      ),
+    );
+  }
+
+  // widget for title text
+  Widget _buildTitleText(BuildContext context, L10n lang, TextTheme textTheme) {
```
kumarpalsinh25, line 64: Repeating, if `BuildContext context` is already passed then no need to pass additional param for `textTheme` as it can be define directly in inner function.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+      textAlign: TextAlign.center,
+    );
+  }
+
+  // widget for "More Details" text
+  Widget _buildMoreDetails(
+    BuildContext context,
+    L10n lang,
+    TextTheme textTheme,
+    WidgetRef ref,
```
kumarpalsinh25, line 88: Same as above for `textTheme`.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
kumarpalsinh25: This is still not addressed. 
```suggestion
        Future<void> Function(bool) onToggle,
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+        basicTelemetryEnabled &&
+        researchEnabled;
 
-    void toggle(bool? input) {
-      setCanReportToSentry(input ?? !allowReportSending);
-      ref.invalidate(allowSentryReportingProvider);
-    }
+    return Column(
+      children: [
+        _buildToggleAllOption(lang.togglleAll, allEnabled, () async {
```
kumarpalsinh25, line 129:
```suggestion
        _buildToggleAllOption(lang.togglleAll, allEnabled, (newVal) async {
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          final newValue = !allEnabled;
```
kumarpalsinh25, line 130:
```suggestion

```

No need of this if we get `newValue` from toggle callback function.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+            basicTelemetry,
+            newValue,
+            ref,
+          );
+          await telemetryAnalyticsNotifier.setPreference(
+            research,
+            newValue,
+            ref,
+          );
+        }, context),
```
kumarpalsinh25, line 154: Put this in separate function call.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+    final allowReportSending = preferences[canReportSentry] ?? isNightly;
+    final matomoAnalyticsEnabled = preferences[matomoAnalytics] ?? isNightly;
+    final basicTelemetryEnabled = preferences[basicTelemetry] ?? isNightly;
+    final researchEnabled = preferences[research] ?? isNightly;
+
+    final allEnabled =
+        allowReportSending &&
+        matomoAnalyticsEnabled &&
+        basicTelemetryEnabled &&
+        researchEnabled;
```
kumarpalsinh25, line 125:
```suggestion

```

Probably this is also not require if we are using callback value of the toggle all switch as mentioned below.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+  }
+
+  // widget for analytics card with separate preference
+  Widget _buildAnalyticsCard(
+    BuildContext context,
+    String title,
+    String subtitle,
+    bool value,
+    Future<void> Function(bool) onToggle,
+  ) {
```
kumarpalsinh25, line 229:
```suggestion
  Widget _buildAnalyticsCard(
  {
    BuildContext context,
    String title,
    String subtitle,
    bool value,
    Future<void> Function(bool) onToggle,
    }
  ) {
```
If there are multiple params in any function then it is better to use named param for better code readability.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+    Future<void> Function() onToggle,
+    BuildContext context,
+  ) {
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.end,
+      children: [
+        Text(text),
+        Transform.scale(
+          scale: 0.6,
+          child: Switch(value: value, onChanged: (_) => onToggle()),
```
kumarpalsinh25, line 216:
```suggestion
          child: Switch(value: value, onChanged: onToggle),
```

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
```diff
@@ -26,59 +26,237 @@ class AnalyticsOptInPage extends ConsumerWidget {
...
+          context,
+          lang.appAnalytics,
+          lang.appAnalyticsInfo,
+          matomoAnalyticsEnabled,
+          (value) async {
+            await ref
+                .read(analyticsPreferencesProvider.notifier)
+                .setPreference(matomoAnalytics, value, ref);
+          },
+        ),
```
kumarpalsinh25, line 187: Looks pretty good now!

app/lib/main.dart
```diff
@@ -75,17 +80,26 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
...
-      // we use the dart-define default env for the default stuff.
-      options.dsn = Env.sentryDsn;
-      options.environment = Env.sentryEnvironment;
-      options.release = Env.sentryRelease;
-
-      // allows us to check whether the user has activated tracing
-      // and prevent reporting otherwise.
-      options.beforeSend = sentryBeforeSend;
-    }, appRunner: () => runApp(wrappedApp));
+  if (withSentry || withMatomo) {
```
kumarpalsinh25, line 83:
```suggestion
 
```

What is the need of this?

kumarpalsinh25 (approved):
Major concerns that I raised are addressed and look good to me.

However there still some code standards related remarks I have added which need to addressed before proceed ahead.

Also I would to also suggest one round of code review from @gnunicorn as well for the concerns that he raised.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: I thought it not needed to be defined again, so I passed it, but I'll remove it now since it's accessible through the existing BuildContext

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: ok

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: ok, will change it

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: thanks

app/lib/main.dart
anisha-e10: `Sentry` has an `appRunner` that runs after initialization, I used that, but `Matomo `doesn't have a similar method, so I added the condition to handle that case. I'll update it based on your suggestion.

app/lib/features/onboarding/pages/analytics_opt_in_page.dart
anisha-e10: ok, let me check and resolve this.

anisha-e10: This variable tracks the states of the four toggle values (true or false). Based on these states, the toggleAll function will either on or off, so this variable is necessary for its functionality.

app/lib/features/onboarding/pages/upload_avatar_page.dart
```diff
@@ -178,7 +179,10 @@ class UploadAvatarPage extends ConsumerWidget {
           await handleCalendarPermission(context);
         }
         if (context.mounted) {
-          context.goNamed(Routes.analyticsOptIn.name);
+          await handleAnalyticsOptIn(context);
```
gnunicorn, line 182: this is getting a lot now, shouldn't we just put that into a single instance somewhere going through all of them properly?

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+
+final basicTelemetryProvider = createAsyncPrefProvider<bool>(
+  prefKey: basicTelemetryKey,
+  defaultValue: false,
+);
+
+final researchProvider = createAsyncPrefProvider<bool>(
+  prefKey: researchKey,
+  defaultValue: false,
+);
```
gnunicorn, line 33: :+1:

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
gnunicorn, line 1: as these also happen to be used in the settings later, I'd suggest moving the entire analytics-things into its own feature `app/lib/features/analytics` (or something)

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+) async {
+  await switch (key) {
+    matomoAnalyticsKey => () async {
+      await ref.read(matomoAnalyticsProvider.notifier).set(value);
+      await setMatomoAnalytics(value, ref);
+    }(),
+    canReportSentryKey => ref.read(canReportSentryProvider.notifier).set(value),
+    basicTelemetryKey => ref.read(basicTelemetryProvider.notifier).set(value),
+    researchKey => ref.read(researchProvider.notifier).set(value),
+    _ => throw ArgumentError('Unknown analytics preference key: $key'),
```
gnunicorn, line 72: if this is a static list, make the key an enum?

that also removes the need for `_ => throw` (which is a runtime error) and would enforce static checking at compile time instead.

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+  if (!canReport) {
+    return null;
+  }
+  return evt;
+}
+
+/// Handles Matomo-specific analytics settings
+Future<void> setMatomoAnalytics(bool value, WidgetRef ref) async {
+  // Update Matomo tracking state
+  MatomoTracker.instance.setOptOut(optOut: value);
```
gnunicorn, line 46: but isn't this the other way round? If `true` we are opt-in, but in this case it would do an opt-out on true.
```suggestion
  MatomoTracker.instance.setOptOut(optOut: !value);
```

app/lib/main.dart
```diff
@@ -75,6 +80,12 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
     child: app,
   );
 
+  if (withMatomo) {
```
gnunicorn, line 83: doesn't this mean we always are opt-in at first? this is against regulations, we might be opt-out first. Please add
```suggestion
  if (withMatomo) {
     MatomoTracker.instance.setOptOut(true);
```

That said... where/when are we reading the setting at startup?

gnunicorn (commented):
Code looking good overall, just a few minor things to fix.

However, there is question whether that is actually then adhered to at startup. I guess we need to put something into a provider that listens to client to ensure we actually update the opt-in and set the right userId ...

that is still missing here.

app/lib/common/utils/telemetry_analytics.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:acter/features/onboarding/pages/analytics_opt_in_page.dart';
+import 'package:flutter/material.dart';
+
+/// Internal function to handle analytics opt-in
+Future<void> handleAnalyticsOptIn(BuildContext context) async {
```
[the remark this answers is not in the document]
anisha-e10: Addressed by this [commit](https://github.com/acterglobal/a3/pull/2801/commits/161feec407e55bd108f7c8b31ece7d823c4de224)

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
anisha-e10: ok, will the entire analytics-things into specific folder in features.
I have already added this telemetry and analytic in settings.

Reference video : 
https://github.com/user-attachments/assets/53967329-f21a-4d31-b7eb-9d0ce9bd4e6c

app/lib/common/utils/telemetry_analytics.dart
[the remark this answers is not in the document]
anisha-e10: Addressed by this [commit](https://github.com/acterglobal/a3/pull/2801/commits/e076b8c69fb1aee3ad3c78d8dc9430ca731f8e01)

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
anisha-e10: Addressed by this [commit](https://github.com/acterglobal/a3/pull/2801/commits/6b251cb3b2ca3f4c07351618fb5f4ce52284728b)

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
anisha-e10: Added in this [commit](https://github.com/acterglobal/a3/pull/2801/commits/ebe1c7c5b208ef8f7256aff56a32b34912a3595b)

app/lib/main.dart
```diff
@@ -42,14 +43,18 @@ Widget makeApp() {
...
   setSpaceOverviewTutorialsAsViewed();
-  return await _startAppInner(app, false);
+  return await _startAppInner(app, false, false);
 }
 
-Future<void> _startAppInner(Widget app, bool withSentry) async {
+Future<void> _startAppInner(
+  Widget app,
+  bool withSentry,
+  bool withMatomo,
```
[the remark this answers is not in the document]
anisha-e10: Addressed by this [commit](https://github.com/acterglobal/a3/pull/2801/commits/6410a5d99358382944e47c4b987ffd277940ef0d)

app/lib/features/onboarding/pages/upload_avatar_page.dart
anisha-e10: I have created provider for that to structure it proper , here is the [commit](https://github.com/acterglobal/a3/pull/2801/commits/7c360ba54d798c9596a8be703979854badd44348)

app/lib/features/onboarding/providers/analytics_preferences_provider.dart
```diff
@@ -0,0 +1,74 @@
...
+  defaultValue: false,
+);
+
+final researchProvider = createAsyncPrefProvider<bool>(
+  prefKey: researchKey,
+  defaultValue: false,
+);
+
+Future<SentryEvent?> sentryBeforeSend(SentryEvent evt, Hint hint) async {
+  final canReport = await mainProviderContainer.read(canReportSentryProvider.future);
```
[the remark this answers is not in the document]
anisha-e10: Sorry, but we can not access this as ref here because it's a `ProviderContainer `actually  either we can do `mainProviderContainer.read(canReportSentryProvider.future)` or we can get value from `sharedPrefs()` simply i think.

video : https://github.com/user-attachments/assets/f56c9593-d513-4fa5-9081-d39ee8dc781b

app/lib/features/auth/providers/post_login_signup_provider.dart
gnunicorn, line 1: this is a good first step, it def goes into the direction I am thinking. What I do not understand is why make this a _provider_?

The steps could be generated by a provider, okay, I can see that. Though I think that it only makes sense if the steps updated according to being done. But that is another thought. What I don't get is why put the stepping into a `Service` that has a `context` attached... We've learnt the hard way that `BuildContext` shouldn't be in provider files. 

I am thinking about a simpler infrastructure, similar to how we generated the `SpaceTabs`: one provider that figures out (via an Enum) the remaining steps. Then a `Wizzard`-widget that reads that provider and shows the "next" page according to the Enum value - I don't think putting the steps into "modals" that need a "show" makes a lot of sense. Each item could either be given a callback to be "done" or just set some flag that removes the entry from the provider and thus automatically switches to the next one. either is fine with me.

This would also allow us to have a neat little steps-counter in the surrounding wizzard, showing the user how many steps are left to complete ...

app/lib/features/onboarding/pages/upload_avatar_page.dart
```diff
@@ -167,19 +165,14 @@ class UploadAvatarPage extends ConsumerWidget {
...
-          await handleNotificationPermission(context);
-        }
-        if (context.mounted) {
-          await handleCalendarPermission(context);
-        }
-        if (context.mounted) {
-          context.goNamed(Routes.analyticsOptIn.name);
-        }
+        if (!context.mounted) return;
+        await ref.read(postLoginSignupProvider).initialize(context);
```
gnunicorn, line 173: I like the centralisation into a service. that's good. But I think also the final `goNamed` should just be dealt with by that thing - and rather than it being a service I think it should be a `startupWizzard`-page with its own route, so we could also just simply route there from here or elsewhere.

See comment on the service file.

app/lib/features/settings/widgets/settings_menu.dart
```diff
@@ -183,6 +184,15 @@ class SettingsMenu extends ConsumerWidget {
                 }
               },
             ),
+            MenuItemWidget(
+              iconData: PhosphorIconsThin.chartLineUp,
+              title: lang.telemetryAnalytics,
+              onTap: () {
+                if (context.mounted) {
+                  showAnalyticsOptIn(context);
```
gnunicorn, line 192: what is that was just a route?

app/lib/main.dart
```diff
@@ -74,8 +78,14 @@ Future<void> _startAppInner(Widget app, bool withSentry) async {
...
     child: app,
   );
+  if (withAnalytics) {
+    await MatomoTracker.instance.initialize(
+      siteId: Env.matomoSiteId,
+      url: Env.matomoUrl,
+    );
 
-  if (withSentry) {
+    MatomoTracker.instance.setOptOut(optOut: true);
```
gnunicorn, line 87: :+1: 

only problem is now: when we do this, we haven't actually figured out if the user logged in. So we still need a listener on `Acter` (below) that listens to the pref-provider and updates it when the client switches (or logs in) etc.

Then all the others have to do it just change the provider switch and not care of `MatomoTracker.instance.setOptOut` at all, as it would just "happen" by the outer listener.

gnunicorn (approved):
Almost there!

Just one thing we need to add, which is a listener for the matomo-flag on the `Acter`-Widget to ensure the opt-in and opt-out works.

I also like the approach for the post-login-signup-stepper, but I would organize that a bit differently. This one, however, I leave in your hands to either do now or later. The opt-in-listener we need to do before merge.

app/lib/features/settings/widgets/settings_menu.dart
anisha-e10: Actually, I used a dialog at the startup of the app, and similarly in the settings page. I've also used this pattern for requesting notification and calendar permissions. If you want to use a route here, i can implement it.

However, just a suggestion — at startup, we're using a dialog, while in the settings we're using a route. This feels a bit inconsistent. I think if we decide to use a route, then we should use it consistently in both places: onboarding and the settings menu.

PR #2804, opened 2025-04-08
fix mention breaking upon message edit

### LINKED ISSUE #2716: Editing breaks mentions
See:

https://github.com/user-attachments/assets/44203df0-c048-416c-b235-ee6f4b46c6e1

gtalha07:
- Fixes #2716

### DIFF
--- a/app/lib/features/chat_ng/actions/send_message_action.dart
+++ b/app/lib/features/chat_ng/actions/send_message_action.dart
@@ -38,9 +38,10 @@ Future<void> sendMessageAction({
     onTyping?.map((cb) => cb(false));
 
     // make the actual draft
+    final chatEditorState = ref.read(chatEditorStateProvider);
     final client = await ref.read(alwaysClientProvider.future);
     late MsgDraft draft;
-    if (html.isNotEmpty) {
+    if (html.isNotEmpty && !chatEditorState.isEditing) {
       final htmlProcessedText = textEditorState.mentionsParsedText(body, html);
       draft = client.textHtmlDraft(htmlProcessedText.$1, bodyProcessedText.$1);
       if (htmlProcessedText.$2.isNotEmpty) {
@@ -58,7 +59,6 @@ Future<void> sendMessageAction({
     }
 
     // actually send it out
-    final chatEditorState = ref.read(chatEditorStateProvider);
     final stream = await ref.read(timelineStreamProvider(roomId).future);
 
     if (chatEditorState.isReplying) {


gnunicorn:
but it's still broken:
image

This is not the pill we have when using the editor for writing. We don't need to support markdown, it should be parsed and shown properly...

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -86,6 +92,91 @@ extension ActerEditorStateHelpers on EditorState {
     return (processedText.trimRight(), mentionAttributes);
   }
 
+  void toMentionPills(String body, Node targetNode) {
+    String text = body;
```
kumarpalsinh25, line 96: ` String text = body;`

Wondering, what is the need of this? If you need `text` as variable name then directly used that from function param name.

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -86,6 +92,91 @@ extension ActerEditorStateHelpers on EditorState {
...
+    final userMatches = userMentionRegExp.allMatches(text);
+    List<(int, int, String, String, MentionType)> allMentions = [];
```
kumarpalsinh25, line 98: `List<(int, int, String, String, MentionType)>`

Ah, this is pretty confusing on which type contains what data. There should proper model class use with relevant variable names and type.

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -86,6 +92,91 @@ extension ActerEditorStateHelpers on EditorState {
     return (processedText.trimRight(), mentionAttributes);
   }
 
+  void toMentionPills(String body, Node targetNode) {
+    String text = body;
+    final userMatches = userMentionRegExp.allMatches(text);
```
kumarpalsinh25, line 97: `final userMatches = userMentionRegExp.allMatches(text);`

I would suggest to also define data type of `userMatches` with final declaration which give better idea data types and better readability.

app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -86,6 +92,91 @@ extension ActerEditorStateHelpers on EditorState {
...
+        markerPositions.add(i);
+      }
+    }
+
+    // now apply attributes
+    if (markerPositions.isNotEmpty) {
+      for (int i = 0; i < allMentions.length; i++) {
+        if (i >= markerPositions.length) break;
+
+        final (_, _, mentionId, displayName, type) = allMentions[i];
```
kumarpalsinh25, line 155: `final (_, _, mentionId, displayName, type) = allMentions[i];`

What exactly this line of code do?

app/test/features/chat_ng/chat_mentions_test.dart
```diff
@@ -1,362 +1,613 @@
...
-        'Hello [@User](https://matrix.to/#/@user123)',
-      );
-      final html = editorState.intoHtml();
-      // verify HTML
-      expect(
-        editorState.mentionsParsedText(markdown, html).$1,
-        '<p>Hello <span><a href="https://matrix.to/#/@user123">@User</a></span></p>',
-      );
-    });
+    group('to mention format text unit tests', () {
```
kumarpalsinh25, line 17: 👏

gtalha07:
> This is not the pill we have when using the editor for writing. We don't need to support markdown, it should be parsed and shown properly...

fixed here [a5eb328](https://github.com/acterglobal/a3/pull/2804/commits/a5eb3286aebd71aa021851336efa3342c613f013)

app/lib/common/widgets/html_editor/html_editor.dart
gtalha07: This is just matching the pattern cases of ```RegExp``` in a text so it's a ```Iterable<RegExp>```.

app/lib/common/widgets/html_editor/html_editor.dart
gtalha07: This is implicitly representation of ```MentionAttributes``` and collects mention objects in a text

app/lib/features/chat_ng/utils.dart
```diff
@@ -2,6 +2,11 @@ import 'package:acter/common/providers/chat_providers.dart';
 import 'package:acter/features/chat_ng/providers/chat_room_messages_provider.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
+const userMentionMarker = '‖';
+final userMentionRegExp = RegExp(
+  r'\[@([^\]]+)\]\(https:\/\/matrix\.to\/#\/(@[^)]+)\)',
+);
```
gnunicorn, line 8: there is also `matrix:`-urls as valid user mentions...

and I thought we were not doing regexps anymore but had an actual DOM model we'd be doing stuff on?

In particular I would expect most of these features to move _into_ the mentions-plugin / the editor side. As from the UI perspective it is just a link that is going to be rendered differently due to the editor doing something different...

gtalha07: Not exactly sure what that DOM model is, can you share reference or source? Here it's only being used because of the message format is in markdown and we could detect mention patterns. I am happy to change it based upon improved and safe version

gnunicorn: I think this is work for a future PR ...

gnunicorn (approved):
Not understanding this in detail enough and wonder if we can extract that into the editor instead, but seems to be fine for now.

PR #2810, opened 2025-04-10
always show the overview section, regardless of whether we  have a topic

gnunicorn:
Fixes the problem that the "Upgrade Button" doesn't show if there is no topic on a non-acter-space.
on top of #2780


kumarpalsinh25:
@gnunicorn 
So this PR is ready for review or it will be in draft till [space tab](https://github.com/acterglobal/a3/pull/2780 ) PR gets merged.

Also not understanding this PR Description.

> Fixes the problem that the "Upgrade Button" doesn't show if there is no topic on a non-acter-space.

Screen of issue point and resolution will be help to get idea about what exactly this PR is doing.

gnunicorn:
I kept it draft to have the other merged first, so that the changes are easier to understand, @kumarpalsinh25 .

PR #2843, opened 2025-04-16
Separating Chat NG widgets

gnunicorn:
As we want to remove legacy chat once chat-ng becomes more reliable, we want to make sure we separate them as much as possible. Due to recent fixes, they have actually become tied up more and more, so this PR refactors these structures to separate the Widget again.

In particular this PR:
 - refactors the router to be more directly tied and more general in use, clean up chatLayoutBuilder while on it.
 - copies over room list related widgets and cleans them up, separating `chatNG` and legacy chat
 - removes all mentions/understanding of chat-ng in legacy chat, moves back to `convo cards` there for simplicity
 - makes AnimatedChatListWidget a generic toolkit widget (`ActerAnimatedListWidget`)

After this PR chat-ng only depends on legacy chat for providers (that will be a separate PR later) and for two widget `RoomAvatar` and `Pill`, which both will become toolkit entries later.  

There is one other weird case where we have a chat list that is being used in the quicksearch with a rather complex wrapper that doesn't seem to be used elsewhere though...

### DIFF
--- a/app/test/features/chat_ng/widgets/chat_item/selected_chat_item_test.dart
+++ b/app/test/features/chat_ng/widgets/chat_item/selected_chat_item_test.dart
@@ -1,15 +1,12 @@
 import 'package:acter/common/providers/chat_providers.dart';
 import 'package:acter/common/providers/notifiers/chat_notifiers.dart';
 import 'package:acter/features/chat_ng/rooms_list/widgets/chat_item_widget.dart';
-import 'package:acter/features/search/widgets/chat_list_widget.dart';
 import 'package:acter/features/chat_ui_showcase/models/convo_showcase_data.dart';
 import 'package:acter/features/datetime/providers/utc_now_provider.dart';
 import 'package:acter/features/labs/model/labs_features.dart';
 import 'package:acter/features/labs/providers/labs_providers.dart';
-import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
-import 'package:riverpod/riverpod.dart';
 
 import '../../../../helpers/mock_event_providers.dart';
 import '../../../../helpers/test_util.dart';


app/lib/features/chat_ng/layout/chat_ng_layout_builder.dart
```diff
@@ -0,0 +1,64 @@
...
+
+class ChatNgLayoutBuilder extends StatelessWidget {
+  final Widget? centerChild;
+  final Widget? expandedChild;
+
+  const ChatNgLayoutBuilder({this.centerChild, this.expandedChild, super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final expanded = expandedChild;
```
gtalha07, line 16: this seems unnecessary to me, a re-initialization for just better wording ?

app/lib/features/chat_ng/layout/chat_ng_layout_builder.dart
```diff
@@ -0,0 +1,64 @@
...
+      return RoomsListNGWidget(
+        onSelected:
+            (String roomId) => context.pushNamed(
+              Routes.chatroom.name,
+              pathParameters: {'roomId': roomId},
+            ),
+      );
+    }
+
+    final pushReplacementRouting = centerChild != null;
```
gtalha07, line 32: yep, use of both but same identity variables just makes this more confusing, would suggest we don't reinitialize widget parameters and use them directly like this...

app/lib/features/chat_ng/rooms_list/widgets/rooms_list.dart
```diff
@@ -0,0 +1,263 @@
...
+import 'package:atlas_icons/atlas_icons.dart';
+import 'package:flutter/material.dart';
+import 'package:acter/l10n/generated/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:go_router/go_router.dart';
+
+typedef RoomSelectAction = Function(String);
+
+class RoomsListNGWidget extends ConsumerStatefulWidget {
```
gtalha07, line 18: assuming this been just moved from elsewhere and not logically modified ?

gtalha07 (approved):
Looks good overall, also reminds me a lot of lines don't cover tests support and should be something I have kept down for later! Also while on that, you might need to update tests as CI fails :)

app/lib/features/chat_ng/rooms_list/widgets/rooms_list.dart
gnunicorn: move and renames, correct.

app/lib/features/chat_ng/rooms_list/widgets/rooms_list.dart
```diff
@@ -0,0 +1,263 @@
...
+                  _isSearchVisible
+                      ? Padding(
+                        padding: const EdgeInsets.symmetric(horizontal: 16),
+                        child: filterBox(context),
+                      )
+                      : const SizedBox.shrink(),
+            ),
+            searchTerms(context),
+            Expanded(
+              child: ChatsListNG(
```
gnunicorn, line 192: only logical difference: the `ChatListNG`-widget here.

app/lib/features/chat_ng/layout/chat_ng_layout_builder.dart
gnunicorn: 1 - this is a copy-paste
2 - remapping avoids the nasty `!` of `if (expandedChild != null) return expandedChild!`  which the static analysis still needs for reason unknown to me.

app/lib/features/chat_ng/layout/chat_ng_layout_builder.dart
gnunicorn: just a copy-paste, didn't change anything here...

app/lib/features/chat_ng/rooms_list/widgets/chats_list.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:acter/features/chat_ng/rooms_list/widgets/chat_item_widget.dart';
+import 'package:flutter/material.dart';
+import 'package:acter/l10n/generated/l10n.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+import 'package:logging/logging.dart';
+
+final _log = Logger('a3::chat_ng::chats_list');
+
+class ChatsListNG extends ConsumerWidget {
```
kumarpalsinh25, line 18: I see that this is copy-past from older chats_list page code but later I would like to do bit of code refactor for better code structure.

kumarpalsinh25 (approved):
Much needed changes at the current point of time as we are working on Chat-NG and very soon planning to remove legacy chat.

kumarpalsinh25:
@gnunicorn 

> There is one other weird case where we have a chat list that is being used in the quicksearch with a rather complex wrapper that doesn't seem to be used elsewhere though...

When I was working on Quick search page, I have refactoring almost all the listing widget like Pin list, event list, space list etc to have same pattern and reused in multiple use cases.

For example,
- PinListWidget is re-used in All Pin List scree, Space Pin List screen, Quick search Pin List screen and even used it as Pin List selector in Boost action.
- Chats_list current code was bit complex and Talha was already working on Chat-NG at the time, so I have created new `ChatListWidget` which is currently used in quick search page and expand to it further screen.

PR #2855, opened 2025-04-21
Activate Matomo tracking

gnunicorn:
Adds Route observing and submitting the UserID if matomo is activated. Development parameter are set now in settings.


gnunicorn:
added in the settings.

app/lib/common/providers/common_providers.dart
```diff
@@ -25,10 +25,12 @@ final eventTypeUpdatesStream = StreamProvider.family<int, String>((
   }
 });
 
+final maybeMyUserIdStrProvider = Provider(
+  (ref) => ref.watch(alwaysClientProvider).valueOrNull?.userId().toString(),
+);
+
```
gnunicorn, line 31: so we can pass `null`

app/lib/main.dart
```diff
@@ -82,10 +79,15 @@ Future<void> _startAppInner(
     await MatomoTracker.instance.initialize(
       siteId: Env.matomoSiteId,
       url: Env.matomoUrl,
+      cookieless: true,
+      pingInterval: null,
```
gnunicorn, line 83: disable features we don't want - like storing the userID locally.

app/lib/main.dart
```diff
@@ -82,10 +79,15 @@ Future<void> _startAppInner(
...
+      dispatchSettings: const DispatchSettings.persistent(
+        dequeueInterval: Duration(seconds: 60),
```
gnunicorn, line 85: and only sent the event every 60s to decrease bandwidth load.

app/lib/router/router.dart
```diff
@@ -179,6 +180,10 @@ final goRouter = GoRouter(
   navigatorKey: rootNavKey,
   initialLocation: Routes.main.route,
   restorationScopeId: 'acter-routes',
+  observers: [
+    // Log the route changes to matomo
+    MatomoGlobalObserver(),
```
gnunicorn, line 185: allows us to track page views / moves.

.github/workflows/build-app.yml
```diff
@@ -301,6 +301,8 @@ jobs:
           RAGESHAKE_APP_VERSION=${{inputs.release_env}}/${{inputs.version}}/${{ matrix.name }}
           DEFAULT_HOMESERVER_URL=${{vars.DEFAULT_HOMESERVER_URL}}
           DEFAULT_HOMESERVER_NAME=${{vars.DEFAULT_HOMESERVER_NAME}}
+          MATOMO_URL=${{ secrets.MATOMO_URL }}
+          MATOMO_SITE_ID=${{ secrets.MATOMO_SITE_ID }}
```
gnunicorn, line 305: passing the matomo information to the build for all targets and build environments (nightly, demo & release)

kumarpalsinh25 (approved):
LGTM!

PR #2815, opened 2025-04-11
Extend timeline event item to cover the room state change

bitfriend:
This PR was split from https://github.com/acterglobal/a3/pull/2751.

Will handle `TimelineItemContent::OtherState` here.

PR #2859, opened 2025-04-21
Extend timeline event item to cover room aliases

bitfriend:
This PR was split from https://github.com/acterglobal/a3/pull/2815.

- Will handle `TimelineItemContent::OtherState` here.
- Make activity cover the room aliases event.


native/test/src/tests/room/room_aliases.rs
```diff
@@ -0,0 +1,116 @@
...
+    .await?;
+
+    let convo = user.convo(room_id.to_string()).await?;
+    let timeline = convo.timeline_stream();
+    let stream = timeline.messages_stream();
+    pin_mut!(stream);
+
+    let aliases = vec!["#friendlyname:server.name".to_owned()];
+    let room_aliases = serde_json::to_string(&aliases)?;
+    let aliases_event_id = convo.set_aliases(room_aliases).await?;
```
bitfriend, line 41: @gnunicorn `server.name` is only placeholder and we have no server that resolve `#friendlyname`.
We should use `m1` dev server for that?
I don't know if github would allow for CI to send request to another machine.

gnunicorn: the ci servers have the servername `localhost`. IF you have the alias set there - which for a test we can do, but probably should pick a random name - then it should be able to resolved it, like `#random:localhost` should resolve.

bitfriend:
`m.room.aliases` was deprecated in the latest version of matrix spec.
Will discard this PR.

PR #2882, opened 2025-04-24
Extend timeline event item to cover the room join rules

bitfriend:
This PR was split from https://github.com/acterglobal/a3/pull/2815.

- Will handle `TimelineItemContent::OtherState` here.
- Make activity cover the room join rules event.


bitfriend:
@gnunicorn Many changes occurred in `acter_flutter_sdk_ffi.dart` due to dart lint formatting.
I think the difference exists between your dev env and my dev env.

PR #2921, opened 2025-04-30
Have a key on swipeTo ensure correct callbacks

### LINKED ISSUE #2840: chat-ng: swipping on a message to reply, shows another message in the UI
it selects the wrong item for the editor. It doesn't always do that, but often when it is the last item.

details: https://github.com/acterglobal/a3-meta/issues/910
and https://github.com/acterglobal/a3-meta/issues/887

gnunicorn:
When the chat ng list of items was updated, swipe-to failed to select the right item and instead would cause the wrong item to be selected. After some debugging the cause turned out to be that swipe-to is stateful and needs to be informed about updates. In listviews this needs to be done by providing a key, as the docs state:

Bildschirmfoto_2025-04-30_18-39-59

And indeed, as soon as I added that, the swipe to works perfectly, even if I sent new messages and thus cause the list view to swap items:

https://github.com/user-attachments/assets/71a60b05-f869-4287-a7c8-662e60f25aa7

Fixes #2840 .

### DIFF
--- a/app/lib/features/chat_ng/widgets/events/message_event_item.dart
+++ b/app/lib/features/chat_ng/widgets/events/message_event_item.dart
@@ -44,6 +44,7 @@ class MessageEventItem extends ConsumerWidget {
     final hasReactions = ref.watch(messageReactionsProvider(item)).isNotEmpty;
     final sendingState = item.sendState();
     return SwipeTo(
+      key: Key(messageId), // needed or swipe doesn't work reliably in listview
       onRightSwipe: (_) => _handleReplySwipe(ref, item),
       child: Column(
         crossAxisAlignment:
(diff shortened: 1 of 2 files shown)


app/lib/features/chat_ng/widgets/events/message_event_item.dart
```diff
@@ -44,6 +44,7 @@ class MessageEventItem extends ConsumerWidget {
     final hasReactions = ref.watch(messageReactionsProvider(item)).isNotEmpty;
     final sendingState = item.sendState();
     return SwipeTo(
+      key: Key(messageId), // needed or swipe doesn't work reliably in listview
```
gtalha07, line 47: Nice, pretty vague to figure out but good catch! 👍🏻

gtalha07 (approved):

PR #2922, opened 2025-04-30
[Chat-NG] fix typing indicator 

### LINKED ISSUE #2907: Typing Indicator Not working in chat room
https://github.com/user-attachments/assets/a5fe2986-a5c4-4205-b94e-2fd6d252cfe1

gtalha07:
- Only changes are refactoring of typing indicator structure, specifically around the provider and state updates. fixes #2907

https://github.com/user-attachments/assets/a5ec1ae8-abd1-4415-b2c8-a89069ee073d

### DIFF
--- a/app/lib/common/widgets/typing_indicator.dart
+++ b/app/lib/common/widgets/typing_indicator.dart
@@ -1,3 +1,5 @@
+import 'package:acter/common/providers/room_providers.dart';
+import 'package:acter/features/chat_ng/providers/chat_room_messages_provider.dart';
 import 'package:flutter/material.dart';
 import 'package:acter/l10n/generated/l10n.dart';
 import 'package:acter_avatar/acter_avatar.dart';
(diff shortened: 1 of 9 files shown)


kumarpalsinh25 (commented):
Code point of view, everything seem to be fine.

What was the root cause of the issue which doesn't showing typing indictor?

kumarpalsinh25 (approved):

gnunicorn:
@gtalha07 I can see a screen-wide refresh happening, flicker of all avatars, when the typing first shows up. Any idea where that comes from and anything we can do about it?

gtalha07:
> Code point of view, everything seem to be fine.
> 
> What was the root cause which was causing typing indictor not showing issue?

I think it's probably the way it was laid out in widget tree before. Having ```ref.watch()``` in widget methods sometimes causes issues with reactive updates over riverpod side...

gtalha07:
> @gtalha07 I can see a screen-wide refresh happening, flicker of all avatars, when the typing first shows up. Any idea where that comes from and anything we can do about it?

Yeah I'd guess it's ```memberAvatarInfoProvider``` getting frequent refreshes. Been used in chain of ```typingEventProvider```, let me check if I can do quick fix on it.

app/lib/common/widgets/typing_indicator.dart
```diff
@@ -6,29 +8,38 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
+    final users = ref.watch(chatTypingEventProvider(roomId)).valueOrNull;
+    final userIds = users?.map((userId) => userId.toString()).toList() ?? [];
+    return userIds
+        .map(
+          (userId) => ref.watch(
+            memberAvatarInfoProvider((userId: userId, roomId: roomId)),
+          ),
+        )
+        .toList();
+  }
```
gnunicorn, line 41: why this complexity if that isn't actually used (here)?

app/lib/common/widgets/typing_indicator.dart
```diff
@@ -45,11 +56,15 @@ class TypingWidget extends StatelessWidget {
   }
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     final l10n = L10n.of(context);
+    final typingUsers = _getTypingUsers(ref);
     final text = _buildTypingText(typingUsers, l10n);
 
+    if (typingUsers.isEmpty) return const SizedBox.shrink();
```
gnunicorn, line 64: I think this can be simplified, no?

```suggestion
    if (ref.watch(chatTypingEventProvider(roomId)).valueOrNull?.isNotEmpty != true) return const SizedBox.shrink();
    
    
    final text = _buildTypingText(typingUsers, l10n);
```

gnunicorn (approved):

app/lib/common/widgets/typing_indicator.dart
gtalha07: I've created separate high level providers to inject them directly where they are needed :)

gtalha07:
> > @gtalha07 I can see a screen-wide refresh happening, flicker of all avatars, when the typing first shows up. Any idea where that comes from and anything we can do about it?
> 
> Yeah I'd guess it's `memberAvatarInfoProvider` getting frequent refreshes. Been used in chain of `typingEventProvider`, let me check if I can do quick fix on it.

 it does seem the problem resides elsewhere, I'll deal that separately and not adding too much complications here 👍🏻

PR #2928, opened 2025-05-01
Html NG

gnunicorn:
replace the matrix html renderer with our own version using our proper internal links.

PR #2932, opened 2025-05-02
HTMLng and Mentions in editor

gnunicorn:
built on top of #2928 this adds support for having the user pill in the mentions editor directly.


app/lib/common/toolkit/html_editor/html_editor.dart
```diff
@@ -578,49 +582,65 @@ class HtmlEditorState extends State<HtmlEditor> {
...
       node,
       index,
       text,
       before,
       after,
     );
+
+    final roomId = widget.roomId;
+
+    final href = attributes[AppFlowyRichTextKeys.href] as String?;
```
gtalha07, line 596: for href we could  use ```defaultTextSpanDecoratorForAttribute``` instead, this inbuilt function handles href attribute also

gnunicorn:
merged into https://github.com/acterglobal/a3/pull/2928

PR #2939, opened 2025-05-05
fix text not visible after adding mentions

gtalha07:
## FIX PREVIEW:

https://github.com/user-attachments/assets/b9ad0e01-6a7a-463c-ba24-db7a0b858fe5

### DIFF
--- a/app/lib/common/widgets/html_editor/html_editor.dart
+++ b/app/lib/common/widgets/html_editor/html_editor.dart
@@ -308,6 +308,10 @@ class HtmlEditorState extends State<HtmlEditor> {
   void initState() {
     super.initState();
     updateEditorState(widget.editorState ?? EditorState.blank());
+    AppFlowyRichTextKeys.partialSliced.addAll([
+      userMentionChar,
+      roomMentionChar,
+    ]);
   }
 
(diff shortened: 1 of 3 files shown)


app/lib/common/widgets/html_editor/html_editor.dart
```diff
@@ -308,6 +308,10 @@ class HtmlEditorState extends State<HtmlEditor> {
   void initState() {
     super.initState();
     updateEditorState(widget.editorState ?? EditorState.blank());
+    AppFlowyRichTextKeys.partialSliced.addAll([
```
gnunicorn, line 311: isn't that a global change on a singleton? Therefore, should that really be happening everytime we construct a new state? rather than somewhere globally?

app/macos/Podfile.lock
gnunicorn, line 1: unrelated podupdate ... nonomnomnom

gnunicorn (approved):
what do I know other than what I can see in the video :D ...

app/macos/Podfile.lock
gtalha07: Ahh geez!

Issue #2626, opened 2025-02-18
Chat NG: Image bubble doesn't look good


kumarpalsinh25:
![Image](https://github.com/user-attachments/assets/5ab5aeaf-35fb-46ac-abc5-17e554028595)
![Image](https://github.com/user-attachments/assets/590f6712-54f6-482b-a0a1-74e74e9e8794)

gtalha07:
@kumarpalsinh25 I don't understand the issue here, are you suggesting UX improvement regarding media loading states ?

kumarpalsinh25:
> @kumarpalsinh25 I don't understand the issue here, are you suggesting UX improvement regarding media loading states ?

@gtalha07 
As shown in reference screenshot, when media is not downloaded (showing media type icon and size info) and media download is in loading state ( Circular progress indicator) , at that time blue bubble chat is not wrapped around it.

Issue #2766, opened 2025-03-28
Complete Chat-NG design spec


gnunicorn:
Write up and potentially designs covering all messages, and state events and their various states and how these should be shown.

- [ ] Overall rules:
   - [ ] General Design rules (mine are in a blue bubble, etc)
- [ ] Regular Messages:
  - [ ] File shared design
  - [ ] Notices message design
- [ ] state events:
   - [ ] user rejected invitation

Ref https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=3613-11395&p=f&t=iSSEnylSAaJehOIt-0

gnunicorn:
maybe we do that with a mocked chat.

Issue #2624, opened 2025-02-18
Chat NG: Image bubble doesn't look good


kumarpalsinh25:
![Image](https://github.com/user-attachments/assets/386bec7b-49ca-4e07-9a57-7305c6ff950c)

gtalha07:
@kumarpalsinh25  Could you confirm if it's still the case ?

PR #2941, opened 2025-05-06
Chat-NG : Adding greatness to Chat Message UI

kumarpalsinh25:
This PR includes numerous minor but precise corrections to the Chat Message UI, significantly enhancing its overall quality and contributing to a more refined user experience.

While it's difficult to recall and describe every minor change in this PR, I’ve made an effort to highlight them through screenshot comparisons along with descriptive text.

- - -
I'll walk through each change with:
🔹 Before vs After screenshot
📝 Brief description of the change
- - -

Screenshot 2025-05-09 at 12 34 37 PM

**🧭 System/State Message Alignment:**
System messages (e.g., "You changed the display name...") are now centered with consistent styling and padding for a cleaner look.

**👥 Grouping Consecutive Messages:**
Messages from the same user are grouped together to reduce redundancy and improve readability.

**📏 Consistent Padding Between Message Bubbles and State Events:**
Previously, padding between grouped user messages and system/state events was inconsistent, leading to a cluttered appearance. This has now been standardized for a cleaner and more structured layout.

**🎨 Gradient in Sender Message Bubble:**
Outgoing messages (in blue) now include a subtle gradient effect, giving a modern and polished appearance.

**🙅‍♂️ Removed Redundant Avatars in DM Chats:**
In one-on-one (DM) conversations, avatars and usernames have been removed from the left side of incoming messages, since the context already makes the sender clear—this reduces clutter and sharpens focus.

**⏰ Improved Message Time Text Color:**
The color of the message timestamp text has been enhanced for better contrast, especially on the blue background of sender message bubbles. This ensures improved readability and accessibility across all message types.

Screenshot 2025-05-09 at 12 39 05 PM

**👤 Improved Username and Avatar Display in Group Chats:**
In group chats, the username is now displayed only on the first message of a consecutive message group, while the user avatar appears on the last message. This reduces visual clutter and makes it easier to follow conversations.

**🌈 Consistent and Dynamic Username Colors:**
Username colors are now dynamically generated based on a consistent logic tied to the username. This ensures that each user always has the same color throughout the chat, improving recognition and readability.

**🧩 Enhanced Bubble Padding for Special Messages:**
The internal padding of message bubbles has been significantly improved, especially for deleted and encrypted messages. This results in a more balanced and visually appealing layout for all message types.

Screenshot 2025-05-09 at 1 10 34 PM

**🖼️ Consistent Media Message Bubbles:**
All media messages—including files, videos, and images—are now presented within a unified chat bubble style, ensuring visual consistency across all message types.

**📎 Unified Grouping Logic for Media Messages:**
File, video, and image messages now follow the same grouping logic as text messages, including avatar and display name management. This means media messages are grouped seamlessly with other messages from the same user, providing a cohesive and organized chat experience.

---
Alongside the changes listed above, this PR includes numerous other minor enhancements to the chat UI that may not be explicitly mentioned here. Additionally, several structural improvements have been made at the code level wherever necessary to support a more maintainable and scalable chat interface.


app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
...
-                            ),
-                          ),
-                        if (timestamp != null)
-                          MessageTimestampWidget(
-                            timestamp: timestamp.expect('should not be null'),
-                          ),
-                      ],
-                    ),
-                  ],
+    return IntrinsicWidth(
```
kumarpalsinh25, line 128: Removed duplicated `Container` and  unwanted `Row` widget from this.

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
...
+      child: Container(
+        margin: const EdgeInsets.symmetric(horizontal: 14),
+        padding: const EdgeInsets.all(10),
+        constraints: BoxConstraints(maxWidth: msgWidth ?? defaultWidth),
+        decoration: decoration,
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            if (displayName != null) ...[
```
kumarpalsinh25, line 138: Added `displayName` widget and logic here in Chat bubble itself to maintain consistency across chat bubble of all the messages type including media messages.

Note: earlier `displayName` widget was only used in TextMessage widget only which was not good.

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
...
+                          Colors.primaries.length],
                 ),
               ),
-            ),
-          ),
-        ],
+              const SizedBox(height: 8),
+            ],
+            bubbleContentWidget,
+            _buildTimestampAndEditedLabel(context),
```
kumarpalsinh25, line 150: Separated timestamp and edited message widget independent widget function for better code separation.

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/profile_changes_event_widget.dart
```diff
@@ -9,11 +9,15 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 class ProfileChangesEventWidget extends ConsumerWidget {
   final String roomId;
   final TimelineEventItem eventItem;
+  final TextStyle? textStyle;
+  final TextAlign? textAlign;
```
kumarpalsinh25, line 13: Add this types of that `ProfileChangesEventWidget` can be use as more generic.

And this allowed me to delete duplicated file as below which contain the same logic and code.
`app/lib/features/chat_ng/widgets/events/profile_update_event.dart`

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/room_membership_event_widget.dart
```diff
@@ -9,11 +9,15 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 class RoomMembershipEventWidget extends ConsumerWidget {
   final String roomId;
   final TimelineEventItem eventItem;
+  final TextStyle? textStyle;
+  final TextAlign? textAlign;
```
kumarpalsinh25, line 13: Add this types of that `RoomMembershipEventWidget` can be use as more generic.

And this allowed me to delete duplicated file as below which contain the same logic and code.
`app/lib/features/chat_ng/widgets/events/member_update_event.dart`

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
       isLastMessageProvider((roomId: roomId, uniqueId: eventId)),
     );
-    final myId = ref.watch(myUserIdStrProvider);
-    final messageId = msg.uniqueId();
-    // FIXME: should check canRedact permission from the room
     final canRedact = item.sender() == myId;
+    final eventType = item.eventType();
 
-    final isMe = myId == item.sender();
+    final eventWidget = switch (eventType) {
```
kumarpalsinh25, line 81: Implemented rendering of different event type of messages here itself which allows me to delete below file.
`app/lib/features/chat_ng/widgets/events/chat_event_item.dart`

Reasons:
- Most of core logical variable declared in this class.
- We already have separate widget for different event type messages.
- So it doesn't make sense to pass all the detail to new class and use if only for switch cases.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
-            () => showMemberInfoDrawer(
-              context: context,
-              roomId: roomId,
-              memberId: userId,
-            ),
-        child: ActerAvatar(
-          options: AvatarOptions.DM(
-            ref.watch(
-              memberAvatarInfoProvider((roomId: roomId, userId: userId)),
+  Widget buildChatBubble(
```
kumarpalsinh25, line 184: Common container widget to wrap different type of event messages.

app/lib/features/chat_ng/widgets/events/message_event_item.dart
```diff
@@ -136,91 +140,86 @@ class MessageEventItem extends ConsumerWidget {
...
+          ref
+              .watch(
+                memberDisplayNameProvider((
+                  userId: senderId,
+                  roomId: letRoomId,
+                )),
+              )
+              .valueOrNull ??
+          senderId;
+    }
```
kumarpalsinh25, line 158: This for `_buildMediaMsgEventContainer` where need to pass it to `ChatBubble` internally.

app/lib/features/chat_ng/widgets/events/message_event_item.dart
```diff
@@ -136,91 +140,86 @@ class MessageEventItem extends ConsumerWidget {
...
-          ref
-              .watch(
-                memberDisplayNameProvider((
-                  userId: senderId,
-                  roomId: letRoomId,
-                )),
-              )
-              .valueOrNull ??
-          senderId;
-    }
```
kumarpalsinh25, line 223: As described earlier, no need to have `displayName` logic for only `buildTextMsgEvent`, it is now generalised in `ChatBubble`.

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/material.dart';
+
+class StateEventContainerWidget extends StatelessWidget {
```
kumarpalsinh25, line 3: ContainerWidget for common UI management for the state messages.

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
```diff
@@ -0,0 +1,37 @@
...
+        constraints: BoxConstraints(
+          maxWidth: MediaQuery.of(context).size.width - 100,
+        ),
+        child: child,
+      ),
+    );
+  }
+}
+
+TextStyle stateEventTextStyle(BuildContext context) {
```
kumarpalsinh25, line 33: Common text style for state messages.

app/lib/features/chat_ng/widgets/events/text_message_event.dart
```diff
@@ -117,21 +111,11 @@ class TextMessageEvent extends ConsumerWidget {
...
       children: [
-        if (dp != null) ...[
-          Text(
-            dp,
-            style: Theme.of(context).textTheme.bodySmall?.copyWith(
-              color: Theme.of(context).colorScheme.tertiary,
-            ),
-          ),
-          const SizedBox(height: 4),
-        ],
```
kumarpalsinh25, line 134: As described earlier, it removed logic from displayName from `TextWidget` and applied it to `ChatBubble`.

gnunicorn:
Fixes #2626, #2766

on #2624 I am not sure.

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
gnunicorn: hmm... I find it always a bit awkward to have "helper" functions that are being given a context to return rendering information... shouldn't this be a widget?

It seems to me to only be used on `Text` (eventually) and one rather pointless `TextSpan` (on encrypted, which I commented on there). I think this should just be a `StateEventTextWidget` contain the style...

Even funnier, this is defined here, in this file, next to a widget which doesn't use it ...

app/lib/features/chat/widgets/messages/encrypted_message.dart
```diff
@@ -28,9 +29,7 @@ class EncryptedMessageWidget extends StatelessWidget {
               ),
               TextSpan(
                 text: L10n.of(context).encryptedChatMessage,
-                style: textTheme.labelSmall?.copyWith(
-                  fontStyle: FontStyle.italic,
-                ),
+                style: stateEventTextStyle(context),
```
gnunicorn, line 32: I understand this isn't directly related to the changes, but as this is the only case where that helper is used without it being a `Text`-Widget directly, I think this rich-text field here in this container is kinda pointless. this is an icon and a text, wrapped in unnecessary complicated widgets and should probably just be refactored. I understand if that shouldn't block the merge though.

app/lib/features/chat/widgets/messages/redacted_message.dart
```diff
@@ -6,11 +8,28 @@ class RedactedMessageWidget extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
     return Container(
-      padding: const EdgeInsets.all(18),
-      child: Text(
-        L10n.of(context).chatMessageDeleted,
-        style: Theme.of(context).textTheme.bodySmall,
+      padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 8),
+      child: Text.rich(
```
gnunicorn, line 14: that gave me pause and I checked again, but this is just a copy from the other widget, isn't it? We can just use a `Row([Icon(..), Text(..)]` here, can't we? If so, I think we should.

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -68,6 +68,34 @@ final animatedListChatMessagesProvider =
           ref.watch(chatMessagesStateProvider(roomId).notifier).animatedList,
     );
 
+final renderableBubbleChatMessagesProvider = StateProvider.autoDispose
```
gnunicorn, line 71: can you add some doc comments to it? I know we have been doing this poorly in the past, but one big thing about chat-ng is do things better as well, so having documentation explaining what this is for, and why this is different than other and relates to other refs and when it can and shouldn't be used would be very useful.

From just reading it I don't know why there exists some filter it relies upon for example...

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
```diff
@@ -91,13 +119,13 @@ final renderableChatMessagesProvider = StateProvider.autoDispose
...
 
-final _getNextMessageProvider = Provider.family<TimelineItem?, RoomMsgId>((
+final _getPreviousMessageProvider = Provider.family<TimelineItem?, RoomMsgId>((
   ref,
   roomMsgId,
 ) {
   final roomId = roomMsgId.roomId;
   final eventId = roomMsgId.uniqueId;
-  final messages = ref.watch(renderableChatMessagesProvider(roomId));
+  final messages = ref.watch(renderableBubbleChatMessagesProvider(roomId));
```
gnunicorn, line 128: hmm... this also changes the logic for the outer providers using this, if I now asked for that value for a state event within state events, even if from the same sender, that will return false, as they are filtered out before. I think we need to add that information as a doc comment on the outer providers (isFirst... and isLast...) to avoid people using if with wrong expectations.

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
 import 'package:acter/common/extensions/options.dart';
 
 class ChatBubble extends StatelessWidget {
-  final Widget child;
+  final Widget bubbleContentWidget;
```
gnunicorn, line 9: why rename from the flutter common `child` if you have a single inner widget to this? I'd prefer to stick with `child`.

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
...
+        padding: const EdgeInsets.all(10),
+        constraints: BoxConstraints(maxWidth: msgWidth ?? defaultWidth),
+        decoration: decoration,
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            if (displayName != null) ...[
+              Text(
+                displayName!,
```
gnunicorn, line 140: not a fan of `!` at all, that being refactored later has bitten us in the path and if it happens is super hard to figure out where, so I _really_ don't like seeing it, especially in rendering code. You probably have it here because there is a lint triggered if you kept it out, but we can easily get around that via re-assignment:

```suggestion
    final name = displayName;
    
    return IntrinsicWidth(
      child: Container(
        margin: const EdgeInsets.symmetric(horizontal: 14),
        padding: const EdgeInsets.all(10),
        constraints: BoxConstraints(maxWidth: msgWidth ?? defaultWidth),
        decoration: decoration,
        child: Column(
          crossAxisAlignment: CrossAxisAlignment.start,
          mainAxisSize: MainAxisSize.min,
          children: [
            if (name != null) ...[
              Text(
                name,
```

app/lib/features/chat_ng/widgets/chat_bubble.dart
```diff
@@ -6,148 +6,180 @@ import 'package:flutter/material.dart';
...
+                style: Theme.of(context).textTheme.bodySmall?.copyWith(
+                  color:
+                      Colors.primaries[displayName!.hashCode.abs() %
```
gnunicorn, line 143: I understand that is meant to be a feature and maybe we keep that discussion for another day, but this is only place where we switch the color of the username at all and there is at least a few colors that have a bad contrast (even the screenshot show that). I am wondering if we shouldn't just pick a color that is nice for our design and stick with that or at maximum "taint" the color with the color derived from the username (so use white and then go 50% towards that color derived) to ensure good contrast ...

secondly, this is derived from the _display_ name and not the userId which means it doesn't stay consistent for the same user but changes on the username ... I am not sure if that's what we want...

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/profile_changes_event_widget.dart
gnunicorn: But the widget `ProfileChangesEventWidget` is only used once - and there it is passed in: why allow it to not be passed in, or even passed in at all and not just put right in here?

I think it is the responsibility of this widget to do the proper rendering and not the outer side to set it .

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/profile_changes_event_widget.dart
gnunicorn, line 1: Why does this file live under `last_message_widgets` if it is being used by `app/lib/features/chat_ng/widgets/events/chat_event.dart` only? Shouldn't this be under `widget/events` as well?

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/room_membership_event_widget.dart
gnunicorn, line 1: same comments as per `ProfileChangesEventWidget`: passing in style doesn't make sense to me and the location of the file doesn't either.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+            isMe: isMe,
+            item: item,
+            roomId: roomId,
+            textStyle: stateEventTextStyle(context),
+            textAlign: TextAlign.center,
+          ),
+        ),
+      _ => StateEventContainerWidget(
+        child: Text(
+          'Unsupported chat event type: $eventType',
```
gnunicorn, line 141: l10n please.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+        StateEventContainerWidget(
+          child: RoomUpdateEvent(
+            isMe: isMe,
+            item: item,
+            roomId: roomId,
+            textStyle: stateEventTextStyle(context),
+            textAlign: TextAlign.center,
+          ),
+        ),
+      _ => StateEventContainerWidget(
```
gnunicorn, line 139: I am not sure we want to always render all unsupported events. It is common practice in matrix to ignore unsupported events in rendering...

I wonder if this might be something we could switch on for developers only...

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+        context,
+        ref,
+        const EncryptedMessageWidget(),
+        item.sender(),
+        isMe,
+        isDM,
+        isFirstMessageBySender,
+        isLastMessageBySender,
+      ),
+      String type when _isSupportedRoomUpdateEvent(type) =>
```
gnunicorn, line 129: oh, this is cool, didn't know we can do that! neat!

app/lib/features/chat_ng/widgets/events/chat_event.dart
gnunicorn: this looks like it should be an internal widget to me. In particular because then it could split out the outer widget refreshing just because the username changed for the inner widget.

accumulating a lot of `ref` on the same widget means unnecessary refreshes.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+                  roomId: letRoomId,
+                )),
+              )
+              .valueOrNull ??
+          senderId;
+    }
+
+    return isMe
+        ? ChatBubble.me(
+          context: context,
```
gnunicorn, line 211: I am just noticing now and it is probably not something for this PR, but why are passing `context` to a Widget Factory? That is clearly wrong. It's context is passed to `build` and that is what it should depend on. I have no idea what side-effect this could have and I really don't want to experience them, as they will be hard to pinpoint to it. This is clearly something wrong we should fix.

Not really part of this PR though. Will make it into a separate ticket afterwards.

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+    BuildContext context,
+    WidgetRef ref,
+    String roomId,
+    String userId,
+    bool isMe,
+    bool isLastMessageBySender,
+    bool isBubbleEvent,
+    bool isDM,
+  ) {
+    if (isLastMessageBySender && isBubbleEvent && !isMe && !isDM) {
```
gnunicorn, line 262: such a nice check, can't we do that before calling this function at all and thus keep the tree clean of useless widgets?

app/lib/features/chat_ng/widgets/events/chat_event.dart
```diff
@@ -51,93 +57,232 @@ class ChatEvent extends ConsumerWidget {
...
+          context: context,
+          isFirstMessageBySender: isFirstMessageBySender,
+          isLastMessageBySender: isLastMessageBySender,
+          displayName: displayName,
+          bubbleContentWidget: child,
+        );
+  }
+
+  bool _isSupportedRoomUpdateEvent(String type) {
+    const supportedRoomUpdateEvents = {
```
gnunicorn, line 227: why not define this `const` at the top of the file?

app/lib/features/chat_ng/widgets/events/message_event_item.dart
```diff
@@ -136,91 +140,86 @@ class MessageEventItem extends ConsumerWidget {
     // shouldn't happen but in case return empty
     if (msgType == null || content == null) return const SizedBox.shrink();
 
+    String? displayName;
+    if (isFirstMessageBySender && !isMe && !isDM) {
+      // FIXME: also ignore in 1-on-1 dm rooms
```
gnunicorn, line 145: I think this is happening now, so this fixme should be removed.
```suggestion
```

app/lib/features/chat_ng/widgets/events/replied_to_event.dart
```diff
@@ -60,10 +60,14 @@ class RepliedToEvent extends StatelessWidget {
...
-        Text(
-          replyProfile.displayName ?? '',
-          style: Theme.of(context).textTheme.bodySmall?.copyWith(
-            color: Theme.of(context).colorScheme.tertiary,
+        Expanded(
+          child: Text(
+            replyProfile.displayName ?? '',
+            style: Theme.of(context).textTheme.bodySmall?.copyWith(
+              color: Theme.of(context).colorScheme.tertiary,
+            ),
```
gnunicorn, line 68: so, we don't calculate the same color for the display name here than we do in the original message? Whatever we agree to do, we probably should though

app/lib/features/chat_ng/widgets/events/room_update_event.dart
```diff
@@ -13,21 +14,26 @@ class RoomUpdateEvent extends ConsumerWidget {
   final bool isMe;
   final TimelineEventItem item;
   final String roomId;
+  final TextStyle? textStyle;
```
gnunicorn, line 17: same as before, it's never used without any, why pass it at all and not put it into this widget?

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
```diff
@@ -0,0 +1,37 @@
...
+
+class StateEventContainerWidget extends StatelessWidget {
+  final Widget? child;
+  const StateEventContainerWidget({super.key, this.child});
+
+  @override
+  Widget build(BuildContext context) {
+    if (child == null || child == const SizedBox.shrink()) {
+      return const SizedBox.shrink();
+    }
```
gnunicorn, line 11: why allow `child` to be optional if this is never being rendered? Just wrapping pointless empty widgets, where an if before would be have been more prudent.

Also, I can't find any use of `StateEventContainerWidget(child: null, ...)` in here.

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
```diff
@@ -0,0 +1,37 @@
...
+        borderRadius: BorderRadius.circular(25),
+      ),
+      child: ConstrainedBox(
+        constraints: BoxConstraints(
+          maxWidth: MediaQuery.of(context).size.width - 100,
+        ),
+        child: child,
+      ),
+    );
+  }
```
gnunicorn, line 30: look how pretty:

```suggestion
  final Widget child;
  const StateEventContainerWidget({super.key, this.child});

  @override
  Widget build(BuildContext context) => Container(
      alignment: Alignment.center,
      padding: const EdgeInsets.symmetric(horizontal: 10, vertical: 5),
      decoration: BoxDecoration(
        color: Theme.of(context).colorScheme.primary.withValues(alpha: 0.1),
        border: Border.all(
          color: Theme.of(context).colorScheme.primary.withValues(alpha: 0.2),
        ),
        borderRadius: BorderRadius.circular(25),
      ),
      child: ConstrainedBox(
        constraints: BoxConstraints(
          maxWidth: MediaQuery.of(context).size.width - 100,
        ),
        child: child,
      ),
    );
```

app/lib/features/chat_ng/widgets/message_timestamp_widget.dart
```diff
@@ -4,8 +4,13 @@ import 'package:jiffy/jiffy.dart';
 
 class MessageTimestampWidget extends StatelessWidget {
   final int timestamp;
+  final Color? textColor;
```
gnunicorn, line 7: similar to the other style passing cases, I think this should be having the information passed in here as well. We are in the `chat`-area of things, so passing `bool isMe` and have the widget decide on the specific design rather than the outer pass in depending on the same flag seems more logical to me.


At least, if I wanted to change the design of the timestamp, this is were I would look and wanted to change it.

app/test/features/chat_ng/messages/chat_message_test.dart
```diff
@@ -85,7 +85,7 @@ void main() {
       test('hides avatar when next message is from same user', () {
```
gnunicorn, line 85: these tests are poorly named. but I understand if  you don't want to touch that.

gnunicorn (approved):
Great work, @kumarpalsinh25, looks very nice.

There are a few structural and consistency issues that need to be addressed, but then it is ready to go.

app/test/features/chat_ng/widgets/chat_bubble_test.dart
```diff
@@ -113,7 +123,9 @@ void main() {
               context: context,
               isEdited: true,
               timestamp: timestamp,
-              child: const Text('My message'),
+              bubbleContentWidget: const Text('My message'),
+              isFirstMessageBySender: true,
+              isLastMessageBySender: true,
```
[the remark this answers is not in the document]
kumarpalsinh25: On bubble shape and corners, still I am thinking that this is not final version and it can be improved. So would add golden or respective test after final version of Chat Bubble.

app/test/features/chat_ng/widgets/chat_bubble_test.dart
```diff
@@ -71,7 +79,7 @@ void main() {
 
         expect(find.text(L10n.of(testContext).edited), findsOneWidget);
 
-        expect(find.text('.'), findsOneWidget);
+        expect(find.text('-'), findsOneWidget);
```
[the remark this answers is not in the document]
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/f3ee5f23aaebf5a5084c81934e10f41c2434317a

app/test/features/chat_ng/widgets/chat_bubble_test.dart
```diff
@@ -124,7 +136,7 @@ void main() {
       final editedText = L10n.of(testContext).edited;
       expect(find.text(editedText), findsOneWidget);
 
-      expect(find.text('.'), findsOneWidget);
+      expect(find.text('-'), findsOneWidget);
```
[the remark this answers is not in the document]
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/f3ee5f23aaebf5a5084c81934e10f41c2434317a

app/lib/features/chat/widgets/messages/encrypted_message.dart
kumarpalsinh25: I understand your point and already tried that approach earlier in my previous PR but here are multiple things are tight up with this logic, and there are different use-cases I observed where TextWidget wan't be sufficient solution. 

Happy to explain different use-cases on call and if we have better solution then happy to implement it on follow-up PR.

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
kumarpalsinh25: Add comment by https://github.com/acterglobal/a3/pull/2941/commits/36b1dcf08ca796c7aa94abb647449441edf1a8bf

app/lib/features/chat_ng/providers/chat_room_messages_provider.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/91272cbb4bd95d36b8f2217cdfd2eb6f4ff3922f

app/lib/features/chat_ng/widgets/chat_bubble.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/da6ae8e6ba3a8ffc5400bb7cc8bfd9a992a2f22f

app/lib/features/chat_ng/widgets/chat_bubble.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/1173c59806fb599e44ebdb3519c9cc59d8e61022

app/lib/features/chat_ng/widgets/chat_bubble.dart
kumarpalsinh25: ### 1) Better color contrast: 
- Fixed by https://github.com/acterglobal/a3/pull/2941/commits/04b05884a36d32084f5d464f7a13544f3dde54bc by replacing `Colors.primaries` with `Colors.accents`.

Screenshot 2025-05-12 at 12 31 27 PM

### 2) display name logic:
- May be confusing is because of variable name but it is actually using and showing username. Check above screenshot for reference.

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/profile_changes_event_widget.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/0e424f6882f672e441a9756bf810996ead54b66a

app/lib/features/chat_ng/widgets/chat_item/last_message_widgets/room_membership_event_widget.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/0e424f6882f672e441a9756bf810996ead54b66a

kumarpalsinh25: For passing style, as mentioned in https://github.com/acterglobal/a3/pull/2941#discussion_r2083895451, I would like to take it on separate discussion.

app/lib/features/chat_ng/widgets/events/chat_event.dart
kumarpalsinh25: Was copied and past from older code so forget it. Anyway it is fixed by https://github.com/acterglobal/a3/pull/2941/commits/60be537c0b179c9d7e7932f09ba40191337cae1b

app/lib/features/chat_ng/widgets/events/chat_event.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/39ebcf9ed2bf931847bf63916942075a1518a034

app/lib/features/chat_ng/widgets/events/chat_event.dart
kumarpalsinh25: Happy to implement improvement related `ChatBubble` widget code but yes on separate PR as I mostly focused on UI for in this PR and did relevant changes only. There are many things in terms of code improvement but I haven't focused it as of now to make this PR as simple as possibile.

app/lib/features/chat_ng/widgets/events/chat_event.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/21bddbd838d6958346ce6df11bc13ec0822d9d51

app/lib/features/chat_ng/widgets/events/chat_event.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/c02097f930435ad0aae36a26b5ce4bb0f38bcac9

app/lib/features/chat_ng/widgets/events/message_event_item.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/aea50731a69a45e827171221c917429aad0efd06

app/lib/features/chat_ng/widgets/events/room_update_event.dart
kumarpalsinh25: Same as above, would like to do separate discussion for all this similar discussion.

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
kumarpalsinh25: Fixed by https://github.com/acterglobal/a3/pull/2941/commits/6fd0a5c83b1e5aedb2dbf7c408984ac6287398e8

app/lib/features/chat_ng/widgets/events/state_event_container_widget.dart
kumarpalsinh25: Need one condtion for `SizedBox.shrink()` check. https://github.com/acterglobal/a3/pull/2941/commits/245c920ea1e8b2abcbcd3ba4ca023a765233901a

So want't possible above inline function suggestion.

app/test/features/chat_ng/messages/chat_message_test.dart
kumarpalsinh25: Not just name but also doesn't understand context of many of test cases in this file. Doesn't wanted to touch or change anything at lease on this PR.

kumarpalsinh25:
> Fixes #2626, #2766
> 
> on #2624 I am not sure.

Yes this are fixed along with this PR.

PR #2949, opened 2025-05-09
Preload activities data

gnunicorn:
So that both link opening preference and activities load faster, right after startup:

https://github.com/user-attachments/assets/322009e3-8872-4be7-bb8f-3dfb9603f9a2


app/lib/features/main/app_shell.dart
```diff
@@ -38,6 +39,28 @@ final _log = Logger('a3::config::home_shell');
 
 final ScreenshotController screenshotController = ScreenshotController();
 
+class _TopLevelShell extends ConsumerWidget {
```
gnunicorn, line 42: put that loading into separate widget that is also "watching", so we keep it in memory.

kumarpalsinh25 (approved):
LGTM!

PR #2976, opened 2025-05-15
move the chat mock data into providers for future live adaption

### LINKED ISSUE #2962: wrap in typing if longer than one line
seeing the typing animation in the sidebar for a chat, it the user has a long display name leads to a three line wrap breaking and expanding of the item beyond the size that other two-line subtitle room items have:

Image

gnunicorn:
- [x] passing in the user-id so we are actually always the right user with the mocked messages
- [x] fixes typing widget associated with the old chat-providers, move them to the new providers
- [x] add fix for super long user names in typing with golden check and mock data, fixes #2962 

Bildschirmfoto_2025-05-15_13-37-06


kumarpalsinh25 (approved):
Very nice changes for making mock data live adaption!

PR #2981, opened 2025-05-16
Selective room updates

### LINKED ISSUE #2961: chat flickers on emoji reaction
All I am doing was tap the react emoji, you see many things (including the user count and avatars) flicker:

https://github.com/user-attachments/assets/fc1b5fd0-6eb1-465f-839a-75307fcf9a3f

gnunicorn:
This adds tests which explain why we see the flicker of room updates when just regular messages are sent to a chat: we trigger the room notifications from the rust-side even though we shouldn't. Next to that test, this also adds some for showing when we do expect notifications (on any state change or joining-state change). And all that is being implemented in the code then.

Long and behold, check this out, [compared to the video in the bug report](https://github.com/acterglobal/a3/issues/2961):

https://github.com/user-attachments/assets/aeb0ef50-5f40-4dd7-9da2-0481c0bb992d

Fixes #2961.

### DIFF
--- a/.changes/29881-flicker.md
+++ b/.changes/29881-flicker.md
@@ -1,0 +1,1 @@
+- [Fix] to only update Room information if an update-causing event has taken place, removes flicker in chat-ng on message reactions


native/test/src/utils.rs
```diff
@@ -263,3 +265,61 @@ pub async fn random_users_with_random_chat_and_space_under_template(
 
     Ok((clients, sync_states, space_id, room_id, engine))
 }
+
```
gnunicorn, line 268: just a move of the various utils into a re-usable place

gnunicorn (commented):
help for the reviewer

bitfriend (approved):
Great work.
I expect this solution can fix blinking of sliding sync too.

PR #2982, opened 2025-05-16
be more expressive when sharing, informing users what happens via sections

gnunicorn:
Adding section titles to share dialog - as I also didn't really remember what each button would cause - as well as dividers and the option to copy an internal link.

See:

https://github.com/user-attachments/assets/ac41ee61-26ed-4f90-82fe-47d87b818fa1


kumarpalsinh25 (approved):
Looks good with code and concept end goal for better expressive.

However UI can be improve, especially on QR and Other options sharing.
(Not needed in this PR if doesn't getting better idea of managing this labels)

PR #2989, opened 2025-05-20
Chat-NG : Beautifying Chat Bubble UI

kumarpalsinh25:
### DM CHAT:
DM Chat

### GROUP CHAT:
Group Chat


gtalha07 (approved):
Just a minor query on theming, else looks good!

PR #2990, opened 2025-05-20
Autofill added for recovery key in onboarding.

anisha-e10:


gnunicorn:
@anisha-e10 so will this only work if the user clicked into any of the fields? I doubt that they will be doing that .... can we auto-focus so they don't have to click?

and there is no way we can just close the keyboard again if they click it? It seems pointless to have it shown at all, though we "need" it for the emulation, I'd rather not have the user distracted by it too much...

anisha-e10:
> @anisha-e10 so will this only work if the user clicked into any of the fields? I doubt that they will be doing that .... can we auto-focus so they don't have to click?

@gnunicorn Here, how the autofocus will work 

https://github.com/user-attachments/assets/60ebc342-434d-4ffd-9f9e-84e7bf74c9e3

Issue #2993, opened 2025-05-20
[chatng] short message overlay is unreachable


gnunicorn:
I am trying to delete my own message, but if it is short, I can't reach it

https://github.com/user-attachments/assets/e596e797-3fb8-44d1-b2a9-3e73fafee453

gnunicorn:
this brings it at least back to the alignment, but is way too small still:

```
diff --git a/app/lib/features/chat_ng/dialogs/message_actions.dart b/app/lib/features/chat_ng/dialogs/message_actions.dart
index 7f4d6924a0..1e75f7917c 100644
--- a/app/lib/features/chat_ng/dialogs/message_actions.dart
+++ b/app/lib/features/chat_ng/dialogs/message_actions.dart
@@ -37,7 +37,8 @@ void messageActions({
         children: [
           _BlurOverlay(animation: animation, child: const SizedBox.shrink()),
           Positioned(
-            left: messagePosition.dx,
+            left: isMe ? messagePosition.dx : 0,
+            right: isMe ? 0 : messagePosition.dx,
             top: 0,
             height: MediaQuery.sizeOf(context).height,
             child: Column(
```

kumarpalsinh25:
@gnunicorn 
Sorry for this glitches, I will take care of it.

Issue #2980, opened 2025-05-16
Chat-NG : Message action - Reaction view doesn't look good for small messages.


kumarpalsinh25:
![Image](https://github.com/user-attachments/assets/aeb66f83-ad88-433d-b41b-9ee7ed00d1d5)

PR #2972, opened 2025-05-14
Chat-NG : Emoji Reactions UI-UX improvement

kumarpalsinh25:
NOTE: ALL CHANGES ARE JUST RELATED UI-UX THING, NO LOGICAL CHANGES.
---
This PR enhance UI-UX of emoji reaction of the chat messages along with the certain bugs fixing as mentioned below.

### Summary:
- [Improvement] : Optimised design of reaction dialog for smaller text messages.
- [Fixes] : Fixes rendering issue when there are multiple reactions on the same Messages.
- [Enhancement] : Overall reactions design is now enhanced with better visibility about your own reaction and other people's reactions.
- [Improvement] : Optimised design of the Reaction details dialog and improved visibility of the selected reaction tab option.

### 1) Reaction on small text message: Fixes https://github.com/acterglobal/a3/issues/2980
Small Message Reaction

### 2) Reaction are overflow issue and UI-UX Improvements: Fixes https://github.com/acterglobal/a3/issues/2970
Emoji Improvement

### 3) Reaction detail dialog: Fixes https://github.com/acterglobal/a3/issues/2974
Reaction Detail Dialog

PR #2995, opened 2025-05-20
Chat-NG : Fixes short message overlay is unreachable

kumarpalsinh25:
Fixes, https://github.com/acterglobal/a3/issues/2993 by bringing older code back.

https://github.com/user-attachments/assets/7fc396e9-2975-4207-9ef4-f1668710b726

But it leads to re-open issue https://github.com/acterglobal/a3/issues/2980

### DIFF
--- a/app/lib/features/chat_ng/dialogs/message_actions.dart
+++ b/app/lib/features/chat_ng/dialogs/message_actions.dart
@@ -39,6 +39,7 @@ void messageActions({
           Positioned(
             left: messagePosition.dx,
             top: 0,
+            width: messageBox.size.width,
             height: MediaQuery.sizeOf(context).height,
             child: Column(
               mainAxisAlignment: MainAxisAlignment.center,


app/lib/features/chat_ng/dialogs/message_actions.dart
```diff
@@ -39,6 +39,7 @@ void messageActions({
           Positioned(
             left: messagePosition.dx,
             top: 0,
+            width: messageBox.size.width,
```
gtalha07, line 42: This looks good to me but additional adjustments are needed over ```message_actions_widget```. 

I tried debugging this PR on desktop with a doubt that this might still cause some overflow and it did.

Screenshot 2025-05-20 at 14 23 36

gtalha07: Also golden tests, if possible.

gtalha07 (dismissed):
This solution looks fine for the message actions placement but inner components cause overflow. Also having UI adherence tests would be very nice to include!

kumarpalsinh25:
> This solution looks fine for the message actions placement but inner components cause overflow. Also having UI adherence tests would be very nice to include!

As I mentioned, I just reverted code to older version which was implemented by you with the known fact of reopening issue https://github.com/acterglobal/a3/issues/2980

I don't wanted to stop this PR due to old issue as https://github.com/acterglobal/a3/issues/2993 is the kind of blocker.

Side-note:
- I did't find solution on your old code so would appreciate you to help in this in separate PR.

app/lib/features/chat_ng/dialogs/message_actions.dart
kumarpalsinh25: @gtalha07
You highlighted know issue which I already raised and re-opened as mentioned in PR description.
https://github.com/acterglobal/a3/issues/2980

I did my try to fix above issue but it was lead to generated new issue mentioned by @gnunicorn  [here](https://github.com/acterglobal/a3/issues/2993) which I am seeing and kind of blocker.

Appreciate your help in this.

Note:
- I tried to implemented Golden test  of Message Action dialog in my last PR https://github.com/acterglobal/a3/pull/2972 but faced bit of difficulties so left with other golden tests.
- Appreciate if you can help here also.

gtalha07 (approved):

PR #2996, opened 2025-05-20
auto-expanding inner HTML editor

gnunicorn:
Fixes [#meta-901](https://github.com/acterglobal/a3-meta/issues/901) .

This moves the Animated expanding box around HtmlEditor into the inner Editor. This is necessary to ensure that the mobile toolbar isn't inside the animated box as that causes overflows in the inner type upon keyboard opens (as there isn't enough space during render).

By moving that login into the inner editor and make it peg against the actual inner size rather than some "guessed" line count, this is more accurately updating the text box while simultanously avoding to cut the mobile toolbar.

Unfortunately the way that the mobile toolbar is wrapped, this fails with a hasSize error on `main` so this is using a slightly modified fork, which allows for an unknown-sized box.

Live in action:

## on desktop

https://github.com/user-attachments/assets/c886c04c-dcdf-4fcf-a901-274d1edd0d7d

## on android (device):

https://github.com/user-attachments/assets/4e0cb54d-7bc7-4238-bd34-d982857ec973

### DIFF
--- a/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
+++ b/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
@@ -1,8 +1,6 @@
 import 'dart:async';
-import 'dart:math';
 
 import 'package:acter/common/providers/chat_providers.dart';
-import 'package:acter/common/providers/keyboard_visbility_provider.dart';
 import 'package:acter/common/themes/colors/color_scheme.dart';
 import 'package:acter/common/utils/utils.dart';
 import 'package:acter/common/widgets/html_editor/html_editor.dart';
--- a/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
+++ b/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
@@ -40,35 +40,23 @@ class ChatEditor extends ConsumerStatefulWidget {
 
 class _ChatEditorState extends ConsumerState<ChatEditor> {
   EditorState textEditorState = EditorState.blank();
-  late EditorScrollController scrollController;
+  // late EditorScrollController scrollController;
   StreamSubscription<EditorTransactionValue>? _updateListener;
   final ValueNotifier<bool> _isInputEmptyNotifier = ValueNotifier(true);
-  final ValueNotifier<double> _contentHeightNotifier = ValueNotifier(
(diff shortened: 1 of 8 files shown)


app/pubspec.yaml
```diff
@@ -61,7 +61,11 @@ dependencies:
       ref: main
 
   flutter_easyloading: ^3.0.5
-  appflowy_editor: ^5.2.0
+  appflowy_editor: #^5.2.0
```
gtalha07, line 64: is something also internally changed within plugin ? An issue within ```appflowy``` ? Or a workaround for our case?

gtalha07 (approved):
Looks good!

app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
```diff
@@ -40,36 +38,24 @@ class ChatEditor extends ConsumerStatefulWidget {
...
     _updateListener = textEditorState.transactionStream.listen((data) {
       _editorUpdate(data.$2);
-      _updateContentHeight();
+      // _updateContentHeight();
     });
 
     WidgetsBinding.instance.addPostFrameCallback((_) => _loadDraft());
 
-    // apply toolbar offset when keyboard is visible
-    ref.listenManual(keyboardVisibleProvider, (prev, next) {
```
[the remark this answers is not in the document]
gnunicorn: yes, as the expanding happens _within_ the editor, we can now just let flutter manage the exact layouting and sizing for us, the toolbar is just within a `column` anyways.

app/pubspec.yaml
gnunicorn: yes. as the mobile toolbar is using `Column` internally, but our expandable editor is within a valuelistener, that doesn't report a size by default but is within a `columns` `expanded` there is competing interest in what to do: squeeze or expand. [The change in the fork](https://github.com/AppFlowy-IO/appflowy-editor/compare/main...acterglobal:appflowy-editor:ben-mobile-expandable-editor) fixes that problem by using `Flexible` rather than `Expanded` and inform the `Column` to squeeze in size if not otherwise noted. That fix is needed for our editor to work on mobile as well.

gnunicorn: I am not yet sure, if that is generic enough to sent upstream, but I will make a PR and discuss that with the team.

PR #2997, opened 2025-05-20
[Chat NG]: selectable text on message events

### LINKED ISSUE #2866: Be able to highlight (and copy) selected text on sent messages

gtalha07:
- Fixes #2866.

### (DESKTOP) :

https://github.com/user-attachments/assets/428798b4-fb3b-4614-8a2e-9e539713815a

### (MOBILE): 

https://github.com/user-attachments/assets/e00296d5-2346-42f2-ad74-f4fb98d00f2f

### DIFF
--- a/app/lib/features/chat_ng/utils.dart
+++ b/app/lib/features/chat_ng/utils.dart
@@ -82,12 +82,3 @@ const List<MaterialAccentColor> chatBubbleDisplayNameColors =
       Colors.orangeAccent,
       Colors.deepOrangeAccent,
     ];
-
-Widget selectableText(BuildContext context, Widget child) => TextSelectionTheme(
-  data: TextSelectionThemeData(
-    selectionColor: Theme.of(
-      context,
-    ).colorScheme.secondary.withValues(alpha: 0.3),
-  ),
-  child: SelectionArea(child: child),
-);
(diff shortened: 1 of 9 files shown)


app/lib/features/chat_ng/utils.dart
```diff
@@ -82,3 +82,12 @@ const List<MaterialAccentColor> chatBubbleDisplayNameColors =
       Colors.orangeAccent,
       Colors.deepOrangeAccent,
     ];
+
+Widget selectableText(BuildContext context, Widget child) => TextSelectionTheme(
```
gnunicorn, line 86: can we move that to `common/toolkit` as a `ActerSelectionArea`-Widget please?!?


generally: helpers that take a `BuildContext` and return a `Widget` and are used more than once, should be a Widget.

gtalha07:
> I guess there is no proper way to provide tests for this, is there? Like drag on it and check for selection doesn't exist?

In regards of a way, I think we can set up tests but not sure about all cases. Let me think of it!

gtalha07:
> Looks good in general.
> 
> One quick question just for confirmation, do this long press to select text is anyway conflicting with long press message action event?

Yeah, good catch. The ```SelectableText``` supports long press gesture, which would need to be disabled to ensure not interfering with message actions. I will make suitable adjustments.

PR #3005, opened 2025-05-22
Implement the missed test case from codecov

bitfriend:
- Implement the missed test case from codecov
- Thus implement some crates for it and extend `api.rsh` too


native/acter/api.rsh
```diff
@@ -1959,10 +1956,27 @@ object Comment {
...
+    /// set description text
+    fn content_text(body: string);
+
+    /// set description html text
+    fn content_formatted(body: string, html_body: string);
+
+    /// send builder update
+    fn send() -> Future<Result<EventId>>;
+}
+
```
gnunicorn, line 1979: Our UI currently doesn't have any notion of updating comments and having updated comments (like we don't have notifications nor activities for those either), I am not sure we want to expose this API yet...

native/test/src/tests/calendar.rs
```diff
@@ -59,9 +63,19 @@ async fn calendar_smoketest() -> Result<()> {
...
     assert_eq!(spaces.len(), 1);
-
     let main_space = spaces.first().expect("main space should be available");
-    assert_eq!(main_space.calendar_events().await?.len(), 3);
+
+    let cal_events = main_space.calendar_events().await?;
+    assert_eq!(cal_events.len(), 3);
+    let main_event = cal_events.first().expect("main event should be available");
+
+    let locations = main_event.locations();
```
gnunicorn, line 72: :+1:

gnunicorn (approved):

native/acter/api.rsh
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/3005/commits/2c274fde939abf4b3d5411880b6d897be630e2ba

Issue #2827, opened 2025-04-15
Typing disappears on selected chat


gnunicorn:
When you are in the selected chat, the typing message disappears:

![Image](https://github.com/user-attachments/assets/1a72487e-b1a3-408f-a0ca-6a261c74a5dc)

kumarpalsinh25:
> [@kumarpalsinh25](https://github.com/kumarpalsinh25) Can you confirm if its fixed?

This issue is still there.

PR #3006, opened 2025-05-22
Typing indicator disappear issue solved in Desktop.

anisha-e10:
Fixes, https://github.com/acterglobal/a3/issues/2827

- This PR addresses issue where the typing indicator would unexpectedly disappear on the desktop app when chat is selected.

### Reference image: 

Screenshot 2025-05-22 at 1 31 42 PM


anisha-e10:
> can we have a (golden) test for this?

I don't have much idea about as i didn't that so need to check that how it's work and how i implement here

gnunicorn:
@anisha-e10 well, no time like the present to learn :D .

app/lib/features/chat_ng/widgets/chat_item/typing_indicator.dart
```diff
@@ -39,7 +40,7 @@ class TypingIndicator extends ConsumerWidget {
         Text(
           typingText,
           style: theme.textTheme.bodySmall?.copyWith(
-            color: theme.colorScheme.primary,
+            color: isSelected ? theme.colorScheme.surfaceTint : theme.colorScheme.primary,
```
gtalha07, line 43: Why don't we just set up a common color if it gets shadowed on selected chat item that gets visible in both cases ? Rather than passing ```isSelected``` ?

app/lib/features/chat_ng/widgets/chat_item/chat_typing_lastmessage_container_widget.dart
```diff
@@ -7,13 +7,14 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
-  const ChatTypingLastMessageContainerWidget({super.key, required this.roomId});
+  final bool isSelected;
+  const ChatTypingLastMessageContainerWidget({super.key, required this.roomId, required this.isSelected});
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final typingUsers = _getTypingUsers(ref);
     return typingUsers.isNotEmpty
-        ? TypingIndicator(roomId: roomId)
+        ? TypingIndicator(roomId: roomId, isSelected: isSelected)
```
gtalha07, line 17: Yeah it seems unnecessary to me. On black background, the ```surfaceTint``` would also work well. @kumarpalsinh25 would like to hear your thoughts ?

kumarpalsinh25: Yes, that's true that `surfaceTint` will work on black and blue both backgrounds, but it will not look great as primary color indication.

app/lib/features/chat_ng/widgets/chat_item/typing_indicator.dart
anisha-e10: I just followed the existing structure for how the selected chat works so I just used that over here.

anisha-e10:
> @anisha-e10 well, no time like the present to learn :D .

 connected with @kumarpalsinh25 and learned how the golden test works.

Issue #3016, opened 2025-05-23
Chat-NG : User avatar position is off when there are some reaction on message.


kumarpalsinh25:
<img width="423" alt="Image" src="https://github.com/user-attachments/assets/d87e71de-5000-457c-933b-7aeb7d6f99c7" />

PR #3017, opened 2025-05-23
Chat-NG : Fixe user avatar position issue when there is reaction on message.

kumarpalsinh25:
Fixes, https://github.com/acterglobal/a3/issues/3016

User avatar position


gtalha07 (approved):
Looks good!

PR #3019, opened 2025-05-23
Blocking spammy accounts right from the invitation

gnunicorn:
See on the demo:

https://github.com/user-attachments/assets/49c77590-ed56-4631-af1c-816df9e8894e

Specifically this:
- [x] refactors invite sections widget into widget, including tests
- [x] adds this section widget into the demo area
- [x] implements new decline and ignore, with tests


kumarpalsinh25 (approved):
LGTM!

PR #3027, opened 2025-05-27
Allow user to reset identity if backup stay incomplete after providing the key

### LINKED ISSUE #2957: providing recovery keys doesn't work
I am using a user with an encryption key backup and am trying to provide the key in the activities, but it doesn't work:

1. things are still not decrypted
2. the UI doesn't disappear
3. no error is shown

see yourself:

https://github.com/user-attachments/assets/8e71d601-8de7-43bd-a24f-cd1f09c8eb2b

gnunicorn:
Fixes #2957 .

After some investigation it turns out that it can some times happen that we run a client into the state where despite having access to the backup, we don't have all encryption keys. In that case we need to "Reset the Encryption Identity" to recover. Ideally this is being done on a device that has the full recovery available, or just rotate the key there. Unfortunately the rust SDK doesn't provide a way for us to distinguish these cases nicely, so this adds an internal tracker, that shows a different UI if we have provided the key but are still marked as `incomplete` by the inner state to inform the users about what they must do.

https://github.com/user-attachments/assets/54f4ded8-da94-4f06-b9e6-97afc964435e

It's a bit complicated but at least it is a way to recover now. Pressing the info icon on the top left, gets the user to the acter.global help-page though:

image

Namely these are:
- [x] updates to latest matrix-sdk, 83e1962ec3de9b474adf74c18619a17d39ba0510
- [x] rust api for identitiy reset (with tests): b2b948d4f4dbf9d9c400527ae602271c25983a70
- [x] and the update to its manager, 8b673b84b6ed2cb7b203dde9593a0a7434df6725
- [x] fixes to showing the recovery on the activities page indepently of the other widget, 11abad5ee31ca27c7eabfd58794cc3b6814362d0
- [x] minor refactors,


app/lib/features/activities/widgets/security_and_privacy_section/store_the_key_securely_widget.dart
```diff
@@ -11,22 +11,19 @@ import 'package:phosphor_flutter/phosphor_flutter.dart';
 class StoreTheKeySecurelyWidget extends ConsumerWidget {
   const StoreTheKeySecurelyWidget({super.key});
 
+  static bool shouldBeShown(WidgetRef ref) =>
+      ref.watch(storedEncKeyProvider).valueOrNull != null;
+
```
kumarpalsinh25, line 16: Any specific reason to declared it as `static` ?

app/lib/features/backups/dialogs/show_confirm_reset_identity.dart
```diff
@@ -0,0 +1,121 @@
...
+
+class _ShowConfirmResetIdentityDialog extends ConsumerStatefulWidget {
+  const _ShowConfirmResetIdentityDialog();
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() =>
+      __ShowConfirmResetIdentityDialogState();
+}
+
+class __ShowConfirmResetIdentityDialogState
```
kumarpalsinh25, line 23: May I know purpose of `__`(Double underscore) in class starting name?

app/lib/features/backups/dialogs/show_confirm_reset_identity.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final lang = L10n.of(context);
+    return Form(
+      key: formKey,
+      child: AlertDialog(
+        title: Text(lang.encryptionBackupResetIdentityKeyTitle),
+        content: Container(
+          constraints: const BoxConstraints(maxWidth: 500),
```
kumarpalsinh25, line 37: If we are just using `constraints` in `Container` then how about using `ConstrainedBox` instead of `Container`?

```suggestion
        content: ConstrainedBox(
          constraints: const BoxConstraints(maxWidth: 500),
```

app/lib/features/backups/dialogs/show_confirm_rotate.dart
```diff
@@ -2,26 +2,27 @@ import 'package:acter/common/toolkit/buttons/danger_action_button.dart';
...
 import 'package:acter/features/backups/providers/backup_manager_provider.dart';
+import 'package:acter/features/backups/providers/backup_state_providers.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_easyloading/flutter_easyloading.dart';
 import 'package:acter/l10n/generated/l10n.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
-class _ShowConfirmResetDialog extends ConsumerWidget {
-  const _ShowConfirmResetDialog();
+class _ShowConfirmRotateDialog extends ConsumerWidget {
```
kumarpalsinh25, line 11: Not understanding working change from `ConfirmReset` to `ConfirmRotate`. Is that intentionally due to key rotation?

app/lib/features/backups/providers/backup_state_providers.dart
```diff
@@ -14,3 +15,8 @@ final backupAreEnabledProvider = FutureProvider((ref) async {
 final backupNeedRecoveryProvider = FutureProvider((ref) async {
   return ref.watch(backupStateProvider) == RecoveryState.incomplete;
 });
+
+final hasProvidedKeyProvider =
+    AsyncNotifierProvider<AsyncPrefNotifier<bool>, bool>(
+      () => AsyncPrefNotifier<bool>('has_provided_backup_key', false),
```
kumarpalsinh25, line 21: Not understanding how this provider works with `has_provided_backup_key`.

app/lib/features/activities/widgets/security_and_privacy_section/store_the_key_securely_widget.dart
gnunicorn: so it can be used on the other Widget from the outside

app/lib/features/backups/dialogs/show_confirm_reset_identity.dart
gnunicorn: was just copy pasted code

app/lib/features/backups/dialogs/show_confirm_rotate.dart
gnunicorn: While working on the code, and more specifically when talking to the sdk team, I learnt that this feature is actually badly named and adapted the code accordingly. Despite that the file had a completely wrong name even for the old feature, this _rotates_ the key rather than "resets" it -- since this PR also introduce "Reset of Identity" I felt it was prudent to fix the old mistake to make them easier to distinguish.

app/lib/features/activities/widgets/security_and_privacy_section/store_the_key_securely_widget.dart
gnunicorn: static isn't a memory allocation. it just means that you can call this function -- associated with this class -- without an instance of that class. it is the same as any other function and doesn't take any more memory. I prefer to have the functions on the classes because then their assocation is clear, if one changes the check and forgets to update it if it is used at several places, that is way less clear. having it defined here and just reusing the check ensures we do the same checks everywhere.

PR #3030, opened 2025-05-28
[smallfix] Miscellaneous fixes

bitfriend:
- Fix typo
- Chain the builder methods sequentially


bitfriend:
> the typo I understand, but on the other change I am not clear: why do we want this change?
> 
> I do remember there being a problem with the full builder pattern with our ffi-gen based system. I think it means we are leaking pointers and then it lead to crashes or something, that's why I opted to always do `&mut self` without returning. So in order to err on the side of caution and not introduce any bugs that could cause crashes, I'd rather not apply this change unless there is a strong argument/need for doing that. But even then it should be having a bunch of manual tests being done to make sure we don't crash it with this bug.

Will revert `&mut Self` about also structs that has `derive(Builder)` flag.

native/acter/src/api/convo.rs
```diff
@@ -419,13 +419,12 @@ impl CreateConvoSettingsBuilder {
...
-            }
+        let user_id = UserId::parse(value)?;
+        if let Some(mut invites) = self.invites.clone() {
+            invites.push(user_id);
+            self.invites = Some(invites);
+        } else {
+            self.invites = Some(vec![user_id]);
         }
         Ok(())
     }
```
bitfriend, line 430: @gnunicorn What do you think about this change?
I think it is reasonable to call `?` operator than using of `if let Ok()`, because this api returns `Result<()>`.

bitfriend:
When using the fixed alias, we were experiencing failure on several times of running on my local.
Because we were trying to register `wombat` that was already registered.

PR #3034, opened 2025-05-28
Chat layout improvement

kumarpalsinh25:
### Chat-NG is getting even better with the greatness of all new chat background graphics and blurred chat header area.

Mobile View
Desktop View

---

### VIDEO REFERENCE:

https://github.com/user-attachments/assets/763d8a5c-feeb-4ff2-924c-209fa437f4ea


app/assets/images/chat_bg.png
gnunicorn, line 1: where does that image come from? what's the license?

app/lib/features/chat_ng/widgets/chat_messages.dart
```diff
@@ -149,7 +149,7 @@ class ChatMessagesConsumerState extends ConsumerState<ChatMessages> {
       key: ref.watch(animatedListChatMessagesProvider(widget.roomId)),
       controller: _scrollController,
       reverse: true,
-      padding: const EdgeInsets.only(bottom: 40),
+      padding: const EdgeInsets.only(bottom: 40, top: 60),
```
gnunicorn, line 152: shouldn't this instead be done in the `Positioned` as `top: 60` and `bottom: 40` in the surrounding element? I find it odd to change the padding of an inner element because the structure outside has been changed...

app/lib/features/chat_ng/widgets/chat_room/app_bar_widget.dart
```diff
@@ -28,14 +29,30 @@ class ChatRoomAppBarWidget extends ConsumerWidget
...
-          _buildRoomAvatar(context),
-          Expanded(child: _buildRoomTitle(context, ref)),
+          // Blur effect
+          BackdropFilter(
+            filter: ImageFilter.blur(sigmaX: 10.0, sigmaY: 10.0),
+            child: Container(
+              color: Colors.black.withValues(alpha: 0.2), // Optional overlay
+            ),
+          ),
+          // AppBar content
```
gnunicorn, line 42: do we really want a blurred background on the appbar? I think the png provided shows that it just makes the content less readable. We are doing a lot of stacking and positioning (I really hope doesn't land on our feet because of unforeseen side-effects) in a rather complex structure to make that happen... for what? that the app bar has less contrast?

app/test/features/chat_ng/chat_room_page/goldens_images/chat_room_page.png
gnunicorn, line 1: I don't find this example very convincing, visually speaking. The blur leads to the content being harder to read and the hard-edge on it means the background seems rather distorted.

gnunicorn (approved):
Is this really better?

I am not standing in the way here, but I feel the need to point out that a) neither WhatsApp nor Signal do blurred AppBars, b) it is a lot of complex widgeting to make that possible for c) a rather underwhelming effect IMHO. I'd drop the the AppBar part and stick with the regular appbar we had before. But if you insist, I only care 6/10 about that.

For the background image, I'd like to ensure we have the proper rights and can proof them.

app/assets/images/chat_bg.png
kumarpalsinh25: This image design by @kokodavid with our custom implementation.
https://www.figma.com/design/M6fUkOq6Dno22PoIjOV8JQ/Acter-3.1---Design-Workspace?node-id=6934-24418&p=f&t=Z8m9t5Xv3Aa5xEyx-0

So there shouldn't be any copyright or license issue.

Moreover I have also confirmed that inner smaller icons are also custom made by us or used open source content.

app/lib/features/chat_ng/widgets/chat_room/app_bar_widget.dart
gtalha07: Though there's not benefit of it in terms of usability and more like enhancing UI, I feel like the blur values are quite strong to create psychological cue of constantly shifting focus more to app bar than messages which should not be the case, at least what I experience. Better would be to tone down blur effect so it doesn't feel much distracting.

app/lib/features/chat_ng/widgets/chat_messages.dart
kumarpalsinh25: Positioned will not work, as I wanted blurred effect with overlay of chat list inner content.

app/lib/features/chat_ng/widgets/chat_room/app_bar_widget.dart
kumarpalsinh25: I have updated `ImageFilter.blur` and `alpha` value of `Container` which make it better and covered the different use-cases.

Image with Blue bubble overlay with App Bar:
Screenshot 2025-06-03 at 8 18 53 PM

Image with White text bubble overlay with App Bar:
Screenshot 2025-06-03 at 8 19 08 PM

app/test/features/chat_ng/chat_room_page/goldens_images/chat_room_page.png
kumarpalsinh25: Here is the update golden image with new code.

image

PR #3039, opened 2025-05-29
Fix bug in locations of calendar event builder

bitfriend:
Will fix bug that adding of location to `CalendarEventBuilder` failed.
Will append the validation of physical location.

### DIFF
--- a/native/acter/api.rsh
+++ b/native/acter/api.rsh
@@ -824,7 +824,7 @@ object EventLocationInfo {
     /// an online link for the location
     fn uri() -> Option<string>;
 
-    /// available for both physical and virtual
+    /// available for physical event only
     fn address() -> Option<string>;
     /// available for both physical and virtual
     fn notes() -> Option<string>;
--- a/native/core/src/events/calendar.rs
+++ b/native/core/src/events/calendar.rs
@@ -37,11 +37,11 @@ pub enum EventLocation {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         uri: Option<String>,
 
-        /// Also physical location can have a website
+        /// Optional address of this physical location
         #[serde(default, skip_serializing_if = "Option::is_none")]
         address: Option<String>,
 
-        /// Also physical location can have a website
+        /// Optional notes of this location
         #[serde(default, skip_serializing_if = "Option::is_none")]
         notes: Option<String>,
     },
@@ -61,7 +61,7 @@ pub enum EventLocation {
         /// URI to this virtual location
         uri: String,
 
-        /// Also physical location can have a website
+        /// Optional notes of this location
         #[serde(default, skip_serializing_if = "Option::is_none")]
         notes: Option<String>,
     },
--- a/packages/rust_sdk/lib/acter_flutter_sdk_ffi.dart
+++ b/packages/rust_sdk/lib/acter_flutter_sdk_ffi.dart
@@ -43895,7 +43895,7 @@ class EventLocationInfo {
     return tmp2;
   }
 
-  /// available for both physical and virtual
+  /// available for physical event only
   String? address() {
     var tmp0 = 0;
     tmp0 = _box.borrow();


native/core/src/events/calendar.rs
```diff
@@ -52,6 +57,17 @@ pub enum EventLocation {
         /// Alternative Icon to show with this location
         #[serde(default, skip_serializing_if = "Option::is_none")]
         icon: Option<Icon>,
+
+        /// URI to this virtual location
+        uri: String,
+
+        /// Also physical location can have a website
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        address: Option<String>,
```
gnunicorn, line 66: but a virtual thing can't have an address...

gnunicorn (changes requested):
Why are you adding a parsing check for coordinates? We are not using those (at all) and I'd rather not have a check for creating them, as we can't control what others are sending, we have to check again anyways at display-time -- and we haven't agreed on what we actually want to do and thus check for. But more over: we are not using them, and the changes should only be about address and notes, the missing fields. Please remove the regexp and the coordination parsing and the superfluous `address`-field on virtual locations.

native/core/src/events/calendar.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/3039/commits/66b29967f17d262e90c07e921fb71589496ab0d8

native/acter/api.rsh
```diff
@@ -812,14 +813,21 @@ object CalendarEventDraft {
...
     fn name() -> Option<string>;
     /// get the location description
     fn description() -> Option<TextMessageContent>;
+
     /// geo uri for the location
     fn coordinates() -> Option<string>;
     /// an online link for the location
     fn uri() -> Option<string>;
+
+    /// available for both physical and virtual
```
gnunicorn, line 827:
```suggestion
    /// available for physical event only
```

native/core/src/events/calendar.rs
```diff
@@ -33,14 +33,19 @@ pub enum EventLocation {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         coordinates: Option<String>,
 
-        /// further Link
+        /// Also physical location can have a website
         #[serde(default, skip_serializing_if = "Option::is_none")]
         uri: Option<String>,
+
+        /// Also physical location can have a website
```
gnunicorn, line 40:
```suggestion
        /// Only physical locations have an address
```

native/core/src/events/calendar.rs
```diff
@@ -33,14 +33,19 @@ pub enum EventLocation {
...
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        address: Option<String>,
+
+        /// Also physical location can have a website
```
gnunicorn, line 44:
```suggestion
        /// Notes are available on both
```

native/core/src/events/calendar.rs
```diff
@@ -52,6 +57,13 @@ pub enum EventLocation {
         /// Alternative Icon to show with this location
         #[serde(default, skip_serializing_if = "Option::is_none")]
         icon: Option<Icon>,
+
+        /// URI to this virtual location
+        uri: String,
+
+        /// Also physical location can have a website
```
gnunicorn, line 64:
```suggestion
        /// All event locations can have additional notes
```

gnunicorn (approved):
some comments are wrong, but good otherwise.

native/core/src/events/calendar.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/3039/commits/bb892bc8a74d09755d221c72d737b6c9886d8136

native/core/src/events/calendar.rs
bitfriend: Fixed by https://github.com/acterglobal/a3/pull/3039/commits/bb892bc8a74d09755d221c72d737b6c9886d8136

PR #3040, opened 2025-05-29
[Chat ng] : editor height fixes

gtalha07:


kumarpalsinh25:
Any way to get fixed on this weird issue that I often seeing in my iPhone?

Screenshot 2025-05-30 at 10 52 45 AM

kumarpalsinh25 (approved):
Not much getting detail idea on core logic however can see improvement from video reference.

gtalha07:
> Any way to get fixed on this weird issue that I often seeing in my iPhone?
> Screenshot 2025-05-30 at 10 52 45 AM

1. The formatting bar being this way, I only saw it on iOS simulator. Though if you encountered it in actual device, I'd suggest opening ticket with video recording reference as it's separate issue.

2. I think this is because of max editor height reached and you can see cut off from above (the inner margins and padding) that are added during auto-expanding [PR](https://github.com/acterglobal/a3/pull/2996).

PR #3054, opened 2025-06-02
Fix tokio runtime error in async fn

bitfriend:
@anisha-e10 reported the following error during calling of `ObjectInvitationsManager::invite()`.
```
[2025-06-02][15:49:41.611914][panic][ERROR] thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime': /Users/dajirajinfotech/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/deadpool-runtime-0.1.4/src/lib.rs:75

[2025-06-02][15:49:41.624095][panic][ERROR] thread '<unnamed>' panicked at 'panic in a destructor during cleanup': /Users/dajirajinfotech/.rustup/toolchains/nightly-2025-04-22-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/panicking.rs:226
```
Will fix it.

### DIFF
--- a/native/acter/src/api/invitations/object_manager.rs
+++ b/native/acter/src/api/invitations/object_manager.rs
@@ -101,24 +101,28 @@ impl ObjectInvitationsManager {
     }
 
     pub async fn reload(&self) -> Result<Self> {
-        let client = self.client.clone();
-        let room = self.room.clone();
-        let event_id = self.inner.event_id();
-        RUNTIME
-            .spawn(async move { ObjectInvitationsManager::new(client, room, event_id).await })
-            .await?
+        ObjectInvitationsManager::new(
+            self.client.clone(),
+            self.room.clone(),
+            self.inner.event_id(),
+        )
+        .await
     }
 }
 
 impl ObjectInvitationsManager {
-    pub async fn new(client: Client, room: Room, event_id: OwnedEventId) -> Result<Self> {
-        let inner =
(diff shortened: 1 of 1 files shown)


bitfriend:
I found `AttachmentsManager::new()` was wrapped by `RUNTIME.spawn()`, but `ObjectInvitationsManager::new()` was not wrapped.

native/acter/src/api/invitations/object_manager.rs
```diff
@@ -101,24 +101,28 @@ impl ObjectInvitationsManager {
...
-            models::InvitationsManager::from_store_and_event_id(client.store(), event_id.as_ref())
-                .await;
-        Ok(ObjectInvitationsManager {
-            client,
-            room,
-            inner,
-        })
+    pub(crate) async fn new(client: Client, room: Room, event_id: OwnedEventId) -> Result<Self> {
+        RUNTIME
+            .spawn(async move {
```
gnunicorn, line 116: let's not have runtime in `new` calls, but instead the thing around it that calls `new`.

gnunicorn (changes requested):
This pretty sure isn't the issue, because a) the reported problem is with `invite`, which this doesn't touch, and b) there is no actual change it isn't missing any runtime, but just moves it around. this isn't fixing it.

PR #3056, opened 2025-06-03
html-ng fixes

### LINKED ISSUE #3060: Chat editor: Edit mode on message with mentions shows markdown text
https://github.com/user-attachments/assets/742366d7-5cdf-42e4-a307-845c92ec2b0d

### LINKED ISSUE #3068: chat editor changes size while typing
Ever increasing....

https://github.com/user-attachments/assets/4fd17a01-97fd-42ea-95fc-b49598358678

gtalha07:
- fixes #3060: Now properly shows formatted content

## PREVIEW:

https://github.com/user-attachments/assets/0baba057-b769-4bf8-a0ab-74d4db21f3d9

- fixes #3068: Editor's size remain unchanged unless going to next line and also shrinks gradually and smooth:

## PREVIEW

https://github.com/user-attachments/assets/394ce75f-a480-4015-b871-0e45895bf577

### DIFF
--- a/app/lib/features/events/pages/create_event_page.dart
+++ b/app/lib/features/events/pages/create_event_page.dart
@@ -64,9 +64,7 @@ class CreateEventPageConsumerState extends ConsumerState<CreateEventPage> {
     // description
     final desc = event.description();
     if (desc != null) {
-      textEditorState.insertTextAtCurrentSelection(
-        desc.formatted() ?? desc.body(),
-      );
+      textEditorState.copyMessageText(desc.body(), desc.formatted());
     } else {
       textEditorState = EditorState.blank();
     }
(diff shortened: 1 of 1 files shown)


app/lib/common/toolkit/html_editor/services/utils.dart
```diff
@@ -0,0 +1,20 @@
+// helper to convert foreign html to appflowy html
+String normalizeToAppflowyHtml(String html) {
```
gtalha07, line 2: It was discovered that any html that comes from outside of appflowy has slightly different formatting due to which it couldn't parse it to document correctly and making edit mode buggy (not pasting content). This function takes cares prior of those conditions and making sure we're keeping it to what appflowy correctly parses.

app/lib/features/events/widgets/add_event_location_widget.dart
```diff
@@ -38,22 +37,14 @@ class _AddEventLocationWidgetState extends ConsumerState<AddEventLocationWidget>
     _selectedType = widget.initialLocation?.type ?? LocationType.virtual;
 
     // Initialize editor states with initial data
-    if (widget.initialLocation == null) {
-      textEditorNoteState = EditorState.blank();
-      textEditorAddressState = EditorState.blank();
-    } else {
-      textEditorNoteState = EditorState(
-        document: ActerDocumentHelpers.parse(
```
gtalha07, line 46: don't need markdown parsers also, as opted out in favor of html which ```copyMessageText``` can take care perfectly!

kumarpalsinh25 (commented):
This PR changes looks quite difficult for me to review due lack of core understanding of Html and Editor related code. So can't make any judgement to approve or reject PR.

However just notice 2 things from give screen-recording as below.

### 1) Red color is applied to "Hello" word in edit mode. Why is the reason of it?
Screenshot 2025-06-17 at 9 03 31 AM

### 2) Editor place-holder text seeing looks like cutting off.
Screenshot 2025-06-17 at 9 03 59 AM

Apart from above 2 things, I would also like to view video to demonstrate how editor behave when we do long paragraph typing without new line. 
(It look good that editor height gets expanded with new line however I just wanted to confirm that same should work with long paragraph text even without new line)

PR #3064, opened 2025-06-04
Pass text style to non-html text messages, too

### LINKED ISSUE #3057: After Html NG PR - 2928 : Different font sizes issue.
After merge of https://github.com/acterglobal/a3/pull/2928, I notice inconsistency in font size of send and receiver text.

Image

gnunicorn:
fixes #3057  .

But unlike what that report stated, the problem wasn't with messages from others, but messages without HTML didn't get the same font settings passed down: 9a7f8f6ecbd8d3ff72022ded59398c2ff627c537

Additionally, f62dd9d7776e64c2cc8324f360a6d7abaf85d5f1 fixes unnecessary columns (with only one item) if we don't have a reply to show either.

Before:

Bildschirmfoto_2025-06-04_11-44-47

After:

after

It's subtle but the font on the left isn't bold anymore now.


kumarpalsinh25 (approved):
Changes looks good to me.

gnunicorn:
@kumarpalsinh25 not actively, no.

PR #3088, opened 2025-06-10
Chain the draft & builder methods sequentially

bitfriend:
Will refactor fn that returns `Result<()>` among draft & builder methods.
Like [this code snippet](https://github.com/acterglobal/a3/blob/c407e06842992fae6a353b015a9ddd75600bb5dd/native/acter/src/api/stories.rs#L472-L478).
Will apply this change to only crates that has `derive(Builder)` flag.

The goal is to use method chaining on both rust and flutter sides.


bitfriend:
If some api returns `&mut Self` on rust side (that means rust side keeps a single object) and referenced object type (ex, `CalendarEventUpdateBuilder`) on flutter side, flutter calls the destructor on every object, so several occurrences of flutter destructor are called about single rust object.
Of course, if rust side returns `Self` not `&mut Self`, it is not problem that the destructor is called several times, because we have several instances on rust side.
But we can chain methods on flutter side too, as following:
```dart
    final updateBuilder =
        calendarEvent.updateBuilder()
          ..title(newName)
          ..descriptionText(newName);
```
As you can see, it is to use `..` (cascade operator).
Instead, we have to return nothing from `api.rsh`, so that destructor can't be called several times.

bitfriend:
Appended small files for download testing.

PR #3091, opened 2025-06-11
Bookmark Icon updates while bookmark a space.

anisha-e10:
Fixes, https://github.com/acterglobal/a3-meta/issues/1018

- This PR solved the bookmark icon issue  by updating the bookmark icon's state in real-time when a user bookmarks or unbookmarks a space. Previously, the icon did not reflect the current state.

### Reference video : 

https://github.com/user-attachments/assets/8355baf7-3982-48c2-b1be-945cefb167dc

### DIFF
--- a/app/lib/common/providers/notifiers/space_notifiers.dart
+++ b/app/lib/common/providers/notifiers/space_notifiers.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:acter/common/providers/space_providers.dart';
 import 'package:acter/features/home/providers/client_providers.dart';
 import 'package:acter_flutter_sdk/acter_flutter_sdk_ffi.dart'
     show Client, Space, SpaceDiff;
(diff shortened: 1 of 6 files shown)


anisha-e10:
> @anisha-e10 Just share video of space bookmark flow for reference. Wanted to confirmed that `ref.invalidate(spaceIsBookmarkedProvider(spaceId));` making difference or not.

I’m still experiencing a delay with the bookmark functionality, so I’m currently working on the `spaceIsBookmarkedProvider`. I’m converting the PR to draft for now, and will request your review once it’s ready.

app/lib/common/providers/notifiers/space_notifiers.dart
```diff
@@ -185,3 +186,24 @@ class SpaceListNotifier extends Notifier<List<Space>> {
...
+  @override
+  Map<String, bool> build() {
+    final spaces = ref.watch(spacesProvider);
+    return Map.fromEntries(
+      spaces.map((space) => MapEntry(
+        space.getRoomIdStr(),
+        space.isBookmarked(),
+      )),
+    );
+  }
```
kumarpalsinh25, line 200: Not quit understanding this logic.

app/lib/features/space/widgets/space_toolbar.dart
```diff
@@ -129,4 +124,9 @@ class SpaceToolbar extends ConsumerWidget {
       ],
     );
   }
+
+  Future<void> _bookmarkSpace(WidgetRef ref) async {
+    final isBookmarked = ref.read(spaceBookmarkProvider.notifier).getBookmark(spaceId);
+    await ref.read(spaceBookmarkProvider.notifier).setBookmark(spaceId, !isBookmarked);
```
kumarpalsinh25, line 130: Why to get `isBookmarked` here and then pass it to same provider? Can't be it in `setBookmark` method itself?

kumarpalsinh25 (approved):
Looks good as functional point of view.

Added minor remarked which need to answer before merge.

app/lib/common/providers/notifiers/space_notifiers.dart
anisha-e10: This method is called when the notifier is first created and it watches the spacesProvider to get all available spaces
For each space, it creates a map entry with:

- The space's room ID as the key
- The space's bookmark status as the value

so, it updates the local state by creating a new map with the updated bookmark status.

`getBookmark() `returns the bookmark status for a given space ID and If the space ID isn't found in the map, it returns false (not bookmarked).

PR #3097, opened 2025-06-12
Now pin icon is updated straight away after change.

anisha-e10:
Fixes https://github.com/acterglobal/a3-meta/issues/1003

**Reference video :** 

https://github.com/user-attachments/assets/e6ebab86-b951-46b5-9528-37b74d6c4226

### DIFF
--- a/.changes/3097-update-pin-icon.md
+++ b/.changes/3097-update-pin-icon.md
@@ -1,0 +1,1 @@
+- [Fixes] : Pin icon updates instantly after a change.


app/lib/common/widgets/acter_icon_picker/acter_icon_widget.dart
```diff
@@ -39,9 +39,11 @@ class _ActerIconWidgetState extends State<ActerIconWidget> {
   }
 
   @override
-  void didChangeDependencies() {
-    super.didChangeDependencies();
-    _setWidgetValues();
+  void didUpdateWidget(ActerIconWidget oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.color != widget.color || oldWidget.icon != widget.icon) {
+      _setWidgetValues();
```
kumarpalsinh25, line 45: So now we are calling `_setWidgetValues()` base on old icon and color condition and earlier it was called every-time.

So not quit understanding root cause of the issue as `_setWidgetValues` every-time earlier then also it was not updating icon.

anisha-e10: The didChangeDependencies method is called, but the widget's state isn't being properly updated that's why replaced didChangeDependencies with didUpdateWidget to properly handle widget property updates and added a condition to only update the values when the color or icon actually changes.

kumarpalsinh25 (approved):

PR #3102, opened 2025-06-17
[Chat Editor]: increase drag handle size for iOS platform

### LINKED ISSUE #2951: selecting text in editor doesn't work
trying to double tap and drag can't change text selection.

gtalha07:
fixes #2951.

## VIDEO PREVIEW

https://github.com/user-attachments/assets/f6d157b8-8945-4c5e-8d95-230da002660c

### DIFF
--- a/.changes/2951-selection-fix.md
+++ b/.changes/2951-selection-fix.md
@@ -1,1 +1,1 @@
-[fix]: (Chat Editor) increased mobile drag handle size for iOS for easier text selection.
+[fix]: (Chat Editor) increased mobile drag handle size on Android/iOS platforms for easier text selection.
--- a/.changes/2951-selection-fix.md
+++ b/.changes/2951-selection-fix.md
@@ -1,0 +1,1 @@
+[fix]: (Chat Editor) increased mobile drag handle size for iOS for easier text selection.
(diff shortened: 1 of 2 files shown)


gnunicorn:
@gtalha07 I don't have iOS, but I reported the bug. thus, this might not be something we only want on iOS?

gtalha07:
> @gtalha07 I don't have iOS, but I reported the bug. thus, this might not be something we only want on iOS?



> Code wise looks fine, however I see the flickering in editor in the PR view at the time between 00:22 to : 0025 which need to fix.

That's not flickering, it looked though like that because of iOS selection bar that appeared on selection (Apple Intelligence suggestions) and I wasn't adding/removing selection due to which it made flicker

gtalha07:
> @gtalha07 I don't have iOS, but I reported the bug. thus, this might not be something we only want on iOS?

The drag handles of android and iOS are different. iOS, by default in Appflowy has small handle size which is made larger. But for usability, the android one isn't looking good either. I had made adjustments 👍🏻 ! 

### APPLIED CHANGES:

| Screenshot 2025-06-18 at 14 20 54 | WhatsApp Image 2025-06-18 at 14 27 32 |

PR #3113, opened 2025-06-19
Message draft fixes

gtalha07:
# DESCRIPTION:

- It has been observed that composer draft is unusual in its states because of some issues regarding it saving procedure. This PR attempts to fix those, as well as some reducing some unnecessary calls over draft to save, along with selection fixes. Most of the changes explanation have been commented out, while some can be also visually reviewed. But the best approach is **_REVIEW BY COMMIT_**.

1. [dc342e7](https://github.com/acterglobal/a3/pull/3113/commits/dc342e7e981f088b94a5f938269f1b1f32978a86) `new` draft not saving fixes:

### BEFORE:

https://github.com/user-attachments/assets/495c3043-416e-4d83-b07b-d18687de4340

### AFTER:

https://github.com/user-attachments/assets/d7d4f9f8-6a25-40e6-ab90-645ae9a61f55

2. [a0a8603](https://github.com/acterglobal/a3/pull/3113/commits/a0a8603b50151cb4b400aa306a1645329d0a7a2a) explained in [comment](https://github.com/acterglobal/a3/pull/3113/files#r2157612910) about dart classic `copyWith()` bug on nullable objects.

3. [da86317](https://github.com/acterglobal/a3/pull/3113/commits/da86317d486af5f70a7499024a0ce90eca367fed) Selection fixes.

### BEFORE: see how going over room would show selection cursor on empty draft which is unintended for desktop.

https://github.com/user-attachments/assets/d5797e8c-a9ce-496c-907d-845c030eb301

### AFTER:  it'd not set selection when draft is empty :)

https://github.com/user-attachments/assets/b349cfcb-a5ae-450d-b7a9-8a9a3225fe69

4. ## COMPLETE PREVIEW showing the fixes over draft and selection states.

https://github.com/user-attachments/assets/6f0147f3-618a-4c71-87b8-20e18c1f0378

### DIFF
--- a/app/lib/features/chat_ng/models/chat_editor_state.dart
+++ b/app/lib/features/chat_ng/models/chat_editor_state.dart
@@ -20,7 +20,7 @@ class ChatEditorState {
     MessageAction? actionType,
   }) {
     return ChatEditorState(
-      selectedMsgItem: selectedMsgItem ?? this.selectedMsgItem,
+      selectedMsgItem: selectedMsgItem,
       actionType: actionType ?? this.actionType,
     );
   }
--- a/app/lib/features/chat_ng/utils.dart
+++ b/app/lib/features/chat_ng/utils.dart
@@ -13,7 +13,6 @@ Future<void> saveMsgDraft(
   final chat = await ref.read(chatProvider(roomId).future);
   final chatEditorState = ref.read(chatEditorStateProvider);
   final messageId = chatEditorState.selectedMsgItem?.eventId();
-
   if (chat != null) {
     if (messageId != null) {
       if (chatEditorState.isEditing) {
--- a/app/lib/features/chat_ng/utils.dart
+++ b/app/lib/features/chat_ng/utils.dart
@@ -14,12 +14,15 @@ Future<void> saveMsgDraft(
   final chatEditorState = ref.read(chatEditorStateProvider);
   final messageId = chatEditorState.selectedMsgItem?.eventId();
 
-  if (chat != null && messageId != null) {
-    if (chatEditorState.isEditing) {
-      await chat.saveMsgDraft(text, htmlText, 'edit', messageId);
-    } else if (chatEditorState.isReplying) {
-      await chat.saveMsgDraft(text, htmlText, 'reply', messageId);
+  if (chat != null) {
+    if (messageId != null) {
+      if (chatEditorState.isEditing) {
+        await chat.saveMsgDraft(text, htmlText, 'edit', messageId);
+      } else if (chatEditorState.isReplying) {
+        await chat.saveMsgDraft(text, htmlText, 'reply', messageId);
+      }
     } else {
+      // no message id, create a unreferenced draft
       await chat.saveMsgDraft(text, htmlText, 'new', null);
     }
   }
--- a/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
+++ b/app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
@@ -37,7 +37,7 @@ class ChatEditor extends ConsumerStatefulWidget {
 }
 
 class _ChatEditorState extends ConsumerState<ChatEditor> {
-  EditorState textEditorState = EditorState.blank();
+  late EditorState textEditorState;
   StreamSubscription<EditorTransactionValue>? _updateListener;
   final ValueNotifier<bool> _isInputEmptyNotifier = ValueNotifier(true);
   Timer? _debounceTimer;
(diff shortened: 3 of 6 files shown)


app/lib/features/chat_ng/models/chat_editor_state.dart
```diff
@@ -20,7 +20,7 @@ class ChatEditorState {
     MessageAction? actionType,
   }) {
     return ChatEditorState(
-      selectedMsgItem: selectedMsgItem ?? this.selectedMsgItem,
```
gtalha07, line 23: this assignment is wrong for a reason that after ```unsetActions``` is called, will still fallback to non-null value. Which means the draft would still consider ```messageId``` to be present even when closing preview and mess with the edit/reply states.

app/lib/features/chat_ng/utils.dart
```diff
@@ -13,13 +13,15 @@ Future<void> saveMsgDraft(
...
   final chatEditorState = ref.read(chatEditorStateProvider);
   final messageId = chatEditorState.selectedMsgItem?.eventId();
-
-  if (chat != null && messageId != null) {
-    if (chatEditorState.isEditing) {
-      await chat.saveMsgDraft(text, htmlText, 'edit', messageId);
-    } else if (chatEditorState.isReplying) {
-      await chat.saveMsgDraft(text, htmlText, 'reply', messageId);
+  if (chat != null) {
+    if (messageId != null) {
```
gtalha07, line 17: also a minor conditional error here, `new` draft doesn't have any referenced id linked for edit/reply so deal it with nested `if`

app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
```diff
@@ -151,6 +163,18 @@ class _ChatEditorState extends ConsumerState<ChatEditor> {
       final fallbackPlain = draft.plainText();
       if (fallbackPlain.trim().isNotEmpty) {
         textEditorState.replaceContent(fallbackPlain, htmlBody);
+      } else {
+        // FOR DESKTOP/SIDE VIEW:
+        // it means the draft is empty, but still valid document (e.g. empty paragraph) for editor to get auto focused
+        // set selection to null to avoid the editor from being focused
```
gtalha07, line 169: so from Appflowy perspective, an empty editor still contains valid empty node, which editor would load initially with selection (auto-focus) by entering room. We simply check here with having desktop (sidebar) view to set selection to null if draft is empty.

app/lib/features/chat_ng/widgets/chat_editor/chat_editor_actions_preview.dart
```diff
@@ -89,22 +88,20 @@ class ChatEditorActionsPreview extends ConsumerWidget {
...
-
-            notifier.unsetActions();
+            final isEdit = ref.read(chatEditorStateProvider).isEditing;
             if (isEdit) {
               textEditorState.clear();
-            } else {
-              final body = textEditorState.intoMarkdown();
-              final bodyHtml = textEditorState.intoHtml();
-
-              await saveMsgDraft(body, bodyHtml, roomId, ref);
```
gtalha07, line 102: removing this entirely, because manual save isn't required where transactional changes below are enough to trigger `transactionStream` listener to update draft state.

app/lib/features/chat_ng/widgets/chat_editor/chat_editor.dart
```diff
@@ -87,10 +87,14 @@ class _ChatEditorState extends ConsumerState<ChatEditor> {
       if (next.isReplying &&
           (next.actionType != prev?.actionType ||
               next.selectedMsgItem != prev?.selectedMsgItem)) {
-        textEditorState.updateSelectionWithReason(
-          null,
-          reason: SelectionUpdateReason.uiEvent,
+        // set selection of editor for composing
+        final t = textEditorState.transaction;
+        t.afterSelection = Selection.single(
```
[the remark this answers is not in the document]
gtalha07: Hmm. Maybe I didn't find the relevant discussion but, here specifically: Appflowy doesn't seem to update cursor position based on the documents changes that are applied here automatically (always) . I've seen examples from the repo that this selection update is also applied when modifying nodes and having a draft state which also lives outside of widget and editor lifecycle where upon re-initialization, cursor state had to be determined manually. 

I think @emilvincentz also mentioned in one of Notion editor bug list, where applying mentions in between text sets wrong position of cursor ( I haven't looked at that code yet) but it seems selection doesn't get updated in that use case.

PR #3120, opened 2025-06-23
Chat: dm read tracking

gtalha07:


gnunicorn:
@gtalha07  in the video, shouldn't there already be one check mark if it has been sent successfully?

gtalha07:
> @gtalha07 in the video, shouldn't there already be one check mark if it has been sent successfully?

I tried it, and it doesn't work well enough. `EventSendState` object gets disposed after being `Sent` and then there's no way to figure out what was the last known state until read. Matrix spec isn't allowing it maybe? This'll work just like we have usual read receipts but with double checks...

Issue #3122, opened 2025-06-23
event locations don't show after edited


gnunicorn:
Love the new event location feature, but noticed that the UI doesn't indicate that any have been added when during the creation process.

**To reproduce**:
1. create a new event
2. select + locations
3. follow the process to add one or two locations
4. close the locations screen

**Expected behavior**
The locations sections has updated, showing the new locations or indicating somehow else that these have been added

**Current result**
The UI stays the same, nothing appears to have changed:

**Screencast**:

https://github.com/user-attachments/assets/0c05057d-4b2c-4595-90ac-0afb9aa667a6

PR #3126, opened 2025-06-24
Event locations is now visible to users at the time of event creation within the same view instead of dialog.

anisha-e10:
Fixes https://github.com/acterglobal/a3/issues/3122

- This PR enhances the event creation experience by displaying the event location in the same view during event creation.

### Reference video : 

https://github.com/user-attachments/assets/e06d7cf2-b1dc-481e-aa29-eb5da1948ea7


kumarpalsinh25 (approved):
LGTM!

PR #3134, opened 2025-06-25
a3 flutter update (3.32.4)

gtalha07:
# Description:

- Upgraded dependencies and removed ```convenient_test_manager``` as not usable in development now. Also updated theme APIs according to new engine.

### AFTER MERGE, SHOULD BE ADVISED TO UPGRADE FLUTTER TO `3.32.4` AS IT'LL BREAK CURRENT SETUP.


kumarpalsinh25 (approved):
Look good overall however I see that some of part of unit tests get removed, any reason for it?

app/pubspec.yaml
```diff
@@ -186,20 +186,17 @@ dev_dependencies:
...
-  convenient_test_dev: ^1.3.7
-  convenient_test_manager_dart:
-    git:
-      url: https://github.com/gnunicorn/flutter_convenient_test/
-      ref: ben-cli-limit-tests
-      path: packages/convenient_test_manager_dart
   msix: ^3.7.0
   uuid: ^4.0.0
-  enough_mail: ^2.1.6
+  # TODO: move to upstream version this once enough_mail is published
```
gnunicorn, line 191: I don't thinks this is used outside of the integration tests - this can be removed as well.

app/test/features/chat/custom_input_test.dart
```diff
@@ -188,102 +187,6 @@ void main() {
         (ref, roomId) => Future.value(roomDrafts[roomId]),
       ),
     ];
```
gnunicorn, line 189: not good to remove unittest but it is only for the legacy chat, which we intend to remove anyways...

app/test/features/chat_ng/messages/chat_message_test.dart
```diff
@@ -101,31 +93,6 @@ void main() {
         expect(result, true);
       });
     });
-
-    group('Image Messages Test', () {
```
gnunicorn, line 105: why is this one removed though?

gnunicorn (approved):

app/test/features/chat/custom_input_test.dart
gtalha07: I can revert it to comment as whole it as it was using `convienient_test_manager`.

PR #3156, opened 2025-07-01
Make HTML Description dialog editable and improve UI styling of title and description dialog.

anisha-e10:
Fixes https://github.com/acterglobal/a3/issues/3155

### **_✨ Editable Description Dialog + UI Improvements_**

- Made the description field editable
- Improved UI styling for title and description dialogs.
- Require users to press Save or Cancel (dialog is not draggable now).

### Reference video: 

https://github.com/user-attachments/assets/4c90c026-de31-4a31-834f-0f2c5b4f3b77

### DIFF
--- a/app/lib/common/themes/components/button_theme.dart
+++ b/app/lib/common/themes/components/button_theme.dart
@@ -6,7 +6,10 @@ import 'package:flutter/material.dart';
 
 ElevatedButtonThemeData elevatedButtonTheme() => ElevatedButtonThemeData(
   style: ElevatedButton.styleFrom(
-    padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
+    padding: EdgeInsets.symmetric(
+      horizontal: 16,
+      vertical: Platform.isAndroid || Platform.isIOS ? 12 : 16,
+    ),
     elevation: 0,
     textStyle: textTheme.titleMedium?.copyWith(
       color: colorScheme.primary,
@@ -48,7 +51,7 @@ OutlinedButtonThemeData outlinedButtonTheme() => OutlinedButtonThemeData(
     shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(10)),
     padding: EdgeInsets.symmetric(
       horizontal: 12,
-      vertical: Platform.isAndroid || Platform.isIOS ? 6 : 12,
+      vertical: Platform.isAndroid || Platform.isIOS ? 6 : 16,
     ),
     textStyle: textTheme.titleMedium?.copyWith(fontSize: 15),
     minimumSize: Size.zero,
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -252,14 +252,10 @@ bool isEmptyHtmlContent(String html) {
 
 /// html requires to have some kind of structure even when document is empty, so check for that
 bool hasValidEditorContent({required String plainText, required String html}) {
-  // If both are completely empty, no valid content
-  if (plainText.trim().isEmpty && html.trim().isEmpty) return false;
-
-  // Check if either plainText has content OR HTML has content
   final hasPlainTextContent = plainText.trim().isNotEmpty;
-  final hasHtmlContent = !isEmptyHtmlContent(html);
+  final hasHtmlTextContent = !isEmptyHtmlContent(html);
   
-  return hasPlainTextContent || hasHtmlContent;
+  return hasPlainTextContent || hasHtmlTextContent;
 }
 
 String formatChatDayDividerDateString(BuildContext context, String dateString) {
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -254,7 +254,6 @@ bool isEmptyHtmlContent(String html) {
 bool hasValidEditorContent({required String plainText, required String html}) {
   final hasPlainTextContent = plainText.trim().isNotEmpty;
   final hasHtmlTextContent = !isEmptyHtmlContent(html);
-  
   return hasPlainTextContent || hasHtmlTextContent;
 }
 
--- a/app/lib/common/utils/utils.dart
+++ b/app/lib/common/utils/utils.dart
@@ -230,20 +230,32 @@ extension ColorUtils on Color {
   }
 }
 
+/// Comprehensive regex for detecting empty HTML tags and structures
+final RegExp emptyHtmlTagsRegex = RegExp(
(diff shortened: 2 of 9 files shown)


app/lib/common/themes/components/button_theme.dart
```diff
@@ -6,7 +6,10 @@ import 'package:flutter/material.dart';
 
 ElevatedButtonThemeData elevatedButtonTheme() => ElevatedButtonThemeData(
   style: ElevatedButton.styleFrom(
-    padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
+    padding: EdgeInsets.symmetric(
+      horizontal: 16,
+      vertical: Platform.isAndroid || Platform.isIOS ? 12 : 16,
```
gtalha07, line 11: I don't understand this change. Different padding for different platforms ? Not getting the motive...

anisha-e10: Yes, the padding is intentionally different for mobile and desktop platforms to optimize the UI for each form factor.

On Android and iOS, we use slightly smaller vertical padding (12) to make better use of limited screen space and align with mobile design guidelines.

On desktop platforms, where there’s generally more screen real estate, we use larger vertical padding (16) for a more spacious and readable layout.

app/lib/common/widgets/edit_html_description_sheet.dart
```diff
@@ -56,29 +66,49 @@ class _EditHtmlDescriptionSheetState
   @override
   void initState() {
     super.initState();
-    textEditorState = ActerEditorStateHelpers.fromContent(
-      widget.descriptionMarkdownValue ?? '',
-      widget.descriptionHtmlValue,
-    );
+
+    final hasValidContent = hasValidEditorContent(
```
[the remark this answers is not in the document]
anisha-e10: It was not working — it's adding empty `<br>` tag or whitespace may be  when the user doesn't enter a description. Then, when trying to open it again for editing, the editor doesn't allow editing and doesn't focus either.

app/lib/common/widgets/edit_html_description_sheet.dart
```diff
@@ -56,29 +66,49 @@ class _EditHtmlDescriptionSheetState
...
             height: 200,
             decoration: BoxDecoration(
               border: Border.all(color: brandColor),
               borderRadius: BorderRadius.circular(10.0),
             ),
-            child: HtmlEditor(editorState: textEditorState, editable: true),
+            child: HtmlEditor(
+                editorState: textEditorState,
+                editable: true,
+                shrinkWrap: false,
```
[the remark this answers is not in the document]
anisha-e10: Agree Addressed by this [commit](https://github.com/acterglobal/a3/pull/3156/commits/bb7c9620cfd3118bd5ded1c9e76922340225db69)

app/lib/common/utils/utils.dart
```diff
@@ -230,20 +230,36 @@ extension ColorUtils on Color {
   }
 }
 
+/// Comprehensive regex for detecting empty HTML tags and structures
+final RegExp emptyHtmlTagsRegex = RegExp(
+  r'<br\s*/?>\s*|<p\s*>(\s|&nbsp;|&#160;|<br\s*/?>)*</p>\s*|<div\s*>(\s|&nbsp;|&#160;)*</div>\s*|<span\s*>(\s|&nbsp;|&#160;)*</span>\s*|<h[1-6]\s*>(\s|&nbsp;|&#160;)*</h[1-6]>\s*|<(strong|b|em|i)\s*>(\s|&nbsp;|&#160;)*</(strong|b|em|i)>\s*|<(ul|ol|li)\s*>(\s|&nbsp;|&#160;)*</(ul|ol|li)>\s*',
+  multiLine: true,
+  caseSensitive: false,
+);
```
kumarpalsinh25, line 238: So I believe empty html tag is the root cause of the editor not work?

It is always bit difficult to judge regex pattern, however I am seeing bit of unit-test cases so that is good.

app/lib/common/utils/utils.dart
```diff
@@ -230,20 +230,36 @@ extension ColorUtils on Color {
...
+
+/// Check if HTML content contains only empty tags or whitespace
+bool isEmptyHtmlContent(String html) {
+  if (html.trim().isEmpty) return true;
+  
+  // Remove all empty tags and structures
+  String cleanedHtml = html
+      .replaceAll(emptyHtmlTagsRegex, '')
+      .replaceAll(RegExp(r'\s+'), ' ') // Normalize whitespace
```
kumarpalsinh25, line 247: Why `.replaceAll(RegExp(r'\s+'), ' ') // Normalize whitespace` is not part of `emptyHtmlTagsRegex` itself?

app/lib/common/utils/utils.dart
```diff
@@ -230,20 +230,36 @@ extension ColorUtils on Color {
...
-          .replaceAll('&nbsp;', ' ')
-          .trim()
-          .isEmpty;
-
-  return !hasOnlyStructure;
+  // If both are completely empty, no valid content
+  if (plainText.trim().isEmpty && html.trim().isEmpty) return false;
+
+  // Check if either plainText has content OR HTML has content
+  final hasPlainTextContent = plainText.trim().isNotEmpty;
```
kumarpalsinh25, line 259: Why to check `plainText` condition twice?

app/lib/common/utils/utils.dart
```diff
@@ -230,20 +230,36 @@ extension ColorUtils on Color {
...
+  final hasHtmlContent = !isEmptyHtmlContent(html);
+  
+  return hasPlainTextContent || hasHtmlContent;
```
kumarpalsinh25, line 262: This can be simplified as below.

```suggestion

  final hasPlainTextContent = plainText.trim().isNotEmpty;
  final hasHtmlTextContent = !isEmptyHtmlContent(html);
  
  return hasPlainTextContent || hasHtmlTextContent;
```

kumarpalsinh25 (approved):
Looks good overall, however added minor remarks which need to address before merge.

app/lib/common/widgets/edit_html_description_sheet.dart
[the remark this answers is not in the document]
anisha-e10: Please check for any Pins or Tasks that already have an empty description. Do not add and then remove the description as shown in the video.

app/lib/common/utils/utils.dart
anisha-e10: They're kept separate because:

1. First remove empty HTML tags, then normalize the remaining whitespace.
2. emptyHtmlTagsRegex can be used elsewhere without always needing whitespace normalization

emptyHtmlTagsRegex is designed to match and remove empty HTML tags and structures like:
```
<br /> tags
Empty <p>, <div>, <span> tags
```
Empty headings, formatting tags, lists, etc.
`.replaceAll(RegExp(r'\s+'), ' ')` serves a different purpose - it normalizes whitespace by collapsing multiple whitespace characters (spaces, tabs, newlines) into single spaces.

app/lib/common/utils/utils.dart
anisha-e10: Addressed by this [commit](https://github.com/acterglobal/a3/pull/3156/commits/db5bac1d585420531a5dfee5afb2e6ef131dbc80)

Issue #3032, opened 2025-05-28
Move off `flutter_adaptive_scaffold`


gnunicorn:
It's been deprecated: https://github.com/flutter/flutter/issues/162965

We currently use that for our dashboard and managing of the outer-appshell with the navigation. Moving off that package isn't going to be easy, but might give us the opportunity to add tests for that and the different sizes and stuff to make sure we don't break it again in the future.

PR #3159, opened 2025-07-02
Add tests for AppShell and sizes to prepare for package migration.

anisha-e10:
Fixes https://github.com/acterglobal/a3/issues/3032

### DIFF
--- a/app/lib/common/utils/breakpoints.dart
+++ b/app/lib/common/utils/breakpoints.dart
@@ -1,0 +1,144 @@
+import 'package:flutter/material.dart';
+
+/// Current breakpoints used across the app - documenting existing behavior
+/// before migration from flutter_adaptive_scaffold
+class ActerBreakpoints {
+  // Current custom breakpoint from acter_build_context.dart
+  static const double largeScreen = 770;
+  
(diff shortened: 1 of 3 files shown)


app/lib/common/utils/breakpoints.dart
```diff
@@ -0,0 +1,143 @@
+import 'package:flutter/material.dart';
+
+
+class ActerBreakpoints {
```
kumarpalsinh25, line 4: Where `ActerBreakpoints` used?
Is this just for declaration and test purpose only or we are `flutter_adaptive_scaffold` package and will use `ActerBreakpoints` ?

kumarpalsinh25 (commented):
I don't quite understand purpose of this PR. Is it just for implementing `ActerBreakpoints` as our custom solution in replacement of `flutter_adaptive_scaffold`?

If yes then I don't see usage of that in this PR.

app/lib/common/utils/breakpoints.dart
gnunicorn: No, sorry, I don't really know what this is about ...

